PR #8, opened 2018-05-25
fixed service.type (grafana, alertmanager)

BWalti:
fixed: using .service.type instead of .server.type for grafana and alertmanager type in questions.yml


guangbochen:
LGTM and merged, thanks.

PR #12, opened 2018-06-21
Fixed ingress for Grafana

DavidZisky:
Grafana is listening on port 3000 (set in grafana-deployment.yaml) so default 80 doesn't work.


guangbochen:
ingress is and always listening to the service port, so no need to change the port number, will close it and thanks.

DavidZisky:
Well... I know, I will double check, but when I deployed EFK from catalog I was not able to access grafana. So I checked config (of grafana) and I saw it's being exposed on port 3000, so then I went to grafana ingress config and I saw it's pointing to grafana service on port 80 and after I changed it to 3000 it started working.

PR #23, opened 2018-08-15
Longhorn v0.3 Chart

LLParse:
https://github.com/rancher/longhorn/issues/200


proposed/longhorn/v0.3.0/templates/deployment-driver.yaml
```diff
@@ -1,25 +1,37 @@
...
           - -d
-          - deploy-flexvolume-driver
+          - deploy-driver
           - --manager-image
           - "{{ .Values.image.longhorn.manager }}:{{ .Values.image.longhorn.managerTag }}"
+          - --manager-url
+          - http://longhorn-backend:9500/v1
+          # manually choose "flexvolume" or "csi"
+          - --driver
+          - csi
```
LLParse: This needs to be parameterized and exposed as a question

proposed/longhorn/v0.3.0/templates/deployment-driver.yaml
```diff
@@ -29,6 +41,16 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: spec.nodeName
+          - name: SERVICE_ACCOUNT
+            valueFrom:
+              fieldRef:
+                fieldPath: spec.serviceAccountName
           - name: FLEXVOLUME_DIR
+            # FOR RKE
```
LLParse: Remove this mnemonic device. Charts aren't aware of what k8s distro is in use

guangbochen:
@LLParse, can we please rename the app structure to
```
proposed/longhorn/latest/ # use this for the latest chart files
proposed/longhorn/v0.1.0/ # move the old version to a new path with the specified version name
```

we will rename all the project charts to the above structure, make it easy to track the code changes, thanks.

LLParse:
@guangbochen Sure thing, done.

proposed/longhorn/latest/questions.yml
```diff
@@ -2,6 +2,7 @@ categories:
 - storage
 labels:
   io.rancher.certified: experimental
+namespace: longhorn-system
 questions:
 - variable: defaultImage
```
LLParse: @yasker IMO we should completely remove this question and the image tag sub-questions. The deployed image versions are not something to be messed with by a regular user; if a developer wants this, they can use their own.

Also note that upgrading to a newer catalog with different defaults for these questions will not change the question answers automatically. Consequently, upgrading from `0.3.0` to theoretical release `0.3.1` will never update the image tags. The upgrade will seem to succeed but no images change.

proposed/longhorn/v0.1.0/Chart.yaml
```diff
@@ -2,4 +2,5 @@ apiVersion: v1
 description: Longhorn is a distributed block storage system for Kubernetes powered by Rancher Labs.
 name: longhorn
 version: 0.1.0
+kubeVersion: 1.8 - 1.10
```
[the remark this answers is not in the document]
yasker: Yeah, I don't think Rancher is using Helm except to generate the yaml for `kubectl apply -f`. Though v1.11 should work now, let's update it to `v1.11` (and have it tested).

proposed/longhorn/latest/questions.yml
yasker: Yeah, if the old value was used as the default engine in the new version of manager, that's really dangerous.

If we can find a way to use the default value (instead of user-specified value) every time when the user selected app, the question will still be useful. Otherwise, let's get rid of this option and hard code it to the chart.

proposed/longhorn/v0.1.0/Chart.yaml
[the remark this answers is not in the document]
LLParse: This is Longhron v0.1.0 chart which has problems with K8s 1.11 - I'll just remove the line altogether since it has no effect.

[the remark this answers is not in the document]
yasker: Sounds good.

LLParse:
No, not really. The only potential solutions I've found, in order from bad to worse are:
1. Rename the chart to something other than `longhorn`.
2. Give the chart a bogus version in `Chart.yml` that is considered an earlier release (semantic versioning)
3. Remove the old chart version altogether

yasker:
Well, probably neither of those are going to happen. What if the user clicks `upgrade`, what would happen? The default namespace is different. Is it going to error out due to that?

yasker:
And v0.3.0 is out. Can we update the chart?

yasker:
@LLParse In fact, we can remove the old version, if that helps.

LLParse:
@yasker Dang, removing the old version didn't block the upgrade path. It does prevent users from getting into the situation in the first place, so I think we should still remove it.

LLParse:
If user clicks `upgrade`, the chart errors out and become a huge mess to clean up.

PR #27, opened 2018-08-22
Added mounted parsers.conf

Razikus:
Was already in configmap, but wasn't even used


cjellick:
Is this still relevant? Even though it's old, seems like it could be.
I'm going to close due to age, but feel free to reopen if you think it's still relevant. Thanks!

PR #32, opened 2018-08-28
Added OpenEBS v0.6.0 Chart.

muratkars:
**What's new in this release:**

- Initial commit to the Rancher Catalog
- Integrate the Volume Snapshots capabilities with Kubernetes Snapshot controller
- Enhance maya-apiserver to use CAS Templates for orchestrating new Storage Engines
- Enhance mayactl to provide additional details about volume such as - replica status, node details where replicas are running.
- Enhance maya-apiserver to schedule Replica Pods on specific nodes using nodeSelector
- Enhance provisioner and maya-apiserver to allow specifying cross AZ scheduling of Replica Pods.
- Support for deploying OpenEBS via Kubernetes stable Helm Charts
- openebs-operator.yaml is modified to run OpenEBS pods in its own namespace openebs
- Enhance e2e tests to simulate chaos at different layers like - CPU, RAM, Disk, Network, and Node

Signed-off-by: Murat Karslioglu <muratkarslioglu@gmail.com>


muratkars:
Hi @guangbochen was it enabled in the values.yaml file? Default is set to policies -> monitoring -> enabled: true

guangbochen:
Hi @muratkars, yes the value is enabled but the `openebs/m-exporter` image is not deployed anywhere on the chart, seems a pod or deployment for this exporter is missing.

kmova:
@guangbochen -after installing the above components and when you create a PVC/PV with openebs storage class - it will create a PV which comprises of controller and replica pods. The m-exporter will be launched as a side-car container to the volume pods. More about the openebs volume can be found at: https://docs.openebs.io/docs/next/storageengine.html

PR #34, opened 2018-08-30
Update Longhorn chart documentation

LLParse:
https://github.com/rancher/longhorn/issues/261

RFR

### DIFF
--- a/proposed/longhorn/latest/README.md
+++ b/proposed/longhorn/latest/README.md
@@ -1,3 +1,7 @@
+# Rancher Longhorn Chart
+
+The following document pertains to running Longhorn from the Rancher 2.0 chart.
+
 ## Source Code
 
 Longhorn is 100% open source software. Project source code is spread across a number of repos:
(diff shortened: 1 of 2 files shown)


proposed/longhorn/latest/README.md
```diff
@@ -9,22 +9,34 @@ Longhorn is 100% open source software. Project source code is spread across a nu
...
 ## Uninstall Longhorn
 
-In order to uninstall Longhorn, user need to remove all the volumes first:
-```
-kubectl -n longhorn-system delete lhv --all
-```
+**Do not attempt to immediately delete Longhorn App from Rancher UI. Longhorn propagates device mounts which can damage the host if not unmounted properly.** Follow this procedure instead.
 
-After confirming all the volumes are removed, then Longhorn can be easily uninstalled using:
+1. To prevent damage to the Kubernetes cluster, we recommend deleting all Kubernetes workloads using Longhorn volumes (Deployments, StatefulSets, DaemonSets, etc).
```
yasker: Need to add a document in the main repo (e.g. `docs/chart.md`). It's better if we just link and copy the document here, rather than let this document live on its own in the chart repo.

proposed/longhorn/latest/README.md
```diff
@@ -9,22 +9,34 @@ Longhorn is 100% open source software. Project source code is spread across a nu
...
+kubectl -n ${NAMESPACE} delete service/csi-provisioner
+kubectl -n ${NAMESPACE} delete daemonset.apps/longhorn-flexvolume-driver
+kubectl -n ${NAMESPACE} delete daemonset.apps/longhorn-manager
+kubectl -n ${NAMESPACE} get nodes.longhorn.rancher.io -o yaml | sed "s/\- longhorn.rancher.io//g" | kubectl apply -f -
+kubectl -n ${NAMESPACE} delete nodes.longhorn.rancher.io --all
+kubectl -n ${NAMESPACE} delete settings.longhorn.rancher.io --all
 ```
 
+3. Delete Longhorn App from Rancher UI. This deletes the remaining chart components.
+
```
yasker: Add one section in the troubleshooting, tell the user how to clean up after hit the uninstall in the UI without properly cleaned up. There will always be users doing that, and they will ask every time. It's better for us to point them to some documents. They may need to manually remove finalizer etc.

proposed/longhorn/latest/README.md
```diff
@@ -9,24 +13,87 @@ Longhorn is 100% open source software. Project source code is spread across a nu
...
+**Do not attempt to immediately delete Longhorn App from Rancher UI. Longhorn propagates device mounts which can damage the host if not unmounted properly.** Follow this procedure instead.
+
+1. To prevent damage to the Kubernetes cluster, we recommend deleting all Kubernetes workloads using Longhorn volumes (PersistentVolumes, PersistentVolumeClaims, Deployments, StatefulSets, DaemonSets, etc).
+
+2. Run all of the following commands and allow them to finish.
 
-After confirming all the volumes are removed, then Longhorn can be easily uninstalled using:
 ```
-kubectl delete -f https://raw.githubusercontent.com/rancher/longhorn/master/deploy/longhorn.yaml
+NAMESPACE=longhorn-system
```
yasker: This seems complex enough to warrant a separate script at least.

PR #62, opened 2018-11-02
Before chart deletion, run job to gracefully cleanup CRD instances

LLParse:
https://github.com/rancher/longhorn/issues/316


yasker:
Helm hook only works after Rancher 2.1. Do we need to specify that? Or it's OK for the user to ignore it for <2.1 and use manual uninstall steps.

guangbochen:
this actually is a missing feature that we are thinking to implement in 2.x https://github.com/rancher/rancher/issues/15014

yasker:
@guangbochen ok... guess we have to live with it now.

yasker:
@LLParse No public endpoint was created for the chart by default?

proposed/longhorn/latest/templates/deployment-ui.yaml
```diff
@@ -14,6 +14,9 @@ spec:
     metadata:
       labels:
         app: longhorn-ui
+        {{- if eq .Values.service.ui.type "Rancher-Proxy" }}
+        kubernetes.io/cluster-service: "true"
+        {{- end }}
```
guangbochen: the `kubernetes.io/cluster-service: "true" ` should be added to the service labels instead of the deployment labels.

proposed/longhorn/latest/templates/deployment-ui.yaml
```diff
@@ -33,7 +36,11 @@ metadata:
   name: longhorn-frontend
```
guangbochen: need to rename the service name to http in order to make rancher-proxy works.

LLParse: Thank you, that did the trick!

yasker:
LGTM

proposed/longhorn/latest/README.md
```diff
@@ -23,12 +23,12 @@ Longhorn is 100% open source software. Project source code is spread across a nu
 
 1. To prevent damage to the Kubernetes cluster, we recommend deleting all Kubernetes workloads using Longhorn volumes (PersistentVolume, PersistentVolumeClaim, StorageClass, Deployment, StatefulSet, DaemonSet, etc).
 
-2. From the `Cluster` tab, click `Launch kubectl` and run this command:
+2. Rancher v2.1+ users should skip this step. From the `Cluster` tab, click `Launch kubectl` and run this command:
```
yasker: `[Rancher v2.0.x users only]`

PR #115, opened 2019-01-29
Enable extra persistence using PVC

dawidmalina:
Aim of this change is to allow additional persistent volume which will be used to other purposes than just buffer which we have currently. As an example we can store part of the log messages on this volume when backed will not be available.


dawidmalina:
@guangbochen can you please review this pull request?

dawidmalina:
@guangbochen I've rebased with latest changes can you please review and if possible merge this?

guangbochen:
LGTM.

Issue #117, opened 2019-01-30
Fluentd launch failed


kingsd041:
Launch `Flunetd ` in the catalog and select `splunk_hec` for `Fluentd Output Type`, fill in the other values correctly and click on the `Lanuch` button

Catalog UI prompt "Failed to install app fluentd-aggregator. Error: YAML parse error on fluentd-aggregator/templates/configmap.yaml: error converting YAML to JSON: yaml: line 97: did not find expected key"

![image](https://user-images.githubusercontent.com/19684717/51974103-208ae000-24ba-11e9-864b-e135b5412cd2.png)

guangbochen:
**Validation Cases**
1. installing the fluentd-aggregator from the catalog apps and config various plugins accordingly, you may reference the https://docs.fluentd.org/v1.0/articles/filter-plugin-overview to find more examples.
2. enabling cluster level logging and set the output to Fluentd, the endpoint could be clusterIP address + default por `24224`
3. ensure the log is parsed and filtered properly according to your plugin configurations, and from the `Output` service e.g., like Splunk or ElasticSearch, you should be able to view those logs.

dawidmalina:
@guangbochen I have quite similar problem and it's must be connected with the way how rancher ui inputs are handled. (maybe it rather connected with helm itself - I don't know)

For example I'm trying import such configuration (which is tested by my):
![zrzut ekranu 2019-02-1 o 18 29 34](https://user-images.githubusercontent.com/5910851/52139270-f4a16300-264f-11e9-8def-b0254619077b.png)

after upgrade I will get such error:
```
Helm template failed. Error: failed parsing --set data: error parsing index: strconv.Atoi: parsing "${thread_name}": invalid syntax : exit status 1
```

So you can say that `${thread_name}` bad syntax... but ;) when I remove this part:
![zrzut ekranu 2019-02-1 o 18 31 11](https://user-images.githubusercontent.com/5910851/52139338-27e3f200-2650-11e9-84ce-5df7b1f6702d.png)

I will get such error:
```
Helm template failed. Error: failed parsing --set data: key "time>\n @type file\n timekey_use_utc true\n path /var/log/fluentd-buffers/${tag}-json-buffer\n flush_interval 1s\n flush_mode interval\n retry_forever true\n </buffer>\n <format>\n @type sprintf\n sprintf_format \"${@timestamp} ${level} ${logger_name} : ${message}n\"\n </format>\n</match>" has no value (cannot end with ,) : exit status 1
```

[...]

PR #119, opened 2019-02-02
update fluentd to fix output plugins

guangbochen:
update fluentd to fix output plugins


guangbochen:
Issue: https://github.com/rancher/charts/issues/117

PR #141, opened 2019-03-19
update drone syntax for 1.x

drpebcak:


drpebcak:
@cloudnautique

drpebcak:
This failure is the same as in the old version of drone.

cloudnautique:
@guangbochen can we get the CI passing on this? We need to move CI environments.

guangbochen:
LGTM

PR #159, opened 2019-04-29
Add StorageOS chart

croomes:
This PR adds the StorageOS chart and catalog integration.  A default deployment installs the StorageOS Operator and then a StorageOS cluster, allowing use of StorageOS for provisioning PVs.

StorageOS has been available on Rancher for some time but not integrated into the Rancher catalog.  Existing docs at: https://docs.storageos.com/docs/platforms/rancher/, which we'll update for the catalog.

### DIFF
--- a/proposed/storageos-operator/templates/cleanup.yaml
+++ b/proposed/storageos-operator/templates/cleanup.yaml
@@ -1,0 +1,133 @@
+{{- if .Values.cluster.create }}
+
+# ClusterRole, ClusterRoleBinding and ServiceAccounts have hook-failed in
+# hook-delete-policy to make it easy to rerun the whole setup even after a
+# failure, else the rerun fails with existing resource error.
+# Hook delete policy before-hook-creation ensures any other leftover resources
+# from previous run gets deleted when run again.
+# The Job resources will not be deleted to help investigage the failure.
+# Since the resources created by the operator are not managed by the chart, each
+# of them must be individually deleted in separate jobs.
+
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: storageos-cleanup-sa
+  namespace: {{ .Values.cluster.namespace }}
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-delete-policy": "hook-succeeded, hook-failed, before-hook-creation"
+    "helm.sh/hook-weight": "1"
+
+---
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: storageos-cleanup-cr
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-delete-policy": "hook-succeeded, hook-failed, before-hook-creation"
+    "helm.sh/hook-weight": "1"
+rules:
+# Using apiGroup "apps" for daemonsets fails and the permission error indicates
+# that it's in group "extensions". Not sure if it's a Job specific behavior,
+# because the daemonsets deployed by the operator use "apps" apiGroup.
+- apiGroups:
+  - "extensions"
+  resources:
+  - "daemonsets"
+  verbs:
+  - "delete"
+- apiGroups:
+  - "apps"
+  resources:
+  - "statefulsets"
+  verbs:
+  - "delete"
+- apiGroups:
+  - rbac.authorization.k8s.io
+  resources:
+  - roles
+  - rolebindings
+  - clusterroles
+  - clusterrolebindings
+  verbs:
+  - delete
+- apiGroups:
+  - storage.k8s.io
+  resources:
+  - storageclasses
+  verbs:
+  - delete
+- apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
+  verbs:
+  - delete
+- apiGroups:
+  - ""
+  resources:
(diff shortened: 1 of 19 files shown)


guangbochen:
Thanks for the PR @croomes, the only issue that I have noticed is that if you deploy the operator app and cluster app to a different namespace(i.e, storageos-operator and kube-system respectively) while removing the app will not delete the cluster app because of multi-namespace restriction of the helm.

Reproduce steps:
1. deploy the storageos operator controller to the `storageos-operator` namespace and install StorageOS cluster
2. remove the SotrageOS in catalog app and it will return `failed: resource's namespace kube-system doesn't match the current namespace storageos-operator`

this is because Helm does not allow multi-namespace resource operation in one Helm release, so either you should set the default cluster namespace same to the operator's namespace or set the default app release namespace to `kube-system` through questions.yml(refer to https://github.com/rancher/charts/blob/dev/proposed/longhorn/latest/questions.yml#L5)

croomes:
Hi @guangbochen, thanks for the review.  We know that it is not common for charts to install in multiple namespaces, but we didn't realise that it was not allowed.

The reason for this approach is that the pods that provide storage to the rest of the cluster should not be preemptible or applications that depend on the storage will also go offline if StorageOS is evicted.  In order for us to mark the StorageOS pods as  cluster-critical they need to be in the `kube-system` namespace.

We wanted to keep the resources in `kube-system` to a minimum, so we defaulted to installing the operator in `storageos-operator` and the cluster in `kube-system`.

To solve the cleanup issue, we added post-delete hooks to delete the resources created by the operator but not managed by the chart: https://github.com/storageos/charts/pull/51/commits/46edd70378146f68b7f63b151091d2a67e492700

Given the error messages you received, there may be an issue with this.  Will you be ok with the approach if we can resolve it?

proposed/storageos-operator/templates/cleanup.yaml
```diff
@@ -0,0 +1,133 @@
...
+# from previous run gets deleted when run again.
+# The Job resources will not be deleted to help investigage the failure.
+# Since the resources created by the operator are not managed by the chart, each
+# of them must be individually deleted in separate jobs.
+
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: storageos-cleanup-sa
+  namespace: {{ .Values.cluster.namespace }}
```
guangbochen: set the namespace to {{ .Release.Namespace }}

proposed/storageos-operator/templates/cleanup.yaml
```diff
@@ -0,0 +1,133 @@
...
+metadata:
+  name: storageos-cleanup-crb
+  annotations:
+    "helm.sh/hook": post-delete
+    "helm.sh/hook-delete-policy": "hook-succeeded, hook-failed, before-hook-creation"
+    "helm.sh/hook-weight": "2"
+subjects:
+- name: storageos-cleanup-sa
+  kind: ServiceAccount
+  namespace: {{ .Values.cluster.namespace }}
```
guangbochen: set the namespace to {{ .Release.Namespace }}

proposed/storageos-operator/templates/cleanup.yaml
```diff
@@ -0,0 +1,133 @@
...
+# Iterate through the Values.cleanup list and create jobs to delete all the
+# unmanaged resources of the cluster.
+
+{{- range .Values.cleanup }}
+
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: "storageos-{{ .name }}-cleanup"
+  namespace: {{ $.Values.cluster.namespace }}
```
guangbochen: set the namespace to {{ .Release.Namespace }}

proposed/storageos-operator/templates/cleanup.yaml
```diff
@@ -0,0 +1,133 @@
...
+    "helm.sh/hook-weight": "3"
+spec:
+  template:
+    spec:
+      serviceAccountName: storageos-cleanup-sa
+      containers:
+      - name: "statefulset-{{ .name }}-cleanup"
+        image: bitnami/kubectl:1.14.1
+        command:
+          - kubectl
```
guangbochen: specify the namespace of deleted resources, so you don't need to deploy job and SA to the `kube-system` as it will cause Helm multi-namespace error.
```xml
          - -n 
          - {{ $.Values.cluster.namespace }}
```

proposed/storageos-operator/templates/cleanup.yaml
croomes: done

proposed/storageos-operator/templates/cleanup.yaml
croomes: done

proposed/storageos-operator/templates/cleanup.yaml
croomes: done

proposed/storageos-operator/templates/cleanup.yaml
croomes: done - thanks, this was a good idea that we hadn't thought of.  Tested and it works well.

croomes:
Hi @guangbochen, thanks for the suggestion, it works well.  Could you have another look please?

guangbochen:
LGTM.

Issue #161, opened 2019-05-06
Wordpress chart doesn't allow existing volume claim for wordpress persistent volume


yasker:
Wordpress: 2.1.10

guangbochen:
You can test with custom catalog using url:  https://github.com/guangbochen/charts.git and `wp` branch before it gets merged into the master, thanks.

yasker:
@guangbochen 

I've tested the chart. Seems the mariadb volume won't be deleted after uninstalling the app? Wordpress pvc would be deleted.

yasker:
@guangbochen for the latest version, seems the `mariadb-password` and `mariadb-root-password` (in secrets) changed every time when the chart was deployed, makes it impossible to reuse the mariadb PVC. Is there a way to set those password to be a fixed value? Or any other way to workaround this suee?

yasker:
@guangbochen Thanks for adding `mariadb-password` and root password as values, it works now.

Though the README about what PVC name should be used for mariadb statefulset looks literally like this now:

`creating PVC with name e.g. data-{$release.name}-{$index}`

Seems there is no replacement in the README file.

guangbochen:
yes, the README file won't be able to get the $release.name and $index of the chart, so we can either provide an example like `data-wordpress-mariadb-0` or a pattern like the above one, thanks.

yasker:
We can add `e.g. data-wordpress-mariadb-0` in the end. Remove the `e.g.` from the original sentence since that's the format must be followed, not an example.

guangbochen:
Done.

yasker:
LGTM

guangbochen:
Merged and close it.

PR #162, opened 2019-05-07
add existing volume claim of the wordpress

guangbochen:
Problem:
https://github.com/rancher/charts/issues/161

Solution: add existing volume claim of the WordPress


yasker:
LGTM.

PR #163, opened 2019-05-16
Bump to storageos-operator 1.2.0 and node 1.2.1

croomes:
We released an new version of the operator and base image yesterday.  The operator supports a new field, `csi.deploymentStrategy` which determines whether the CSI helpers are deployed by the Operator as StafefulSets (previous) or Deployments (new default).  The cleanup jobs have been amended to support both and ignore whichever wasn't created.

### DIFF
--- a/proposed/storageos-operator/templates/rbac.yaml
+++ b/proposed/storageos-operator/templates/rbac.yaml
@@ -22,6 +22,7 @@ rules:
   resources:
   - statefulsets
   - daemonsets
+  - deployments
   verbs:
   - "*"
 - apiGroups:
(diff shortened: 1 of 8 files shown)


proposed/storageos-operator/templates/rbac.yaml
```diff
@@ -99,7 +100,20 @@ rules:
...
+  - securitycontextconstraints
+  verbs:
+  - create
+  - delete
+  - update
+  - get
+  - use
+  resourceNames:
+  - privileged
+
```
croomes: Including this allows us to maintain a single upstream repo

guangbochen:
LGTM.

PR #172, opened 2019-05-27
[openebs]: remove pre-install hook from secret and pod labels

prateekpandey14:
- Updated README
 - Updated chart version
 - Updated values.yaml
 - Update pod labels for deployments

 `pre-install` hook: Executes after templates are rendered,
  but before any resources are created in Kubernetes.

  This hook has been added to the secret , will only be generated
  on chart install, to prevent overriding the certs anytime we upgrade
  the chart’s released instance.

  Thus `pre-install hook` prevents the secret creation if someone uses
  the older version of charts and upgraded to new version.

Signed-off-by: prateekpandey14 <prateek.pandey@openebs.io>


guangbochen:
since the `v0.9.0` version is only merged into the `dev` branch so we don't need to add a backup version of it. LGTM.

prateekpandey14:
@guangbochen can we merge this PR in dev  and then to master. thanks

guangbochen:
LGTM.

PR #173, opened 2019-05-29
Remove wp chart version from deployment selector to allow upgrades

guangbochen:
**Solution:**
Remove wp chart version from deployment selector to allow upgrades

**Related Issue:**
https://github.com/rancher/rancher/issues/20523


guangbochen:
will not bump the chart version as it means to fix the existing chart of deployment upgrade.

charts/wordpress/v2.1.10/templates/deployment.yaml
```diff
@@ -13,7 +13,6 @@ spec:
     metadata:
       labels:
         app: {{ template "fullname" . }}
-        chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
```
alena1108: Can we alter the existing chart? Or do we have to create a new version/folder. Assuming some people can already use it.

guangbochen:
close in favor of https://github.com/rancher/charts/pull/174

Issue #179, opened 2019-06-06
cert-manager upgrade to version 8


daxmc99:
I was able to install cert-manager version 8 with these steps. 

1.     Add https://charts.jetstack.io to global catalogs
2.     Enter into the system project
3.     Install cert-manager v8 into the 'kube-system' namespace.

We should see about creating a questions.yaml file for this chart to allow for installation into 'kube-system' more easily.

tlitovsk:
Hi.

Let's encrypt plans to block all cert managers less than version 8 starting November 1.
Can we please have an updated chart in the curated library.

Best Regards

cesarandreslopez:
This chart needs an upgrade soon, please. From a Let's encrypt email:

"we plan to start blocking all traffic from cert-manager versions less than 0.8.0 (the 
current semver minor release), as of November 1, 2019. Please upgrade 
all of your cert-manager instances before then."

guangbochen:
Yes, we have planned to upgrade those curated charts to align with Rancher 2.3 releases(mid-sep.),  will add a reply to this question once it is released, thanks.

dramich:
The fix for this needs to ensure it's getting deployed properly when moving to version 9. Per the [upgrade docs](https://docs.cert-manager.io/en/latest/tasks/upgrading/upgrading-0.8-0.9.html) the correct way to do that is to delete the current deployment and re-deploy. QA work for this would need to validate this step is being done.

amagura:
@guangbochen The Rancher 2.3 release came and went--do you have any update on when this chart will be updated to version 0.8.0 or 0.9.0?  November 1st is little more than a week away!

dramich:
@amagura see the [forum post](https://forums.rancher.com/t/update-on-cert-manager-application-in-the-catalog/15598)

amagura:
@dramich This doesn't work for my setup.  We're not using helm and are entirely dependent on Rancher's app catalog.  Also I took some time and looked at the code for cert-manager; there's a branch that has a `v0.9.1` directory in it--is that just empty then?  Are y'all not going to add support for installing this via Rancher's app catalog ever?  @guangbochen

dramich:
@amagura from the forum post:

> Jetstack is rapidly approaching a 1.0 release of cert-manager, after which the pace of changes will hopefully subside. When we’re able to ensure that your experience with cert-manager in the App Catalog is easy and painless, we’ll look into bringing it back into the App Catalog.

PR #206, opened 2019-08-15
Bump cert-manager to v0.9.0

guangbochen:
**Issue:**
1. Let's encrypt plans to block all cert managers less than version 8 starting November 1.
2. auto-create cert-manager `CRDs` if not exist

**Solution:**
1. bump cert-manager to v0.9.0
2. add `crds-hooks` to cert-manager chart

**Related Issue:**
https://github.com/rancher/charts/issues/179


charts/cert-manager/v0.9.0/templates/clusterissuer.yaml
```diff
@@ -0,0 +1,88 @@
...
+        server: "https://acme-v02.api.letsencrypt.org/directory"
+        {{- else if eq .Values.ingressShim.defaultIssuerName "letsencrypt-staging" }}
+        server: "https://acme-staging-v02.api.letsencrypt.org/directory"
+        {{- else }}
+        server: {{ .Values.letsencrypt.server }}
+        {{- end }}
+        email: {{ .Values.letsencrypt.email }}
+        privateKeySecretRef:
+          name: {{ .Values.ingressShim.defaultIssuerName }}-account-key
+        http01: {}
```
oskapt: I mentioned this in [the docs PR](https://github.com/rancher/docs/pull/1691#issuecomment-526662805), and it's even more relevant here.

Basically, there's a [new solver format in 0.8](https://docs.cert-manager.io/en/latest/tasks/upgrading/upgrading-0.7-0.8.html) that deprecates the old way of doing it. The old way still works, but it'll be removed before 1.0, and they're about to roll out with 1.0-alpha. Users who don't upgrade the ClusterIssuer will be surprised.

It's a bit chicken/egg. `ingress-shim` certs will only use the new solver if the old config is removed. Prematurely removing the old config can break certs not issued by `ingress-shim`. So perhaps we can check if they have non-`ingress-shim` certs, and if not, make the update to the ClusterIssuer for them? 

What happens if someone installs our app, manually updates the ClusterIssuer, and then upgrades from the app to a newer version? Will their solver update be overwritten?

There was also [an issue with the upgrade to 0.9](https://github.com/jetstack/cert-manager/issues/1927) that required `--force` to recreate the deployments because they couldn't be patched with new ClusterRoles and ClusterRoleBindings being installed at the same time. Is our app immune to that?

Finally, in my work to create [the video for this](https://youtu.be/nF_k1YZYjH0), I found that certs generated by `ingress-shim` wouldn't update until the secret was removed. Then they updated and issued a new secret. They _might_ update automatically at the next renewal.

oskapt:
They're up to `0.10` now, so whenever this gets merged, we should update it to use the latest stable release.

guangbochen:
Thanks, will look into the new version and `ingress-shim` case u have mentioned above.

dramich:
@guangbochen Does this delete the current deployment before proceeding? As @oskapt mentioned there is a requirement to do so in order to get the upgrade to version 9 as seen in the docs [here](https://docs.cert-manager.io/en/latest/tasks/upgrading/upgrading-0.8-0.9.html).

dramich:
Another note, the currently released version is 0.9.1.

guangbochen:
The helm upgrade will delete the old version and roll out the new deployment version with v0.9.1 @dramich 
image
Updated images to v0.9.1.

jaredkipe:
Would love to get this in Rancher 2.2.x please (it appears to me to be 2.3 target)

guangbochen:
@jaredkipe this upgrade will be available in 2.2.x release, the 2.3 is just a milestone target, thanks

> Would love to get this in Rancher 2.2.x please (it appears to me to be 2.3 target)

drpebcak:
@guangbochen what is the status of this? It would be good to get this merged and available soon.

guangbochen:
merged into dev and will sync to the master once QA validates it, thanks.

guangbochen:
@deniseschannon chart 2.3 upgrade.

Moumouls:
Hi @guangbochen , first of all thanks for this PR !
I have a quick question here, when do you think the new version of `cert-manager' will be available on `master'?

guangbochen:
@Moumouls thanks for using the catalog apps, in regards to the `cert-manger` upgrade case please refer to the discussion in  https://github.com/rancher/charts/issues/179 and solution at https://forums.rancher.com/t/update-on-cert-manager-application-in-the-catalog/15598, thanks.

PR #220, opened 2019-08-30
Bump datadog chart to v1.33.0

guangbochen:
Bump datadog chart to v1.33.0 with k8s 1.16 compatibility.


deniseschannon:
@guangbochen this one needs to be rebased

PR #224, opened 2019-09-02
Bump prometheus chart to v9.1.0

guangbochen:
Bump Prometheus chart to v9.1.0 with k8s v1.16 capabiity.


deniseschannon:
@guangbochen this one needs to be rebased

PR #235, opened 2019-09-03
Add KubeMQ message broker

Ido-Sheffer:
added kubemq\latest message broker to the dev branch.


charts/kubemq/latest/Chart.yaml
```diff
@@ -0,0 +1,16 @@
...
+- event
+- queue
+home: https://kubemq.io/
+version: 0.1.0
+sources:
+- https://github.com/kubemq-io/docs
+maintainers:
+- name: ido-sheffer
+  email: ido.sheffer@kubemq.io
+icon: https://account.kubemq.io/assets/img/logo1.png
```
guangbochen: would prefer to add the icon to the parent directory and refer it by: `icon: file://../logo.png`, so that air-gapped user can also view it correctly, thanks.

Ido-Sheffer: hello update a new pull request "Added KubeMQ message broker #256" https://github.com/rancher/charts/pull/256

PR #237, opened 2019-09-05
Update storageos-operator to version 0.2.12

darkowlzz:


croomes:
LGTM (StorageOS approval)

proposed/storageos-operator/v0.2.12/templates/psp.yaml
```diff
@@ -0,0 +1,29 @@
+{{- if .Values.podSecurityPolicy.enabled }}
+
+apiVersion: extensions/v1beta1
```
guangbochen: can we use `policy/v1beta1` instead for k8s v1.16 support.

darkowlzz: sure.

PR #261, opened 2019-10-02
Add instana-agent v1.0.16

dlbock:
This adds the Instana agent helm chart v1.0.16 adapted for Rancher


charts/instana-agent/v1.0.15/templates/podsecuritypolicy.yaml
```diff
@@ -0,0 +1,52 @@
+{{- if .Values.rbac.create -}}
+{{- if (and .Values.podSecurityPolicy.enable (not .Values.podSecurityPolicy.name)) -}}
+kind: PodSecurityPolicy
+apiVersion: extensions/v1beta1
```
guangbochen: can we use `policy/v1beta1` instead, thanks

charts/instana-agent/v1.0.15/Chart.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v1
+name: instana-agent
+version: 1.0.15
+appVersion: 1.0
+description: Instana Agent for Kubernetes
+home: https://www.instana.com/
+icon: http://instana-management-assets.s3-website-eu-west-1.amazonaws.com/stan_icon_front_black_big.png
```
guangbochen: would prefer to add the logo to the parent directory and use it via `icon: file://../logo.png` for air-gapped users.

dlbock:
@guangbochen Thanks for the review. Since some of the changes you've requested will also have to be made on our helm chart in https://github.com/helm/charts/tree/master/stable/instana-agent, I will have to bump this version to v1.0.16, just a heads up.

guangbochen:
Thanks, LGTM and merged.

dlbock:
@guangbochen thanks for the review & merge! When should I expect this chart to appear in the Rancher default catalog?

guangbochen:
We have the sync PR in https://github.com/rancher/charts/pull/246, and after doing a few testing and verification it will get merged, maybe in couple of days, thanks.

dlbock:
awesome thank you!

PR #264, opened 2019-10-09
Add max rancher version label to Istio charts

guangbochen:
Add max rancher version label to Istio charts for not showing from 2.3.0 onwards.
 

**Related Issue:**
https://github.com/rancher/rancher/issues/23168#issuecomment-539819542


loganhz:
lgtm, build failed due to we have to change the existing charts

deniseschannon:
@loganhz what does that mean? where are changes that need to happen?

guangbochen:
@deniseschannon need to add `max_rancher_version` in `questions.yaml` to those existing istio charts, it will cause CI failure but I think we can ignore in this case.

PR #270, opened 2019-10-12
Add min and max version filter to cert-manger

guangbochen:
Add min and max version filter to cert-manger


charts/cert-manager/v0.9.1/questions.yml
```diff
@@ -1,4 +1,5 @@
 namespace: kube-system
+rancher_min_version: 2.3.0-rc1
```
deniseschannon: This is unnecessary as min/max was only added as of v2.3.0

guangbochen:
Done

PR #290, opened 2019-10-17
Update magent v5.1.4 depdency files

guangbochen:
Update magent v5.1.4 depdency files with k8s 1.16 capability support

**Related Issue:**
https://github.com/rancher/rancher/issues/23459


deniseschannon:
@guangbochen this needs to be rebased

PR #304, opened 2019-10-22
Added Kubemq message broker

Ido-Sheffer:


deniseschannon:
We only need this PR in the `partners` branch for testing and then it'll be brought over to `master`. The `master` branch is what is shipped with Rancher.

PR #313, opened 2019-10-31
Add helm chart for NeuVector 2.5.x release support

becitsthere:


guangbochen:
@becitsthere can u please help to split the commit into 2, thanks.
1. base copy NeuVector `v1.2.3` to `v1.3.4`.
2. add update commit.

selvamt94:
@guangbochen I am an engineer from Neuvector, can you please give more detail about how to split the commit. We want to commit the newer version of our helm chart instead of this version. Shall we cancel this commit and make new commit instead. Thanks, Selvam

PR #319, opened 2019-11-07
Add partner labels and fix neuvector logo

guangbochen:
**Problem:**
Add missing partner label to `redskyops` and `traefik`, add neuvector logo using local path since the current URL is not showable.


guangbochen:
@deniseschannon Can u PTAL, thanks.

PR #324, opened 2019-11-11
Add CockroachDB

keith-mcclellan:
Replicates latest Helm chart and adds app-readme.md and questions.yml with initial support for Rancher advanced config

### DIFF
--- a/charts/cockroachdb/19.1.5/app-readme.md
+++ b/charts/cockroachdb/19.1.5/app-readme.md
@@ -1,0 +1,11 @@
+# CockroachDB Chart
+
+CockroachDB is a Distributed SQL database that runs natively in Kubernetes. It gives you resilient, horizontal scale across multiple clouds with always-on availability and data partitioned by location.
+
+CockroachDB scales horizontally without reconfiguration or need for a massive architectural overhaul. Simply add a new node to the cluster and CockroachDB takes care of the underlying complexity.
+
+    - Scale by simply adding new nodes to a CockroachDB cluster
+    - Automate balancing and distribution of ranges, not shards
+    - Optimize server utilization evenly across all nodes
+
+
--- a/charts/cockroachdb/19.1.5/questions.yml
+++ b/charts/cockroachdb/19.1.5/questions.yml
@@ -1,0 +1,12 @@
+labels: 
+  - io.rancher.certified: partner
+categories:
+  - Config
+questions: 
+    default: 100Gi
+    description: "Size of volume for each CockroachDB Node/Pod"
+    group: Config
+    label: "Storage per Node/Pod"
+    required: true
+    type: string
+    variable: Storage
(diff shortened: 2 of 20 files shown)


charts/cockroachdb/19.1.5/questions.yml
```diff
@@ -0,0 +1,10 @@
+categories:
+- Config
```
guangbochen: can u please help to add partner label as below:
```
labels:
  io.rancher.certified: partner
```

charts/cockroachdb/19.1.5/questions.yml
```diff
@@ -0,0 +1,10 @@
...
+questions:
+- variable: Storage
+default:  "100Gi"
```
guangbochen: The tabs indentation of the following variables are incorrect, can u please help to update it, thanks.
image

charts/cockroachdb/19.1.5/app-readme.md
```diff
@@ -0,0 +1,381 @@
+# CockroachDB Helm Chart
```
guangbochen: since `app-readme.md` is a file that provides descriptive text in the chart’s UI header.  so we don't need a full doc copy of README.

charts/cockroachdb/19.1.5/questions.yml
guangbochen: It seems storage is required of the CockroachDB app, so `StorageClass` should be a required setting.
you can refer https://github.com/rancher/charts/blob/master/charts/docker-registry/v1.8.1/questions.yml#L141 as an example.

keith-mcclellan: Technically, if a customer wants to manually provision PVs they do not need to set a storageclass, which is why we don't set it as required.

keith-mcclellan: I did however reformat the yaml file

charts/cockroachdb/19.1.5/questions.yml
```diff
@@ -0,0 +1,15 @@
...
+  - Config
+default: 100Gi
+description: "Size of volume for each CockroachDB Node/Pod"
+group: Config
+label: "Storage per Node/Pod"
+labels: 
+  - 
+    io.rancher.certified: partner
+questions: 
+  -
```
guangbochen: these formats are incorrect, can u please refer to the existing chart as an example to fix them, thanks.

keith-mcclellan: hopefully this latest format change is what you're looking for

keith-mcclellan:
squashed my changes into a single commit for cleanliness upstream

guangbochen:
Fixed `questions.yml` format error with https://github.com/rancher/charts/pull/324/commits/d74a391dd075791f38e1659e3cfae7d80498590f

PR #331, opened 2019-11-18
Added KubeMQ message broker

Ido-Sheffer:


guangbochen:
close as the same PR is already merged.

PR #342, opened 2019-11-29
Updated kubeflow chart to v0.6.2

rawmind0:
Fix issue https://github.com/rancher/rancher/issues/22636


proposed/kubeflow/v0.6.2/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+appVersion: "1.0"
+description: Machine Learning Toolkit for Kubernetes
+name: kubeflow
+version: 0.6.2
+icon: file://../icon.jpg
+maintainers:
+- name: guangbochen
```
guangbochen: I think u can also add yourself as a maintainer, thanks.

proposed/kubeflow/v0.6.2/questions.yml
```diff
@@ -0,0 +1,158 @@
+categories:
+- machine learning
+rancher_max_version: 2.3.3
+labels:
+  io.rancher.certified: experimental
+namespace: kubeflow
+questions:
+- variable: profiles.admin
+  default: johnDoe@acme.com
```
guangbochen: is the admin required, if not would prefer to set the default value an empty string.

proposed/kubeflow/v0.6.2/questions.yml
```diff
@@ -0,0 +1,158 @@
...
+  description: "Argo artifact repository access key secret"
+  type: string
+  label: Argo artifact repository access key secret
+- variable: argo.artifactRepositorySecretKeySecretName
+  default: "mlpipeline-minio-artifact"
+  description: "Argo artifact repository secret key name"
+  type: string
+  label: Argo artifact repository secret key name
+- variable: argo.artifactRepositorySecretKeySecretKey
+  default: "secretkey"
```
guangbochen: are those `secretkey` and `acceskey` exist? if not and when those variables are not required would prefer to leave it as an empty string, also the `questions.yml` support variable type `secret`.

proposed/kubeflow/v0.6.2/Chart.yaml
rawmind0: Ok, updated.

proposed/kubeflow/v0.6.2/questions.yml
rawmind0: Yes, it's required. Updated setting required true.

proposed/kubeflow/v0.6.2/questions.yml
rawmind0: Ok, updated to type `password`

rawmind0:
Thanks for the review @guangbochen .

Have you tried to deploy in k8s 1.16?? As specified in README.md, kubeflow is just compatible with k8s 1.14 and 1.15, https://github.com/rancher/charts/pull/342/files#diff-c88cf87969a411a10e36bc77e1aee95eR8 . Could you please try to deploy in one of these versions?? https://www.kubeflow.org/docs/started/k8s/overview/#minimum-system-requirements

guangbochen:
LGTM, thanks

PR #347, opened 2019-12-17
Add delay during postStart if directory is not yet created by flexvolume container

shivamerla:
During post-start hook, we expect that flexvolume exec directory is already created, but it might not be if flexvolume container is not yet initialized. Hence add a delay loop to wait until the directory is created.
Signed-off-by: Shiva Krishna, Merla <shivakrishna.merla@hpe.com>


shivamerla:
@guangbochen this was a minor fix in post-start hook to workaround a timing issue. Can you please check if version bump is really necessary?

guangbochen:
since this fix contains changes of the main templates so a version bump is needed, otherwise existing users won't get an updated version of this fix, thanks.

shivamerla:
closing this as another PR #356 was opened with version bump.

PR #373, opened 2020-02-12
Chart version 1.0.22

dlbock:
Changes included in this PR:
- Allow users to specify any environment variable for the agent. This is to avoid having to create a new helm value whenever we need to set an environment variable in the agent, which is hard to maintain, not to mention confusing for our users.
- Increase liveness probe timeouts as we have had quite a few instances of overloaded agents getting killed on start up because the liveness probe is not generous enough. Therefore do the following for both the `instana-agent` and `leader-elector` containers:
  - Increase initial delay to 5 minutes
  - Increase timeout for each probe to 3 seconds
  - Increase period to once every 10 seconds (the default)


dlbock:
cc @guangbochen

guangbochen:
LGTM

PR #376, opened 2020-02-18
Revert "px release 1.0.7"

trierra:
Reverts rancher/charts#375

adding new configurations has lead to the error:
```
Failed to install app portworx. Error: release portworx failed: resource's namespace portworx doesn't match the current namespace kube-system
```


harsh-px:
@deniseschannon / @guangbochen looks like our previous merge 1.0.7  broke the chart and installs are failing with the message in the description.

This is because the 1.0.7 change added a new namespace called `portworx` in the templates itself and helm doesn't like.

Can you help merge this revert?

harsh-px:
No longer needed as we have hidden it

harsh-px:
This PR can be closed.

PR #407, opened 2020-03-19
Update mariadb chart to v7.3.14

PennyScissors:


charts/mariadb/v7.3.14/values.yaml
```diff
@@ -18,7 +18,7 @@
 ##
 image:
   registry: docker.io
-  repository: bitnami/mariadb
```
deniseschannon: Curious at why we have one commit that moves them to `ranchercharts/bitnami-` and then another commit that reverts it back to the `bitnami` repo?

PennyScissors: I changed it to use `ranchercharts/bitnami-` and it got stuck at deploying.
I then reverted and added in the commit message that we need to update the docker hub images since the version mismatch is probably the cause.

charts/mariadb/v7.3.14/app-readme.md
```diff
@@ -5,5 +5,5 @@
 MariaDB is developed as open source software and as a relational database it provides an SQL interface for accessing data. The latest versions of MariaDB also include GIS and JSON features.
 
 ## Notable changes
-### v6.7.4
+### v7.3.14
 **Warning:** Upgrade from previous version is currently not supported.
```
deniseschannon: If upgrade works, then this line should be removed. 

Is there anything else that was found in the release notes of 7.3.14?

charts/mariadb/v7.3.14/app-readme.md
```diff
@@ -0,0 +1,9 @@
+# MariaDB
+
+[MariaDB](https://mariadb.org) is one of the most popular database servers in the world. It’s made by the original developers of MySQL and guaranteed to stay open source. Notable users include Wikipedia, Facebook and Google.
+
+MariaDB is developed as open source software and as a relational database it provides an SQL interface for accessing data. The latest versions of MariaDB also include GIS and JSON features.
+
+## Notable changes
```
deniseschannon: Let's also update this to `## Changelog`

charts/mariadb/v7.3.14/questions.yml
```diff
@@ -23,7 +23,7 @@ questions:
     type: string
     label: MariaDB Image Name
   - variable: image.tag
-    default: "10.3.17-debian-9-r0"
+    default: "10.3.22-debian-10-r44"
```
guangbochen, line 26: you also need to update the repository to `bitnami/mariadb` if u want to use the origin repo.

charts/mariadb/v7.3.14/questions.yml
```diff
@@ -34,7 +34,7 @@ questions:
     label: Metrics Image Name
     show_if: "metrics.enabled=true"
   - variable: metrics.image.tag
-    default: "0.12.1-debian-9-r2"
+    default: "0.12.1-debian-10-r42"
```
guangbochen, line 37: same as above, need to update the repo

charts/mariadb/v7.3.14/questions.yml
PennyScissors: Updated

PR #409, opened 2020-03-19
Update wordpress chart to v9.0.3

PennyScissors:


charts/wordpress/v9.0.3/templates/secrets.yaml
```diff
@@ -2,15 +2,11 @@ apiVersion: v1
...
-  labels:
-    app: "{{ template "wordpress.fullname" . }}"
-    chart: "{{ template "wordpress.chart" . }}"
-    release: {{ .Release.Name | quote }}
-    heritage: {{ .Release.Service | quote }}
+  labels: {{- include "wordpress.labels" . | nindent 4 }}
 type: Opaque
 data:
   {{- if .Values.wordpressPassword }}
-  wordpress-password: {{ default "" .Values.wordpressPassword | b64enc | quote }}
```
deniseschannon: What implications does removing the `default ""` have in the template?

charts/wordpress/v9.0.3/values.yaml
```diff
@@ -13,8 +13,8 @@
 ##
 image:
   registry: docker.io
-  repository: ranchercharts/bitnami-wordpress
-  tag: 5.2.3-debian-9-r5
+  repository: bitnami/wordpress
```
deniseschannon: @loganhz @guangbochen Do you know why we put the images of this chart into `ranchercharts` in Dockerhub?

loganhz: We were trying to use the dockerhub download numbers to see how many times they are deployed from Rancher catalogs

guangbochen: @deniseschannon yes, we have mirrored the images to the `ranchercharts` so we can understand the popularity of the catalog app based on the image downloads.
here is the repo that I'm using for the image mirroring. https://github.com/guangbochen/charts-image-mirros
image

guangbochen: we have added this on k8s v1.17 migration, but we can either keep doing image-mirrors to track the app downloads or change it back to use the official library.

charts/wordpress/v9.0.3/questions.yml
```diff
@@ -19,7 +19,7 @@ questions:
     type: string
     label: WordPress Image Name
   - variable: image.tag
-    default: "5.2.3-debian-9-r5"
+    default: "5.3.2-debian-10-r43"
```
guangbochen, line 22: update the repo name to `bitnami/wordpress`

PennyScissors: This was updated in the last commit>

PR #418, opened 2020-04-05
Uploaded Alcide Chart files

alonalcide:


charts/Alcide/v1/Chart.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+name: alcide-advisor-cronjob
+description: Alcide Advisor CronJob Helm Chart
+keywords:
+  - security
+  - vulnerability scan
+  - guardrails
+type: application
+version: 2.0.1
```
PennyScissors, line 9: Please rename the parent folder to v2.0.1/
It is named after the version of the chart stated here.

Also, a "home" key is required for the chart test to pass (currently failing).

PR #425, opened 2020-04-13
Team support for FAH Client

wjimenez5271:


charts/folding-at-home/v0.1.0/templates/deployment.yaml
```diff
@@ -0,0 +1,151 @@
...
+    <config>
+      <!--
+        To set your user name, team and passkey just edit the text
+        in quotes below.
+      -->
+
+      <!-- User Information -->
+      <user value=""/> <!-- Enter your user name here -->
+      <team value=""/>         <!-- Your team number -->
+      <passkey value=""/>       <!-- 32 hexadecimal characters if provided -->
```
davidnuzik, line 126: Nit: passkey is optional. Think perhaps should tweak the comment to make this more clear?
Do we need to explain what the passkey is/does? People shouldn't need it unless when they created their (also optional) user they opted into making one.
IIRC the passkey is essentially something you can opt into creating for your user so others can't claim your work (not sure how that could happen though).

wjimenez5271: good point, i'll add that.

charts/folding-at-home/v0.2.0/questions.yaml
```diff
@@ -0,0 +1,59 @@
...
+    default: ""
+    description: "Folding at Home Teamname"
+    type: string
+    label: FAH Teamname (optional)
+    group: "Folding @ Home Team Info"
+  - variable: fah.passkey
+    default: ""
+    description: "Folding at Home Passkey"
+    type: string
+    label: FAH Passkey if you set one. FAH accounts do not require passkeys, this is optional
```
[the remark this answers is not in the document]
davidnuzik: Fine by me. Thanks!

davidnuzik:
> Also FYI you were commenting in the 0.1 version of the chart, 0.2 has the variables supported

Ah my bad. Curious, why do I see so many files changed?

wjimenez5271:
@deniseschannon ready for your review

PR #439, opened 2020-04-22
Move longhorn from proposed to charts

PennyScissors:
- Move all longhorn charts from proposed/ to charts/
- Remove experimental tag from all charts
- Add rancher max version 2.3.1 to charts older than v0.7.0
- Test chart installed from proposed/ can be upgraded
to chart in charts/ successfully.


PennyScissors:
@yasker @deniseschannon Not sure if one of you should review this or someone else.

deniseschannon:
Can this be broken into more commits?

Commit 1 - Moving the charts
Commit 2 - Changes (like removing the tag and adding max version)

PennyScissors:
> Can this be broken into more commits?
> 
> Commit 1 - Moving the charts
> Commit 2 - Changes (like removing the tag and adding max version)

@deniseschannon Sorry about that, lesson learned. It should be good now.

PR #443, opened 2020-05-03
Add openfaas chart

PennyScissors:


charts/openfaas/v5.7.2/Chart.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
+description: Enable Kubernetes as a backend for OpenFaaS (Functions as a Service)
+name: openfaas
+version: 5.7.2
+sources:
+- https://github.com/openfaas/faas
+- https://github.com/openfaas/faas-netes
+home: https://www.openfaas.com
+icon: https://raw.githubusercontent.com/openfaas/media/master/OpenFaaS_logo_stacked_opaque.png
```
PennyScissors, line 10: @guangbochen Is a link to an icon like this okay, or should I copy them into the chart folder so we can reference them locally?

deniseschannon: Reference them locally @PennyScissors as an air gapped setup wouldn't be able to grab that image.

charts/openfaas/v5.7.2/values.yaml
```diff
@@ -0,0 +1,197 @@
+functionNamespace: ""  # Multiple namespaces is not supported in Rancher
```
guangbochen, line 1: I think the fuctionNS is not for multi-apps deployment but a namespace declaration of openfass functionless apps, so it should be compatible with Rancher's catalog?

PennyScissors: I don't think it is for multi-app deployment either, but when I tried deploying it with a separate namespace for functions I got the following error:
`Failed to install app openfaas. Error: release openfaas failed: resource's namespace openfaas-fn doesn't match the current namespace openfaas`
It seems Rancher is not creating the "openfaas-fn" namespace. Unless Rancher expects me to configure something that I missed?

charts/openfaas/v5.7.2/Chart.yaml
PennyScissors: Done

charts/openfaas/v5.7.2/values.yaml
guangbochen: I see, thanks for the update.

PR #446, opened 2020-05-07
Add kubecost chart

PennyScissors:


charts/cost-analyzer/v1.57.2/Chart.yaml
```diff
@@ -4,3 +4,10 @@ description: A Helm chart that sets up Kubecost, Prometheus, and Grafana to moni
   cloud costs.
 name: cost-analyzer
 version: 1.57.2
+icon: file://../logo.png
+home: http://kubecost.com/
+sources:
+- https://github.com/kubecost/cost-analyzer-helm-chart
+maintainers:
+- name: PennyScissors
```
deniseschannon, line 12: Can you put your actual name?

PennyScissors: I tried before and it didn't work. It kept saying it can't find me. I'll try something else to see if it works.

PR #463, opened 2020-06-01
add orion vgpu chart

wpfnihao:
Add Orion VGPU chart to rancher charts.


wpfnihao:
submitted to the wrong catalog.

PR #464, opened 2020-06-04
Add Sysdig Agent chart v1.7.20

airadier:
Big bump on the chart version and Sysdig Agent version to sync with latest official release.

Added four categories in the questions.yml:

- Monitoring
- Security
- Compliance
- Vulnerability Scan

Signed-off-by: Alvaro Iradier <alvaro.iradier@sysdig.com>

### DIFF
--- a/charts/sysdig/v1.7.20/CHANGELOG.md
+++ b/charts/sysdig/v1.7.20/CHANGELOG.md
@@ -1,0 +1,393 @@
+# Chart: Sysdig
+
+## Change Log
+
+This file documents all notable changes to Sysdig Helm Chart. The release
+numbering uses [semantic versioning](http://semver.org).
+
+## v1.17.20
+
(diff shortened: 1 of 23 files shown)


airadier:
I noticed that the chart testing is failing because the linter reports this assignment:

```
{{- $repositoryName = printf "%s-%s" .Values.image.repository "slim" -}}
```

but that assignment is [valid since go 1.11](https://golang.org/doc/go1.11#text/template) and supported in slightly newer versions of the chart-testing tool (currently the CI seems to be using ranchercharts/chart-testing:v2.0.2-rancher3).

Please let us know if a newer chart-testing version could be used. Otherwise we will change the way that variable is handled.

Thanks.

airadier:
Replaced with https://github.com/rancher/charts/pull/468

PR #480, opened 2020-06-30
Add Portworx-Essentials offerring to the rancher catalog

sharma-tapas:
This chart adds Portworx-Essentials offerring to the Rancher catalog. With this the user can now install the free version of Portworx.


charts/portworx-essentials/v1.0.0/Chart.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - amd64
+  - portworx essentials
+  - free
+home: https://portworx.com/
+maintainers:
+  - name: sharma-tapas
+    email: tapas@portworx.com
+sources:
+  - https://github.com/portworx/helm
+icon: https://raw.githubusercontent.com/portworx/px-logos/master/Portworx_Product_Logos/Portworx_Essentials/essentials_full_color_logo/png/portworx-essentials-horizontal-color-500px.png
```
PennyScissors, line 22: Please keep a copy of the logo local in the `portworx-essentials` parent folder and reference it here as `file://../logo.png`. This will allow users running Rancher in an airtight environment to see it.

charts/portworx-essentials/v1.0.0/questions.yml
```diff
@@ -0,0 +1,856 @@
...
+################################### Essentials options ################################
+- variable: essentialID
+  type: string
+  required: true
+  default: "none"
+  label: "Essentials Entitlement ID"
+  description: "Get your free essentials entitlement ID from https://central.portworx.com/profile"
+  group: "license parameters"
+
+################################### KVDB options ################################
```
PennyScissors, line 17: This should be removed since it shows up as a dropdown with a single option, and it's already being set in `values.yml`.

charts/portworx-essentials/v1.0.0/README.md
```diff
@@ -0,0 +1,36 @@
+# Portworx
```
PennyScissors, line 1: As a completely new user of Portworx, I had some trouble figuring out the bare minimum needed to get the chart installed. It would be nice to include here or in app-readme (if brief), some instructions to at least get a basic setup installed. Some of the things I had to figure out were:
1. This chart needs to be installed in the system part of your cluster since the chart is locked to the "kube-system" namespace and default has no access to that namespace.
2. My understanding is you can have a variety of different configurations for Portworx, but the bare minimum that will get you a full green everything active install is having a cluster with 3 nodes, and at least 1 volume in one of them. It would be helpful to document this somewhere since as a new user, even though Portworx can work with less, I would still wonder if is working correctly and spend time figuring out why Stork is not installing everything correctly.
3. The default username/password for the UI should be here.

charts/portworx-essentials/v1.0.0/Chart.yaml
sharma-tapas: Added the local copy of the logo.

charts/portworx-essentials/v1.0.0/questions.yml
sharma-tapas: Removed this option completely.

charts/portworx-essentials/v1.0.0/README.md
sharma-tapas: Added more details to the README.md

sharma-tapas:
@PennyScissors Please review the updates.

PennyScissors:
lgtm

harsh-px:
Thanks for the re-review @PennyScissors . Can we get this merged now?

PR #485, opened 2020-07-08
splicemachine: Add SpliceMachine Database splice-helm v0.1.5

splicemaahs:


charts/splice-helm/v0.1.5/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+version: 0.1.5
+appVersion: "3.0"
+name: splice-helm
+description: A Helm chart for launching Splice Machine on Kubernetes
+home: https://www.splicemachine.com
+icon: https://splicemachine.github.io/sm.logo.800x200px_180509.svg
```
guangbochen, line 7: Please make a local copy of the logo in the parent folder `splice-helm/` and reference the image here as file://../splice-logo.png. This way the logo can be displayed in airtight environments.

charts/splice-helm/v0.1.5/questions.yml
```diff
@@ -0,0 +1,85 @@
+labels:
+  io.cattle.role: project # options are cluster/project
```
guangbochen, line 2: can u please add the partner label of `io.rancher.certified: partner`

splicemaahs:
I will take care of these.

splicemaahs:
I streamlined the base install, hopefully this will ensure the install test will work.

charts/splice-helm/v0.1.7/Chart.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+appVersion: "3.0"
+description: Splice Machine data platform is a scale-out SQL RDBMS, data warehouse
+  and machine learning management solution in one.
+home: https://www.splicemachine.com
+icon: file://../splice-logo.jpg
```
guangbochen, line 6: should be .png?

splicemaahs: It most certainly should have been, thanks.

guangbochen (approved):
LGTM

splicemaahs:
> LGTM



> @splicemaahs you can use `.tgz` file for all dependency charts, and this will speed-up user git clone and downloading with catalog apps, thanks.
> please refer to https://github.com/rancher/charts/tree/master/charts/wordpress/v9.0.3/charts as an example, thanks.

It turns out our build process was using helm package and we were extracting that for the Rancher helm build, apparently that process extracts the tgz files before building the top level tgz.

guangbochen:
you can run `helm package` on each sub-charts and only keep the `.tgz` file of sub-charts when submitting to the rancher's catalog, it should be working as expected, this is nice to have as both of them works.

splicemaahs:
> you can run `helm package` on each sub-charts and only keep the `.tgz` file of sub-charts when submitting to the rancher's catalog, it should be working as expected, and this is nice to have as both of them works.

I ran a helm dependency build at the top level chart, which created the .tgz files for the sub-charts.  I ran the `helm package` on a few of the sub-charts and compared the resulting .tgz files with the ones created by `helm dependency build` on the parent chart and the resulting .tgz file contents are the same.

guangbochen (approved):
LGTM, thanks

PR #487, opened 2020-07-10
Add externalIP option to traefik

wabscale:
I was having issues with using the rancher UI to deploy the traefik chart on custom infrastructure. There was no way to specify an external IP in the rancher UI. There was a way to specify an external IP in the values.yaml file, but it broke the `spec.serviceType`. I added some conditions to make the `spec.serviceType` to be handled correctly when there is an externalIP specified.

I've added the option to specify an externalIP in the traefik questions.yml. Would this merit a patch version bump? 1.0.0 -> 1.0.1?


PennyScissors:
Closing this as we have migrated to use [rancher/partner-charts](https://github.com/rancher/partner-charts) for all partner charts moving forward.

PR #497, opened 2020-07-16
Chart version 1.0.29

dlbock:
#### What this PR does / why we need it:

Expose new configuration options:
- `leaderElector.image.name`
- `leaderElector.image.tag`

Changed the existing `agent.leaderElectorPort` configuration value to `leaderElector.port` to be consistent with the other leader elector configurations.


dlbock:
cc @PennyScissors @deniseschannon

PR #500, opened 2020-07-21
Deprecate aws-event-sources chart v0.1.0

PennyScissors:
**Problem:**
The aws-event-sources chart is a Helm 3 chart and its CRDs will not install correctly when using Helm 2.

**Solution:**
Move chart to the [rancher/helm3-charts](https://github.com/rancher/helm3-charts) repository.

**Related PR:** https://github.com/rancher/helm3-charts/pull/21


deniseschannon:
We can't delete it in case someone has already deployed it. The only opton is to set a Rancher min and max version to the same so that it doesn't show up as an option but is available for anyone who is still using it.

PR #504, opened 2020-07-30
Move dev-charts to charts

cbron:
1st commit is a copy of current dev-charts
2nd commit is logos
3rd commit is changes for url changes


packages/canal/charts/Chart.yaml
```diff
@@ -7,7 +7,7 @@ home: https://www.projectcalico.org/
 keywords:
   - canal
 sources:
-  - https://github.com/rancher/charts
+  - https://github.com/rancher/dev-charts
```
StrongMonkey: why changing it back?

StrongMonkey: nvm

PR #479, opened 2020-06-23
Adding TiDB Operator helm chart 

Yisaer:
Here's some detail for TiDB Operator https://github.com/pingcap/tidb-operator

PR #505, opened 2020-07-31
Add tidb-operator v1.1.3

cofyc:
previous: https://github.com/rancher/charts/pull/479

v1.1.2 is skipped as we always recommend users deploy latest stable version.


cofyc:
@guangbochen @deniseschannon can you take a look?

guangbochen:
sure, @cofyc can u please help to split this PR into 2 commits, one with a base copy from the previous version, and the 2nd commit to adding the updated changes, thanks.

you can refer to an example PR in [here](https://github.com/rancher/charts/pull/511).

cofyc:
> sure, @cofyc can u please help to split this PR into 2 commits, one with a base copy from the previous version, and the 2nd commit to adding the updated changes, thanks.
> 
> you can refer to an example PR in [here](https://github.com/rancher/charts/pull/511).

done

PR #507, opened 2020-07-31
Minor updates to rancher-monitoring and rancher-pushprox chart

aiyengar2:
This PR encompasses several minor commits that are being introduced into the `rancher-monitoring` and `rancher-pushprox` charts to address feedback.

Related Issue(s): https://github.com/rancher/rancher/issues/28098


packages/rancher-pushprox/charts/templates/pushprox-clients.yaml
```diff
@@ -26,8 +26,11 @@ spec:
       {{- end }}
       containers:
       - name: pushprox-client
-        image: arvindiyengar/pushprox-linux-amd64:add_flag_for_token_path
-        command: [ "/app/pushprox-client" ]
+        image: {{ .Values.clients.image }}
```
deniseschannon, line 29: While we're making changes, we should add in support for a private registry. Also, I've seen some charts separate out between image and tag in their templating. Should we do the same?

aiyengar2: I agree with separating out the image tag vs. the templating; will update the PR accordingly for that.

Also, there are a set of minor changes that also need to be addressed like private registry, updating resource limits, etc. so I'm going to change this PR into one that addresses several minor fixes for the `rancher-monitoring` package.

Will be reflected on next push

aiyengar2: Pending in latest push: private registry changes

packages/rancher-monitoring/overlay/CHANGELOG.md
```diff
@@ -10,8 +10,8 @@ All notable changes from the upstream Prometheus Operator chart will be added to
...
+- Added `monitoring-admin`, `monitoring-edit`, and `monitoring-view` default `ClusterRoles` to allow admins to assign roles to users to interact with Prometheus Operator CRs. These can be enabled by setting `.Values.global.rbac.userRoles.create` (default: `true`). In a typical RBAC setup, you might want to assign specific users `monitoring-edit` or `monitoring-view` within a specific namespace to allow them to set up `ServiceMonitors` / `PodMonitors` that only monitor resources within that namespace. If `.Values.global.rbac.userRoles.aggregateRolesForRBAC` is enabled, these ClusterRoles will aggregate into the respective default ClusterRoles provided by Kubernetes
+- Added `grafana-config-edit` and `grafana-config-view` default `ClusterRoles` to allow admins to assign roles to users to interact with Secrets or ConfigMaps utilized by Grafana. These can be enabled by setting `.Values.global.rbac.userRoles.create` (default: `true`). In a typical RBAC setup, you might want to assign the following users with these permissions:
```
aiyengar2, line 14: missing changes to the `CHANGELOG.md` for all the minor commits in this PR

packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -8,7 +8,7 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-monitoring/charts-original/Cha
 -  and management of Prometheus instances.
 +annotations:
 +  catalog.cattle.io/certified: rancher
-+  catalog.cattle.io/namespace: monitoring-system
++  catalog.cattle.io/namespace: cattle-monitoring-system
```
aiyengar2, line 11: needs to also update namespaceOverride in values.yaml and prometheus operator URL

PR #508, opened 2020-08-01
Added a 'required' option for user to specify PostgreSQL password. 

vinayagg:
Without specifying a password, the charts were auto generating a password. This made upgrades and reinstalls fail as that generated a new password and the data saved on persistent volume was not readable.
Now we are giving the user to specify a password. So on upgrades and reinstalls they can specify the same password and upgrades and reinstalls will suceed.
Note: To reinstall without reusing the data, user needs to delete install, delete PVs, then reinstall. In this case, the data will be removed and the user can also specify a different password if desired.

### DIFF
--- a/charts/artifactory-ha/v2.5.3/questions.yml
+++ b/charts/artifactory-ha/v2.5.3/questions.yml
@@ -261,7 +261,7 @@ questions:
   - variable: postgresql.postgresqlPassword
     default: ""
     description: "PostgreSQL password"
-    type: string
+    type: password
     required: true
     label: PostgreSQL Password
     group: "Database Settings"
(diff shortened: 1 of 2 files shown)


charts/artifactory-ha/v2.5.3/questions.yml
```diff
@@ -258,6 +258,14 @@ questions:
   group: "Database Settings"
   show_subquestion_if: "true"
   subquestions:
+  - variable: postgresql.postgresqlPassword
+    default: ""
+    description: "PostgreSQL password"
+    type: string
```
PennyScissors, line 264: Using `type: password` might be better suited for these as it will hide the password in the text field and accomplish the same results.

vinayagg: Thank you for your review. I have updated all passwords to type "password".

PR #517, opened 2020-08-03
Add generateCRDChart flag

aiyengar2:
This PR is a continuation of https://github.com/rancher/dev-charts/pull/52.

It should address all the pending comments on the original PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,14 +70,62 @@ The step includes:
    
 5. Run `make CHART={CHART_NAME} clean`
    
-   This will clean up the `charts` directory so that it won't committed.
+   This will clean up the `charts` directory so that it won't be committed.
 
 This repo provides a [workflow](./.github/workflows) that automatically uploads patch files and tarball of charts. Commit will only need to update `package/${chart-name}/charts` and make sure patches are 
 up-to-date with the latest chart. It also automatically build github pages to serve `index.yaml` and artifacts of charts.
 
+### Experimental: Splitting CRDs from an upstream package into a separate package
+
+There are cases in which upstream charts import CRDs into a cluster using the Helm 3 `crd/` directory, which allows a user to first install the CRDs before rendering the templates created by the chart. However, using this approach has certain caveats [as documented by Helm](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/), such as an inability to upgrade / delete those CRDs and or use the `--dry-run` flag on the package before installing the CRDs. As a result, it may be advised to move those CRDs into a separate chart under the `templates/` directory so that the lifecycle of those CRDs can continue to be managed by Helm.
+
+However, in the current `rancher/charts` model, this would require deleting the CRDs from the upstream chart (which introduces significant changes to the package's patch) and maintaining a separate CRD chart that also needs to be kept up to date with the upstream chart. This poses several challenges, including but not limited to:
+- Keeping the version of the Rancher chart and the CRD chart consistent
+- Keeping the annotations added to the CRD chart in line with those added to the Rancher chart
+- Adding a validation YAML to the Rancher chart to direct the user to install the CRD chart before installing the Rancher chart if CRDs do not currently exist on the server
+- Viewing the patch between the CRDs introduced by the upstream chart and the CRDs within the CRD chart
+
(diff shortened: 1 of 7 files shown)


aiyengar2:
I created a separate branch that has the result of running the following commands using these scripts:

```
make CHART=rancher-monitoring prepare
make CHART=rancher-monitoring patch
```

[Here](https://github.com/aiyengar2/charts/tree/add_split_crd_flag_example/packages/rancher-monitoring) is what this would look like from the developer's perspective. Please let me know if you have any comments regarding the annotations (@cbron) or any additional things that need to be addressed!

scripts/prepare-crds
```diff
@@ -0,0 +1,107 @@
...
+apiVersion: ${api_version}
+version: ${chart_version}
+description: A Rancher chart that creates ${name} CRDs within a cluster.
+name: ${name}-crd
+type: application
+EOF
+
+# Add annotations to charts/Chart.yaml pointing to CRD chart
+# Only add one GVR as the annotation
+yq w -i ${f}/charts/Chart.yaml "annotations[catalog.cattle.io/auto-install-gvr]" "${crd_gvrs[0]}"
```
cbron, line 66: If another CRD is added to the folder it could change which CRD is used, but I suppose as long as they change at the same time its ok because this should only be used at install time. Alternative would be to have user provide them.

packages/rancher-monitoring/package.yaml
```diff
@@ -1,2 +1,3 @@
 url: https://kubernetes-charts.storage.googleapis.com/prometheus-operator-8.16.1.tgz
 packageVersion: 00
+splitCRDsIntoSeparatePackage: true
```
cbron, line 3: nit: somewhat verbose, maybe createCRDChart or something, but its fine

README.md
```diff
@@ -70,14 +70,61 @@ The step includes:
...
+
+4. On running `make CHART={CHART_NAME} charts`
+
+   A tarball for both the original chart and the CRD chart will be generated.
+
+Some more considerations when migrating to using this flag:
+- After adding this flag to a chart, you will have to look through the upstream chart and manually remove any CRD build specific code from the upstream chart (i.e. removing `helm.sh/hook: crd-install` from the CRD files, removing any cleanup Jobs introduced by the upstream chart to automatically delete CRDs on uninstall, etc.)
+- The CRDs moved to their own chart must not contain any code that was pulled from helper templates located within the main chart. If it is found that this is necessary for any chart, please submit a feature request.
+
+See `packages/rancher-monitoring` for an example of a chart that currently uses this flag.
```
cbron, line 122: Great job adding docs

packages/rancher-monitoring/package.yaml
aiyengar2: Should be a quick change, will update in the next commit!

scripts/prepare-crds
aiyengar2: I think it makes more sense for the user to provide it. Updated the fields provided by the user if they want to use the flag to instead be the following:
```
createCRDChart:
  enabled: true
  providesGVR: prometheuses.monitoring.coreos.com/v1
```

This way the user has control over what the value of the annotation will be so that it doesn't change on an upgrade.

aiyengar2: Updated [the example repo with `rancher-monitoring` changes](https://github.com/aiyengar2/charts/tree/add_split_crd_flag_example/packages/rancher-monitoring) based on the new flag used

cbron: Looks good, thanks.

packages/rancher-monitoring/package.yaml
PennyScissors: This looks pretty cool. I like `generateCRDChart.enabled`, but either works.

aiyengar2: I like that better too. Updated in the latest commit

StrongMonkey: IMO this doesn't need to dot-splited words. Feel like the single word is enough since we are not adopting the helm value style for this package.yaml.

aiyengar2, line 3: needs to be updated, will be done in the next commit

aiyengar2: updated

PR #520, opened 2020-08-03
Update version to 1.6.7

brendarearden:
Adding newest version of istio
Changing default values for components that should not be on during default install

### DIFF
--- a/packages/rancher-istio/charts/values.yaml
+++ b/packages/rancher-istio/charts/values.yaml
@@ -1,7 +1,7 @@
 revision: "" # leave blank for no canary deployment, or set to given revision, eg: "1-7-2"
 overlayFile: ""
-tag: 1.6.6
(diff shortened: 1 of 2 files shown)


packages/rancher-istio/charts/values.yaml
```diff
@@ -32,11 +32,11 @@ pilot:
     image: pilot
 
 policy:
-    enabled: true
+    enabled: false
```
cbron, line 35: Why did these all get switched to false ? Egress can stay false, but are policy and telemetry off by default ?

brendarearden: I turned them off by default since we don't have some of the other addons  - the demo profile also has them off by default. I went ahead and changed them to true again since we will likely switch them back in the next few prs

PR #521, opened 2020-08-04
Remove CRD code from rancher-monitoring

aiyengar2:
Removes the code that supports the Helm 2 hack for crd/ (`prometheus-operator/cleanup-crds.yaml` and `prometheus-operator/crds.yaml`) and removes crd-install hooks from crds.

Also updates the README.md and CHANGELOG.md accordingly.


aiyengar2:
docs/ directory should also be removed as of https://github.com/rancher/charts/pull/524

aiyengar2:
Just updated the tgz commit to point to the right files since https://github.com/rancher/charts/pull/524 was merged

PR #522, opened 2020-08-04
added kafka, splunk, syslog output support

paynejacob:


packages/rancher-logging/overlay/templates/outputs/elasticsearch/flow.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: logging.banzaicloud.io/v1beta1
```
brendarearden, line 2: do the `if .Values.elasticsearch.enabled` need to be here since they were removed from the other template?

paynejacob: Yes logging can be created regardless of of we have an output or not, the flow should only be created if an output exists for it. in this case this flow belongs to es and should only be created for es.

paynejacob:
Verified es deployment works `helm install --generate-name --set elasticsearch.enabled=true --set elasticsearch.host=jpayne.do.rancher.space --set elasticsearch.port=7200 projects/rancher/charts/assets/rancher-logging/rancher-logging-3.4.00000.tgz`

packages/rancher-logging/rancher-logging.patch
```diff
@@ -6,9 +6,10 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-logging/charts-original/Chart.
  appVersion: 3.4.0
 -description: A Helm chart to install Banzai Cloud logging-operator
 -name: logging-operator
+-version: 3.4.0
 +description: A Helm chart to install Banzai Cloud logging-operator and a cluster logging pipeline.
 +name: rancher-logging
- version: 3.4.0
++version: 3.4.000
```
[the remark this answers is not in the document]
paynejacob: not sure what happened here will fix

packages/rancher-logging/overlay/templates/outputs/elasticsearch/flow.yaml
cbron: Oh nvm, it already is

packages/rancher-logging/overlay/templates/outputs/syslog/deployment.yaml
```diff
@@ -0,0 +1,30 @@
...
+    matchLabels:
+      output: syslog
+  template:
+    metadata:
+      labels:
+        output: syslog
+    spec:
+      containers:
+      - name: fluentbit
+        image: paynejacob/fluent-bit-out-syslog:latest
```
cbron, line 20: needs to get moved to rancher, can be done after

packages/rancher-logging/overlay/templates/outputs/syslog/secret.yaml
```diff
@@ -0,0 +1,28 @@
...
+        Name              forward
+        Port              24224
+
+    [OUTPUT]
+        Name              syslog
+        InstanceName      syslog-output
+        Match             *
+        Addr              {{ .Values.syslog.address }}
+        Cluster           {{ .Values.syslog.cluster }}
+        Buffer_Chunk_Size 64000
```
cbron, line 21: is this going to be adjustable ?

PR #526, opened 2020-08-06
Add upstream Longhorn chart

ttpcodes:
This PR adds the chart for Longhorn as an Rancher modified chart as requested in rancher/rancher#27725 and longhorn/longhorn#1043.

### DIFF
--- a/packages/longhorn/longhorn.patch
+++ b/packages/longhorn/longhorn.patch
@@ -1,0 +1,12 @@
+diff -x '*.tgz' -x '*.lock' -uNr packages/longhorn/charts-original/Chart.yaml packages/longhorn/charts/Chart.yaml
+--- packages/longhorn/charts-original/Chart.yaml
++++ packages/longhorn/charts/Chart.yaml
+@@ -25,3 +25,8 @@
+ - https://github.com/longhorn/longhorn-ui
+ - https://github.com/longhorn/longhorn-tests
+ version: 1.0.1
++annotations:
++  catalog.cattle.io/certified: rancher
++  catalog.cattle.io/namespace: longhorn-system
++  catalog.cattle.io/release-name: longhorn
++  catalog.cattle.io/ui-component: longhorn
--- a/packages/longhorn/package.yaml
+++ b/packages/longhorn/package.yaml
@@ -1,0 +1,2 @@
+url: https://github.com/longhorn/charts/releases/download/longhorn-1.0.1/longhorn-1.0.1.tgz
+packageVersion: 00


packages/longhorn/longhorn.patch
```diff
@@ -0,0 +1,12 @@
+diff -x '*.tgz' -x '*.lock' -uNr packages/longhorn/charts-original/Chart.yaml packages/longhorn/charts/Chart.yaml
+--- packages/longhorn/charts-original/Chart.yaml
++++ packages/longhorn/charts/Chart.yaml
+@@ -25,3 +25,8 @@
+ - https://github.com/longhorn/longhorn-ui
+ - https://github.com/longhorn/longhorn-tests
+ version: 1.0.1
+++annotations:
++  catalog.rancher.io/certified: rancher
++  catalog.rancher.io/experimental: true
```
deniseschannon, line 10: Is it really still experimental @yasker?

packages/longhorn/longhorn.patch
```diff
@@ -0,0 +1,12 @@
...
++  catalog.rancher.io/namespace: longhorn-system
++  catalog.rancher.io/release-name: longhorn
```
deniseschannon, line 12: Assuming you'll want a custom UI form to launch longhorn instead of relying only on questions.yml, you will need to add in the 
```
catalog.cattle.io/ui-component:
```

and create an issue in github.com/rancher/dashboard.

packages/longhorn/longhorn.patch
yasker: No, it's no longer experimental.

packages/longhorn/longhorn.patch
yasker: I think we can deal with this part later since we don't have custom UI form now?

deniseschannon: If you plan on leveraging questions.yml, then you should add that file in the overlay folder. Otherwise you will not have any UI and only ability to edit yaml.

yasker: I see. Is there any way to test this?

yasker: In fact we should add 
```
```

See https://rancher.atlassian.net/wiki/spaces/EN/pages/563609933/Helm+Charts+Requirements

yasker: Let's skip questions.yml for now, since we will have the UI with Rancher.

If we need it later, we can always add that.

We still need to add `catalog.cattle.io/ui-component: longhorn` label.

packages/longhorn/longhorn.patch
```diff
@@ -0,0 +1,375 @@
...
++  - https://github.com/longhorn/longhorn
++  - https://github.com/longhorn/longhorn-engine
++  - https://github.com/longhorn/longhorn-instance-manager
++  - https://github.com/longhorn/longhorn-manager
++  - https://github.com/longhorn/longhorn-ui
++  - https://github.com/longhorn/longhorn-tests
+ version: 1.0.1
++annotations:
++  catalog.cattle.io/certified: rancher
++  catalog.cattle.io/experimental: false
```
deniseschannon, line 47: Removing the annotation would be better.

packages/longhorn/package.yaml
```diff
@@ -0,0 +1,5 @@
+url: https://github.com/longhorn/charts/releases/download/longhorn-1.0.1/longhorn-1.0.1.tgz
```
aiyengar2, line 1: We should use the git format for the `package.yaml` for this:

```suggestion
url: https://github.com/longhorn/charts.git
subdirectory: charts/longhorn
type: git
commit: longhorn-1.0.1
```

packages/longhorn/longhorn.patch
```diff
@@ -0,0 +1,374 @@
...
++  - https://github.com/longhorn/longhorn-tests
+ version: 1.0.1
++annotations:
++  catalog.cattle.io/certified: rancher
++  catalog.cattle.io/namespace: longhorn-system
++  catalog.cattle.io/release-name: longhorn
++  catalog.cattle.io/ui-component: longhorn
+diff -x '*.tgz' -x '*.lock' -uNr packages/longhorn/charts-original/crds/crds.yaml packages/longhorn/charts/crds/crds.yaml
+--- packages/longhorn/charts-original/crds/crds.yaml
++++ packages/longhorn/charts/crds/crds.yaml
```
aiyengar2, line 52: The `prepare-crd` script needs currently does not expect multiple CRDs to be located within the same file due to the way we collect the CRDs to be added to the validation script: https://github.com/rancher/charts/blob/531420dcf5ed3089d3dbca6f312304cf853b9b6d/scripts/prepare-crds#L48-L54

We should split each CRD into its own file for now. 

Created an issue to track the change required to support this: https://github.com/rancher/rancher/issues/28320

yasker: I am still unclear on why we need to include the CRD separately. Pinged you separately for a few questions.

yasker:
Per discussion with @aiyengar2 , there is no need to separate CRDs in Longhorn's case since:
1.  Longhorn doesn't not install CRs with CRDs.
2. For now, there is no point keep CRDs around after Longhorn being uninstalled - volumes will be gone when uninstalled, so does the configuration information. So Longhorn will remove the CRDs as a part of uninstallation.

So it’s fine to keep it that way, to install/uninstall CRDs using Longhorn chart instead of relying on Rancher to do it.

We can remove the third commit from the PR.

cbron:
Dashboard will only have helm3 support, so you should migrate [the chart](https://github.com/longhorn/longhorn/blob/master/chart/Chart.yaml#L1) to helm 3 or ensure its helm 3 compatible. I think you can keep existing setup, or move to a CRDs folder and let charts repo split it for you.

meldafrawi:
Longhorn v1.0.1 is compatible with Helm v3, it has been verified.

yasker:
Close this in favor of #546 .

PR #516, opened 2020-08-03
Add gatekeeper crd chart

PennyScissors:
Chart for rancher-gatekeeper's crds

**Related PR:** https://github.com/rancher/charts/pull/515

PR #527, opened 2020-08-06
Update gatekeeper chart

PennyScissors:
This PR accomplishes the same that was done in the PR's below, but using the CRD chart generator instead.

**Old PRs:**
https://github.com/rancher/charts/pull/515
https://github.com/rancher/charts/pull/516

**Related PRs:**
image-mirror opa gatekeeper bump: rancher/image-mirror#15


packages/rancher-gatekeeper/rancher-gatekeeper.patch
```diff
@@ -17,8 +17,30 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-gatekeeper/charts-original/Cha
...
+-    home: https://github.com/open-policy-agent/gatekeeper
+-    keywords:
+-    - open policy agent
+-    name: gatekeeper
+-    sources:
+-    - https://github.com/open-policy-agent/gatekeeper.git
+-    urls:
+-    - https://raw.githubusercontent.com/open-policy-agent/gatekeeper/v3.1.0-beta.11/charts/gatekeeper/gatekeeper-v3.1.0-beta.11.tgz
+-    version: v3.1.0-beta.11
+-generated: "2020-07-24T16:56:23.669766565-07:00"
```
aiyengar2: belongs in the next commit

packages/rancher-gatekeeper/rancher-gatekeeper.patch
```diff
@@ -301,4 +339,4 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-gatekeeper/charts-original/val
 +  systemDefaultRegistry: ""
 +  kubectl:
 +    repository: rancher/istio-kubectl
-+    tag: 1.4.6
++    tag: 1.5.8
```
aiyengar2: belongs in the next commit

aiyengar2:
I had two concerns about this PR as a whole:

#### Regarding [moving CRDs into a crd/ folder](https://github.com/rancher/charts/pull/527/commits/d143c250e35b8338acb8641d62b1503397d89a02):
I think this looks good for now, but as part of changes to the build scripts for CRD charts, it might make sense to add an additional flag `crdFiles` to the `generateCRDChart` flag in `package.yaml` to list filepaths to move into the CRD folder before running all the CRD scripts, e.g.
```
generateCRDChart:
  enabled: true
  providesGVR: configs.config.gatekeeper.sh/v1alpha1
  crdFiles:
  - templates/config-customresourcedefinition.yaml
  - templates/constrainttemplate-customresourcedefinition.yaml
  - templates/constrainttemplatepodstatus-customresourcedefinition.yaml
```

This would be useful for other packages that are currently still using Helm 2. It will also unclutter the patch file and allow us to identify specific changes that we added to the CRDs (such as removing the helm hooks) just by looking at the patch file.

#### Regarding [upgrading the opa-gatekeeper chart manually](https://github.com/rancher/charts/pull/527/commits/4fa519e3ff49f5f3575411c07a3bee7c314f2f6d):
I talked to @PennyScissors about this offline and it seems like the upstream chart made a lot of structural changes (i.e. going from a single YAML to 8 split YAMLs and several changes on top of that) that made it difficult to transition changes from the current chart to the new upstream chart version since the patch is based off of the older version of the chart.

Adapting the patch to be based off the new upstream would involve redoing all the changes that were made to the chart till date onto the new upstream base and creating a brand new patch for it. While this may be possible at this point, this might be a bigger concern in the future for this chart and other charts in this repo when / if several significant changes are introduced by Rancher (i.e. in the case of `rancher-monitoring`) and we encounter this situation again while trying to update the chart to the latest upstream version.

@StrongMonkey do you have any thoughts on this?

packages/rancher-gatekeeper/overlay/CHANGELOG.md
```diff
@@ -0,0 +1,21 @@
...
+- Updated for Helm 3 compatibility
+    - Moved crds to `crds` directory
+    - Removed `crd-install` hooks and templates from crds
+
+### Removed
+- Removed the following files as the `gatekeeper-webhook-service` was removed in our previous version of the chart
+    - `gatekeeper-validating-webhook-configuration-validatingwebhookconfiguration.yaml`
+    - `gatekeeper-webhook-service-service.yaml`
+- Removed `gatekeeper-system-namespace.yaml` as Rancher handles namespaces for chart installation
+- Removed unnecessary `index.yaml` as we package and host our charts
```
cbron, line 21: Good changelog, I think the modified/removed sections are valuable to end users.

packages/rancher-gatekeeper/rancher-gatekeeper.patch
```diff
@@ -1,103 +1,479 @@
...
  keywords:
    - open policy agent
--version: v3.1.0-beta.7
-+version: 0.1.0
- home: https://github.com/open-policy-agent/gatekeeper
+ version: v3.1.0-beta.11
+@@ -8,3 +8,9 @@
  sources:
    - https://github.com/open-policy-agent/gatekeeper.git
- appVersion: v3.1.0-beta.7
```
cbron, line 16: you could remove old version to clean things up since the chart hasn't gone live yet, but not necessary

packages/rancher-gatekeeper/rancher-gatekeeper.patch
```diff
@@ -259,34 +1148,33 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-gatekeeper/charts-original/tem
...
 +++ packages/rancher-gatekeeper/charts/values.yaml
-@@ -1,12 +1,12 @@
- replicas: 1
+@@ -1,5 +1,5 @@
+ replicas: 3
 -auditInterval: 60
 +auditInterval: 300
  constraintViolationsLimit: 20
  auditFromCache: false
+ disableValidatingWebhook: false
```
cbron, line 1160: Is this what got disabled ? Here it looks like you are re-enabling it.

StrongMonkey:
@aiyengar2 Agreed with what Caleb said, it is just a maintaining effort we have to constistantly put. There is no magic behind patch file, and when the change in upstream is large it serves no purpose as an automation script. But without the patch file I don't see how possible to rebase the chart with upstream unless the developer remembers all the change he did, which is the same problem with istio rancher integration before 1.4. So patch file is just changelog of what we changed between our chart and upstream, and when the change is small you can apply the patch directly, and when change is large you have to manually resolve the diff and apply the diff.

packages/rancher-gatekeeper/rancher-gatekeeper.patch
PennyScissors: That was added as part of the patch and I missed it. However, thanks for pointing it out because I could just disable the webhook validation from here and keep the yaml which will clean up the patch a bit by not showing the deletion.

packages/rancher-gatekeeper/rancher-gatekeeper.patch
PennyScissors: @cbron remove the old version as in remove the previous patch?

packages/rancher-gatekeeper/rancher-gatekeeper.patch
cbron: That sounds cleaner.

packages/rancher-gatekeeper/rancher-gatekeeper.patch
PennyScissors: I'll remove it.

packages/rancher-gatekeeper/rancher-gatekeeper.patch
```diff
@@ -259,28 +208,30 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-gatekeeper/charts-original/tem
 +          def_msg := sprintf("Label <%v: %v> does not satisfy allowed regex: %v", [key, value, expected.allowedRegex])
 +          msg := get_message(input.parameters, def_msg)
 +        }
-\ No newline at end of file
 diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-gatekeeper/charts-original/values.yaml packages/rancher-gatekeeper/charts/values.yaml
 --- packages/rancher-gatekeeper/charts-original/values.yaml
 +++ packages/rancher-gatekeeper/charts/values.yaml
-@@ -1,12 +1,12 @@
- replicas: 1
+@@ -1,15 +1,15 @@
```
StrongMonkey, line 1090: Why is replicas defaults to 3?

PennyScissors: @StrongMonkey The opa chart went through big changes in version `v3.1.0-beta.9`. 
This is being picked up from those changes after updating to `v3.1.0-beta.11`.

PennyScissors: That change is being picked up from updating to v3.1.0-beta.11 since there were big changes in the upstream v3.1.0-beta.9 we currently don't have.

StrongMonkey:
@PennyScissors Can we move all the add-on files into overlays files?(like allowedrepos.yaml, job-constraints-crd.yaml and requiredlabels.yaml)

packages/rancher-gatekeeper/overlay/CHANGELOG.md
```diff
@@ -0,0 +1,23 @@
+# Changelog
+All notable changes from the upstream OPA Gatekeeper chart will be added to this file
+
+## [Package Version 00] - 2020-07-27
+### Added
+- Enabled the CRD chart generator in `package.yaml`
+
+### Modified
+- Updated chart version in `Chart.yaml` to follow the upstream's format `v3.1.0-beta.X`
```
deniseschannon, line 9: I feel like the changelog should only have notes on what we changed from upstream.

PennyScissors: Okay, I'll edit the changelog.

aiyengar2:
> Assuming that we may do @aiyengar2 suggestion of listing CRD files later, but doing via patch file for now.
> LGTM, but let's get another review on it before merging.

Issue to track this: https://github.com/rancher/rancher/issues/28320

maggieliu:
related to: https://github.com/rancher/rancher/issues/28243

maggieliu:
related to: https://github.com/rancher/rancher/issues/27378

PR #531, opened 2020-08-07
Istio Phase 2: Change namespace, add forceInstall flag

brendarearden:
Change namespace from istio-system to cattle-istio system
 -- Left option to default to istio-system if being installed on own an not through rancher

Change istioctl-installer job to delete on success (to ensure namespace is clear after uninstall)

Add value to set for forceInstall to bypass version checks and analyze checks in installer script

Added back in addon components since those are still available until 1.7.x (and we are currently at 1.6.7)

### DIFF
--- a/packages/rancher-istio/charts/README.md
+++ b/packages/rancher-istio/charts/README.md
@@ -4,16 +4,17 @@ A Rancher created chart that packages the istioctl binary to install via a helm 
 
 # Installation
 
-### pre-requisites
+### Requirements
 
-This chart depends on the rancher-kiali-server-crd chart. 
+This chart depends on the rancher-kiali-server-crd chart.
 
-It also depends on rancher-monitoring being installed with default values for nameOverride, namespaceOverride, and the prometheus.service.port.
+It also depends on the `rancher-monitoring` chart being installed with default values for `nameOverride`, `namespaceOverride`, and `prometheus.service.port`.
 If those values are modified on the rancher-monitoring deployment, please adjust the `kiali.external_services.prometheus` url settings:
 ```
 http://{{ .Values.nameOverride }}-prometheus.{{ .Values.namespaceOverride }}.svc:{{ prometheus.service.port }}
 ```
 
-### installation
-
+### Installation
+```
 helm install rancher-istio ./ --create-namespace -n cattle-istio-system
+```
--- a/packages/rancher-istio/charts/configs/istio-base.yaml
+++ b/packages/rancher-istio/charts/configs/istio-base.yaml
@@ -67,6 +67,7 @@ spec:
         name: istio-ingressgateway
         type: {{ .Values.ingressGateways.type }}
     global:
+      istioNamespace: {{ template "istio.namespace" . }}
       proxy:
         image: {{ .Values.global.proxy.image }}
       proxy_init:
--- a/packages/rancher-istio/charts/samples/overlay-example.yaml
+++ b/packages/rancher-istio/charts/samples/overlay-example.yaml
@@ -1,0 +1,37 @@
+apiVersion: install.istio.io/v1alpha1
+kind: IstioOperator
+spec:
+  components:
+    ingressGateways:
+    - enabled: true
+      name: ilb-gateway
+      namespace: user-ingressgateway-ns
+      k8s:
+        resources:
+          requests:
+            cpu: 200m
+        service:
+          ports:
+          - name: tcp-citadel-grpc-tls
+            port: 8060
+            targetPort: 8060
+          - name: tcp-dns
+            port: 5353
+        serviceAnnotations:
+          cloud.google.com/load-balancer-type: internal
+    - enabled: true
+      name: other-gateway
+      namespace: cattle-istio-system
+      k8s:
+        resources:
+          requests:
+            cpu: 200m
+        service:
+          ports:
+          - name: tcp-citadel-grpc-tls
+            port: 8060
+            targetPort: 8060
+          - name: tcp-dns
+            port: 5353
+        serviceAnnotations:
+          cloud.google.com/load-balancer-type: internal
--- a/packages/rancher-istio/charts/templates/admin-role.yaml
+++ b/packages/rancher-istio/charts/templates/admin-role.yaml
@@ -4,8 +4,8 @@ kind: ClusterRole
 metadata:
   labels:
     rbac.authorization.k8s.io/aggregate-to-admin: "true"
-  namespace: istio-system
   name: istio-admin
+  namespace: {{ template "istio.namespace" . }}
 rules:
   - apiGroups:
       - config.istio.io
--- a/packages/rancher-istio/charts/templates/istio-install-job.yaml
+++ b/packages/rancher-istio/charts/templates/istio-install-job.yaml
@@ -2,13 +2,11 @@ apiVersion: batch/v1
 kind: Job
 metadata:
   name: istioctl-installer
-  namespace: istio-system
+  namespace: {{ template "istio.namespace" . }}
   annotations:
     "helm.sh/hook": post-install,post-upgrade
     "helm.sh/hook-weight": "-5"
-    # Todo: use following line instead
-    # "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
-    "helm.sh/hook-delete-policy": before-hook-creation
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
 spec:
   backoffLimit: 1
   template:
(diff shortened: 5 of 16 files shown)


packages/rancher-istio/charts/templates/istio-install-job.yaml
```diff
@@ -2,27 +2,29 @@ apiVersion: batch/v1
...
-    "helm.sh/hook-delete-policy": before-hook-creation
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
 spec:
   backoffLimit: 1
   template:
     spec:
       containers:
         - name: istioctl-installer
-          image: rancher/istio-installer:{{ .Values.installerVersion }}
+          image: brendarearden/istio-test:4.1.6
```
brendarearden, line 16: will update when new rancher/istio-installer is available prior to merging this pr

packages/rancher-istio/charts/templates/istio-uninstall-job.yaml
```diff
@@ -12,10 +12,12 @@ spec:
     spec:
       containers:
         - name: istioctl-uninstaller
-          image: rancher/istio-installer:{{ .Values.installerVersion }}
+          image: brendarearden/istio-test:4.1.6
```
brendarearden, line 15: will update when new rancher/istio-installer is available prior to merging this pr

packages/rancher-istio/charts/templates/istio-install-job.yaml
rmweir: just a reminder to do this

packages/rancher-istio/charts/values.yaml
```diff
@@ -2,7 +2,21 @@ revision: "" # leave blank for no canary deployment, or set to given revision, e
 overlayFile: ""
 tag: 1.6.7
 installerVersion: 0.1.5
+forceInstall: false
 
+# these addon values will be depricated starting in 1.7.0
```
aiyengar2, line 7: nit:

```suggestion
# these addon values will be deprecated starting in 1.7.0
```

packages/rancher-istio/charts/templates/admin-role.yaml
```diff
@@ -4,7 +4,7 @@ kind: ClusterRole
 metadata:
   labels:
     rbac.authorization.k8s.io/aggregate-to-admin: "true"
-  namespace: istio-system
+  namespace: {{ .Values.global.istioNamespace | default "istio-system" }}
```
aiyengar2, line 7: Should this be changed from `istio-system` to `cattle-istio-system`?

```suggestion
  namespace: {{ .Values.global.istioNamespace | default "cattle-istio-system" }}
```

packages/rancher-istio/charts/configs/istio-base.yaml
```diff
@@ -57,6 +65,9 @@ spec:
         name: istio-ingressgateway
         type: {{ .Values.ingressGateways.type }}
     global:
+      {{- if .Values.global.istioNamespace }}
+      istioNamespace: {{ .Values.global.istioNamespace }}
+      {{- end }}
```
aiyengar2, line 70: Could we use `{{ .Values.global.istioNamespace | default "cattle-istio-system" }}` here?

packages/rancher-istio/charts/values.yaml
```diff
@@ -47,6 +61,7 @@ sidecarInjectorWebhook:
   rewriteAppHTTPProbe: true
 
 global:
+  istioNamespace: cattle-istio-system
```
aiyengar2, line 64: Based on how this is being used to override the default namespace we deploy Istio in, I think that we should call this `istioNamespaceOverride` instead of `istioNamespace`

packages/rancher-istio/charts/templates/admin-role.yaml
brendarearden: by default istio deploys to istio-system if the istioNamespace value is not provided, so the roles and services accounts need to do the same

packages/rancher-istio/charts/configs/istio-base.yaml
brendarearden: the reason I opted not to is if we aren't changing the namespace we don't need to set this value at all as istio will deploy in the istio-system namespace.

brendarearden: I take that back, since we are using this flag in installer image, we will want to ensure its always set - I went ahead to updated it with your suggestion

packages/rancher-istio/charts/values.yaml
brendarearden: istioNamespace is the istio way to override the default namespace so I propose we leave it to make it easier when comparing with istio documentation. But this does bring up a good question - do we want to fall back to the default istio-system or do we want to enforce the cattle-istio-system which might be a good reason to add that additional override value so it would be {{ <override value> | cattle-istio-system | istio-system }}. allowing it to default to istio-system when istioNamespace isn't set makes the chart more appealing outside of rancher.

aiyengar2: Hmm I agree with keeping this inline with istio documentation by having `istioNamespace: istio-system`.

I also like the idea of having an additional override value like `istioNamespaceOverride: cattle-istio-system` that overrides the `istioNamespace` value provided in the `values.yaml`. That way a user who wants to use this chart outside of Dashboard can just set `istioNamespaceOverride: ""` if required and fallback to `istioNamespace: istio-system` like they would normally expect without any more changes.

@cbron do you have any thoughts on this?

packages/rancher-istio/charts/configs/istio-base.yaml
aiyengar2: we may want to update this based on the resolution in https://github.com/rancher/charts/pull/531#discussion_r469574900

packages/rancher-istio/charts/templates/admin-role.yaml
aiyengar2: This may not be consistent with the default set here right to fall back to `cattle-istio-system`? https://github.com/rancher/charts/pull/531#discussion_r468087217.

packages/rancher-istio/charts/values.yaml
cbron: I like that approach too since basically no one installs the standard istio platform outside of istio-system NS. They wouldn't expect it to go to cattle NS. Either partial or default func is fine, I guess partial is cleaner.

cbron:
@brendarearden rebase from upstream and CI should pass

packages/rancher-istio/charts/samples/overlay-example.yaml
```diff
@@ -0,0 +1,39 @@
...
+        service:
+          ports:
+          - name: tcp-citadel-grpc-tls
+            port: 8060
+            targetPort: 8060
+          - name: tcp-dns
+            port: 5353
+        serviceAnnotations:
+          cloud.google.com/load-balancer-type: internal
+      name: ilb-gateway
```
cbron, line 22: Can you put this name and namespace up on first line where ` - enabled: true` is ? I always find name at bottom to be confusing.

packages/rancher-istio/charts/samples/overlay-example.yaml
```diff
@@ -0,0 +1,39 @@
...
+            cpu: 200m
+        service:
+          ports:
+          - name: tcp-citadel-grpc-tls
+            port: 8060
+            targetPort: 8060
+          - name: tcp-dns
+            port: 5353
+        serviceAnnotations:
+          cloud.google.com/load-balancer-type: internal
```
cbron, line 37: two is probably enough, the 3rd one looks the same ?

brendarearden: I will remove the first one and the leave the two that look the same (they are, but the namespace is different and I want to highlight)

packages/rancher-istio/charts/README.md
```diff
@@ -6,7 +6,7 @@ A Rancher created chart that packages the istioctl binary to install via a helm
 
 ### pre-requisites
```
aiyengar2, line 7: nit:
```suggestion
### Requirements
```

packages/rancher-istio/charts/README.md
```diff
@@ -6,7 +6,7 @@ A Rancher created chart that packages the istioctl binary to install via a helm
...
 
-This chart depends on the rancher-kiali-server-crd chart. 
+This chart depends on the rancher-kiali-server-crd chart.
 
 It also depends on rancher-monitoring being installed with default values for nameOverride, namespaceOverride, and the prometheus.service.port.
 If those values are modified on the rancher-monitoring deployment, please adjust the `kiali.external_services.prometheus` url settings:
```
aiyengar2, line 12: nit:
```suggestion
It also depends on the `rancher-monitoring` chart being installed with default values for `nameOverride`, `namespaceOverride`, and `prometheus.service.port`.
If those values are modified on the rancher-monitoring deployment, please adjust the `kiali.external_services.prometheus` url settings:
```

packages/rancher-istio/charts/README.md
aiyengar2: nit: 

Also, if we are updating the README, on the later parts of the `README.md`, installation should be capitalized and the helm install instruction should be encapsulated with a ``` like below:

```bash
helm install rancher-istio ./ --create-namespace -n cattle-istio-system
```

PR #532, opened 2020-08-08
rancher-kiali-server

cbron:


scripts/prepare-crds
```diff
@@ -74,10 +78,10 @@ fi
...
-	if [[ ${a} == "catalog.cattle.io/release-name" ]]; then
-		v="${v}-crd"
-	fi
 	if ! [[ -z ${v} ]]; then
+		if [[ ${a} == "catalog.cattle.io/release-name" ]]; then
+			v="${v}-crd"
+		fi
 		yq w -i ${f}/charts-crd/Chart.yaml "annotations[${a}]" "${v}"
 	fi
 done
```
cbron, line 87: @aiyengar2 would be good for you to review whole PR, but especially this part ^

packages/rancher-istio/charts/requirements.lock
```diff
@@ -0,0 +1,6 @@
+dependencies:
```
aiyengar2, line 1: I think `requirements.lock` and the binary in `charts/` should be omitted from this PR with a `.gitignore`

packages/rancher-istio/charts/templates/service-monitors.yaml
```diff
@@ -0,0 +1,51 @@
...
+kind: ServiceMonitor
+metadata:
+  name: envoy-stats-monitor
+  namespace: istio-system
+  labels:
+    monitoring: istio-proxies
+spec:
+  selector:
+    matchExpressions:
+    - {key: istio-prometheus-ignore, operator: DoesNotExist}
```
aiyengar2, line 12: nit:
```suggestion
    - key: istio-prometheus-ignore
      operator: DoesNotExist
```

packages/rancher-istio/charts/templates/service-monitors.yaml
```diff
@@ -0,0 +1,51 @@
...
+  name: istio-component-monitor
+  namespace: istio-system
+  labels:
+    monitoring: istio-components
+spec:
+  jobLabel: istio
+  targetLabels: [app]
+  selector:
+    matchExpressions:
+    - {key: istio, operator: In, values: [pilot]}
```
aiyengar2, line 45: nit:
```suggestion
    - key: istio
      operator: In
      values: [pilot]
```

packages/rancher-istio/charts/values.yaml
```diff
@@ -51,3 +51,27 @@ global:
...
+  auth:
+    # todo: what auth strategy ?
+    # strategy: token # this is default, you can use a servic-account or kubeconfig token
+    strategy: anonymous
+  deployment:
+    ingress_enabled: false
+  external_services:
+    prometheus:
+      custom_metrics_url: "http://monitoring-rancher-monitor-prometheus.monitoring-system.svc:9090"
+      url: "http://monitoring-rancher-monitor-prometheus.monitoring-system.svc:9090"
```
aiyengar2, line 71: We should add a comment here (or to the `README.md` for this field) that this default depends on `rancher-monitoring` also being installed with the default values for `nameOverride`, `namespaceOverride`, and the `prometheus.service.port`.

If modified on the `rancher-monitoring` deployment, the appropriate value here should be:
`http://{{ .Values.nameOverride }}-prometheus.{{ .Values.namespaceOverride }}.svc:{{ prometheus.service.port }}` based on the values set on the `rancher-monitoring` chart.

packages/rancher-istio/charts/Chart.yaml
```diff
@@ -9,3 +9,5 @@ annotations:
   catalog.cattle.io/namespace: istio-system
   catalog.cattle.io/release-name: rancher-istio
   catalog.cattle.io/ui-component: istio
+  catalog.cattle.io/requires-gvr: prometheuses.monitoring.coreos.com/v1
+  catalog.cattle.io/auto-install-gvr: monitoringdashboards.monitoring.kiali.io/v1alpha1
```
aiyengar2, line 13: At the moment, this will be overridden based on https://github.com/rancher/charts/blob/531420dcf5ed3089d3dbca6f312304cf853b9b6d/scripts/prepare-crds#L67-L71

While making the changes to the `prepare-crd` script, can you add some logic to make sure we don't override the values if they are already declared in the Chart.yaml? e.g.
```bash
if ! [[ -z ${providesGVR} ]]; then
        if [[ -z $(yq r ${f}/charts/Chart.yaml "annotations[catalog.cattle.io/auto-install-gvr]") ]]; then
	    yq w -i ${f}/charts/Chart.yaml "annotations[catalog.cattle.io/auto-install-gvr]" "${providesGVR}"
        fi
	yq w -i ${f}/charts-crd/Chart.yaml "annotations[catalog.cattle.io/provides-gvr]" "${providesGVR}"
	yq w -i ${f}/charts-crd/Chart.yaml "annotations[catalog.cattle.io/hidden]" "true"
fi
```

scripts/prepare-crds
```diff
@@ -48,8 +48,12 @@ chart_version=$(cat ${f}/charts/Chart.yaml | yq r - 'version')
 crd_apis=()
 for crd_yaml in ${f}/charts-crd/templates/*; do
 	crd_group=$(yq r ${crd_yaml} 'spec.group')
-	crd_version=$(yq r ${crd_yaml} 'spec.version')
 	crd_kind=$(yq r ${crd_yaml} 'spec.names.kind')
+	crd_version=$(yq r ${crd_yaml} 'spec.version')
+	if test -z "$crd_version"
+	then
```
aiyengar2, line 54: nit:
```suggestion
	if [[ -z "$crd_version" ]]; then
```

scripts/prepare-crds
aiyengar2: Good catch, this would be a bug if the annotation was not set as it would have resulted in:

```yaml
annotations:
    catalog.cattle.io/release-name: "-crd"
```

Looks good to me, thanks for the change!

packages/rancher-istio/charts/configs/istio-base.yaml
```diff
@@ -4,6 +4,14 @@ spec:
...
     istiocoredns:
       enabled: {{ .Values.istiocoredns.enabled }}
+    prometheus:
+      enabled: false
+    grafana:
+      enabled: false
+    kiali:
+      enabled: false
+    tracing:
+      enabled: false
```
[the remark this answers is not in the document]
cbron: They will conflict, I'd go with the original except that we don't want people turning this on right now as they are all deprecated. If they really want to turn on they can force override via Overlay.

packages/rancher-istio/charts/templates/service-monitors.yaml
cbron: this is an imported file so I'd prefer to leave as-is to simplify future imports

packages/rancher-istio/charts/values.yaml
cbron: added

packages/rancher-istio/charts/Chart.yaml
cbron: added below

packages/rancher-istio/charts/requirements.lock
cbron: fixed

cbron: Removed the tgz actually failed the generate-charts CI step: 
> Error: found in Chart.yaml, but missing in charts/ directory: rancher-kiali-server

cbron: Figured this out, the `helm dependency update` call doesn't get called for non package.yaml charts. Think we can just add that in after any sub-charts get prepared.

cbron: Added a fix for this, although not entirely sure we want to go down that road. And another issue with crd generation getting double versions.

scripts/prepare
```diff
@@ -61,5 +61,16 @@ for f in packages/*; do
...
+# if any sub-charts were prepared pull them in
+for f in packages/*; do
+	if [[ ! -f ${f}/package.yaml && -f ${f}/charts/requirements.yaml ]]; then
+		ls ${f}/charts/requirements.yaml
+		pwd=$(pwd)
+		cd ${f}/charts
+		helm dependency update
+		cd $pwd
+	fi
 done
```
cbron, line 76: @StrongMonkey can you review this idea ^. Basically I am packaging a chart (kiali) separately in charts and now need to call `helm dependency update` in a non upstream chart to pull it in after it gets generated.

StrongMonkey: Looks good to me.

aiyengar2: @cbron I believe we should just move the following lines of code outside the `if [[ -z $CHART || $CHART == $(basename -- ${f}) ]]; then` statement:

https://github.com/rancher/charts/blob/f6f0c3ddbe3c56e62f3b9c32b4964a1c5fe230f5/scripts/prepare#L42-L62

aiyengar2: The `copied_dependencies` logic will store nothing for charts without `overlay/` directories (i.e. for non upstream charts) since it checks to see if the overlay chart exists before doing the `cp -R` or adding it to the list of `copied_dependencies`. 

So the code path should be the same regardless of a `package.yaml` exists.

aiyengar2: One more change will also be required: we need to flip these two if statements

https://github.com/rancher/charts/blob/f6f0c3ddbe3c56e62f3b9c32b4964a1c5fe230f5/scripts/prepare#L8-L9

The logic I mentioned above will be within the`if [[ -z $CHART || $CHART == $(basename -- ${f}) ]]; then` statement but outside the `if [[ -f ${f}/package.yaml ]]; then` statement. This will avoid running the `helm dependency update` on all charts when a specific `CHART` is selected via env var.

packages/rancher-istio/charts/requirements.lock
aiyengar2: I think this route makes sense; without the tgz files, developers won't be able to do `helm install` either on the chart. The solution when working with a chart that did have a `package.yaml` involved creating these tgz files and [ignoring them on a patch](https://github.com/rancher/charts/blob/f6f0c3ddbe3c56e62f3b9c32b4964a1c5fe230f5/scripts/generate-patch#L65), but this should have been done for all packages (see comments below).

For the `generate-charts` step, even if the tgz is removed when committing this change it should still be automatically generated during CI since we always run the `prepare` script before running `generate-charts`: https://github.com/rancher/charts/blob/f6f0c3ddbe3c56e62f3b9c32b4964a1c5fe230f5/Makefile#L10-L11

scripts/prepare
cbron: The problem with that approach is that the generated charts need to exist before they can be vendored in via `helm dependency`. So if we loop the charts and kiali comes after istio, then kiali won't exist when istio tries to vendor. 

But on 3rd part yes I need to wrap in `$CHART` check.

packages/rancher-istio/charts/requirements.lock
cbron: Right, for these kind of charts the generate charts step won't work unless prepare is run first. It does make dev slightly harder but they can just run `helm dependency` manually to test un-committed changes.

scripts/prepare
cbron: added that check

aiyengar2: Good catch; then I agree we should use two for loops, but we can just move https://github.com/rancher/charts/blob/f6f0c3ddbe3c56e62f3b9c32b4964a1c5fe230f5/scripts/prepare#L42-L62 into the second for loop instead of having the same logic twice.

Neither for loop here should impact charts that do not use package.yaml since such charts will not have an `overlay/` folder, which means no `copied_dependencies` will ever be added due to this: https://github.com/rancher/charts/blob/f6f0c3ddbe3c56e62f3b9c32b4964a1c5fe230f5/scripts/prepare#L49

cbron: updated, @StrongMonkey want to have a look again

scripts/prepare-crds
```diff
@@ -48,8 +48,11 @@ chart_version=$(cat ${f}/charts/Chart.yaml | yq r - 'version')
 crd_apis=()
 for crd_yaml in ${f}/charts-crd/templates/*; do
 	crd_group=$(yq r ${crd_yaml} 'spec.group')
-	crd_version=$(yq r ${crd_yaml} 'spec.version')
 	crd_kind=$(yq r ${crd_yaml} 'spec.names.kind')
+	crd_version=$(yq r ${crd_yaml} 'spec.version')
+	if test -z "$crd_version"; then
```
aiyengar2, line 53: nit for consistency:

```suggestion
	if [[ -z ${crd_version} ]]; then
```

cbron: fixed

PR #535, opened 2020-08-10
added rke support

paynejacob:
This pr adds configuration for rke1 log aggregation.  By default rke 1 log aggregation is disabled.

### DIFF
--- a/packages/rancher-logging/rancher-logging.patch
+++ b/packages/rancher-logging/rancher-logging.patch
@@ -27,7 +27,7 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-logging/charts-original/values
  
  resources: {}
    # We usually recommend not to specify default resources and to leave this as a conscious
-@@ -76,4 +76,44 @@
+@@ -76,4 +76,48 @@
  monitoring:
    # Create a Prometheus Operator ServiceMonitor object
    serviceMonitor:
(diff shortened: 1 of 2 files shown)


packages/rancher-logging/overlay/templates/logging-rke.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if .Values.additionalLoggers.RKE1Enabled }}
+apiVersion: logging.banzaicloud.io/v1beta1
+kind: Logging
+metadata:
+  name: {{ .Release.Name }}-rke
+  namespace: {{ .Release.Namespace }}
+  labels:
+{{ include "logging-operator.labels" . | indent 4 }}
```
aiyengar2, line 8: nit:
```suggestion
  labels: {{ include "logging-operator.labels" . | nindent 4 }}
```

packages/rancher-logging/rancher-logging.patch
```diff
@@ -74,4 +74,6 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-logging/charts-original/values
 +  address: ""
 +  cluster: true
 +  root_ca: ""
-\ No newline at end of file
++
++additionalLoggers:
++  RKE1Enabled: false
```
aiyengar2, line 79: Since we are creating an additional Logging component with this flag enabled, I think we should name it with `Logging` rather than `Logger`. Also, it may be possible that more fields other than `enabled` are exposed here and a comment would be helpful to explain what this field does.

Here's my suggestion for reformatting it:
```yaml
## Configure which default Logging CRs are deployed to capture logs from other locations
##
additionalLoggingSources:
    rke:
        enabled: true
```

PR #539, opened 2020-08-12
use secrets for certs and passwords

paynejacob:
Problem:
We want to avoid passing sensitive data in values.yaml

Solution:
All sensitive fields use a secret now


packages/rancher-logging/overlay/templates/outputs/elasticsearch/output.yaml
```diff
@@ -12,32 +12,32 @@ spec:
   {{- if .Values.elasticsearch.user }}
 user: {{ .Values.elasticsearch.user }}
   {{- end}}
-  {{- if .Values.elasticsearch.password }}
+{{- if .Values.elasticsearch.password.secret_name }}
 password:
```
cbron, line 16: What happens if elastic-search is enabled but password isn't passed, it just fails ?

packages/rancher-logging/overlay/templates/outputs/elasticsearch/secret.yaml
```diff
@@ -1,14 +0,0 @@
-{{- if .Values.elasticsearch.enabled }}
```
cbron, line 1: Are we just going to document how users should set these secrets up ? We could deploy an empty secret but I'm guessing by then its too late.

packages/rancher-logging/overlay/templates/outputs/elasticsearch/output.yaml
paynejacob: es does not require a password so it depends on the setup, if users do not provide a password to a secured cluster they will get errors in their fluentd.

packages/rancher-logging/overlay/templates/outputs/elasticsearch/secret.yaml
paynejacob: The ui is supposed to create a secret for them, we will need to document it for cli users.

cbron: Do we know the default secret name ? Should put that in the default values ? Or not because that would assume user has authentication turned on at ES side.

paynejacob: I am in favor of not defaulting a name because it isn't normally required.

PR #540, opened 2020-08-12
added toggle for logging persistence

paynejacob:


packages/rancher-logging/rancher-logging.patch
```diff
@@ -82,3 +82,5 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-logging/charts-original/values
 +additionalLoggingSources:
 +  rke1:
 +    enabled: false
++
++disablePvc: true
```
cbron, line 86: Weren't we going to add a buffer size in here too via the UI ? When we turned this on or off the buffer size would show up or not. I get that might be part of different CRD so maybe in different area ?

paynejacob: per our conversation earlier that will be a separate pr https://github.com/rancher/rancher/issues/28401

PR #544, opened 2020-08-13
rke2-monitoring

MonzElmasry:
Enable the rancher dashboard to monitor rke2 clusters by adding support for `rke2ControllerManager`, `rke2Scheduler`, `rke2Proxy`, and `rke2Etcd`
issue :https://github.com/rancher/rke2/issues/33
canal and nginx will be supported in a different chart.


cjellick:
Why is there so much stuff changing in this PR that is unrelated to your exact change? We should strive to make these patch PRs as small as possible

MonzElmasry:
@cjellick looks like blocks of code are moving around in the patch file when i make a patch, these changes are generated

aiyengar2:
> @cjellick looks like blocks of code are moving around in the patch file when i make a patch, these changes are generated

Hmm this is weird, I haven't experienced code blocks moving around when trying to generate changes on the patch before. Let's take a look at this when we meet later today @MonzElmasry

StrongMonkey:
I don't have to a good way to fix this right now. It appears that changes are introduced because of the diff on GNU and BSD produces different output. By taking close look at changes, it is swapping the positions so nothing critically gets changed.

MonzElmasry:
https://github.com/rancher/charts/compare/dev-v2.5...MonzElmasry:new-something
generated changes on make patch

packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -1657,6 +1668,60 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-monitoring/charts-original/val
...
++      - effect: "NoSchedule"
++        operator: "Exists"
++
++rke2Proxy:
++  enabled: false
++  metricsPort: 10249
++  component: kube-proxy
++  clients:
++    port: 10013
++    useLocalhost: true
```
cbron, line 1710: Do we not need the tolerations settings here ?

MonzElmasry: In my testing it didnt need any, but added anyway just in case

cbron: @brandond or @cjellick do we know which parts we do or don't want tolerations on ?

brandond: Are we talking about the bootstrap flag on the HelmChart that adds tolerations? If so, I think we should only put those on things that need to run as part of cluster bootstrap. Other stuff like monitoring can wait until after the core functionality is up.

brandond: Oh, I see. The these are just to let monitoring run on masters or nodes that have been tainted NoExecute/NoSchedule? Seems fine to me.

cbron: Ok, if we are good with that. I was more concerned about users being able to edit this setting via values.

cjellick:
🚢 🇮🇹

PR #546, opened 2020-08-14
Add upstream Longhorn chart

shuo-wu:
This PR is from https://github.com/rancher/charts/pull/526
Issue: https://github.com/rancher/rancher/issues/27725


yasker:
@shuo-wu Can you update this PR so it can be merged? Also, why we're removing the `labels`?

shuo-wu:
According to [this doc](https://rancher.atlassian.net/wiki/spaces/EN/pages/563609933/Helm+Charts+Requirements#Questions.yml-Deprecation), the field `Label` will be deprecated from `Questions.yml`. But I am not sure what `converted to annotations on the chart` is.

And how can I test this PR? by using Rancher `v2.5.0-alpha1`?

maggieliu:
@brendarearden Can you provide @shuo-wu with the steps needed to test his chart? I will also start a slack group chat if more details are needed.

packages/longhorn/package.yaml
```diff
@@ -0,0 +1,5 @@
+url: https://github.com/longhorn/charts.git
+subdirectory: charts/longhorn
+type: git
+commit: 0a88eaae4d7bdce762af942ea490d72aee338bb1
+packageVersion: 00
```
aiyengar2, line 5: @cbron @vincent99 does Longhorn need to split into CRD charts? In https://github.com/rancher/charts/pull/526#issuecomment-673114537 we decided to not split it into a separate CRD chart, but just wanted to confirm that that's still the case.

aiyengar2 (changes requested):
Need to drop the last commit from this PR since that's automatically done by Git hooks on merge.

packages/longhorn/package.yaml
cbron: We do need to split it into CRD chart, as we now don't want to remove CRDs on chart uninstall. That was new development from last week. So if you uninstall the longhorn chart it removes everything but CRDs/CRs, and you also have to uninstall the longhorn-crd chart to remove the CRDs/CRs.

aiyengar2: Hmm so @shuo-wu you'll need to add the following to the `package.yaml`:
```yaml
generateCRDChart:
  enabled: true
```

You'll also need to split the [`crds.yaml`](https://github.com/longhorn/charts/blob/master/charts/longhorn/templates/crds.yaml) file into separate files per CRD via the patch and relocate those files to the `crd/` directory for the scripts to pick it up correctly.

Unfortunately, support for 1) pointing to a combined CRD file and 2) a set of CRDs outside the `crd/` directory is not supported but tracked in https://github.com/rancher/rancher/issues/28320.

shuo-wu: I am a little bit confused about the longhorn-crd chart uninstallation. Right now Longhorn uses a separate job to clean up the workloads as well as all CRs. And this uninstall job is launched during the longhorn chart uninstallation. If we split the longhorn crd/cr chart uninstallation from the longhorn chart uninstallation, how can we guarantee all of the CRs can be removed during the longhorn crd chart uninstallation? By launching the job again?

cbron: @shuo-wu is that uninstall job just part of this helm chart or is it wider job thats always used with longhorn ? 

Ideally all charts in the rancher/charts repo follows the same pattern so we can make assumptions based on that. But if this uninstall script is integral to how longhorn works then we might have to make an exception. The typical behavior is that we can uninstall the main chart which leaves behind the CRs and CRDs, and leave it up to user to delete those by uninstalling the CRD chart.

shuo-wu: @cbron I am not sure what do you mean by `wider job`

This job is responsible for:
  1. Delete the workloads dynamically deployed by Longhorn system. (instead of YAML files)
  2. Cleaning up CRs.

If I remove the uninstaller job from the main/feature chart, those dynamically deployed cannot be removed. And deleting the crd chart somehow cannot clean up all CRDs (may due to the finalizers of the related CRs)
If I retain the uninstaller job in the main chart, the uninstaller will get stuck there when it tries to handle CRs.

Right now we need to either modify the Longhorn uninstaller (to make it clean up extra workloads only) or make an exception for the charts like Longhorn.

packages/longhorn/longhorn.patch
```diff
@@ -0,0 +1,366 @@
+diff -x '*.tgz' -x '*.lock' -uNr packages/longhorn/charts-original/Chart.yaml packages/longhorn/charts/Chart.yaml
```
cbron, line 1: The binary files and changes in yaml files above should not be added, those will be added by the CI process.

packages/longhorn/package.yaml
```diff
@@ -0,0 +1,8 @@
...
+generateCRDChart:
+  enabled: true
+  providesGVR: longhorn.io/v1beta1
```
cbron, line 8: providesGVR was removed from the package.yaml, please add to chart patch annotations above as `catalog.cattle.io/provides-gvr`

packages/longhorn/package.yaml
cbron: Sounds like we need to lean towards making an exception, since everyone with a current longhorn chart installed would be affected. I will ask.

cbron (changes requested):

packages/longhorn/longhorn.patch
cbron: Now that I look at this, there are quite a few things that need to be done in order to have this qualify in the dashboard UI. I will send you a link of how to implement, but things that are missing are: 

* add custom private registry settings
* re-format image repository/tag so they get picked up for airgap
* add RBAC ClusterRoles for aggregating to default k8s clusterRoles

maggieliu:
related to: https://github.com/rancher/rancher/issues/27725

packages/longhorn/package.yaml
cbron: We talked about this in the architecture meeting and would like to split the CRDs into 2nd chart, and also move the uninstall script to the CRD chart.

packages/longhorn/longhorn.patch
shuo-wu: 1. What do you mean by `The binary files and changes in yaml files above should not be added, those will be added by the CI process.`?
2. I still have no idea about how to modify the RBAC part after reading the doc. Is there an example available?

packages/longhorn/package.yaml
shuo-wu: Do you mean creating a separate CRD chart manually? Then how about deleting the workloads dynamically deployed by Longhorn system?

aiyengar2: @shuo-wu yes, I believe @cbron is referring to taking an approach similar to [`fleet`](https://github.com/rancher/fleet/tree/master/charts), which has both a `fleet` chart and a `fleet-crd` chart. 

Ideally, the behavior of the CRD chart should be to be able to have sole control over 1) installing CRDs 2) upgrading CRDs 3) removing CRDs. The main chart should not be able to do anything CRD-related, but should still be able to do anything CR / workload related (i.e. creating or cleaning up CRs and workloads, which is part of what the current uninstall script does).

w.r.t. which of the two charts should be responsible for deleting the workloads dynamically, is there a way for the uninstaller chart to be given a flag to only delete the workloads and not clean up the CRDs?

If this is possible, the main chart should be patched to pass in this flag to the uninstall job so that it no longer is doing anything CRD related and you can likely use the logic described [here](https://github.com/rancher/charts/pull/546#discussion_r489052974) in order to modify the longhorn chart to be able to be split into 2 charts.

If this is not possible, I think the best move would be to move the uninstall script to the CRD chart as Caleb recommended. This does mean that the CRD chart for Longhorn specifically will have one additional responsibility not found in any other CRD charts (i.e. it also cleans up workloads and CRs in addition to the CRDs on uninstall), but I think that was called out in the architecture meeting based on [Caleb's comment](https://github.com/rancher/charts/pull/546#discussion_r491224165).

packages/longhorn/longhorn.patch
aiyengar2: 1. He's referring to all the changes in `index.yaml` and `assets/*` that have been included in this PR; all those files changes should be removed. This PR should only be committing the `longhorn.patch` and the `package.yaml` since the remaining files are automatically generated by CI on push.
2. This just involves adding a label to your `ClusterRoles` in order to aggregate them into the default cluster roles, mentioned in the [k8s docs](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles). The [labels](https://github.com/rancher/charts/blob/f53ab23985965dd79b6d09d5771d1dd23de88716/packages/rancher-logging/overlay/templates/userroles.yaml#L6) added to the `ClusterRoles` created by the `rancher-logging` chart are a great example of this.

cbron: Part 2 we should talk about in a design discussion. The goal is that a user with an existing clusterRole of edit in the installation namespace would get whatever privileges are relevant for them.

shuo-wu: The changes/conflicts in `index.yaml` and `assets/*` are introduced by the automatically generated commit. I didn't touch that part.

packages/longhorn/package.yaml
shuo-wu: The Longhorn uninstaller is responsible for removing the CRs (objects) rather than CRDs. If the CRs can be cleaned up during the main chart uninstallation, the Longhorn uninstaller should be fine.

packages/longhorn/longhorn.patch
shuo-wu: I just add labels in the roles in the chart for the ClusterRole `admin` and `edit` aggregation, add launch an extra role in `/overlay` for the ClusterRole `view` aggregation. Let me know if there is anything not right.

cbron: The generated changes are happening because this is on branch named dev-v2.5, we should fix that on our side but for now can you change the branch name.

packages/longhorn/overlay/templates/userroles.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: ClusterRole
+metadata:
+  name: "longhorn-view"
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+rules:
+  - apiGroups: [ "longhorn.io" ]
+    resources: [ "volumes", "volumes/status", "engines", "engines/status", "replicas", "replicas/status", "settings",
+                 "engineimages", "engineimages/status", "nodes", "nodes/status", "instancemanagers", "instancemanagers/status" ]
+    verbs: [ "get", "list", "watch" ]
```
cbron, line 11: Can we also get the edit and admin roles added to this same file. This view one looks good. The only real difference between edit and admin is that admin can modify roles/role-bindings, so since this chart doesn't need to do that they should be the same. 

Just as an overview, the goal here is to give a user with the relevant ClusterRole like `edit` all the edit permissions they would need to configure the resources around the chart. And that happens automatically via the k8s ClusterRole aggregation through the label.

packages/longhorn/package.yaml
cbron: We talked about this with Sheng and since the uninstaller has to run while the operator is still installed in the cluster, we can't move it to the CRD chart or it would have no effect. So we'll have to leave in the primary chart. But we do still want to split the CRD chart out, and that chart would only control the CRDs not any CRs.

packages/longhorn/overlay/templates/userroles.yaml
cbron: Oh, I see your other comment that you added the admin/edit in original ClusterRole. We'd prefer to have these clusterRoles only include what is necessary for the user to interact with the chart/longhorn. Not everything needed for the operator to run. So ideally they only include the longhorn CRDs.

packages/longhorn/package.yaml
shuo-wu: Sure, then it seems that I need to create a new PR with a new PR name.

What's the difference between manually splitting the CRD chart and using `catalog.cattle.io/provides-gvr: longhorn.io/v1beta1` generate the hidden CRD chart?

packages/longhorn/overlay/templates/userroles.yaml
shuo-wu: How about some specifical k8s defined resources like `coordination.k8s.io/lease`? It's not included  in those ClusterRole by default.

packages/longhorn/package.yaml
cbron: If the CRD's were in the CRD folder you can split them automatically with package.yaml using `generateCRDChart:  enabled: true` and that will handle all annotations. You can move them with a patch to the CRD directory, the patch is applied in `make prepare` before the CRD split so that should work. 

You can also split them manually. On the CRD chart you'll need to provide all the primary required annotations and the `hidden` annotation, and on the primary chart the `auto-install: ${original_chart_name}-crd=match`.

`provides-gvr` should be set manually on the primary chart either way, it doesn't relate to the crd chart, but instead allows another chart that needs longhorn to function to require it via the UI. Thats unlikely right now but its a standard we like to set.

packages/longhorn/overlay/templates/userroles.yaml
cbron: Would the user need to configure the lease to configure longhorn ? Or does longhorn take care of the lease for them ? 

If they don't need access to it then leave it out. If they do need access to it then we should talk about it. We have another case where we created a separate Role (not a ClusterRole) to specify non-chart resources. That way users don't accidentally assign out ClusterRole with ClusterRoleBinding and give cluster wide access to basic resources.

shuo-wu: Longhorn itself will need to configure the lease during the launching. Then it seems that I don't need to create one more Role for non-chart resources.

shuo-wu:
Due to the PR naming that will trigger the commit auto-generating, we will track the issue in https://github.com/rancher/charts/pull/692 instead.

PR #548, opened 2020-08-14
Create chart for backup-restore-operator

mrajashree:
This commit creates a chart for the [backup-restore-operator](https://github.com/rancher/backup-restore-operator)
It contains the serviceaccount for the operator and clusterrolebinding binding it to cluster-admin role.
It also contains the manifests for the three CRDs: Backup, Restore and ResourceSet
Note: The serviceaccount gets cluster-admin role because this chart's controllers need permissions to CRUD any resource. Also the controller can get, upload and delete backups to s3, so only the cluster-admin role should be able to do this

https://github.com/rancher/rancher/issues/28368


mrajashree:
Closing in favor of https://github.com/rancher/charts/pull/589

PR #554, opened 2020-08-17
added image overrides

paynejacob:


packages/rancher-logging/overlay/templates/outputs/syslog/deployment.yaml
```diff
@@ -17,7 +17,7 @@ spec:
     spec:
       containers:
       - name: fluentbit
-        image: paynejacob/fluent-bit-out-syslog:latest
+        image: "{{ .Values.images.syslog_forwarder.repository }}:{{ .Values.images.syslog_forwarder.tag }}"
```
cbron, line 20: can you add the registry helper here

packages/rancher-logging/overlay/templates/logging-rke1.yaml
```diff
@@ -8,6 +8,9 @@ metadata:
 spec:
   controlNamespace: {{ .Release.Namespace }}
   fluentbit:
+    image:
+      repository: {{ .Values.images.fluentbit.repository }}
```
cbron, line 12: Since this is in a template, we can just prepend registry helper to the repository, otherwise we have to update the banzai template to do it. And then follow that pattern everywhere else.

PR #555, opened 2020-08-17
Update commit example of overriding a chart

StrongMonkey:


StrongMonkey:
Updated the 404 link to an actual example.

PR #557, opened 2020-08-18
Change to rancher images

brendarearden:
Updated images to use rancher images

Issue:
https://github.com/rancher/rancher/issues/27943


packages/rancher-istio/charts/values.yaml
```diff
@@ -61,18 +77,15 @@ meshConfig:
 kiali:
   enabled: true
   auth:
-    # todo: what auth strategy ?
-    # strategy: token # this is default, you can use a service-account or kubeconfig token
     strategy: anonymous
   deployment:
+    ##TODO: update kiali to allow for repository and image
     ingress_enabled: false
+    image_name: docker.io/rancher/kiali-kiali
```
aiyengar2, line 84: Is there a reason why the Kiali image does not use the global hub value?

packages/rancher-istio/charts/values.yaml
```diff
@@ -61,18 +77,15 @@ meshConfig:
...
+    image_version: v1.22.1
   external_services:
     prometheus:
       custom_metrics_url: "http://rancher-monitoring-prometheus.monitoring-system.svc:9090"
       url: "http://rancher-monitoring-prometheus.monitoring-system.svc:9090"
```
aiyengar2, line 89: I don't think I caught this in a previous PR, but these URLs need to be updated based on the recent `rancher-monitoring` namespace changes:
```suggestion
      custom_metrics_url: "http://rancher-monitoring-prometheus.cattle-monitoring-system.svc:9090"
      url: "http://rancher-monitoring-prometheus.cattle-monitoring-system.svc:9090"
```
Not sure if this can be addressed in this PR or if we want to address this in a separate PR. It also might require changes to the `README.md` iirc for Istio.

packages/rancher-istio/charts/values.yaml
brendarearden: I will be making all the kiali updates in a separate pr - since its a dependent chart I will have to patch the rancher-kiali-servcer chart to use the global hub value.

packages/rancher-istio/charts/values.yaml
brendarearden: I went ahead and added it to this pr, we currently have it in the readme as ```http://{{ .Values.nameOverride }}-prometheus.{{ .Values.namespaceOverride }}.svc:{{ prometheus.service.port }}``` so I thiiink we should be okay

brendarearden:
changes since approvals:  get changes from upstream dev-v2.5 and deleted 1.6.8tgz so these changes will be available in the ui

packages/rancher-istio/charts/configs/istio-base.yaml
```diff
@@ -52,7 +52,7 @@ spec:
       enabled: {{ .Values.policy.enabled }}
     telemetry:
       enabled: {{ .Values.telemetry.enabled }}
-  hub: docker.io/istio
+  hub: {{ .Values.global.hub | default "docker.io/istio" }}
```
cbron, line 55: Values.global.hub should be the system_default_registry helper ? and default to docker.io/rancher

packages/rancher-istio/charts/configs/istio-base.yaml
```diff
@@ -67,25 +67,31 @@ spec:
         name: istio-ingressgateway
         type: {{ .Values.ingressGateways.type }}
     global:
+      hub: {{ .Values.global.hub | default "docker.io/istio" }}
```
cbron, line 70: same

packages/rancher-istio/charts/configs/istio-base.yaml
```diff
@@ -67,25 +67,31 @@ spec:
...
       proxy:
-        image: {{ .Values.global.proxy.image }}
+        image: {{ template "system_default_registry" . }}{{ .Values.global.proxy.repository }}:{{ .Values.tag }}
       proxy_init:
-        image: {{ .Values.global.proxy_init.image}}
+        image: {{ template "system_default_registry" . }}{{ .Values.global.proxy_init.repository}}:{{ .Values.tag }}
     istiocoredns:
-      coreDNSImage: {{ .Values.istiocoredns.image.repository }}
-      coreDNSPluginImage: {{ .Values.istiocoredns.pluginImage }}
+      coreDNSImage: {{ template "system_default_registry" . }}{{ .Values.istiocoredns.image.repository }}
```
cbron, line 78: needs tag

brendarearden: coredns is set like this in the manifest: 
```
image: {{ .Values.istiocoredns.coreDNSImage }}:{{ .Values.istiocoredns.coreDNSTag }}
```
since it doesn't check for the slash like pilot does, for example, I can't provide `<hub>/<registy>:<tag>`  as the value for the image. The tag is set separately on line 80 `coreDNSTag: {{ .Values.istiocoredns.image.tag }}`
```
{{- if contains "/" .Values.pilot.image }}
          image: "{{ .Values.pilot.image }}"
{{- else }}
          image: "{{ .Values.pilot.hub | default .Values.global.hub }}/{{ .Values.pilot.image | default "pilot" }}:{{ .Values.pilot.tag | default .Values.global.tag }}"
{{- end }}
```

packages/rancher-istio/charts/configs/istio-base.yaml
brendarearden: You are correct, I will update to use the registry helper. I guess I am not convinced on what the default should be - does it matter if they want to use the chart outside of rancher and it still point to our repo? I guess they could always change it with an overlay file of their own 🤔

cbron: Because of our setup it will always use our orgs images, they likely would never specify a registry so that would just be empty string

packages/rancher-istio/charts/configs/istio-base.yaml
cbron: Ahh, gotcha, as long as it’s specified in the values it should be fine

packages/rancher-istio/charts/configs/istio-base.yaml
brendarearden: instead of using the `default` templating option, I updated the helper method to set the default value instead of being empty that way all the logic is in one spot

cbron: Hmm, I suppose thats ok. Could see this having weird consequences but we can try it out.

packages/rancher-istio/charts/values.yaml
cbron: Yes you'll need to add the registry helper to kiali, and that to the image spec, and it should flow through from global var here overriding kiali global var. So you wouldn't specify hub at this point, but we can leave for now.

packages/rancher-istio/charts/configs/istio-base.yaml
cbron: Actually that won't work because this hub value is supposed to have both registry and org in it. Our helper should only do the registry, otherwise you'd get: `docker.io/rancher/rancher/kiali-kiali`

brendarearden: The way hub is used throughout the manifests is `{{ .Values.global.hub }}{{ .Values.pilot.image }}:{{ .Values.tag }}` which gives us the flexibility to have 1. hub be `<registry>` and repository be `<org/image>` or 2. hub be <registry/org> and repository be `<image>` so it might come down to what is easier for us to traverse in the values.yaml for the image.txt file. 

Right now we are doing 1. so it feels weird to have hub be `<registry>` but then default to `<registry/org>` because that will also result in `docker.io/rancher/rancher/kiali-kiali` if registry is not set. If we stick with 1. I will default to `docker.io` and I will set that in the registry helper, since it is just setting the registry

PR #569, opened 2020-08-20
Remove canary and set version to 1.6.7 for upgrade testing

brendarearden:
Needed to create a an older version of istio with recent changes (name change, remove canary, manifest changes) for upgrade testing


brendarearden:
They are already removed per this pr https://github.com/rancher/charts/pull/557/files and with the rename from rancher-istio-installer to rancher-istio, you will see the rancher-istio-installer for the deleted version

PR #574, opened 2020-08-21
Istio: Add default pod disruption budget 

brendarearden:
enabled the default pod disruption budget
https://github.com/rancher/rancher/issues/27943


brendarearden:
I lost tag version changes in the rebase and kept the 1.6.7 instead of 1.6.8 changes. I have corrected this and all tags should be 1.6.8

packages/rancher-istio/charts/configs/istio-base.yaml
```diff
@@ -72,6 +72,10 @@ spec:
         image: {{ template "system_default_registry" . }}{{ .Values.global.proxy.repository }}:{{ .Values.global.proxy.tag }}
       proxy_init:
         image: {{ template "system_default_registry" . }}{{ .Values.global.proxy_init.repository}}:{{ .Values.global.proxy_init.tag }}
+      {{- if .Values.global.defaultPodDisruptionBudget.enabled }}
+      defaultPodDisruptionBudget:
+        enabled: {{ .Values.global.defaultPodDisruptionBudget.enabled }}
```
cbron, line 77: I thought this was a min/max setting, what exactly does this do ?

brendarearden:
```
  # enable pod disruption budget for the control plane, which is used to
  # ensure Istio control plane components are gradually upgraded or recovered.
  defaultPodDisruptionBudget:
    enabled: true
    # The values aren't mutable due to a current PodDisruptionBudget limitation
    # minAvailable: 1
```

which enabled this https://github.com/istio/istio/blob/master/manifests/charts/gateways/istio-ingress/templates/poddisruptionbudget.yaml

PR #576, opened 2020-08-24
Update default images on monitoring charts to Rancher mirrors

aiyengar2:
- [x] Dependent on merging in https://github.com/rancher/image-mirror/pull/22 first in order to create the images.

Issue: https://github.com/rancher/rancher/issues/28441


packages/rancher-pushprox/charts/values.yaml
```diff
@@ -60,6 +63,10 @@ clients:
     tag: v0.1.0-rancher1-client
   command: ["pushprox-client"]
 
+  copyCertsImage:
+    repository: rancher/rke-tools
```
aiyengar2, line 67: Is there a better Rancher image to use for this initContainer? All it does is run a simple Bash script does extremely basic operations (`echo`, `find`, `test`, `cp` and `chmod`) that can be performed by any lightweight image.

packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -1332,6 +1426,11 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-monitoring/charts-original/val
 +    # Change this if you change the namespaceOverride or nameOverride of prometheus-operator
 +    url: http://rancher-monitoring-prometheus.cattle-monitoring-system.svc
 +    port: 9090
++  image:
++    repository: directxman12/k8s-prometheus-adapter-amd64
```
brendarearden, line 1430: and this one is going to be updated in the next pr that you mentioned in the issue?

aiyengar2: Yeah, this will be updated in the next PR based on the [comment](https://github.com/rancher/rancher/issues/28441#issuecomment-678939966) on the issue as you mentioned!

cbron: this will later be updated to `rancher/directxman12-k8s-prometheus-adapter-amd64` ?

packages/rancher-pushprox/charts/values.yaml
cbron: I'd say rke-tools is too heavy-duty, and might have stuff on there we don't want in cluster from security perspective. We prepend busybox to our images.txt list, but I wonder if that will fail the list generation because its not in the rancher org. We have some checks around that. `rancher/kubectl` should work I think. But I like that idea that longterm we have a box that has those tools plus some networking ones like curl/wget/dig etc..

packages/rancher-monitoring/rancher-monitoring.patch
aiyengar2: Yes, in this PR: https://github.com/rancher/charts/pull/582/commits/7752a3978ad3d86b5eaf17ca10081e520e087675#diff-ed5b5b87cf149adc6a6408bf81993efdR1703

packages/rancher-pushprox/charts/values.yaml
aiyengar2: Changed to `rancher/kubectl` in the latest commit!

aiyengar2:
Double checked that the images exist, removed the tgz files, and rebased to dev-v2.5. PR should be ready to merge

PR #579, opened 2020-08-24
Allow patches on remote deps with fixed versions

aiyengar2:
This commit adds the ability to add patches on subcharts in `rancher/charts` that meet the following constraints:
- Must be a remote chart (i.e. the repo url cannot point to a `file://<path>`)
- Must have a fixed version number (since we need a consistent base to apply the patch on)

The following changes have been made to the scripts to faciliate this feature:

**prepare**

Only run the patch after preparing the subchart

**prepare-subcharts:**

Inital checks:
1. We only prepare the subcharts if the chart has a `requirements.yaml`.
2. Extract the patches that would be applied to the requirements.yaml and only apply that

Note: in the future, we need to add support for dependencies outlined in the Chart.yaml directory, which is currently the recommended approach for Helm 3 charts that use `apiVersion: v2`, but since none of our charts currently use that it is assumed that they will encode their dependencies in a `requirements.yaml` at the moment.

Generate special dependencies (overlay, package.yaml):
1. For dependencies that are mirrored (i.e. rancher-kiali-server, where the dependency itself is stored as a `package.yaml`) that have not generated a `charts/` directory yet, we recursively run the `prepare` script for them in order to generate the directories so that `helm dependency update` can pull them in. The assumption here is that there will be no circular dependencies, although there is no explicit check for this.
2. For dependencies that are overlaid, we perform the overlay so that `helm dependency update` operates as expected.

Pull in latest subcharts and update files:
1. Perform `helm dependency update` to update the `requirements.yaml` and `requirements.lock` with the latest subchart versions

Cleanup (overlay, package.yaml, patched requirements)
1. Remove overlaid dependencies since they won't be tracked in the patch
2. Run the `clean` script on the mirrored dependencies
3. Revert the patch on the requirements.yaml

Apply logic to enable / disable patching:
1. If a subchart is a local chart, leave it as a tgz so it will be ignored on patch
2. If a subchart does not have a fixed version number, leave it as a tgz so it will be ignored on patch
3. If neither of these are true, unarchive the tgz for the subchart and delete the tgz so it will be tracked by the patch

NOTE: since the dependency still exists as either a tgz or a directory, `generate-charts` will not break from this change.

**generate-patch:**

Loop through the local `charts/requirements.yaml` to update the `charts-original/charts` directory:
1. If a subchart is not a local chart, remove it as we don't need to check for patches against it
2. If a subchart does not have a fixed version number,  remove it as we don't need to check for patches against it
3. If neither of these are true, pull in the version number specified from the remote helm repo to track patches against it.

**Why does the version number need to be fixed?**:

In general, Helm chart owners can use wildcards within a `requirements.yaml` to specify the version of a subchart when a specific patch version does not necessarily need to be fixed. This is usually a great feature since that means that parent charts do not need to be updated in order for a `helm dependency update` to automatically pull in the latest versions of a dependency that do not break the parent chart; however, this causes an issue with patching subcharts since the base that the patch is applied on is not fixed.

Therefore, this commit only allows users to apply patches on subcharts that have a fixed subchart version specified in their `requirements.yaml`.

**Why can't we track local charts?**

If the chart is local to this repo, the assumption is tha

### DIFF
--- a/scripts/generate-patch
+++ b/scripts/generate-patch
@@ -58,16 +58,41 @@ for f in packages/*; do
 				mkdir -p ${f}/charts-original
 				curl -sLf ${url} | tar xvzf - -C ${f}/charts-original --strip ${fields} ${subdirectory} > /dev/null 2>&1
 			fi
+			if [[ -f ${f}/charts/requirements.yaml ]]; then
+				i=0
+				while ! [[ -z $(yq r ${f}/charts/requirements.yaml "dependencies[${i}]") ]]; do
+					subchart=$(yq r ${f}/charts/requirements.yaml "dependencies[${i}].name" | head -n 1)
+					version=$(yq r ${f}/charts/requirements.yaml "dependencies[${i}].version" | head -n 1)
+					helm_repo=$(yq r ${f}/charts/requirements.yaml "dependencies[${i}].repository" | head -n 1)
+					if [[ ${helm_repo} == file://* ]] || [[ ${version} == *'*'* ]]; then
+						# Subchart is tracked locally; patches will not be tracked
+						rm -rf ${f}/charts-original/charts/${subchart}
+						((i=i+1))
+						continue
+					fi
+					# Subchart is tracked in remote Helm repo with a fixed version number; we grab the download URL and get the local chart archive
+					echo "Comparing current changes to ${subchart}-${version} against upstream found at ${helm_repo}"
+					download_url=$(curl -sLf ${helm_repo}index.yaml | cat | yq r - "entries.${subchart}.(version == ${version}).urls[0]")
+					curl -sLf ${download_url} | tar xvzf - -C ${f}/charts-original/charts > /dev/null 2>&1
+					((i=i+1))
+				done
+			fi
 			if [[ -d ${f}/charts ]]; then
 				split_crds=$(cat ${f}/package.yaml | yq r - generateCRDChart.enabled)
 				if [[ "${split_crds}" == "true" ]]; then
 					revert_crd_changes ${f}
 				fi
 				diff -x *.tgz -x *.lock -uNr ${f}/charts-original ${f}/charts > ${f}/$(basename -- ${f}).patch || true
+				if ! [[ -s ${f}/$(basename -- ${f}).patch ]]; then
+					# If no changes exist in patch then remove it
+					rm ${f}/$(basename -- ${f}).patch
+				fi
+				if [[ -f ${f}/$(basename -- ${f}).patch ]]; then
+					remove_timestamp_from_diff
+				fi
 				if [[ "${split_crds}" == "true" ]]; then
 					./scripts/prepare-crds ${f}
 				fi
-				remove_timestamp_from_diff
 			fi
 			rm -rf ${f}/charts-original
 		fi
--- a/scripts/prepare-subcharts
+++ b/scripts/prepare-subcharts
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+set -e
+
+if [[ -z $1 ]]; then
+	echo "No directory provided to pull in subcharts"
+	exit 1
+fi
+
+f=$1
+
+copied_dependencies=()
+if [[ -f ${f}/charts/requirements.yaml ]]; then
+	repos=$(cat ${f}/charts/requirements.yaml | yq r - "dependencies.*.repository")
+	while IFS= read -r repo; do
+		if [[ ${repo} == file://* ]]; then
+			charts_path=${repo/file:\/\//${f}/charts/}
+			overlay_path=${repo/file:\/\//${f}/overlay/}
+			if [[ ! -d $charts_path ]] && [[ -d $overlay_path ]]; then
+				cp -R $overlay_path $charts_path
+				copied_dependencies+=("${charts_path}")
+			fi
+		fi
+	done < <(echo "${repos}")
+fi
+
+pwd=$(pwd)
+cd ${f}/charts
+helm dependency update
+cd $pwd
+
+for dep in "${copied_dependencies[@]}"; do
+	rm -rf $dep
+done
--- a/scripts/prepare-subcharts
+++ b/scripts/prepare-subcharts
@@ -1,6 +1,22 @@
 #!/usr/bin/env bash
 set -e
 
+# Pull in the subcharts required by a chart
+#
+# At the end of this process, all the required subcharts will exist in the charts/charts directory.
+# Charts will either exist as chart archives (i.e. a .tgz file) if they are not eligible for patches or
+# as a local chart archive (i.e. a directory) if they are eligible for patches (i.e. remote subchart with fixed version number)
+#
+# In order to prepare sub-charts so that we can patch them:
+# - We apply any patches on the requirements.yaml to track any new subcharts added via patch or updates to version numbers of the subcharts
+# - We pull in any subchart that is a feature chart by running prepare on it
+# - We copy in any local subcharts that are located in the overlay/ directory
+# - We call `helm dependency update` to generate chart archives (.tgz files) for local and remote charts
+# - We clean up any copied charts from the overlay/ directory
+# - We clean up any feature charts that were recursively prepared
+# - We unarchive the chart archive if and only if its a remote subchart with a fixed version number (which means it can be patched)
+# - We revert the initial patch we applied on the requirements.yaml
+
 if [[ -z $1 ]]; then
 	echo "No directory provided to pull in subcharts"
 	exit 1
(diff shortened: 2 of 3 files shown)


aiyengar2:
@ibuildthecloud re: your previous comments on the complexity of the bash scripts in another PR, I think we should switch the go based scripts soon

aiyengar2:
Using a single loop in the `prepare` script fixed the build issue that we encountered before the most recent push to this PR.

#### Why make the switch?

We initially introduced multiple for loops was to handle a special case in adding support for `rancher-kiali-server` as a dependency of `rancher-istio` as part of https://github.com/rancher/charts/pull/532

The previous version of the `prepare script` ran steps to `create` -> `apply patch` -> `pull subcharts` -> `prepare crds`. To be able to support `rancher-istio` and similar charts like it, we needed to be able to run `create` + `apply patch` on all charts first so that `pull subcharts` did not fail.

However, in this PR, we switch the order to `create` -> `pull subcharts` -> `apply patch` -> `prepare crds` and, during `pull subcharts`, recursively call prepare on each of the charts. Therefore, the new process looks something like:

`rancher-istio create`
`rancher-istio pull-subcharts`

- `rancher-kiali-server create`
- `rancher-kiali-server pull-subcharts`
- `rancher-kiali-server apply patch`
- `rancher-kiali-server prepare crds`

`rancher-istio apply patch`
`rancher-istio prepare crds`

So now, there's no longer a need to have multiple for loops to run certain steps for all charts before running the steps for an individual chart.

scripts/prepare-subcharts
```diff
@@ -0,0 +1,127 @@
...
+		fi
+		if [[ -f ${charts_path%/*}/package.yaml ]]; then
+			# Prepare the subchart and pull it in
+			subchart=$(basename -- ${charts_path%/*})
+			mirrored_dependencies+=("${subchart}")
+			prepare_subchart () {
+				local CURRENT_CHART=${CHART}
+				local CURRENT_INDEX=${i}
+				CHART=${subchart}
+				./scripts/prepare
```
aiyengar2, line 47:
```suggestion
				CHART=${subchart} ./scripts/prepare
```

scripts/prepare-subcharts
```diff
@@ -0,0 +1,127 @@
...
+for dep in "${overlaid_dependencies[@]}"; do
+	rm -rf $dep
+done
+
+# Cleanup any dependencies that came from mirrored package.yamls
+for subchart in "${mirrored_dependencies[@]}"; do
+	cleanup_subchart () {
+		local CURRENT_CHART=${CHART}
+		CHART=${subchart}
+		./scripts/clean
```
aiyengar2, line 79:
```suggestion
		CHART=${subchart} ./scripts/clean
```

aiyengar2:
https://github.com/rancher/charts/pull/582 demonstrates successful patches using this process on subcharts

scripts/prepare-subcharts
```diff
@@ -0,0 +1,125 @@
...
+	((i=i+1))
+done
+
+# Update the current requirements.lock and requirements.yaml to keep track of the overlay/ dependencies
+pwd=$(pwd)
+cd ${f}/charts
+helm dependency update
+cd $pwd
+
+# Remove any dependencies that came from the overlay/ directory
```
cbron, line 68:
```suggestion
# Remove any dependencies that came from the overlay/ directory, we don't track patches for overlay files
```

scripts/prepare-subcharts
```diff
@@ -0,0 +1,125 @@
...
+for dep in "${overlaid_dependencies[@]}"; do
+	rm -rf $dep
+done
+
+# Cleanup any dependencies that came from mirrored package.yamls
```
cbron, line 73: This is only for local charts that have package.yaml right ? 

```suggestion
# Cleanup any dependencies that came from another local chart, any changes should be made there. 
```

scripts/prepare-subcharts
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/env bash
+set -e
+
+# Pull in the subcharts required by a chart and apply any patches on subcharts with fixed version numbers
```
cbron, line 4: I think the flow and goal of this script is somewhat confusing. Can you add a high level overview of how it works. Something like: 

```
# In order to prepare sub-charts so that we can patch them: 
#  - We pull in any sub-chart that is a feature chart by running prepare on it
#  - We then call `helm dependency update` to pull in any other sub-chart
#  - And last we clean up any sub-chart that shouldn't have a patch applied to it.
```

scripts/prepare-subcharts
```diff
@@ -0,0 +1,125 @@
...
+		rm ${f}/charts/charts/${subchart_tgz}
+	fi
+	((i=i+1))
+done
+
+# Undo the patch on the requirements.yaml
+if [[ -f ${f}/requirements.patch ]] && [[ -s ${f}/requirements.patch ]]; then
+	patch -R -E -p3 -d ${f}/charts < ${f}/requirements.patch
+	rm ${f}/requirements.patch
+fi
```
cbron, line 125: Everything in the sub-chart folder already gets cleaned up by `make clean` right ?

scripts/generate-patch
```diff
@@ -58,16 +58,41 @@ for f in packages/*; do
...
+				done
+			fi
 			if [[ -d ${f}/charts ]]; then
 				split_crds=$(cat ${f}/package.yaml | yq r - generateCRDChart.enabled)
 				if [[ "${split_crds}" == "true" ]]; then
 					revert_crd_changes ${f}
 				fi
 				diff -x *.tgz -x *.lock -uNr ${f}/charts-original ${f}/charts > ${f}/$(basename -- ${f}).patch || true
+				if ! [[ -s ${f}/$(basename -- ${f}).patch ]]; then
+					# File is empty; skip creating a patch for it
```
cbron, line 87:
```suggestion
					# If no changes exist in patch then remove it
```

scripts/generate-patch
```diff
@@ -58,16 +58,41 @@ for f in packages/*; do
...
+					rm ${f}/$(basename -- ${f}).patch
+				fi
 				if [[ "${split_crds}" == "true" ]]; then
 					./scripts/prepare-crds ${f}
 				fi
-				remove_timestamp_from_diff
+				if [[ -f ${f}/$(basename -- ${f}).patch ]]; then
+					remove_timestamp_from_diff
+				fi
```
cbron, line 95: Why is this part after the prepare-crds part ? Can't it go above it ?

scripts/prepare-subcharts
```diff
@@ -0,0 +1,125 @@
...
+
+f=$1
+
+if ! [[ -f ${f}/charts/requirements.yaml ]]; then
+	echo "Chart ${CHART} has no dependencies; skipping step to pull in dependencies"
+	exit 0
+fi
+
+# Apply the patch on the requirements.yaml
```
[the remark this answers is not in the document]
aiyengar2: Charts like `rancher-monitoring` that introduce new subcharts like Prometheus Adapter [as part of the patch](https://github.com/rancher/charts/blob/6293280499af87bbe4905c15d8e938d1c273ac4e/packages/rancher-monitoring/rancher-monitoring.patch#L291-L384) need to be able to pull in the added dependencies as part of preparing the subcharts; therefore, `requirements.yaml` needs to be patched before running the prepare on the subcharts.

An initial solution I had was to just run the entire patch first before trying to pull in dependencies, but there's an edge case in the following scenario:
1. I run `make prepare` on the chart with the current set of scripts
1. I make some patches to the added subcharts (i.e. Prometheus Adapter) and run `make patch` to generate the patch. Patch is successfully generated
1. I run `make prepare` again and the patch is rejected since, before any dependencies are pulled in, the upstream chart has no folder for my added subchart (i.e. `charts/charts/prometheus-adapter` does not exist). Therefore, it can't apply a patch to my non-existent subchart and fails.

It seemed like the easiest way to resolve this was to just apply any patches required on the `requirements.yaml` only, pull in the subcharts (so `charts/charts/prometheus-adapter` now exists), unapply the `reqiuirements.yaml`-only patch, and then continue the `prepare` process normally to apply the patch as expected.

scripts/prepare-subcharts
```diff
@@ -0,0 +1,125 @@
...
+for file in $(find ./${f} -type f -name "*.patch"); do
+	awk '/diff.*requirements.yaml/{flag=1;next}/diff/{flag=0}flag' $file >> ${f}/requirements.patch
```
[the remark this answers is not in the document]
aiyengar2: Totally agree with how this code is unreadable / hard to understand. This definitely highlights why we need to move to using Go scripts instead of using Bash scripts at some point in the near future.

For this release, I'm not sure if there's any better way to simplify / refactor this awk statement; do you think it would be easier to understand if I added a comment? i.e.
```suggestion
        # This awk statement extracts the contents between the first line that matches the regex `diff.*requirements.yaml`
        # and the first line that matches the regex `diff` after it. This corresponds to the contents of any patches on the requirements.yaml file.
        # For more information about awk flags, see https://www.thelinuxrain.com/articles/how-to-use-flags-in-awk
	awk '/diff.*requirements.yaml/{flag=1;next}/diff/{flag=0}flag' $file >> ${f}/requirements.patch
```

scripts/prepare-subcharts
```diff
@@ -0,0 +1,125 @@
...
+			continue
+		fi
+		if [[ -f ${charts_path%/*}/package.yaml ]]; then
+			# Prepare the subchart and pull it in
+			subchart=$(basename -- ${charts_path%/*})
+			mirrored_dependencies+=("${subchart}")
+			prepare_subchart () {
+				local CURRENT_CHART=${CHART}
+				local CURRENT_INDEX=${i}
+				CHART=${subchart} ./scripts/prepare
```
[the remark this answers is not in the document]
aiyengar2: Yes; if we are running with `CHART=` or when `CHART` is unset, this would double call the scripts / prepare  if it's already another chart in the repo  (i.e. for our CI scripts). For example, this would prepare something like `rancher-kiali-server` twice.

However, for the average dev that uses `CHART=<chart-name>` when running the scripts, this recursive prepare is required in order to ensure any subcharts are also prepared / packaged (i.e. running `make CHART=rancher-istio prepare` needs to prepare `rancher-kiali-server` first).

One thing to note: there's no explicit check here about a circular dependency on prepares (i.e. `rancher-istio` depends on `rancher-kiali-server` depends on `rancher-istio` resulting in an eternal loop of prepares), but the assumption is that the developer won't run into this issue when designing charts.

scripts/prepare-subcharts
aiyengar2: will update in latest push

scripts/prepare-subcharts
aiyengar2: Yes, this is intended to clean up charts like `rancher-kiali-server` that we ran a prepare for when trying to prepare `rancher-istio` only.

Will update the comment like you mentioned.

scripts/prepare-subcharts
aiyengar2: Yes, but this isn't located in the subchart folder; this is the extracted patch specifically created from the awk statement (explained in https://github.com/rancher/charts/pull/579#discussion_r481832652) in order to support patches on added dependencies as described in https://github.com/rancher/charts/pull/579#discussion_r481825133.

Since this is created by the prepare-subcharts scripts, it's also deleted by the prepare-subcharts script.

scripts/generate-patch
aiyengar2: will update in next commit

scripts/generate-patch
aiyengar2: `remove_timestamp_from_diff` and the prepare-crds part are independent of each other since they touch different files, so it should be fine to place this above. Will update in the next commit!

scripts/prepare-subcharts
aiyengar2: Modified this a little bit to add a more step-by-step overview:
```
# Pull in the subcharts required by a chart
#
# At the end of this process, all the required subcharts will exist in the charts/charts directory.
# Charts will either exist as chart archives (i.e. a .tgz file) if they are not eligible for patches or 
# as a local chart archive (i.e. a directory) if they are eligible for patches (i.e. remote subchart with fixed version number)
#
# In order to prepare sub-charts so that we can patch them: 
# - We apply any patches on the requirements.yaml to track any new subcharts added via patch or updates to version numbers of the subcharts
# - We pull in any subchart that is a feature chart by running prepare on it
# - We copy in any local subcharts that are located in the overlay/ directory
# - We call `helm dependency update` to generate chart archives (.tgz files) for local and remote charts
# - We clean up any copied charts from the overlay/ directory
# - We clean up any feature charts that were recursively prepared
# - We unarchive the chart archive if and only if its a remote subchart with a fixed version number (which means it can be patched)
# - We revert the initial patch we applied on the requirements.yaml
```

Let me know if you have any thoughts if this is too detailed or unclear!

scripts/prepare-subcharts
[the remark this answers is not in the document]
cbron: Can you add just the first 2 comment lines.

scripts/prepare-subcharts
cbron: If we need to edit the main chart, and requirements, and a sub-chart, all those patches would just exist in the primary patch file right ? This is just temporary.

cbron (commented):
One or two minor things. Need to squash at least some of the commits here.

scripts/prepare-subcharts
aiyengar2: Yeah, all of those would be in the same patch file. This just temporarily strips the requirements.yaml changes out.

scripts/prepare-subcharts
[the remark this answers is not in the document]
aiyengar2: will be updated in next push

cbron (approved):

PR #586, opened 2020-08-25
CIS v2 charts

prachidamle:
Updated:
--------------
- NO separate CRD chart. Combine CRDs into rancher-cis-benchmark/crds chart itself.
- Add a package.yaml under rancher-cis-benchmark chart
- Added the linux nodeSelector and tolerations

---------------
Adding two charts for CIS v2 feature:

- rancher-cis-benchmark-crds - CRD chart
- rancher-cis-benchmark - chart deploying rancher/cis-operator

Commands to install the charts:
helm install clusterscan-operator ./rancher-cis-benchmark/charts/ --create-namespace=true -n cis-operator-system --
kubeconfig <kubeconfig>

Commands to uninstall the charts:
helm uninstall clusterscan-operator -n cis-operator-system --kubeconfig <kubeconfig>

### DIFF
--- a/packages/rancher-cis-benchmark/charts/Chart.yaml
+++ b/packages/rancher-cis-benchmark/charts/Chart.yaml
@@ -1,0 +1,13 @@
+apiVersion: v1
+appVersion: v0.0.1
+description: The cis-operator enables running CIS benchmark security scans on a kubernetes cluster
+name: rancher-cis-benchmark
+version: 0.0.1
+keywords:
+- security
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/namespace: cis-operator-system
+  catalog.cattle.io/release-name: rancher-cis-benchmark
+  catalog.cattle.io/ui-component: rancher-cis-benchmark
+  catalog.cattle.io/provides-gvr: cis.cattle.io.clusterscans/v1
--- a/packages/rancher-cis-benchmark/charts/README.md
+++ b/packages/rancher-cis-benchmark/charts/README.md
@@ -1,0 +1,14 @@
+# Rancher CIS Benchmark Chart
+
+The cis-operator enables running CIS benchmark security scans on a kubernetes cluster and generate compliance reports.
+
+# Installation
+
+### Requirements
+
+This chart depends on the rancher-cis-benchmark-crd chart.
+
+### Installation
+```
+helm install rancher-cis-benchmark ./ --create-namespace -n cis-operator-system
+```
--- a/packages/rancher-cis-benchmark/charts/templates/_helpers.tpl
+++ b/packages/rancher-cis-benchmark/charts/templates/_helpers.tpl
@@ -1,0 +1,23 @@
+{{/* Ensure namespace is set the same everywhere */}}
+{{- define "cis.namespace" -}}
+  {{- .Release.Namespace | default "cis-operator-system" -}}
+{{- end -}}
+
+{{- define "system_default_registry" -}}
+{{- if .Values.global.cattle.systemDefaultRegistry -}}
+{{- printf "%s/" .Values.global.cattle.systemDefaultRegistry -}}
+{{- else -}}
+{{- "" -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Windows cluster will add default taint for linux nodes,
+add below linux tolerations to workloads could be scheduled to those linux nodes
+*/}}
+{{- define "linux_node_tolerations" -}}
+- key: "cattle.io/os"
+  value: "linux"
+  effect: "NoSchedule"
+  operator: "Equal"
+{{- end -}}
--- a/packages/rancher-cis-benchmark/charts/templates/configmap.yaml
+++ b/packages/rancher-cis-benchmark/charts/templates/configmap.yaml
@@ -1,0 +1,11 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: default-clusterscanprofiles
+  namespace: {{ template "cis.namespace" . }}
+data:
+  # Default ClusterScanProfiles per cluster provider type
+  rke: "rke-profile-permissive"
+  eks: "eks-profile"
+  gke: "gke-profile"
+  default: "cis-1.5-profile"
--- a/packages/rancher-cis-benchmark/charts/templates/deployment.yaml
+++ b/packages/rancher-cis-benchmark/charts/templates/deployment.yaml
@@ -1,0 +1,46 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: cis-operator
+  namespace: {{ template "cis.namespace" . }}
+  labels:
+    cis.cattle.io/operator: cis-operator
+spec:
+  selector:
+    matchLabels:
+      cis.cattle.io/operator: cis-operator
+  template:
+    metadata:
+      labels:
+        cis.cattle.io/operator: cis-operator
+    spec:
+      serviceAccountName: cis-operator-serviceaccount
+      containers:
+      - name: cis-operator
+        image: '{{ template "system_default_registry" . }}{{ .Values.image.cisoperator.repository }}:{{ .Values.image.cisoperator.tag }}'
+        imagePullPolicy: Always
+        env:
+        - name: SECURITY_SCAN_IMAGE
+          value: {{ .Values.image.securityScan.repository }}
+        - name: SECURITY_SCAN_IMAGE_TAG
+          value: {{ .Values.image.securityScan.tag }}
+        - name: SONOBUOY_IMAGE
+          value: {{ .Values.image.sonobuoy.repository }}
+        - name: SONOBUOY_IMAGE_TAG
+          value: {{ .Values.image.sonobuoy.tag }}
+        resources:
+          {{- toYaml .Values.resources | nindent 12 }}
+      nodeSelector:
+        kubernetes.io/os: linux
+      {{- with .Values.nodeSelector }}
+      {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.affinity }}
+      affinity:
+      {{- toYaml . | nindent 8 }}
+      {{- end }}
+      tolerations:
+      {{- include "linux_node_tolerations" . | nindent 8}}
+      {{- with .Values.tolerations }}
+      {{- toYaml . | nindent 8 }}
+      {{- end }}
--- a/packages/rancher-cis-benchmark/charts/templates/rbac.yaml
+++ b/packages/rancher-cis-benchmark/charts/templates/rbac.yaml
@@ -1,0 +1,43 @@
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRole
+metadata:
+  labels:
+    app.kubernetes.io/name: rancher-cis-benchmark
+    app.kubernetes.io/instance: release-name
+  name: cis-operator-role
+rules:
+- apiGroups:
+  - '*'
+  resources:
+  - '*'
+  verbs:
+  - '*'
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  labels:
+    app.kubernetes.io/name: rancher-cis-benchmark
+    app.kubernetes.io/instance: release-name
+  name: cis-operator-rolebinding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cis-operator-role
+subjects:
+- kind: ServiceAccount
+  name: cis-serviceaccount
+  namespace: {{ template "cis.namespace" . }}
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: cis-operator-installer
+subjects:
+- kind: ServiceAccount
+  name: cis-operator-serviceaccount
+  namespace: {{ template "cis.namespace" . }}
+roleRef:
+  kind: ClusterRole
+  name: cluster-admin
+  apiGroup: rbac.authorization.k8s.io
--- a/packages/rancher-cis-benchmark/charts/templates/serviceaccount.yaml
+++ b/packages/rancher-cis-benchmark/charts/templates/serviceaccount.yaml
@@ -1,0 +1,14 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  namespace: {{ template "cis.namespace" . }}
+  name: cis-operator-serviceaccount
+---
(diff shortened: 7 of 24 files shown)


packages/rancher-cis-benchmark-crds/charts/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+appVersion: v0.0.1
+description: Helm chart for installing the CRDs used by rancher cis-operator than runs security scans on a kubernetes cluster
+name: rancher-cis-benchmark-crds
+version: 0.0.1
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/namespace: cis-operator-system
+  catalog.cattle.io/release-name: rancher-cis-benchmark-crds
+  catalog.cattle.io/provides-gvr: clusterscans.cis.cattle.io/v1
```
StrongMonkey, line 10: does CRD chart also need `provides-gvr` labels? Would that be a self-conflict condition?

packages/rancher-cis-benchmark/charts/templates/rbac.yaml
```diff
@@ -0,0 +1,43 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: cis-operator-installer
+subjects:
+- kind: ServiceAccount
+  name: cis-operator-serviceaccount
+  namespace: {{ template "cis.namespace" . }}
+roleRef:
+  kind: ClusterRole
+  name: cluster-admin
```
StrongMonkey, line 42: Assume this is already discussed, unless we need to switch to a fine-grain role.

packages/rancher-cis-benchmark/charts/values.yaml
```diff
@@ -0,0 +1,35 @@
+# Default values for rancher-cis-benchmark.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+image:
+  cisoperator:
+    repository: rancher/cis-operator
+    tag: v0.0.2
+  securityScan:
+    repository: prachidamle/security-scan
```
StrongMonkey, line 10: hmmm, this is not pushed under `rancher`

prachidamle: Need this PR merged for building the rancher/security-scan image https://github.com/rancher/security-scan/pull/23

packages/rancher-cis-benchmark/charts/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,14 @@
...
+  name: cis-operator-serviceaccount
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  namespace: {{ template "cis.namespace" . }}
+  labels:
+    app.kubernetes.io/name: rancher-cis-benchmark
+    app.kubernetes.io/instance: release-name
+  name: cis-serviceaccount
```
StrongMonkey, line 14: what is this serviceaccount used for? I see it is referenced by clusterrolebinding but not actuallly used by any of pods.

packages/rancher-cis-benchmark-crds/charts/Chart.yaml
prachidamle: I think UI needs this annotation to be present on the separate CRD chart  @cbron can you confim pls.

packages/rancher-cis-benchmark/charts/templates/serviceaccount.yaml
prachidamle: It is being used by the pods that get launched by the operator for every CIS scan created by user

prachidamle: https://github.com/rancher/cis-operator/blob/master/pkg/securityscan/job/job.go#L72

packages/rancher-cis-benchmark/charts/Chart.yaml
```diff
@@ -0,0 +1,12 @@
...
+appVersion: v0.0.1
+description: Helm chart for installing the rancher clusterscan-operator than runs security scans on a kubernetes cluster
+name: rancher-cis-benchmark
+version: 0.0.1
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/namespace: cis-operator-system
+  catalog.cattle.io/release-name: rancher-cis-benchmark
+  catalog.cattle.io/ui-component: rancher-cis-benchmark
+  catalog.cattle.io/requires-gvr: clusterscans.cis.cattle.io/v1
```
cbron, line 11: remove this one, you only need either requires or auto-install

packages/rancher-cis-benchmark-crds/charts/Chart.yaml
cbron: Lets add provides to both

packages/rancher-cis-benchmark/charts/templates/rbac.yaml
prachidamle: yes, CIS is cluster-admin only feature. We also have fine grain roles added to share access to other users.

packages/rancher-cis-benchmark-crds/charts/Chart.yaml
prachidamle: done

packages/rancher-cis-benchmark/charts/templates/_helpers.tpl
```diff
@@ -0,0 +1,31 @@
...
+*/}}
+{{- define "linux_node_tolerations" -}}
+- key: "cattle.io/os"
+  value: "linux"
+  effect: "NoSchedule"
+  operator: "Equal"
+{{- end -}}
+
+{{- define "linux_node_selector" -}}
+{{- if semverCompare "<1.14-0" .Capabilities.KubeVersion.GitVersion -}}
```
cbron, line 26: We can skip this because 2.5 won't ship with even 1.15. See the backup and restore PR for example. Think it makes sense to skip helper and just put directly on deployment since its a single line now.

packages/rancher-cis-benchmark/charts/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+# Installation
+
+### Requirements
+
+This chart depends on the rancher-cis-benchmark-crd chart.
+
+### Installation
+```
+helm install rancher-cis-benchmark ./ --create-namespace -n cis-operator-system
```
cbron, line 13: Nit: add in CRD chart install here as well.

packages/rancher-cis-benchmark/charts/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+appVersion: v0.0.1
+description: Helm chart for installing the rancher clusterscan-operator than runs security scans on a kubernetes cluster
+name: rancher-cis-benchmark
+version: 0.0.1
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/namespace: cis-operator-system
```
cbron, line 8: We aren't prepending the NS with cattle ? I know that was recommendation at one point, we can talk about it.

packages/rancher-cis-benchmark/charts/templates/_helpers.tpl
cbron: Oh, looks like its already implemented that way below, so we can just remove this helper.

packages/rancher-cis-benchmark/charts/templates/rbac.yaml
```diff
@@ -0,0 +1,43 @@
...
+    app.kubernetes.io/name: rancher-cis-benchmark
+    app.kubernetes.io/instance: release-name
+  name: cis-operator-role
+rules:
+- apiGroups:
+  - '*'
+  resources:
+  - '*'
+  verbs:
+  - '*'
```
cbron, line 14: Is this right ? It gets * on * ?

packages/rancher-cis-benchmark/charts/templates/view-role.yaml
```diff
@@ -0,0 +1,17 @@
...
+  name: cis-view
+rules:
+  - apiGroups:
+      - cis.cattle.io
+    resources:
+      - clusterscanbenchmarks
+      - clusterscanprofiles
+      - clusterscans
+      - clusterscanreports
+    verbs: ["get", "watch", "list"]
```
cbron, line 17: nit: its nice to have the admin/edit/view all in the same file, that way if you have to change a resource down the line you don't forget about one.

packages/rancher-cis-benchmark/charts/templates/configmap.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: v1
+metadata:
+  name: default-clusterscanprofiles
+  namespace: {{ template "cis.namespace" . }}
+data:
+  # Default ClusterScanProfiles per cluster provider type
+  rke: "rke-profile-permissive"
+  eks: "eks-profile"
+  gke: "gke-profile"
+  default: "cis-1.5-profile"
```
cbron, line 11: How is this file used ? I don't see it referenced anywhere.

packages/rancher-cis-benchmark/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,46 @@
...
+  template:
+    metadata:
+      labels:
+        cis.cattle.io/operator: cis-operator
+    spec:
+      serviceAccountName: cis-operator-serviceaccount
+      containers:
+      - name: cis-operator
+        image: '{{ template "system_default_registry" . }}{{ .Values.image.cisoperator.repository }}:{{ .Values.image.cisoperator.tag }}'
+        imagePullPolicy: Always
```
cbron, line 21: Any reason to use Always and not `IfNotPresent` ?

cbron (commented):
Didn't review CRDs that closely as I assume that was all similar to previous setup.

packages/rancher-cis-benchmark/charts/templates/configmap.yaml
prachidamle: It is referenced by the cis-operator [here](https://github.com/rancher/cis-operator/blob/master/pkg/securityscan/scanHandler.go#L118-L130) - instead of hardcoded default profiles, we decided to maintain it as a configmap in the chart so that if the defaults change we can upgrade the chart.

packages/rancher-cis-benchmark/charts/templates/deployment.yaml
prachidamle: Wont `Always` ensure that the latest image is pulled always even when chart is upgraded? Also actually I just carried forward this from the CIS v1 rancher-cis-benchmark system-charts https://github.com/rancher/system-charts/blob/dev-v2.4/charts/rancher-cis-benchmark/0.1.2/values.yaml#L31

packages/rancher-cis-benchmark/charts/Chart.yaml
prachidamle: In the feature channel it was actually suggested to not use cattle prefix - and also that UI is just looking for `-system ` suffix

packages/rancher-cis-benchmark/charts/README.md
prachidamle: There is no separate CRD chart - its all in the single `crds` folder now. So I dont think there is a CRD chart install command needed - correct?

packages/rancher-cis-benchmark/charts/templates/rbac.yaml
prachidamle: This is carried forward from the CIS v1 system-chart https://github.com/rancher/system-charts/blob/dev-v2.4/charts/rancher-cis-benchmark/0.1.2/templates/rbac.yaml#L11-L26

packages/rancher-cis-benchmark/charts/README.md
cbron: For manual installs there is assuming they are using the published version. If they use the vanilla git version it's all one command.

packages/rancher-cis-benchmark/charts/templates/view-role.yaml
cbron: This looks good now but can you re-name the file from `admin-roles` to just `roles` or something generic.

packages/rancher-cis-benchmark/charts/templates/deployment.yaml
cbron: When the chart is upgraded the version tag should change and new image would get pulled. But looking into it, `always` just does a check against cached image, so this should be fine.

packages/rancher-cis-benchmark/charts/templates/view-role.yaml
prachidamle: done - renamed as cis-roles.yaml

packages/rancher-cis-benchmark/charts/values.yaml
cbron: Need to get this in ASAP as it will block airgap testing

cbron (approved):
lgtm, we can merge but we need to get that security scan image moved over asap.

packages/rancher-cis-benchmark/charts/package.yaml
```diff
@@ -0,0 +1,3 @@
+providesGVR: clusterscans.cis.cattle.io/v1
+generateCRDChart:
+  enabled: true
```
cbron, line 3: This file should be above charts folder

cbron (changes requested):

packages/rancher-cis-benchmark/charts/package.yaml
prachidamle: moved the file up, now the CI validate step fails. Is there anything missing @cbron

cbron: CI failure is known separate issue being worked on

cbron (approved):
lgtm, just need to wait until CI fix is in then rebase.

packages/rancher-cis-benchmark/charts/package.yaml
prachidamle: CI issue is resolved now

packages/rancher-cis-benchmark/charts/Chart.yaml
deniseschannon: @prachidamle For our charts, let's make sure to prefix it with `cattle`

prachidamle: Need to clarify with @vincent99 and @ibuildthecloud if we are still using "cattle-*-system" convention for namespace - or just '*-system' is good enough.

PR #587, opened 2020-08-25
removed logging output and flow configs

paynejacob:


paynejacob:
https://github.com/rancher/rancher/issues/28511

Because the helm chart will no longer be responsible for managing outputs or flows we will remove them.  Going forward this chart is only responsible for loggings and the operator.

cbron:
Need to remove the tgz

PR #589, opened 2020-08-25
Add chart for backup-restore-operator

mrajashree:
This chart contains manifests for:
* Creating a ServiceAccount for the backup-restore-operator
* ClusterRoleBinding to grant cluster-admin role to the serviceAccount. This operator
needs cluster-admin role because it should have CRUD permissions on all resources.
* Deployment for launching a pod for the operator. The above serviceAccount is assigned to
the deployment. We also set env vars for default s3 store and default pvc details if they are set
* If s3 details are specified in values.yaml, the s3secret.yaml will be used for creating that secret

**NOTE**
1. The PVC input fields are still not final, QA wants this chart to be merged and they are aware that they can't use the default PVC configuration with this version of the chart. So please ignore those fields as they will change in the next version (in a day or two)
2. The CRD manifests are all autogenerated from wrangler's CRD & openapi schema generator. So changes related to field descriptions can be done, but the rest are all autogenerated
3. Also I'll add in additionalPrinterColumns in the next version

### DIFF
--- a/packages/backup-restore-operator/charts/Chart.yaml
+++ b/packages/backup-restore-operator/charts/Chart.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+appVersion: v0.0.1-rc4
+description: Helm chart for installing the rancher backup-restore-operator
+name: backup-restore-operator
+version: 0.0.1
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/namespace: cattle-resources-system
+  catalog.cattle.io/release-name: backup-restore-operator
+  catalog.cattle.io/ui-component: backup-restore-operator
--- a/packages/backup-restore-operator/charts/examples/s3values.yaml
+++ b/packages/backup-restore-operator/charts/examples/s3values.yaml
@@ -1,0 +1,14 @@
+image: 
+  repository: rancher/backup-restore-operator
+  tag: v0.0.1-rc4
+debug: false            # Enable debug logging in controller
+# Set as empty object {} if no s3 bucket needs to be configured for the operator
+s3: 
+  credentialSecretName: creds
+  region: us-west-2
+  bucketName: rajashree-backup-test
+  folder: ecm1
+  endpoint: s3.us-west-2.amazonaws.com
+
+global:
+  systemDefaultRegistry: ""
--- a/packages/backup-restore-operator/charts/templates/backup.yaml
+++ b/packages/backup-restore-operator/charts/templates/backup.yaml
@@ -1,0 +1,92 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: backups.resources.cattle.io
+spec:
+  group: resources.cattle.io
+  names:
+    kind: Backup
+    plural: backups
+  scope: Namespaced
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      properties:
+        spec:
+          properties:
+            encryptionConfigName:
+              description: 'Name of secret containing the encryption config, must
+                be in the namespace of the chart: cattle-resources-system'
+              type: string
+            resourceSetName:
+              description: Name of resourceSet CR to use for backup, must be in the
+                same namespace
+              type: string
+            retentionCount:
+              type: integer
+            schedule:
+              description: Cron schedule for recurring backups
+              type: string
+            storageLocation:
+              nullable: true
+              properties:
+                s3:
+                  nullable: true
+                  properties:
+                    bucketName:
+                      type: string
+                    credentialSecretName:
+                      type: string
+                    endpoint:
+                      type: string
+                    endpointCA:
+                      type: string
+                    folder:
+                      type: string
+                    insecureTLSSkipVerify:
+                      type: boolean
+                    region:
+                      type: string
+                  type: object
+              type: object
+          required:
+          - resourceSetName
+          type: object
+        status:
+          properties:
+            conditions:
+              items:
+                properties:
+                  lastTransitionTime:
+                    type: string
+                  lastUpdateTime:
+                    type: string
+                  message:
+                    type: string
+                  reason:
+                    type: string
+                  status:
+                    type: string
+                  type:
+                    type: string
+                type: object
+              nullable: true
+              type: array
+            lastSnapshotTs:
+              type: string
+            nextSnapshotAt:
+              type: string
+            numSnapshots:
+              type: integer
+            observedGeneration:
+              type: integer
+            summary:
+              type: string
+          type: object
+      type: object
+  version: v1
+  versions:
+  - name: v1
+    served: true
+    storage: true
--- a/packages/backup-restore-operator/charts/templates/clusterrolebinding.yaml
+++ b/packages/backup-restore-operator/charts/templates/clusterrolebinding.yaml
@@ -1,0 +1,12 @@
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: backup-restore-operator-installer
+subjects:
+- kind: ServiceAccount
+  name: backup-restore-operator-serviceaccount
+  namespace: {{ .Release.Namespace }}
+roleRef:
+  kind: ClusterRole
+  name: cluster-admin
+  apiGroup: rbac.authorization.k8s.io
--- a/packages/backup-restore-operator/charts/templates/deployment.yaml
+++ b/packages/backup-restore-operator/charts/templates/deployment.yaml
@@ -1,0 +1,53 @@
+apiVersion: apps/v1 
+kind: Deployment
+metadata:
+  name: backup-restore-operator
+  namespace: {{ .Release.Namespace }}
+  labels:
+    resources.cattle.io/operator: backup-restore
+spec:
+  selector:
+    matchLabels:
+      resources.cattle.io/operator: backup-restore
+  template:
+    metadata:
+      labels:
+        resources.cattle.io/operator: backup-restore
+    spec:
+      serviceAccountName: backup-restore-operator-serviceaccount
+      containers:
+      - name: backup-restore-operator
+        image: {{ template "system_default_registry" . }}{{ .Values.image.repository }}:{{ .Values.image.tag }}
+        imagePullPolicy: Always
+        env:
+        - name: CHART_NAMESPACE
+          value: {{ .Release.Namespace }}
+          {{- if .Values.pvc }}
+        - name: DEFAULT_PVC_BACKUP_STORAGE_LOCATION
+          value: {{ .Release.Name }}-pvc
+          {{- end }}
+          {{- if .Values.s3 }}
+        - name: DEFAULT_S3_BACKUP_STORAGE_LOCATION
+          value: {{ .Release.Name }}-s3
+          {{- end }}
+        {{- if .Values.pvc }}
+        volumeMounts:
+        - mountPath: "/var/lib/backups"
+          name: pv-storage
+      volumes:
+        - name: pv-storage
+          persistentVolumeClaim:
+            claimName: {{ .Values.pvcName }}
+          {{- end}}
+    {{- with .Values.nodeSelector }}
+    nodeSelector:
+    {{- toYaml . | nindent 8 }}
+    {{- end }}
+    {{- with .Values.affinity }}
+    affinity:
+    {{- toYaml . | nindent 8 }}
+    {{- end }}
+    {{- with .Values.tolerations }}
+    tolerations:
+    {{- toYaml . | nindent 8 }}
+    {{- end }}
--- a/packages/backup-restore-operator/charts/templates/resourceset.yaml
+++ b/packages/backup-restore-operator/charts/templates/resourceset.yaml
@@ -1,0 +1,94 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: resourcesets.resources.cattle.io
+spec:
+  group: resources.cattle.io
+  names:
+    kind: ResourceSet
+    plural: resourcesets
+  scope: Namespaced
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      properties:
+        controllerReferences:
+          items:
+            properties:
+              apiVersion:
+                type: string
+              name:
+                type: string
+              namespace:
+                type: string
+              replicas:
+                type: integer
+              resource:
+                type: string
+            type: object
+          nullable: true
+          type: array
+        resourceSelectors:
+          items:
+            properties:
+              apiVersion:
+                type: string
+              kinds:
+                items:
+                  type: string
+                nullable: true
+                type: array
+              kindsRegexp:
+                type: string
+              labelSelectors:
+                nullable: true
+                properties:
+                  matchExpressions:
+                    items:
+                      properties:
+                        key:
+                          type: string
+                        operator:
+                          type: string
+                        values:
+                          items:
+                            type: string
+                          nullable: true
+                          type: array
+                      type: object
+                    nullable: true
+                    type: array
+                  matchLabels:
+                    additionalProperties:
+                      type: string
+                    nullable: true
+                    type: object
+                type: object
+              namespaceRegexp:
+                type: string
+              namespaces:
+                items:
+                  type: string
+                nullable: true
+                type: array
+              resourceNameRegexp:
+                type: string
+              resourceNames:
+                items:
+                  type: string
+                nullable: true
+                type: array
+            type: object
+          nullable: true
+          required:
+          - apiVersion
+          type: array
+      required:
+      - resourceSelectors
+      type: object
+  version: v1
+  versions:
+  - name: v1
+    served: true
+    storage: true
--- a/packages/backup-restore-operator/charts/templates/restore.yaml
+++ b/packages/backup-restore-operator/charts/templates/restore.yaml
@@ -1,0 +1,87 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: restores.resources.cattle.io
+spec:
+  group: resources.cattle.io
+  names:
+    kind: Restore
+    plural: restores
+  scope: Namespaced
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      properties:
+        spec:
+          properties:
+            backupFilename:
+              type: string
+            deleteTimeout:
+              maximum: 5
+              type: integer
+            encryptionConfigName:
+              type: string
+            prune:
+              nullable: true
+              type: boolean
+            storageLocation:
+              nullable: true
+              properties:
+                s3:
+                  nullable: true
+                  properties:
+                    bucketName:
+                      type: string
+                    credentialSecretName:
+                      type: string
+                    endpoint:
+                      type: string
+                    endpointCA:
+                      type: string
+                    folder:
+                      type: string
+                    insecureTLSSkipVerify:
+                      type: boolean
+                    region:
+                      type: string
+                  type: object
+              type: object
+          required:
+          - backupFilename
+          type: object
+        status:
+          properties:
+            conditions:
+              items:
+                properties:
+                  lastTransitionTime:
+                    type: string
+                  lastUpdateTime:
+                    type: string
+                  message:
+                    type: string
+                  reason:
+                    type: string
+                  status:
+                    type: string
+                  type:
+                    type: string
+                type: object
+              nullable: true
+              type: array
+            numRetries:
+              type: integer
+            observedGeneration:
+              type: integer
+            restoreCompletionTs:
+              type: string
+            summary:
+              type: string
+          type: object
+      type: object
+  version: v1
+  versions:
+  - name: v1
+    served: true
+    storage: true
--- a/packages/backup-restore-operator/charts/templates/s3secret.yaml
+++ b/packages/backup-restore-operator/charts/templates/s3secret.yaml
@@ -1,0 +1,18 @@
+{{- if .Values.s3 -}}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ .Release.Name }}-s3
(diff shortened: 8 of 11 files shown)


packages/backup-restore-operator/charts/examples/s3values.yaml
```diff
@@ -0,0 +1,13 @@
+image: rancher/backup-restore-operator
```
StrongMonkey, line 1: I think this shoud follow the format as 
```
image:
  repository:
  tag: 
```

packages/backup-restore-operator/charts/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,12 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: backup-restore-operator-installer
+subjects:
+- kind: ServiceAccount
+  name: backup-restore-operator-serviceaccount
+  namespace: {{ .Release.Namespace }}
+roleRef:
+  kind: ClusterRole
+  name: cluster-admin
```
StrongMonkey, line 11: It shouldn't bind to cluster-admin instead using specific rules that your controller uses

packages/backup-restore-operator/charts/templates/resourceset.yaml
```diff
@@ -0,0 +1,94 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
StrongMonkey, line 1: I think CRD should defines in `crds` folder?

packages/backup-restore-operator/charts/templates/clusterrolebinding.yaml
mrajashree: @StrongMonkey this chart needs to be able to modify any resouces. Hence the cluster-admin role was decided in the design meeting

packages/backup-restore-operator/charts/templates/resourceset.yaml
mrajashree: @StrongMonkey I think there was something wrong with adding them in crds folder, where uninstall/upgrade didn't work, and it was decided to add them in templates folder?

mrajashree: @StrongMonkey I think adding crds in `crds` folder didn't take care of uninstall/upgrade, hence it was decided to put them in templates folder as per the `Helm Chart Requirements` doc:
```
Put the CRD’s in the main template folder, that way they are cleaned up on delete.
```

StrongMonkey: yeah, but arvind did came up with a plan so that crd can be made into a separate chart.  @aiyengar2 Not sure if you were going to use this.

mrajashree: I thought that was more for a chart that needed to create a CR for one of the CRDs it installs? And since backup-restore-operator doesn't need to create a CR during installation I thought it'd be fine? if not I'll change it

deniseschannon: My understanding is that how @mrajashree approached this implementation is correct. If you aren't creating a CR from a custom CRD, you move the CRDs to the templates folder so that can be cleaned up. 

Has that changed @aiyengar2 @cbron?

aiyengar2: > I thought that was more for a chart that needed to create a CR for one of the CRDs it installs? And since backup-restore-operator doesn't need to create a CR during installation I thought it'd be fine? if not I'll change it

Yes this is correct. 

@StrongMonkey the separate CRD chart is only required if you need to both install the CRDs and deploy resources of that kind into a cluster in the same chart since Helm will not be able to identify a resource of the CRD's kind before the CRD is installed to the cluster. However, if you aren't trying to deploy resources of that kind into a cluster, just keeping the CRDs in the `templates/` folder of the main chart should be sufficient.

aiyengar2: Question to @cbron and @deniseschannon though; since Rajashree is not using a separate CRD chart, does she need to add the annotation for `catalog.cattle.io/provides-gvr`?

Technically, this chart is providing CRDs that other charts can rely on, but has the unintended consequences of also installing a whole host of things other than CRDs. I'm assuming the answer here is no, but just wanted to call it out to confirm what the expectation should be for similar charts (I think Longhorn might fall under the same use case in https://github.com/rancher/charts/pull/526)

packages/backup-restore-operator/charts/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+appVersion: v0.0.1-rc4
+description: Helm chart for installing the rancher backup-restore-operator
+name: backup-restore-operator
+version: 0.0.1
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/namespace: cattle-resources-system
```
aiyengar2, line 8: nit: is there any reason why we chose `cattle-resources-system`? I would have assumed the namespace here should be `cattle-backup-restore-system` or something similar that is tied to the chart name

packages/backup-restore-operator/charts/templates/backup.yaml
```diff
@@ -0,0 +1,92 @@
...
+    openAPIV3Schema:
+      properties:
+        spec:
+          properties:
+            encryptionConfigName:
+              description: 'Name of secret containing the encryption config, must
+                be in the namespace of the chart: cattle-resources-system'
+              type: string
+            resourceSetName:
+              description: Name of resourceSet CR to use for backup, must be in the
```
aiyengar2, line 23: nit:

```suggestion
              description: Name of ResourceSet CR to use for backup, must be in the
```

packages/backup-restore-operator/charts/templates/backup.yaml
```diff
@@ -0,0 +1,92 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: backups.resources.cattle.io
+spec:
+  group: resources.cattle.io
+  names:
+    kind: Backup
+    plural: backups
```
aiyengar2, line 9:
```suggestion
    listKind: BackupList
    plural: backups
    singular: backup
```

packages/backup-restore-operator/charts/templates/backup.yaml
```diff
@@ -0,0 +1,92 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: backups.resources.cattle.io
+spec:
```
aiyengar2, line 5: Are we including any [`additionalPrinterColumns`](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/) or is that a later change?

packages/backup-restore-operator/charts/templates/resourceset.yaml
```diff
@@ -0,0 +1,94 @@
...
+kind: CustomResourceDefinition
+metadata:
+  name: resourcesets.resources.cattle.io
+spec:
+  group: resources.cattle.io
```
aiyengar2, line 6: Are we including any [`additionalPrinterColumns`](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/) or is that a later change?

packages/backup-restore-operator/charts/templates/resourceset.yaml
```diff
@@ -0,0 +1,94 @@
...
+  names:
+    kind: ResourceSet
+    plural: resourcesets
```
aiyengar2, line 9:
```suggestion
    listKind: ResourceSetList
    plural: resourcesets
    singular: resourceset
```

packages/backup-restore-operator/charts/templates/backup.yaml
```diff
@@ -0,0 +1,92 @@
...
+  scope: Namespaced
+  subresources:
+    status: {}
+  validation:
+    openAPIV3Schema:
+      properties:
+        spec:
+          properties:
+            encryptionConfigName:
+              description: 'Name of secret containing the encryption config, must
```
aiyengar2, line 19: nit:
```suggestion
              description: 'Name of Secret containing the encryption config, must
```

packages/backup-restore-operator/charts/templates/backup.yaml
```diff
@@ -0,0 +1,92 @@
...
+              description: Name of resourceSet CR to use for backup, must be in the
+                same namespace
+              type: string
+            retentionCount:
+              type: integer
+            schedule:
+              description: Cron schedule for recurring backups
+              type: string
+            storageLocation:
+              nullable: true
```
aiyengar2, line 32: What happens if the storageLocation is `null`?

packages/backup-restore-operator/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,53 @@
...
+          value: {{ .Release.Name }}-s3
+          {{- end }}
+        {{- if .Values.pvc }}
+        volumeMounts:
+        - mountPath: "/var/lib/backups"
+          name: pv-storage
+      volumes:
+        - name: pv-storage
+          persistentVolumeClaim:
+            claimName: {{ .Values.pvcName }}
```
aiyengar2, line 40: Not sure how you have this structured, but this should probably be:
```suggestion
            claimName: {{ .Values.pvc.name }}
```

packages/backup-restore-operator/charts/templates/restore.yaml
```diff
@@ -0,0 +1,87 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: restores.resources.cattle.io
+spec:
+  group: resources.cattle.io
```
aiyengar2, line 6: Are we including any [`additionalPrinterColumns`](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/) or is that a later change?

packages/backup-restore-operator/charts/templates/restore.yaml
```diff
@@ -0,0 +1,87 @@
...
+  names:
+    kind: Restore
+    plural: restores
```
aiyengar2, line 9:
```suggestion
    listKind: RestoreList
    plural: restores
    singular: restore
```

packages/backup-restore-operator/charts/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,12 @@
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: backup-restore-operator-installer
```
aiyengar2, line 4: I think all mentions of `backup-restore-operator` should be changed to `{{ .Chart.Name }}`

packages/backup-restore-operator/charts/templates/s3secret.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if .Values.s3 -}}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ .Release.Name }}-s3
```
aiyengar2, line 5: Should this be changed to the following?

```suggestion
  name: {{ .Chart.Name }}-s3
```

packages/backup-restore-operator/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,53 @@
...
+    {{- toYaml . | nindent 8 }}
+    {{- end }}
+    {{- with .Values.affinity }}
+    affinity:
+    {{- toYaml . | nindent 8 }}
+    {{- end }}
+    {{- with .Values.tolerations }}
+    tolerations:
+    {{- toYaml . | nindent 8 }}
+    {{- end }}
```
aiyengar2, line 53: Instead of using `with`, I think it would be easier to use `if` and just `toYaml` the entire `.Values.<selector>` like in `rancher-pushprox` here:

https://github.com/rancher/charts/blob/98d70cee35a19afd217633b5283e21a0dd0b8296/packages/rancher-pushprox/charts/templates/pushprox-clients.yaml#L16-L18

packages/backup-restore-operator/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,53 @@
+apiVersion: apps/v1 
+kind: Deployment
+metadata:
+  name: backup-restore-operator
+  namespace: {{ .Release.Namespace }}
+  labels:
+    resources.cattle.io/operator: backup-restore
```
aiyengar2, line 7: @deniseschannon @cbron should these new Rancher charts be updated to have the recommended labels from Helm 3 described [here](https://helm.sh/docs/chart_best_practices/labels/)? It's not a major change and could be pretty easily introduced with a helper like in `rancher-pushprox` here:

https://github.com/rancher/charts/blob/98d70cee35a19afd217633b5283e21a0dd0b8296/packages/rancher-pushprox/charts/templates/_helpers.tpl#L11-L15

This would then be changed to:
```suggestion
  labels: {{ include "backupRestore.labels" . | nindent 4 }}
```

packages/backup-restore-operator/charts/templates/clusterrolebinding.yaml
aiyengar2, line 11: We should create a ClusterRole CR to this and add a label to aggregate to `cluster-admin` instead of binding directly to the Kubernetes `cluster-admin`.

Here's an example from `rancher-monitoring`:

https://github.com/rancher/charts/blob/98d70cee35a19afd217633b5283e21a0dd0b8296/packages/rancher-monitoring/rancher-monitoring.patch#L1170-L1207

packages/backup-restore-operator/charts/templates/backup.yaml
```diff
@@ -0,0 +1,92 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
aiyengar2, line 1: To keep the CRDs separated from the actual elements deployed onto the cluster, we should either prepend the YAML names like `crd-<crd-name>.yaml` or move these into `templates/crds/backup.yml`.

packages/backup-restore-operator/charts/Chart.yaml
mrajashree: we chose `cattle-resources-system` as namespace and `resources` as apiversion because it was discussed in the design meeting there's a chance we can use ResourceSet CRD for cleaning up CRDs of uninstalled charts.

packages/backup-restore-operator/charts/templates/backup.yaml
mrajashree: All CRD manifests were generated by the wrangler openapi schema generator, is there a reason this should be changed?

packages/backup-restore-operator/charts/templates/backup.yaml
mrajashree: that will come later

packages/backup-restore-operator/charts/templates/backup.yaml
mrajashree: The operator will try to use either the s3 backup location or pvc configured at the operator level. If not it errors. Describe this in more detail here https://rancher.atlassian.net/wiki/spaces/EN/pages/494501903/Backup+restore+design+for+docs+testing

packages/backup-restore-operator/charts/templates/backup.yaml
aiyengar2: Oh, I wasn't aware that these CRD definitions were auto-generated. I've seen these fields in a couple of the Prometheus Operator CRDs, wasn't sure if they would be required here. 

i.e. with this CRD installed, can you do `kubectl get backup` or do you need to do `kubectl get backups`?

packages/backup-restore-operator/charts/templates/clusterrolebinding.yaml
mrajashree: This operator should be able to modify/create/delete any resources, and the ClusterRole from monitoring example specifies rules on particular resources. Since we want it to be able to target resources from all apigroups, I don't see how we'll know what resources to add to this clusterRole

packages/backup-restore-operator/charts/templates/s3secret.yaml
mrajashree: why can't we use the release name?

packages/backup-restore-operator/charts/templates/backup.yaml
mrajashree: This won't have any impact on the actual chart installation, is this separation to make the chart structure more clear?

packages/backup-restore-operator/charts/templates/deployment.yaml
mrajashree: The PVC approach hasn't been finalized, I forgot to mention in the description. QA has started testing my branch for everything else, and they want this to get merged even without PVC changes. So I'll update this with my next PR

packages/backup-restore-operator/charts/templates/backup.yaml
aiyengar2: > This won't have any impact on the actual chart installation, is this separation to make the chart structure more clear?

Yes

packages/backup-restore-operator/charts/templates/resourceset.yaml
cbron: Templates folder and no provides annotation is fine

packages/backup-restore-operator/charts/templates/backup.yaml
mrajashree: I can do both right now

aiyengar2: No need to make these changes then, will resolve 👍

packages/backup-restore-operator/charts/templates/s3secret.yaml
aiyengar2: `.Chart.Name` is globally fixed, whereas the `.Release.Name` is determined by the user input.

Unless there's a use case for a user deploying multiple backup operators onto the same cluster with different Release names (which might require some more changes due to CRDs being in the `templates/` directory instead of the `crd/` directory or a separate CRD chart), I think it would be preferable to use the `.Chart.Name`

packages/backup-restore-operator/charts/templates/deployment.yaml
mrajashree: Isn't this almost similar? 😅 And `with` allows setting the scope so it's easier to use I think

mrajashree: This is almost similar, and `with` allows setting scope so it's easier to use I think, I don't think I need to change this

packages/backup-restore-operator/charts/examples/s3values.yaml
```diff
@@ -0,0 +1,14 @@
+image: 
+  repository: rancher/backup-restore-operator
+  tag: v0.0.1-rc4
+debug: false            # Enable debug logging in controller
+# Set as empty object {} if no s3 bucket needs to be configured for the operator
+s3: 
+  credentialSecretName: creds
+  region: us-west-2
+  bucketName: rajashree-backup-test
```
StrongMonkey, line 9: Assume this has to change to replace your username? :smile:

mrajashree: yeah that was in the examples folder for QA, I'll remove that examples folder 😅 And I think I have to update the commented section in actual values.yaml too

packages/backup-restore-operator/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,53 @@
...
+spec:
+  selector:
+    matchLabels:
+      resources.cattle.io/operator: backup-restore
+  template:
+    metadata:
+      labels:
+        resources.cattle.io/operator: backup-restore
+    spec:
+      serviceAccountName: backup-restore-operator-serviceaccount
```
StrongMonkey, line 17: Nit: can this name be `backup-restore-operator`?

packages/backup-restore-operator/charts/templates/deployment.yaml
deniseschannon: Probably but not a blocker for being merged.

packages/backup-restore-operator/charts/templates/backup.yaml
deniseschannon: I agree with renaming them cause it makes it clear.

packages/backup-restore-operator/charts/templates/deployment.yaml
aiyengar2: Would be nice if we can use `{{ .Chart.Name }}` to avoid hard-coding it

mrajashree:
This needs to be split into two charts, CRD and main, because it was decided in today's design meeting to create ResourceSet CR for rancher's backups. I have finalized fields for the PVC input, and Chart's name is being used for naming resources. I think these changes are quite big and I'm opening a new PR to include them. Will address all "unresolved" relevant comments in this new PR https://github.com/rancher/charts/pull/594

packages/backup-restore-operator/charts/templates/resourceset.yaml
aiyengar2: > > I thought that was more for a chart that needed to create a CR for one of the CRDs it installs? And since backup-restore-operator doesn't need to create a CR during installation I thought it'd be fine? if not I'll change it
> 
> Yes this is correct.
> 
> @StrongMonkey the separate CRD chart is only required if you need to both install the CRDs and deploy resources of that kind into a cluster in the same chart since Helm will not be able to identify a resource of the CRD's kind before the CRD is installed to the cluster. However, if you aren't trying to deploy resources of that kind into a cluster, just keeping the CRDs in the `templates/` folder of the main chart should be sufficient.
> 
> Related comment / discussion: [#526 (comment)](https://github.com/rancher/charts/pull/526#issuecomment-673114537)

This explanation has been added to the `README.md` in https://github.com/rancher/charts/pull/595 to clarify this requirement for any future contributors.

PR #591, opened 2020-08-26
support configurable docker root

paynejacob:
Because users can change the docker root we need to support changing the container log path

### DIFF
--- a/packages/rancher-logging/rancher-logging.patch
+++ b/packages/rancher-logging/rancher-logging.patch
@@ -56,7 +56,7 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-logging/charts-original/values
  
  resources: {}
    # We usually recommend not to specify default resources and to leave this as a conscious
-@@ -76,4 +76,35 @@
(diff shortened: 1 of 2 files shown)


packages/rancher-logging/rancher-logging.patch
```diff
@@ -143,4 +143,6 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-logging/charts-original/values
 +
 +global:
 +  systemDefaultRegistry: ""
++
++dockerRoot: "/var/lib/docker/"
```
brendarearden, line 147: nit: a helper to ensure it ends with a slash incase a user forgets it

brendarearden (dismissed):
lgtm

packages/rancher-logging/rancher-logging.patch
paynejacob: That is a good idea

paynejacob:
This is incomplete and most likely not compatible with recent chart changes.

PR #592, opened 2020-08-26
Switch PushProx image to rancher/busybox

aiyengar2:
rancher/kubectl does not have sh.

Switching to recently created busybox image that has all the necessary commands installed:

```bash
arvindiyengar: ~/Rancher/charts/src/github.com/rancher/charts 
$ docker run rancher/busybox:1.31.1 sh -c "ls /bin" | egrep "echo|find|test|cp|chmod"              
acpid
chmod
cp
cpio
dhcprelay
echo
find
findfs
rpm2cpio
tcpsvd
test
udhcpc
udhcpc6
udhcpd
```

Issue: https://github.com/rancher/rancher/issues/28441


aiyengar2:
Related conversation that led to the initial use of `rancher/kubectl` can be found here: https://github.com/rancher/charts/pull/576#discussion_r475369524

I assumed that `rancher/kubectl` had the necessary commands without checking 😅

PR #594, opened 2020-08-27
Add chart for backup-restore-operator

mrajashree:
This chart needs to create a CR when its installed. So as per offline discussion of making original charts similar in structure to upstream chart, this chart has all its CRDs in the /crds folder. And the CR it should create is in the templates folder.
Carrying over all relevant comments from https://github.com/rancher/charts/pull/589

This chart creates:
1. ServiceAccount bound to the cluster-admin role, which is used for launching the operator deployment. This operator needs CRUD permissions on all resources in the cluster. Hence only cluster-admin should be able to operate it. That's why it was decided in design meetings to assign it the cluster-admin role and have no other roles for it.
2. The CRDs in crds folder are generated using wrangler's CRD definition and openapi schema package. [This](https://rancher.atlassian.net/wiki/spaces/EN/pages/494469147/Backup+Restore+Rancher+MCM+app) design doc covers the explanation behind each field
3. User can either configure a s3 bucket OR a PVC to store backups in. The operator will error if both are specified. This helm chart also uses the `fail` template function to prevent chart installation if both are enabled.
4. If s3.enabled is set to true, the chart creates a secret in the chart's namespace that contains all details required for uploading/retrieving backups to/from s3 compatible stores.
5. If persistence.enabled is set to true, the operator creates a PVC in the chart's namespace and mounts it at /var/lib/backups. We want the PVC's access mode to be RWO, hence that is not accepted from the user. The user can specify storageClass, volumeName and size. The fields accepted for this PVC and the comments surrounding them in the values.yaml file follow the pattern that most of the upstream Helm charts follow. 
6. The `checksum/secret` annotation for s3-secret in the deployment spec will re-configure the pod to use the updated s3 secret if the user changes their s3 configuration, fixing https://github.com/rancher/rancher/issues/28607

https://github.com/rancher/rancher/issues/28640

NOTE: The "url" field in package.yaml is empty since its an original chart, will change it based on the new logic for generating CRDs.

https://github.com/rancher/rancher/issues/28368

### DIFF
--- a/packages/backup-restore-operator/charts/templates/_helpers.tpl
+++ b/packages/backup-restore-operator/charts/templates/_helpers.tpl
@@ -1,0 +1,76 @@
+{{- define "system_default_registry" -}}
+{{- if .Values.global.cattle.systemDefaultRegistry -}}
+{{- printf "%s/" .Values.global.cattle.systemDefaultRegistry -}}
+{{- else -}}
+{{- "" -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Windows cluster will add default taint for linux nodes, 
+add below linux tolerations to workloads could be scheduled to those linux nodes
+*/}}
+{{- define "linux-node-tolerations" -}}
+- key: "cattle.io/os"
+  value: "linux"
+  effect: "NoSchedule"
+  operator: "Equal"
+{{- end -}}
+
+
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+*/}}
+{{- define "backupRestore.fullname" -}}
+{{- .Chart.Name | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Create chart name and version as used by the chart label.
+*/}}
+{{- define "backupRestore.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Common labels
+*/}}
+{{- define "backupRestore.labels" -}}
+helm.sh/chart: {{ include "backupRestore.chart" . }}
+{{ include "backupRestore.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
+
+{{/*
+Selector labels
+*/}}
+{{- define "backupRestore.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "backupRestore.fullname" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+resources.cattle.io/operator: backup-restore
+{{- end }}
+
+
+{{/*
+Create the name of the service account to use
+*/}}
+{{- define "backupRestore.serviceAccountName" -}}
+{{ include "backupRestore.fullname" . }}
+{{- end }}
+
+
+{{- define "backupRestore.s3SecretName" -}}
+{{- printf "%s-%s" .Chart.Name "s3" | replace "+" "_" | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{- define "backupRestore.pvcName" -}}
+{{ include "backupRestore.fullname" . }}
+{{- end }}
+
+{{- define "backupRestore.nfsPVName" -}}
+{{ include "backupRestore.fullname" . }}
+{{- end }}
--- a/packages/backup-restore-operator/charts/templates/deployment.yaml
+++ b/packages/backup-restore-operator/charts/templates/deployment.yaml
@@ -1,0 +1,58 @@
+{{- if and .Values.s3.enabled .Values.persistence.enabled }}
+{{- fail "\n\nCannot configure both s3 and PV for storing backups" }}
+{{- end }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "backupRestore.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "backupRestore.labels" . | nindent 4 }}
+spec:
+  selector:
+    matchLabels:
+      {{- include "backupRestore.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      labels:
+        {{- include "backupRestore.selectorLabels" . | nindent 8 }}
+      annotations:
+        checksum/secret: {{ include (print $.Template.BasePath "/s3-secret.yaml") . | sha256sum }}
+    spec:
+      serviceAccountName: {{ include "backupRestore.serviceAccountName" . }}
+      containers:
+      - name: {{ .Chart.Name }}
+        image: {{ template "system_default_registry" . }}{{ .Values.image.repository }}:{{ .Values.image.tag }}
+        imagePullPolicy: Always
+        env:
+        - name: CHART_NAMESPACE
+          value: {{ .Release.Namespace }}
+          {{- if .Values.s3.enabled }}
+        - name: DEFAULT_S3_BACKUP_STORAGE_LOCATION
+          value: {{ include "backupRestore.s3SecretName" . }}
+          {{- end }}
+          {{- if .Values.persistence.enabled }}
+        - name: DEFAULT_PERSISTENCE_ENABLED
+          value: "persistence-enabled"
+        volumeMounts:
+        - mountPath: "/var/lib/backups"
+          name: pv-storage
+      volumes:
+        - name: pv-storage
+          persistentVolumeClaim:
+            claimName: {{ include "backupRestore.pvcName" . }}
+          {{- end }} 
+      nodeSelector:
+        kubernetes.io/os: linux
+      {{- with .Values.nodeSelector }}
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      tolerations:
+      {{- include "linux-node-tolerations" . | nindent 8}}
+      {{- with .Values.tolerations }}
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
--- a/packages/backup-restore-operator/charts/templates/s3-secret.yaml
+++ b/packages/backup-restore-operator/charts/templates/s3-secret.yaml
@@ -1,0 +1,33 @@
+{{- if .Values.s3.enabled -}}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "backupRestore.s3SecretName" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "backupRestore.labels" . | nindent 4 }}
+type: Opaque
+stringData:
+  {{- with .Values.s3 }}
+  {{- if .credentialSecretName }}
+  credentialSecretName: {{ .credentialSecretName }} 
+  {{- end }}
+  {{- if .credentialSecretNamespace }}
+  credentialSecretNamespace: {{ .credentialSecretNamespace }} 
+  {{- end }}
+  {{- if .region }}
+  region: {{ .region }}
+  {{- end }}
+  bucketName: {{ .bucketName }}
+  {{- if .folder }}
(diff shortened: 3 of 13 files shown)


packages/backup-restore-operator/charts/values.yaml
```diff
@@ -0,0 +1,46 @@
...
+## Any Backups created using this resourceSet must be created in the same namespace.
+rancherResourceSetNamespace: ""
+
+## Default s3 bucket for storing all backup files created by the backup-restore-operator
+s3:
+  enabled: false
+  ## credentialSecretName if set, should be the name of the Secret containing AWS credentials. It should be created in the Chart's namespace and can be created after the chart is installed. 
+  ## To use IAM Role, don't set this field
+  credentialSecretName: creds 
+  region: us-west-2 (optional)
```
StrongMonkey, line 15: Do you need to comment this out `(optional)`?

packages/backup-restore-operator/charts/values.yaml
```diff
@@ -0,0 +1,46 @@
...
+  bucketName: bucket-name
+  folder: base folder (optional)
```
StrongMonkey, line 17: same above

packages/backup-restore-operator/package.yaml
```diff
@@ -0,0 +1,5 @@
+url:
```
StrongMonkey, line 1: Yeah you need to fix CI so that this doesn't error out.

packages/backup-restore-operator/package.yaml
```diff
@@ -0,0 +1,5 @@
+url: 
+packageVersion: 00
```
StrongMonkey, line 2: Also don't have to set packageVersion in here

packages/backup-restore-operator/package.yaml
mrajashree: yup, I don't think it's decided yet what the expected value should be?

mrajashree: yes, I don't think its decided yet what the value for `url` should be for the scripts in case of original chart

mrajashree: yes, it's being tracked in this issue https://github.com/rancher/rancher/issues/28628

packages/backup-restore-operator/charts/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "backupRestore.serviceAccountName" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "backupRestore.labels" . | nindent 4 }}
```
aiyengar2, line 7: nit, `nindent` should automatically indent the first line iirc. Happens a couple of times throughout the PR.

```suggestion
  labels: {{- include "backupRestore.labels" . | nindent 4 }}
```

packages/backup-restore-operator/charts/templates/s3-secret.yaml
```diff
@@ -0,0 +1,30 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "backupRestore.s3SecretName" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "backupRestore.labels" . | nindent 4 }}
+type: Opaque
+stringData:
+  {{- with .Values.s3}}
```
aiyengar2, line 11: nit:
```suggestion
  {{- with .Values.s3 }}
```

packages/backup-restore-operator/charts/templates/s3-secret.yaml
```diff
@@ -0,0 +1,30 @@
...
+  {{- if .folder }}
+  folder: {{ .folder }} 
+  {{- end }}
+  endpoint: {{ .endpoint }}
+  {{- if .endpointCA }} 
+  endpointCA: {{ .endpointCA }} 
+  {{- end }}
+  {{- if .insecureTLSSkipVerify }}
+  insecureTLSSkipVerify: {{ .insecureTLSSkipVerify }} 
+  {{- end }}
```
aiyengar2, line 28: Should this be set to a default value or does the fact that it doesn't exist imply that it's false?

i.e.
```suggestion
  insecureTLSSkipVerify: {{ default false .insecureTLSSkipVerify }} 
```

packages/backup-restore-operator/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,53 @@
...
+        - name: DEFAULT_PVC_NAME
+          value: {{ include "backupRestore.pvcName" . }}
+        volumeMounts:
+        - mountPath: "/var/lib/backups"
+          name: pv-storage
+      volumes:
+        - name: pv-storage
+          persistentVolumeClaim:
+            claimName: {{ include "backupRestore.pvcName" . }}
+          {{- end}}
```
aiyengar2, line 41: nit:
```suggestion
          {{- end }}
```

packages/backup-restore-operator/charts/templates/s3-secret.yaml
mrajashree: yup the fact that it doesn't exist implies that it's false

packages/backup-restore-operator/charts/templates/serviceaccount.yaml
mrajashree: yeah I was using the sample chart helm3 creates and it added the labels in this format (next line) so thought I'd let it be 😅

packages/backup-restore-operator/charts/templates/_helpers.tpl
```diff
@@ -57,4 +57,31 @@ Create the name of the service account to use
 
 {{- define "backupRestore.pvcName" -}}
 {{ include "backupRestore.fullname" . }}
-{{- end }}
\ No newline at end of file
+{{- end }}
+
+{{- define "windows-node-selector" -}}
+{{- if semverCompare "<1.14-0" .Capabilities.KubeVersion.GitVersion -}}
```
cbron, line 67: I know this was in example, but by 2.5.0 we will be shipping 1.16+, so that greatly simplifies it. We could break out of helper, up to you.

packages/backup-restore-operator/charts/templates/deployment.yaml
```diff
@@ -38,16 +38,19 @@ spec:
         - name: pv-storage
           persistentVolumeClaim:
             claimName: {{ include "backupRestore.pvcName" . }}
-          {{- end }}
-      {{- with .Values.nodeSelector }}
+          {{- end }}      
       nodeSelector:
+{{- include "windows-node-selector" . | nindent 8 }}
+{{- include "linux-node-selector" . | nindent 8 }}
```
cbron, line 52: I thought you could only do one nodeSelector in a pod spec. Are you trying to allow this to go on either type of node ?

packages/backup-restore-operator/package.yaml
```diff
@@ -1,4 +1,3 @@
-url: 
 generateCRDChart:
   enabled: true
   providesGVR: resourcesets.resources.cattle.io/v1
```
cbron, line 3: Recent development, can you put this line up a level where url was.

cbron (changes requested):
CI will be fixed once Brenda's PR is in, we should wait to merge until then or it will break all other chart generation.

packages/backup-restore-operator/charts/templates/deployment.yaml
mrajashree: yeah even with both nodeSelectors the chart was able to run the pod on a linux cluster node. Although after it was created the pod only had the `kubernetes.io/os: linux` nodeSelector. So looks like with both selectors specified it allows the pod to go on either type of node. But I haven't tested it for a windows node, so I removed that selector

cbron: I think it just overwrites with whatever was set last. If you want to test you could flip them and try to deploy to linux and see if it fails.

cbron: Does this need to run on windows for 2.5.0 ? Might be able to solve this with nodeAffinity, we can talk about it.

cbron (commented):
Are we not doing the User RBAC roles for admin/view/edit ? Not sure if it makes sense for your use case. Outside of that and windows node selector it lgtm.

mrajashree:
@cbron yes individual roles for admin/edit/view won't be applicable, because each restore operation can modify/delete any resources. I have removed the windows node selector based on our discussion

packages/backup-restore-operator/charts/templates/deployment.yaml
mrajashree: based on offline discussion, we don't need to add it in for windows right now

cbron (approved):
lgtm, just wait for other PR to merge, rebase to re-run CI to make sure it passes

packages/backup-restore-operator/charts/values.yaml
```diff
@@ -0,0 +1,61 @@
...
+    size: 2Gi
+
+  nfs:
+    enabled: false
+   
+    ## Path that is exported by the NFS server. More info: https://kubernetes.io/docs/concepts/storage/volumes#nfs
+    path: 
+    
+    ## Server is the hostname or IP address of the NFS server. More info: https://kubernetes.io/docs/concepts/storage/volumes#nfs
+    server:
```
aiyengar2, line 49: nit:
```suggestion
    ## Path that is exported by the NFS server. More info: https://kubernetes.io/docs/concepts/storage/volumes#nfs
    path: ""
    
    ## Server is the hostname or IP address of the NFS server. More info: https://kubernetes.io/docs/concepts/storage/volumes#nfs
    server: ""
```

packages/backup-restore-operator/charts/templates/nfs-pv.yaml
```diff
@@ -0,0 +1,18 @@
...
+  {{- with .Values.persistence.nfs }}
+  capacity:
+    storage: {{ .size | quote }}
+  accessModes:
+    - ReadWriteOnce
+  nfs:
+    server: {{ .server }}
+    path: {{ .path | quote }}
+  {{- end }}
+{{- end }}
```
aiyengar2, line 18: nit: newline

packages/backup-restore-operator/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,63 @@
...
+        kubernetes.io/os: linux
+      {{- with .Values.nodeSelector }}
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      tolerations:
+{{- include "linux-node-tolerations" . | nindent 8}}
```
aiyengar2, line 60: nit: the unindented include looks weird. We should either indent it or move it back to the previous line like:
```suggestion
      tolerations: {{- include "linux-node-tolerations" . | nindent 8}}
```

packages/backup-restore-operator/charts/templates/_helpers.tpl
```diff
@@ -0,0 +1,75 @@
...
+
+{{- define "backupRestore.nfsPVName" -}}
+{{ include "backupRestore.fullname" . }}
+{{- end }}
+
+{{/*
+Windows cluster will add default taint for linux nodes, 
+add below linux tolerations to workloads could be scheduled to those linux nodes
+*/}}
+{{- define "linux-node-tolerations" -}}
```
aiyengar2, line 70: nit: inconsistent casing

`linux-node-tolerations` uses a different format than the other helpers in this file. Should it be `backupRestore.linuxNode.tolerations`? Would also be nice it we do something similar with `kubernetes.io/os: linux` for `backupRestore.linuxNode.selector`.

Alternatively, since this is something Rancher-specific but not backup-restore specific, maybe we should add it to the top of the file under `system_default_registry` as `linux_node_tolerations`.

aiyengar2 (approved):
nit comments, LGTM

StrongMonkey (approved):

packages/backup-restore-operator/charts/templates/_helpers.tpl
mrajashree: the node selector was just one line, and from a previous comment I thought we could break out of the helper for that. I can rename the tolerations helper though

packages/backup-restore-operator/charts/values.yaml
mrajashree: not applicable anymore, after some discussion we decided to remove nfs part

PR #595, opened 2020-08-27
Add generateCRDChart.assumeOwnershipOfCRDs flag

aiyengar2:
This commit adds a new flag to the experimental feature of generating a CRD chart for charts that need to be able to assume the ownership of any existing CRDs within a cluster.

Feature charts with this flag enabled will differ from the normal CRD chart in the following ways:
- Instead of having CRDs from `crd/` in `templates/`, they will be relocated to `crd-manifest/`.
- On render, the CRDs in `crd-manifest` are placed into a ConfigMap that will be deployed on the cluster.
- On install / upgrade / rollback, a pre-install / pre-upgrade / pre-rollback hook Job that does a `kubectl apply -f` on the manifest within the crd-manifest ConfigMap (with appropriate RBAC credentials via a ServiceAccount, CRB, and ClusterRole) will install the CRDs onto the cluster.
- On uninstall, a delete hook Job does a `kubectl delete -f` on the manifest within the crd-manifest ConfigMap (with the same RBAC credentials) to remove the CRDs from the cluster.

At the moment, this will only be used by the `rancher-monitoring` chart.

Related Issue: https://github.com/rancher/rancher/issues/28326


scripts/prepare-crds
```diff
@@ -115,4 +115,165 @@ done
...
+# the assumeOwnershipOfCRDs flag is enabled
+assume_ownership=$(cat ${f}/package.yaml | yq r - generateCRDChart.assumeOwnershipOfCRDs)
+if [[ "${assume_ownership}" != "true" ]]; then
+	exit 0
+fi
+
+mv ${f}/charts-crd/templates ${f}/charts-crd/crd-manifest
+mkdir -p ${f}/charts-crd/templates
+
+cat << EOF > ${f}/charts-crd/values.yaml
```
aiyengar2, line 130: Would it make sense to invest a little bit of time in creating something like a `chart-templates/` directory at the root of this repository with a `chart-templates/charts-crd/` and `chart-templates/chart-crd-assume-ownership` folder that has all these files within it? 

Using `cat << EOF > $PATH` to create all these files as part of a Bash script seems clunky, but I wasn't sure if it was worth investing the time to rewrite the `prepare-crd` script to do this (as opposed to maybe just doing it later when we switch to Go scripts)

README.md
```diff
@@ -77,6 +77,11 @@ up-to-date with the latest chart. It also automatically build github pages to se
 
 ### Experimental: Splitting CRDs from an upstream package into a separate package
 
+**Note to Contributers:** This flag can only be used to create separate CRD charts if your chart meets the following three requirements:
+- The chart defines one or more CRDs
+- The chart is based on an upstream chart and includes a `package.yaml` (if this is not the case, you will need to manually create a separate CRD chart)
+- The chart tries to install some default CRs based on the CRDs that it defines (if this is not the case, you should place the CRDs directly within the `templates/` directory of the chart; using a CRD chart is only necessary since rendering the chart will fail since the `kind` of the default CRs cannot be found in the cluster as the CRD is not installed yet).
```
aiyengar2, line 83: Creating a link back to the original discussion for this documentation: https://github.com/rancher/charts/pull/589#discussion_r476931372

scripts/prepare-crds
```diff
@@ -115,4 +115,165 @@ done
...
+  name: {{ .Chart.Name }}-manifest
+data:
+  crd-manifest.yaml: |
+    {{- \$currentScope := . -}}
+    {{- \$crds := (.Files.Glob  "crd-manifest/**.yaml") -}}
+    {{- range \$path, \$_ :=  \$crds -}}
+    {{- with \$currentScope -}}
+    {{ .Files.Get \$path | nindent 4 }}
+    ---
+    {{- end -}}{{- end -}}
```
aiyengar2, line 204: I thought a ConfigMap was fine for storing the CRD manifest since there's no sensitive data here; does anyone feel strongly about making this a Secret instead?

packages/rancher-monitoring/package.yaml
```diff
@@ -2,4 +2,5 @@ url: https://kubernetes-charts.storage.googleapis.com/prometheus-operator-8.16.1
 packageVersion: 00
 generateCRDChart:
   enabled: true
-  providesGVR: prometheuses.monitoring.coreos.com/v1
\ No newline at end of file
+  providesGVR: prometheuses.monitoring.coreos.com/v1
+  assumeOwnershipOfCRDs: true
```
cbron, line 6: Since the bash scripts are getting so complicated already, I wonder if this sort of thing should just exist in overlays in the monitoring chart, as its a fairly unique situation. Only a few cases could possible exist.

scripts/prepare-crds
aiyengar2: Sounds good, I can edit the PR to use that instead.

It shouldn't take too much longer, but it would be editing the logic of previously written code (as opposed to the current PR, which only add / modifies an additional set of files to the existing `prepare-crds` script). Should be relatively easy to do though.

scripts/prepare-crds
```diff
@@ -115,4 +115,165 @@ done
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ .Chart.Name }}-manager
+  labels:
+    app: {{ .Chart.Name }}-manager
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ .Chart.Name }}-manager
```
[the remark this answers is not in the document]
aiyengar2: Since the Chart Name here would be the CRD chart's name and not the upstream chart's name, `{{ .Chart.Name }}-manager` would already be something like `rancher-monitoring-crd-manager` so I don't think we need to append another `crd-` to it.

scripts/prepare-crds
```diff
@@ -115,4 +115,165 @@ done
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ .Chart.Name }}-manager
+EOF
+
+cat << EOF > ${f}/charts-crd/templates/manifest.yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Chart.Name }}-manifest
```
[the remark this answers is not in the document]
aiyengar2: good catch, updating in latest commit

scripts/prepare-crds
```diff
@@ -115,4 +115,165 @@ done
...
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ .Chart.Name }}-delete
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ .Chart.Name }}
+  annotations:
+    "helm.sh/hook": pre-delete
```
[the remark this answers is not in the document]
aiyengar2: This is done pre-delete since the ClusterRole, ClusterRoleBinding, and ServiceAccount still need to exist when executing the Job, so post-delete would fail.

scripts/prepare-crds
```diff
@@ -115,4 +115,165 @@ done
...
+cat << EOF > ${f}/charts-crd/templates/jobs.yaml
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ .Chart.Name }}-create
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ .Chart.Name }}
+  annotations:
+    "helm.sh/hook": post-install, post-upgrade, post-rollback
```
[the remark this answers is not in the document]
aiyengar2: Hmm, so this comment brought another question to my mind: do we even need a separate CRD chart if, using this approach, we're going to install the CRDs using a Job and clean them up afterwards anyways? 

There's no reason why we need a completely separate CRD chart if we aren't using the `templates/` directory to actually contain CRDs; at the moment, the only reason why I think having a separate CRD chart for monitoring would still be useful is for feature charts that depend on just the CRDs but not the operator (i.e. Logging V2 might install the CRD chart in order to deploy ServiceMonitors / PodMonitors onto the cluster, but there's no need for the resources to actually monitor anything until the user turns on Monitoring V2).

@ibuildthecloud do you have any thoughts on this as well?

scripts/prepare-crds
```diff
@@ -115,4 +115,165 @@ done
...
+          image: {{ template "system_default_registry" . }}{{ .Values.image.repository }}:{{ .Values.image.tag }}
+          imagePullPolicy: IfNotPresent
+          command:
+          - /bin/sh
+          - -c
+          - kubectl apply -f /etc/config/crd-manifest.yaml
+          volumeMounts:
+            - name: crd-manifest
+              readOnly: true
+              mountPath: /etc/config
```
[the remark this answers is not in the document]
aiyengar2: Not particularly. Is there a better place to place it?

I chose `/etc/` primarily cause that's usually where config files are stored.

packages/rancher-monitoring/package.yaml
aiyengar2: The issue with using `overlay/` is that these files aren't being added to the main chart but rather to a separate CRD chart. However, if the resolution for https://github.com/rancher/charts/pull/595#discussion_r480376364 is to avoid creating a separate CRD chart and instead just have the Job exist in the main chart, I think this can be done as a one-off for monitoring.

scripts/prepare-crds
```diff
@@ -115,4 +115,165 @@ done
...
+cat << EOF > ${f}/charts-crd/values.yaml
+# Default values for ${name}-crd.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+global:
+  systemDefaultRegistry: ""
+
+image:
+  repository: rancher/hyperkube
```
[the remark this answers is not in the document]
aiyengar2: No particular reason; I can make the switch to kubectl as part of the next revision to this PR

scripts/prepare-crds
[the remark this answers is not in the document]
cbron: One chart is going to be cleaner and avoid issues like the CRD chart not getting called on update.
I still don't think that use-case of another chart requiring CRD's only is a valid use-case. When a chart requires another chart it gets kicked out of deploy page over to the other chart where user has to install that one. User doesn't get an option to deploy only the CRD's of that required chart.

aiyengar2:
TLDR: Based on the discussion on this ticket and discussion offline, I'm closing out this PR in favor of another one that reverts to using the CRD approach currently used by Prometheus upstream.

#### Summary of Discussion

The initial purpose of switching to using CRD charts was due to the following 3 requirements that CRD charts with CRDs in the `templates/` directory met:
1. CRDs needed to be installed onto the cluster on the first `helm install` only if they aren't currently installed on the cluster
1. CRDs needed to be updated on the cluster on `helm upgrade`
1. CRDs needed to be deleted from the cluster along with the chart of `helm uninstall` regardless of whether we installed them or not (i.e. "assumed ownership")

However:
- Requirement #1 is already done by Prometheus upstream through using the `crd/` directory
- @cbron followed up with @vincent99 about https://github.com/rancher/charts/pull/595#discussion_r480258634 and it seems like we aren't reinstalling CRD charts on an upgrade. This eliminates the ability for CRD charts to satisfy requirement #2 so there's no reason why Monitoring alone needs to support this use case.
- Requirement #3 is already done by the Prometheus upstream chart if a particular flag, `.prometheusOperator.cleanupCustomResource` which [triggers a Job to delete the CRDs](https://github.com/helm/charts/blob/master/stable/prometheus-operator/templates/prometheus-operator/cleanup-crds.yaml) just like this PR seeks to achieve

Therefore, since minor changes to Prometheus upstream now seems to meet all the requirements that we would potentially need to support the assumeOwnership approach for Rancher Monitoring, no script changes should be required at the moment since no charts need it. I'll be addressing the monitoring-specific changes in another PR.

aiyengar2:
Reopened this PR based on the new requirement that CRDs / CRs should not be auto-uninstalled on primary chart uninstall. This necessitates the CRD chart approach for monitoring once again to decouple CRDs from the primary chart.

Also, we will be upgrading CRD charts on a primary chart upgrade (requirement 2 from above).

cc: @cbron

packages/rancher-monitoring/package.yaml
aiyengar2: Based on https://github.com/rancher/charts/pull/595#issuecomment-685861749, the overlay/ solution is no longer possible.

scripts/prepare-crds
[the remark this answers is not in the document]
aiyengar2: > We need to double check that this chart would get auto-installed during an upgrade. I'll ask. If not then this whole thing has to get moved back into the primary chart.

Based on new requirements detailed in https://github.com/rancher/charts/pull/595#issuecomment-685861749, this seems to be the case. We'll have to use a separate CRD chart for this.

scripts/prepare-crds
aiyengar2: @cbron I added this change separately to the last commit on the PR. Let me know if the changes look good to you and I can accordingly squash them into my other commits.

scripts/generate-patch
```diff
@@ -20,9 +20,14 @@ function revert_crd_changes {
...
-	# Move charts-crd/templates back into charts/crd/
+	# Move CRDs back into charts/crd/
 	mkdir -p ${f}/charts/crds/
-	mv ${f}/charts-crd/templates/* ${f}/charts/crds/
+	if [[ -d ${f}/charts-crd/crd-manifest ]]; then
+		mv ${f}/charts-crd/crd-manifest/* ${f}/charts/crds/
+	else
+		mv ${f}/charts-crd/templates/* ${f}/charts/crds/
+	fi
+
```
aiyengar2, line 30: Based on discussion in https://github.com/rancher/charts/pull/607#discussion_r483288658, the work to split this into a `clean-crds` script that is called on clean and before patch is applied should be done as part of this PR.

Will update in the next commit.

chart-templates/crd-base/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: ${api_version}
+version: ${chart_version}
+description: A Rancher chart that creates ${name} CRDs within a cluster.
+name: ${name}-crd
+type: application
```
aiyengar2, line 5: Depending on which PR gets merged in first, we may need to update the `Chart.yaml` here based on https://github.com/rancher/charts/pull/604#discussion_r483348846.

chart-templates/crd-assume-ownership/templates/_helpers.tpl
```diff
@@ -0,0 +1,7 @@
+# Rancher
+
+{{- define "system_default_registry" -}}
+{{- if .Values.global.systemDefaultRegistry -}}
```
cbron, line 4: Recently changed to `global.cattle.systemDefaultRegistry`

packages/rancher-monitoring/package.yaml
```diff
@@ -2,4 +2,5 @@ url: https://kubernetes-charts.storage.googleapis.com/prometheus-operator-8.16.1
 packageVersion: 00
 generateCRDChart:
   enabled: true
-  providesGVR: prometheuses.monitoring.coreos.com/v1
\ No newline at end of file
+  providesGVR: prometheuses.monitoring.coreos.com/v1
```
cbron, line 5: Provides has moved outside of generateCRDChart

chart-templates/crd-base/overlay-upstream/templates/validate-install-crd.yaml
```diff
@@ -0,0 +1,12 @@
...
+# {{- range .Capabilities.APIVersions -}}
+# {{- if hasKey \$found (toString .) -}}
+# 	{{- set \$found (toString .) true -}}
+# {{- end -}}
+# {{- end -}}
+# {{- range \$_, \$exists := \$found -}}
+# {{- if (eq \$exists false) -}}
+# 	{{- required "Required CRDs are missing. Please install the ${name}-crd chart before installing this chart." "" -}}
+# {{- end -}}
+# {{- end -}}
```
cbron, line 12: I'm not a fan of a few things with this part: 
* it disable's `helm template` without finding the API's, and finding them is not practical, so it slows down development and debugging
* it necessitates a second "overlay" concept, the more concepts we have the harder it is to understand and maintain. 
* Removing this section also removes the concept of setting CRDs to environment variables in the script.
* While the error message is helpful, we should simply be documenting this info in the README/overview pages, and the existing helm error message should be somewhat similar.

scripts/prepare-crds
```diff
@@ -56,16 +59,40 @@ for crd_yaml in ${f}/charts-crd/templates/*; do
...
+# Copy base template and apply variables to the template
+mkdir -p ${f}/charts-crd
+cp -R ./chart-templates/crd-base/* ${f}/charts-crd
+assume_ownership=$(cat ${f}/package.yaml | yq r - generateCRDChart.assumeOwnershipOfCRDs)
+if [[ ${assume_ownership} == "true" ]]; then
+    cp -fR ./chart-templates/crd-assume-ownership/* ${f}/charts-crd
+fi
+for template_file in $(find ${f}/charts-crd -type f); do
+  # Applies any environment variables currently set onto your template file
+  eval "echo \"$(sed 's/"/\\"/g' ${template_file})\"" > ${template_file}
```
cbron, line 78: This is basically a templating function ?

chart-templates/crd-assume-ownership/templates/jobs.yaml
```diff
@@ -0,0 +1,73 @@
...
+      serviceAccountName: {{ .Chart.Name }}-manager
+      containers:
+        - name: kubectl
+          image: {{ template "system_default_registry" . }}{{ .Values.image.repository }}:{{ .Values.image.tag }}
+          imagePullPolicy: IfNotPresent
+          command:
+          - /bin/kubectl
+          - delete
+          - -f
+          - /etc/config/crd-manifest.yaml
```
cbron, line 64: We may need to have more advanced script in here long term if finalizer blocks the delete. https://github.com/kubernetes/kubernetes/issues/60538#issuecomment-369099998

chart-templates/crd-base/Chart.yaml
cbron: We can do this now.

scripts/generate-patch
cbron: We can follow up on this now, as that is merged

cbron (changes requested):
Concept looks good, few things still

scripts/prepare-crds
aiyengar2: Yes, this command basically runs `eval` on the contents of the template file, which automatically subs out the global variables. The `sed` command within it just ensures that double quotes don't get lost in translation

chart-templates/crd-assume-ownership/templates/_helpers.tpl
aiyengar2: will be updated in the latest commit

packages/rancher-monitoring/package.yaml
aiyengar2: Will be updated in the latest commit

chart-templates/crd-base/overlay-upstream/templates/validate-install-crd.yaml
aiyengar2: I agree with all the points you've listed, but I'm not sure if we want to remove this part of the script. Since using the charts directly as Helm charts (rather than installing via Dashboard) is still in scope for users, removing this script would rely on Helm's rendering engine to figure out if CRDs aren't installed on render (with a less direct error message) instead of enforcing that all the CRDs mentioned need to exist on an install. 

However, Helm's rendering engine will only fail if the chart tries to create a custom resource that does not exist on chart deploy; for example, an ECM cluster that has CRDs for Prometheus, Alertmanager, ServiceMonitors, and PrometheusRules will successfully deploy the `rancher-monitoring` chart, but will technically lack full functionality since PodMonitors and ThanosRulers do not exist.

aiyengar2: To workaround the `helm template` issue I usually just use `helm install <args> --debug --dry-run` or just specify the CRDs like ` --api-versions monitoring.coreos.com/v1/Alertmanager` on `helm template`. One option would be to modify this script to output the necessary helm template command for a dev to use as part of the error message if the CRDs don't exist.

chart-templates/crd-assume-ownership/templates/jobs.yaml
aiyengar2: Hmm this one is a little tricky. If we want to do the patch based logic that is described in the comment mentioned above, we'll need to probably switch this PR to using the `rancher/hyperkube` image instead the `rancher/kubectl` image, since the `rancher/kubectl` image does not come with `/bin/sh` installed iirc (preventing us from running multiple `kubectl` commands unless we use one initContainer per CRD to uninstall).

Also, in order to get the CRD names to apply the finalizer patches, we would have to continue to use the concept of setting CRDs to env vars in the script (flagged in https://github.com/rancher/charts/pull/595#discussion_r483970192 as something we want to avoid).

As of now, I'm going to just add an initContainer that does a `kubectl patch -f /etc/config/crd-manifest.yaml` before running `kubectl delete -f /etc/config/crd-manifest`, which will effectively remove any declared finalizer blocks by "resetting" the CRD before deleting it; hopefully, this should resolve any issues brought up by https://github.com/kubernetes/kubernetes/issues/60538#issuecomment-369099998.

scripts/generate-patch
aiyengar2: Updated in latest push

chart-templates/crd-base/Chart.yaml
aiyengar2: Updated in latest push

scripts/clean-crds
```diff
@@ -0,0 +1,37 @@
...
+    exit 0
+fi
+
+# Remove the validate-install-${name}-crd.yaml
+name=$(cat ${f}/charts/Chart.yaml | yq r - 'name')
+rm ${f}/charts/templates/validate-install-crd.yaml
+# Remove additional annotations added to original chart if added
+if [[ "$(yq r ${f}/charts/Chart.yaml 'annotations[catalog.cattle.io/auto-install]')" == "${name}-crd=match" ]]; then
+    yq d -i ${f}/charts/Chart.yaml 'annotations[catalog.cattle.io/auto-install]'
+fi
```
aiyengar2, line 37: Modified the dependency logic here based on https://github.com/rancher/charts/pull/604, which changes `annotations[catalog.cattle.io/auto-install-gvr]` to `annotations[catalog.cattle.io/auto-install]` and also allows a user to submit a non-default value for the annotation.

scripts/generate-patch
```diff
@@ -88,7 +65,13 @@ for f in packages/*; do
...
 				split_crds=$(cat ${f}/package.yaml | yq r - generateCRDChart.enabled)
 				if [[ "${split_crds}" == "true" ]]; then
-					revert_crd_changes ${f}
+					./scripts/clean-crds ${f}
+				fi
+				providesGVR="$(cat ${f}/package.yaml | yq r - providesGVR)"
+				if ! [[ -z ${providesGVR} ]]; then
+					if [[ "$(yq r ${f}/charts/Chart.yaml 'annotations[catalog.cattle.io/provides-gvr]')" == "${providesGVR}" ]]; then
+						yq d -i ${f}/charts/Chart.yaml "annotations[catalog.cattle.io/provides-gvr]"
+					fi
```
aiyengar2, line 74: Another change from needed to fix https://github.com/rancher/charts/pull/604 to ensure that the flag is removed on generating a patch.

chart-templates/crd-assume-ownership/templates/jobs.yaml
cbron: Yea I'm not sure around likelihood of this actually happening with our charts.

scripts/clean-crds
cbron: This is assuming user would be maintaining custom CRD chart outside of primary chart folder, thats ok assumption for now.

chart-templates/crd-base/overlay-upstream/templates/validate-install-crd.yaml
cbron: We can leave for now and see if it turns into a headache. I think we should re-asses on rewrite.

cbron (approved):
Think this works. Would be good for @StrongMonkey to review.

aiyengar2:
The original second half of this PR that moved `revert_crd_changes` into a new script was moved to a separate PR https://github.com/rancher/charts/pull/620. The first three commits of this PR now match those 3 commits, so you can ignore them on the review.

cbron:
And there are new changes ?

chart-templates/crd-base/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: ${api_version}
+version: ${chart_version}
+description: A Rancher chart that creates ${name} CRDs within a cluster.
```
StrongMonkey, line 3: Probably need to change `A Rancher chart`. I think @cbron mentioned this in slack.

chart-templates/crd-base/overlay-upstream/templates/validate-install-crd.yaml
```diff
@@ -0,0 +1,12 @@
+# {{- \$found := dict -}}
+${set_found_crd}
```
StrongMonkey, line 2: What is this used for?

chart-templates/crd-base/Chart.yaml
aiyengar2: Good catch, will update this in the next commit:
```suggestion
description: Installs the CRDs for ${name}.
```

chart-templates/crd-base/overlay-upstream/templates/validate-install-crd.yaml
aiyengar2: This comes directly from our original scripts at https://github.com/rancher/charts/blob/153171167ff52f0311b9a4d9968d33742d248101/scripts/prepare-crds#L97-L112, but the purpose of the `validate-install-crd.yaml` is to cause the original chart to fail on rendering (i.e. `helm install --debug --dry-run`) if the CRDs are not currently installed onto the cluster.

Some of the caveats with this approach are detailed in https://github.com/rancher/charts/pull/595#discussion_r483970192.

StrongMonkey: but do you still these commented lines? if not I think we should remove it

StrongMonkey (commented):
Also Can we put chart-template folder into `scripts/`?

chart-templates/crd-base/overlay-upstream/templates/validate-install-crd.yaml
aiyengar2: We need the commented lines since Helm doesn't know how to deal with an empty file that is generated if the CRDs are installed onto the cluster (i.e. since the file is empty, Helm will complain it is invalid YAML). However, a file which renders comments doesn't have this issue.

aiyengar2:
> Also Can we put chart-template folder into `scripts/`?

@StrongMonkey updated!

StrongMonkey (approved):

PR #597, opened 2020-08-28
Remove extra end to fix parse error

brendarearden:
**Problem**
When removing `.Release.Names` from the helper, I did not remove enough `{{end}}`s at the end of the statement resulting in a parse error

**Solution**
Remove the extra `{{end}}`

**Issue**
https://github.com/rancher/rancher/issues/28147


packages/rancher-kiali-server/rancher-kiali-server.patch
```diff
@@ -51,8 +51,11 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-kiali-server/charts-original/t
 +    {{- printf "%s" $name | trunc 63 | trimSuffix "-" }}
```
rmweir, line 51: is this here on purpose?

brendarearden: Yes, but only because I left the option for $name to also be set by `.Values.nameOverride`. If someone decides to use an override name, we want to make sure it meets some basic dns validation requirements. I wasn't sure how common it is to override the name, so I wasn't sure about removing it. We don't need it for any of out current flows, so would it make sense to just stick with chart.name?

brendarearden: not sure whats going on with that.....https://github.com/rancher/charts/blob/677dfada2ece6331e0cb45b2a6d7fb419b1e59f1/packages/rancher-kiali-server/rancher-kiali-server.patch#L51 its already in the dev-v2.5 branch

brendarearden:
closing this and combining in this pr https://github.com/rancher/charts/pull/610

PR #598, opened 2020-08-28
Provides must exist on both parent and CRD chart

cbron:
If chart X requires chart Y, but chart Y is split into crd/main chart,
then provides must exist on both so that X auto-installs the main
chart which auto-installs the CRD chart. Without this you would get
Y CRDs but no operator.

### DIFF
--- a/scripts/prepare-crds
+++ b/scripts/prepare-crds
@@ -72,6 +72,7 @@ if ! [[ -z ${providesGVR} ]]; then
 		yq w -i ${f}/charts/Chart.yaml "annotations[catalog.cattle.io/auto-install-gvr]" "${providesGVR}"
 	fi
 	if [[ -z "$(yq r ${f}/charts/Chart.yaml 'annotations[catalog.cattle.io/provides-gvr]')" ]]; then
+		yq w -i ${f}/charts/Chart.yaml "annotations[catalog.cattle.io/provides-gvr]" "${providesGVR}"
 		yq w -i ${f}/charts-crd/Chart.yaml "annotations[catalog.cattle.io/provides-gvr]" "${providesGVR}"
 	fi
 	if [[ -z "$(yq r ${f}/charts/Chart.yaml 'annotations[catalog.cattle.io/hidden]')" ]]; then


scripts/prepare-crds
```diff
@@ -72,6 +72,7 @@ if ! [[ -z ${providesGVR} ]]; then
 		yq w -i ${f}/charts/Chart.yaml "annotations[catalog.cattle.io/auto-install-gvr]" "${providesGVR}"
 	fi
 	if [[ -z "$(yq r ${f}/charts/Chart.yaml 'annotations[catalog.cattle.io/provides-gvr]')" ]]; then
+		yq w -i ${f}/charts/Chart.yaml "annotations[catalog.cattle.io/provides-gvr]" "${providesGVR}"
 		yq w -i ${f}/charts-crd/Chart.yaml "annotations[catalog.cattle.io/provides-gvr]" "${providesGVR}"
```
aiyengar2, line 76: The logic of both the Parent and CRD chart having `provides-gvr` confuses me a little bit. Based on this logic, a scenario would be:

1. I tell Dashboard I want to install `rancher-istio`. Dashboard initiates an install.
1. Dashboard observes that `rancher-istio` has the `requires-gvr` flag, which means that it needs to install a package with `provides-gvr` **(Question 1: how does it know whether to install `rancher-monitoring-crd` or `rancher-monitoring` if both have the the `provides-gvr` flag?)**
1. Dashboard initiates an install of `rancher-monitoring`.
1. Dashboard observes that `rancher-monitoring` has the `auto-install-gvr` flag, which means that it needs to install a package with `provides-gvr` **(Question 2: how does it know whether to install `rancher-monitoring-crd` or `rancher-monitoring` if both have the `provides-gvr` flag?)**
1. Dashboard initiates an install of `rancher-monitoring`.
1. Dashboard completes the install of `rancher-monitoring-crd`
1. Dashboard completes the install of `rancher-monitoring`
1. Dashboard completes the install of `rancher-istio`

It seems like to resolve the questions above, we should introduce a new structure for the flags that differentiates between requiring just the CRDs vs. requiring the entire operator to be installed; i.e. Istio requires the Operator for monitoring but something like Logging might only require the CRDs to create the default Service Monitors / Pod Monitors (since a user does not have to enable monitoring for logging to work). Something like:
- `provides-gvr`: only on CRD charts, says that it provides just the CRDs
- `provides-operator`: on operator charts, says that it provides the operator, which implies providing the CRDs
- `auto-install-gvr`: on operator charts that *own* the CRDs, says that we must auto-install a chart that has the appropriate `provides-gvr` flag
- `requires-gvr`: on operator charts that *rely* on the CRDs, says that we must have the CRD specified in the cluster for this chart to be installed (i.e. `rancher-logging`)
- `requires-operator`: on operator charts that *rely* on the existence of the operator itself, says that we must have the specified chart installed before trying to install this chart (i.e. `rancher-istio` requiring `rancher-monitoring`)

What are you thoughts @cbron ?

cbron: Hmm, we need to get @vincent99 involved here too. My initial thought was that auto-install would look for provides-gvr with hidden==true, and requires looks for provides-gvr with hidden==false. 

I know Darren specifically doesn't want to go the specifying-chart route because its too brittle, we need to focus on gvrs/crds. I could see how the addition of `provides-gvr-operator` would fix both problems though. Ie: 
* monitoring-crd: `provides-gvr: the-crd`  (this annotation is only ever on crd charts)
* monitoring: `provides-gvr-operator: the-crd`  (this is on just operator charts or operator+crd charts)
* istio: `requires-gvr: the-crd`

Then the UI will look for provides-gvr if auto-install, and provides-gvr-operator if requires.

I believe that solves all cases, except a chart that only wants another charts crds and not its operator, although I'm not entirely sure thats a use case we want to solve.

aiyengar2: I agree with the approach described above of splitting the `provides-gvr` flag.

However, I forgot about the `hidden` flag; based on that, maybe it isn't necessary to introduce a `provides-operator` flag. Instead, like you imply, we are effectively saying:
- `provides-gvr: the-crd` + `hidden: false` => `provides-gvr-operator: the-crd`
- `provides-gvr: the-crd` + `hidden: true` => `provides-gvr: the-crd`

But in my opinion, I think we should remove the `hidden` flag entirely and replace it with splitting the `provide-gvr` into `gvr` and `operator` since, afaik, there's no reason we would have for having a chart be `hidden` unless its a CRD chart. It also makes the annotations on a chart more intuitive.

cbron:
Approaching different way

PR #604, opened 2020-09-01
Update how we manage dependency annotations

cbron:


packages/rancher-istio/charts/Chart.yaml
```diff
@@ -10,4 +10,4 @@ annotations:
   catalog.cattle.io/release-name: rancher-istio
   catalog.cattle.io/ui-component: istio
   catalog.cattle.io/requires-gvr: prometheuses.monitoring.coreos.com/v1
-  catalog.cattle.io/auto-install-gvr: monitoringdashboards.monitoring.kiali.io/v1alpha1
+  catalog.cattle.io/auto-install: rancher-kiali-server-crd=1.22.001
```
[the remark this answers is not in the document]
brendarearden: For each update, that file has to change anyway so changing that value when needed won't be too much extra effort - but yes, not ideal

packages/rancher-gatekeeper/package.yaml
```diff
@@ -2,6 +2,6 @@ url: https://github.com/open-policy-agent/gatekeeper.git
 subdirectory: charts/gatekeeper
 type: git
 commit: 58ae7eacd2f93831b55557d0cd56ffdc3e04ab9a
+providesGVR: configs.config.gatekeeper.sh/v1alpha1
 generateCRDChart:
   enabled: true
```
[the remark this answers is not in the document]
brendarearden: I prefer the future proofing for more CRD specific values.

scripts/prepare-crds
```diff
@@ -65,18 +65,13 @@ name: ${name}-crd
...
-		yq w -i ${f}/charts/Chart.yaml "annotations[catalog.cattle.io/auto-install-gvr]" "${providesGVR}"
-	fi
-	if [[ -z "$(yq r ${f}/charts/Chart.yaml 'annotations[catalog.cattle.io/provides-gvr]')" ]]; then
-		yq w -i ${f}/charts-crd/Chart.yaml "annotations[catalog.cattle.io/provides-gvr]" "${providesGVR}"
-	fi
-	if [[ -z "$(yq r ${f}/charts/Chart.yaml 'annotations[catalog.cattle.io/hidden]')" ]]; then
-		yq w -i ${f}/charts-crd/Chart.yaml "annotations[catalog.cattle.io/hidden]" "true"
-	fi
+# Make the primary chart auto-install the CRD chart
+if [[ -z "$(yq r ${f}/charts/Chart.yaml 'annotations[catalog.cattle.io/auto-install]')" ]]; then
```
brendarearden, line 69: Since this one is changed to update the primary chart, should it also be moved to the scripts/prepare file before or after providesGVR check?

cbron: auto-install would only exist if there is a sub-chart

scripts/prepare
```diff
@@ -39,6 +39,13 @@ for f in packages/*; do
 			if [[ "${split_crds}" == "true" ]]; then
 				./scripts/prepare-crds ${f}
 			fi
+			# Add provides GVR annotation, this lets other charts require it
+			providesGVR="$(cat ${f}/package.yaml | yq r - providesGVR)"
+			if ! [[ -z ${providesGVR} ]]; then
+				if [[ -z "$(yq r ${f}/charts/Chart.yaml 'annotations[catalog.cattle.io/provides-gvr]')" ]]; then
```
aiyengar2, line 45: hmm, do we need the `provides-gvr` in the `package.yaml` anymore? 

We initially kept the `provides-gvr` annotation value in the `package.yaml` since the same value needed to exist in both the original chart and the CRD chart. But now we only use that annotation in the original chart, so maybe it should just belong in the patch.

scripts/prepare-crds
```diff
@@ -65,18 +65,13 @@ name: ${name}-crd
...
+	yq w -i ${f}/charts/Chart.yaml "annotations[catalog.cattle.io/auto-install]" "${name}-crd=match"
+fi
+# Hide the CRD chart
+if [[ -z "$(yq r ${f}/charts-crd/Chart.yaml 'annotations[catalog.cattle.io/hidden]')" ]]; then
+	yq w -i ${f}/charts-crd/Chart.yaml "annotations[catalog.cattle.io/hidden]" "true"
```
aiyengar2, line 74: If we can merge this PR with https://github.com/rancher/charts/pull/595/commits/80d71abe311594a5f235f2987a032528188690a6, this can just be directly added into the `chart-template/chart-crd-base/Chart.yaml` instead of introducing a new if condition.

aiyengar2 (changes requested):

scripts/prepare
cbron: It's no longer necessary, but it's easier to adjust in package.yaml than adjusting everyone's annotations in their patches. We could fix in future and switch it everywhere.

scripts/prepare-crds
cbron: We can handle that later / separately. This is a blocker for UI so need to get it moving, if we handle in both places thats ok.

aiyengar2 (approved):

aiyengar2:
@cbron do we need to remove the .tgz from `assets/` of all of the packages for the new changes?

cbron:
dependencies updated

brendarearden (approved):
Looks good! And readme will be updated in a separate pr, correct?

cbron:
@brendarearden I removed provides-gvr stuff from docs.

brendarearden (approved):

PR #607, opened 2020-09-03
Add generate CRD for charts that don't have a url in package.yaml

brendarearden:
**Problem**
The ability to generate CRD charts is tied to the `package.yaml` file existing and the` url` key value being set. This does not support CRD generation when the chart has locally defined crds(usually found at the following path `<path>/<chart>/charts/crds`)

_Note: If the `package.yaml` doesn't exist, then no crd generation or patching occurs_ 

**Solution**
When a chart has a `package.yaml`  file and if no ` url` key value is present, the local chart's crds directory will be generated. This occurs by checking that the package.yaml file exists and only runs url specific commands if the url also exits. 

Updated clean command to only run if the charts-crd directory exists for charts that have a `package.yaml` file

Generate patch is also skipped for charts that have a `package.yaml`  file and if no ` url` key value present. This is because there are no upstream changes to apply a patch on; all changes occur locally. 

**Issue**
https://github.com/rancher/rancher/issues/28628

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,18 @@ packages/${CHART_NAME}/
     values.yaml
 ```
 
+For **Rancher original** charts with a crds directory, it should have the following tree structure
+
+```text
+packages/${CHART_NAME}/
+  package.yaml              # metadata manifest to enable or disable crds generation
+  charts/                   # regular helm chart directory
+    templates/
+    crds/
+    Chart.yaml
+    values.yaml
+```
+
 For **Rancher modified** charts, it should have the following tree structure
 
 ```text
@@ -34,11 +46,13 @@ packages/${CHART_NAME}/
 A regular `package.yaml` will have the following content:
 
 ```yaml
-url: https://charts.bitnami.com/bitnami/external-dns-2.20.10.tgz # url to fetch upstream chart
+url: https://charts.bitnami.com/bitnami/external-dns-2.20.10.tgz # url to fetch upstream chart, omit this field for a rancher original chart
 packageVersion: 00 # packageVersion of modified charts, producing a $version-$packageVersion chart. For example, if istio 1.4.7 is modified with changes, rancher produces a 1.4.700 chart version that includes the modification rancher made on top of upstream charts.
+generateCRDChart:
+  enabled: true
 ```
 
-Here is an example of upstream chart based on git repository
+Here is an **example** of upstream chart based on git repository
 
 ```yaml
 url: https://github.com/open-policy-agent/gatekeeper.git  # Url to fetch upstream chart from git
@@ -46,7 +60,16 @@ subdirectory: chart/gatekeeper-operator # Sub directory for helm charts in git r
 type: git # optinal, indicate that upstream chart is from git
 commit: v3.1.0-beta.8 # the revision of git repo
 packageVersion: 00 # package version
-``` 
+generateCRDChart:
+  enabled: true
+```
+
+Here is an **example** of local chart with a crds directory
+
+```yaml
+generateCRDChart:
+  enabled: true
+```
 
 ### Workflow
 
--- a/scripts/clean
+++ b/scripts/clean
@@ -1,12 +1,16 @@
 #!/usr/bin/env bash
 
 for f in packages/*; do
-  if [[ -f ${f}/package.yaml ]]; then
-  	if [[ -z $CHART || $CHART == $(basename -- ${f}) ]]; then
-  		rm -rf ${f}/charts
-		rm -rf ${f}/charts-crd
-  	fi
-  fi
+	if [[ -f ${f}/package.yaml ]]; then
+		if [[ -z $CHART || $CHART == $(basename -- ${f}) ]]; then
+			if [[ $(cat ${f}/package.yaml | yq r - url) ]]; then
+				rm -rf ${f}/charts
+			elif [ -d ${f}/charts-crd/templates ]; then
+				mv ${f}/charts-crd/templates ${f}/charts/crds
+			fi
+			rm -rf ${f}/charts-crd
+		fi
+	fi
 done
 
 for f in packages/*; do
--- a/scripts/generate-patch
+++ b/scripts/generate-patch
@@ -37,6 +37,9 @@ for f in packages/*; do
   if [[ -f ${f}/package.yaml ]]; then
   	if [[ -z $CHART || $CHART == $(basename -- ${f}) ]]; then
 			url=$(cat ${f}/package.yaml | yq r - url)
+			if [[ -z ${url} ]]; then
+				continue
+			fi
 			subdirectory=$(cat ${f}/package.yaml | yq r - subdirectory)
 			type=$(cat ${f}/package.yaml | yq r - type)
 			commit=$(cat ${f}/package.yaml | yq r - commit)
--- a/scripts/prepare
+++ b/scripts/prepare
@@ -4,8 +4,8 @@ set -e
 # Pull in the upstream charts
 for f in packages/*; do
 	if [[ -z $CHART || $CHART == $(basename -- ${f}) ]]; then
-		if [[ -f ${f}/package.yaml ]]; then
-			url=$(cat ${f}/package.yaml | yq r - url)
+		url=$(cat ${f}/package.yaml | yq r - url)
+		if [[ -f ${f}/package.yaml && -n ${url} ]]; then
 			subdirectory=$(cat ${f}/package.yaml | yq r - subdirectory)
 			type=$(cat ${f}/package.yaml | yq r - type)
 			fields=$(echo ${subdirectory} | awk -F'/' '{print NF}')


scripts/prepare
```diff
@@ -6,35 +6,37 @@ set -e
...
+				fi
+				for file in $(find ./${f} -type f -name "*.patch"); do
+					basename=$(basename -- ${file})
+					patch -E -p3 -d ${f}/charts <${f}/$basename
+				done
 			fi
-			for file in $(find ./${f} -type f -name "*.patch"); do
-				basename=$(basename -- ${file})
-				patch -E -p3 -d ${f}/charts < ${f}/$basename
-			done
```
cbron, line 37: Git's diff here is hard to read, are the only changes the wrapping in the conditional on line 10, and moving line 34-37 up ?

brendarearden: yes

scripts/clean
```diff
@@ -1,12 +1,17 @@
...
-  fi
+	if [[ -f ${f}/package.yaml ]]; then
+		if [[ -z $CHART || $CHART == $(basename -- ${f}) ]]; then
+			if [[ $(cat ${f}/package.yaml | yq r - url) ]]; then
+				rm -rf ${f}/charts
+				rm -rf ${f}/charts-crd
+			else
+				[ -e ${f}/charts-crd ] && mv ${f}/charts-crd/templates ${f}/charts/crds
+				rm -rf ${f}/charts-crd
+			fi
```
aiyengar2, line 12: nit:
```suggestion
			elif [ -d ${f}/charts-crd/templates ]; then
			    mv ${f}/charts-crd/templates ${f}/charts/crds
			fi
			rm -rf ${f}/charts-crd
```

scripts/clean
```diff
@@ -1,12 +1,17 @@
...
-		rm -rf ${f}/charts-crd
-  	fi
-  fi
+	if [[ -f ${f}/package.yaml ]]; then
+		if [[ -z $CHART || $CHART == $(basename -- ${f}) ]]; then
+			if [[ $(cat ${f}/package.yaml | yq r - url) ]]; then
+				rm -rf ${f}/charts
+				rm -rf ${f}/charts-crd
+			else
+				[ -e ${f}/charts-crd ] && mv ${f}/charts-crd/templates ${f}/charts/crds
```
aiyengar2, line 10: This logic will need to be modified when https://github.com/rancher/charts/pull/595 gets merged in since the CRDs that need to be moved back will be in the `crd-manifest/` directory instead of the `template/` directory.

Since the logic for reverting a CRD chart might need to be exposed here, perhaps it might make sense to take the function currently located in generated-patch and make it its own script (i.e. `clean-crds`) and call that here instead:
https://github.com/rancher/charts/blob/4e030b0ca7ae349ea3b28c4593b00d142da2b80e/scripts/generate-patch#L18-L33

Would also make sense along the way to add in this check directly in that script:
https://github.com/rancher/charts/blob/4e030b0ca7ae349ea3b28c4593b00d142da2b80e/scripts/generate-patch#L81-L84

README.md
```diff
@@ -34,19 +46,30 @@ packages/${CHART_NAME}/
 A regular `package.yaml` will have the following content:
 
 ```yaml
-url: https://charts.bitnami.com/bitnami/external-dns-2.20.10.tgz # url to fetch upstream chart
+url: https://charts.bitnami.com/bitnami/external-dns-2.20.10.tgz # url to fetch upstream chart, remove this key:value for a rancher original chart
```
aiyengar2, line 49: nit:
```suggestion
url: https://charts.bitnami.com/bitnami/external-dns-2.20.10.tgz # url to fetch upstream chart, omit this field for a rancher original chart
```

README.md
```diff
@@ -34,19 +46,30 @@ packages/${CHART_NAME}/
...
 packageVersion: 00 # packageVersion of modified charts, producing a $version-$packageVersion chart. For example, if istio 1.4.7 is modified with changes, rancher produces a 1.4.700 chart version that includes the modification rancher made on top of upstream charts.
+generateCRDChart:
+  enabled: true
```
aiyengar2, line 52: these examples should include the providesGVR field here: https://github.com/rancher/charts/blob/4e030b0ca7ae349ea3b28c4593b00d142da2b80e/packages/rancher-monitoring/package.yaml#L5

scripts/generate-patch
```diff
@@ -34,8 +34,8 @@ function revert_crd_changes {
 
 
 for f in packages/*; do
-  if [[ -f ${f}/package.yaml ]]; then
-  	if [[ -z $CHART || $CHART == $(basename -- ${f}) ]]; then
+	if [[ (-f ${f}/package.yaml) && ($(cat ${f}/package.yaml | yq r - url)) ]]; then
+		if [[ -z $CHART || $CHART == $(basename -- ${f}) ]]; then
 			url=$(cat ${f}/package.yaml | yq r - url)
```
aiyengar2, line 39: Instead of checking it in the for loop above, maybe we should just check something like:
```suggestion
	if [[ -f ${f}/package.yaml ]]; then
		if [[ -z $CHART || $CHART == $(basename -- ${f}) ]]; then
			url=$(cat ${f}/package.yaml | yq r - url)
			if [[ -z ${url} ]]; then
			    continue
			fi
```

scripts/prepare
```diff
@@ -6,35 +6,37 @@ set -e
...
-				mkdir -p /tmp/tmp-charts
-				git clone --depth=1 --no-tags $url /tmp/tmp-charts
-				pwd=$(pwd)
-				cd /tmp/tmp-charts
-				git fetch origin $commit
-				git checkout $commit
-				cd $pwd
+		if [[ (-z $CHART || $CHART == $(basename -- ${f})) ]]; then
+			if [[ $(cat ${f}/package.yaml | yq r - url) ]]; then
+				url=$(cat ${f}/package.yaml | yq r - url)
```
aiyengar2, line 11: Same comment as above; since at this part of the loop we know that `[[ -f ${f}.package.yaml ]]`, let's just bring the `url=` outside / above this if statement, change the if statement to `if [[ -z ${url} ]]`, and do a `continue` if that the `url` is null. That way we don't have to nest all the other stuff below it in the if statement itself.

aiyengar2: this logic might slightly change once we rebase on https://github.com/rancher/charts/pull/579, but we can revisit this once the rebase is complete.

aiyengar2 (changes requested):
Since https://github.com/rancher/charts/pull/579 is merged in, we'll need to rebase and re-evaluate if the new `prepare-subcharts` script needs any changes.

Does `prepare-crds` need any changes as well?

scripts/clean
cbron: I thought that crd-manifest thing was optional ? We need to talk about that in other PR, I don't think all of them should follow that route. So lets just merge this PR before that one.

README.md
cbron: providesGVR moved outside of `generateCRDChart` and is no longer related to the CRD installation, since CRD installation is handled by a direct chart reference `auto-install: chart=v1.2.3`

cbron: #604

scripts/prepare
```diff
@@ -4,8 +4,8 @@ set -e
 # Pull in the upstream charts
 for f in packages/*; do
 	if [[ -z $CHART || $CHART == $(basename -- ${f}) ]]; then
-		if [[ -f ${f}/package.yaml ]]; then
-			url=$(cat ${f}/package.yaml | yq r - url)
+		url=$(cat ${f}/package.yaml | yq r - url)
+		if [[ ( -f ${f}/package.yaml ) && ! ( -z ${url} ) ]]; then
```
brendarearden, line 8: continue would skip the rest of the steps in the steps, so prepare-crds would never get called. I opted to use some of your suggestion to still set url once, and then check if its null as part of the if statement.

scripts/generate-patch
```diff
@@ -37,6 +37,9 @@ for f in packages/*; do
   if [[ -f ${f}/package.yaml ]]; then
   	if [[ -z $CHART || $CHART == $(basename -- ${f}) ]]; then
 			url=$(cat ${f}/package.yaml | yq r - url)
+			if [[ -z ${url} ]]; then
```
brendarearden, line 40: continue works here because we do not want to generate a patch at all for charts that do not have a url

scripts/clean
aiyengar2: The `crd-manifest` stuff is optional, not all charts will follow this route but either this PR or https://github.com/rancher/charts/pull/595 will need to contain this change. If we plan to merge this PR in before, I can add the necessary changes to the other PR instead

scripts/prepare
aiyengar2: Sounds good to me. Resolving both this and https://github.com/rancher/charts/pull/607#discussion_r483292418 since they are linked

scripts/generate-patch
aiyengar2: Sounds good to me, this is linked to https://github.com/rancher/charts/pull/607#discussion_r483290617 so I'll resolve both conversations.

scripts/prepare
aiyengar2, line 8: nit: don't need to create subshells here, we should omit the parenthesis
```suggestion
		if [[ -f ${f}/package.yaml && -n ${url} ]]; then
```

aiyengar2 (approved):
One nit change, LGTM otherwise

cbron (approved):

aiyengar2 (approved):

StrongMonkey (approved):

PR #608, opened 2020-09-03
upgraded upstream chart version

paynejacob:


paynejacob:
https://github.com/rancher/rancher/issues/28682

brendarearden:
I know that we will only be releasing a single version of the chart for 2.5.0, so deleting the old tgz is good, but did you need to leave it for now for upgrade testing at all?

brendarearden (approved):
lgtm

prachidamle (approved):

packages/rancher-logging/rancher-logging.patch
```diff
@@ -3,12 +3,12 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-logging/charts-original/Chart.
 +++ packages/rancher-logging/charts/Chart.yaml
 @@ -1,5 +1,13 @@
  apiVersion: v1
- appVersion: 3.4.0
+ appVersion: 3.6.0
 -description: A Helm chart to install Banzai Cloud logging-operator
 -name: logging-operator
-+description: A Helm chart to install Banzai Cloud logging-operator and a cluster logging pipeline.
++description: Helm chart to install Banzai Cloud logging-operator and a cluster logging pipeline.
```
cbron, line 9: Goal here is to describe what the chart does not what it is. So like: 
> The Banzai operater collects logs blah blah

packages/rancher-logging/rancher-logging.patch
```diff
@@ -3,12 +3,12 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-logging/charts-original/Chart.
 +++ packages/rancher-logging/charts/Chart.yaml
 @@ -1,5 +1,13 @@
  apiVersion: v1
```
cbron, line 5: keywords need to be updated

packages/rancher-logging/rancher-logging.patch
```diff
@@ -3,12 +3,12 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-logging/charts-original/Chart.
...
 -description: A Helm chart to install Banzai Cloud logging-operator
 -name: logging-operator
-+description: A Helm chart to install Banzai Cloud logging-operator and a cluster logging pipeline.
++description: Helm chart to install Banzai Cloud logging-operator and a cluster logging pipeline.
 +name: rancher-logging
- version: 3.4.0
+ version: 3.6.0
 +icon: https://charts.rancher.io/assets/logos/logging.svg
 +annotations:
 +  catalog.cattle.io/certified: rancher
```
cbron, line 14: Further down the gvr needs to be update to correct format

cbron (changes requested):

packages/rancher-logging/rancher-logging.patch
```diff
@@ -1,22 +1,25 @@
...
-+description: A Helm chart to install Banzai Cloud logging-operator and a cluster logging pipeline.
++description: Collects and filter logs using highly configurable CRDs.  Powered by Banzai Cloud Logging Operator.
 +name: rancher-logging
- version: 3.4.0
+ version: 3.6.0
 +icon: https://charts.rancher.io/assets/logos/logging.svg
++keywords:
++  - logging
++  - monitoring
++  - security
```
cbron, line 16: Not sure what right answer is for keywords, we can fix later.

cbron (approved):
lgtm, no tgz purge since new version

PR #609, opened 2020-09-03
Remove nginx-ingress

cbron:
It was part of rke2 and all of those charts have moved elsewhere.

Darren/Craig approved.


paynejacob (approved):

paynejacob:
Do we need to remove nginx from index.yaml?

brendarearden:
should the entire patch file be removed if you are removing an upstream url to patch over?

paynejacob (approved):

brendarearden (approved):

PR #614, opened 2020-09-07
Adding bluespire messaging suite app to rancher charts

bluespireinc:
Pull Request for adding a partner app "bluehub" from Blue Spire


charts/bluehub/v0.1.0/questions.yaml
```diff
@@ -0,0 +1,261 @@
+labels:
+ io.rancher.certified: partner
+questions:
+- variable: zookeeper.enabled
+  default: "false"
+  type: boolean
+  label: "If true, installs the Zookeeper"
```
PennyScissors, line 7: Can you update all of these questions to have a short brief `label` like "Install Zookeeper" and use the `description` key for more detailed descriptions of what the value does?

charts/bluehub/v0.1.0/questions.yaml
```diff
@@ -0,0 +1,261 @@
...
+  group: "Zookeeper Settings"
+  show_subquestion_if: true
+  subquestions:
+    - variable: zookeeper.imgVersion
+      required: true
+      default: "3.5.8"
+      type: string
+      label: "Zookeeper Version:"
+      group: "Zookeeper Settings"
+    - variable: zookeeper.type
```
PennyScissors, line 17: This should be an enum as following:
```yaml
- variable: zookeeper.type
  label: Service Type
  required: true
  type: enum
  default: NodePort
  group: "Zookeeper Settings"
  options:
  - "ClusterIP"
  - "NodePort"
  - "LoadBalancer"
```

charts/bluehub/v0.1.0/questions.yaml
```diff
@@ -0,0 +1,261 @@
...
+  type: boolean
+  label: "If true, installs the Zookeeper"
+  group: "Zookeeper Settings"
+  show_subquestion_if: true
+  subquestions:
+    - variable: zookeeper.imgVersion
+      required: true
+      default: "3.5.8"
+      type: string
+      label: "Zookeeper Version:"
```
PennyScissors, line 15: Remove colons from labels

charts/bluehub/v0.1.0/Chart.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
+name: bluehub
+home: https://thebluespire.com/kafka-distribution
+icon: file://./bluehub.png
```
PennyScissors, line 4: Move the icon one level up to the `bluehub` directory and fix reference here. This way you can use the same icon for future chart versions.

PennyScissors:
You can use `helm package` on each subchart to create tgz files and remove the chart directories, and this will speed-up user git clone and downloading with catalog apps.
Please refer to https://github.com/rancher/charts/tree/master/charts/wordpress/v9.0.3/charts as an example.

PennyScissors:
What steps did you take to install this chart in Rancher? I tried installing it with the default values, and other configurations, but couldn't do it successfully.

PennyScissors (changes requested):
What steps did you take to install this chart in Rancher? I tried installing it with the default values, and other configurations, but couldn't do it successfully.

charts/bluehub/v0.1.0/questions.yaml
bluespireinc: changes made as suggested

charts/bluehub/v0.1.0/questions.yaml
bluespireinc: changes made as suggested

charts/bluehub/v0.1.0/questions.yaml
bluespireinc: changes made as suggested

charts/bluehub/v0.1.0/Chart.yaml
bluespireinc: changes made as suggested

bluespireinc:
> What steps did you take to install this chart in Rancher? I tried installing it with the default values, and other configurations, but no success.

Secrets are needed for most of these components to be deployed. As these secrets can be stored outside in vault, DB or some configuration management tool, we dint want it to be bundled along with the product. However, have enabled an option for default secrets in questions.yaml with a note to use it for testing only. These have self signed certificates and other security parameters. this should allow you to deploy the product with all default values

charts/bluehub/v0.1.0/questions.yaml
```diff
@@ -0,0 +1,329 @@
+labels:
+ io.rancher.certified: partner
+questions:
+- variable: zookeeper.enabled
```
PennyScissors, line 4: If Zookeeper, Kafka and Security Manager are required as a bare minimum, their values shouldn't be subquestions wrapped by a variable that enables them since users can choose to not install them and the installation will fail.

Please remove those from the enabled wrapper, and reorder question groups as following so required groups are at the top.

Zookeeper
Kafka
Security Manager
Schema Registry
Cluster Manager Settings
Kproducer

bluespireinc: The changes are committed as suggested

PennyScissors (approved):
lgtm

PR #618, opened 2020-09-08
Update provides/requires gvr, added keywords

brendarearden:
**Problem**
requires gvr is using the crd name and not the steve crd value so its not able to determine if the crd is installed

chart.yaml needs keywords field with allowed values in it

**Solution**
Change the provides and requires gvr to match the steve crd value

Add keywords field to chart.yaml

**Issue**
https://github.com/rancher/rancher/issues/28795


packages/rancher-istio/charts/Chart.yaml
```diff
@@ -4,10 +4,13 @@ description: Helm chart for installing istio components with the istioctl
...
+keywords:
+- networking
+- infrastructure
 annotations:
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/namespace: istio-system
   catalog.cattle.io/release-name: rancher-istio
   catalog.cattle.io/ui-component: istio
-  catalog.cattle.io/requires-gvr: prometheuses.monitoring.coreos.com/v1
+  catalog.cattle.io/requires-gvr: monitoring.coreos.com.prometheus/v1
```
cbron, line 15:
```suggestion
  catalog.cattle.io/requires-gvr: monitoring.coreos.com.prometheuses/v1
```

cbron (changes requested):
one thing

cbron (approved):

PR #620, opened 2020-09-09
Deprecate providesGVR flag and update validate script

aiyengar2:
This PR deprecates the provideGVR flag and updates the validate script to unblock CI issues for local charts that use the generateCRDChart flag.

Related to build failures in https://github.com/rancher/charts/pull/594

### DIFF
--- a/scripts/clean
+++ b/scripts/clean
@@ -3,7 +3,8 @@
 for f in packages/*; do
 	if [[ -f ${f}/package.yaml ]]; then
 		if [[ -z $CHART || $CHART == $(basename -- ${f}) ]]; then
-			if [[ -d ${f}/charts-crd ]]; then
+			split_crds=$(yq r ${f}/package.yaml generateCRDChart.enabled)
+			if [[ "${split_crds}" == "true" && -d ${f}/charts-crd ]]; then
 				./scripts/clean-crds ${f}
 			fi
 			if [[ $(cat ${f}/package.yaml | yq r - url) ]]; then
--- a/scripts/clean
+++ b/scripts/clean
@@ -3,12 +3,12 @@
 for f in packages/*; do
 	if [[ -f ${f}/package.yaml ]]; then
 		if [[ -z $CHART || $CHART == $(basename -- ${f}) ]]; then
+			if [[ -d ${f}/charts-crd ]]; then
+				./scripts/clean-crds ${f}
+			fi
 			if [[ $(cat ${f}/package.yaml | yq r - url) ]]; then
 				rm -rf ${f}/charts
-			elif [ -d ${f}/charts-crd/templates ]; then
-				mv ${f}/charts-crd/templates ${f}/charts/crds
 			fi
-			rm -rf ${f}/charts-crd
 		fi
 	fi
 done
--- a/scripts/clean
+++ b/scripts/clean
@@ -7,7 +7,8 @@ for f in packages/*; do
 			if [[ "${split_crds}" == "true" && -d ${f}/charts-crd ]]; then
 				./scripts/clean-crds ${f}
 			fi
-			if [[ $(cat ${f}/package.yaml | yq r - url) ]]; then
+			url=$(yq r ${f}/package.yaml url)
+ 			if [[ -n $url ]]; then
 				rm -rf ${f}/charts
 			fi
 		fi
--- a/scripts/clean-crds
+++ b/scripts/clean-crds
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+set -e
+
+# Reverts changes done to a chart from the prepare-crd script to go back to using one chart
+if [[ -z $1 ]]; then
+    echo "No directory provided to revert charts-crd changes within"
+	exit 1
+fi
+
+f=$1
+
+if ! [[ -d ${f}/charts-crd ]]; then
+    echo "Chart does not have a charts-crd/ directory to revert changes from"
+    exit 1
+fi
+
+# Move CRDs back into ${f}/charts/crd/ and remove ${f}/charts-crd
+mkdir -p ${f}/charts/crds/
+mv ${f}/charts-crd/templates/* ${f}/charts/crds/
+rm -rf ${f}/charts-crd
+
+if ! [[ -d ${f}/charts ]]; then
+    # Main chart has already been deleted; no need to modify contents
+    exit 0
+fi
+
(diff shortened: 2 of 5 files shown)


scripts/validate
```diff
@@ -5,12 +5,46 @@ cd $(dirname $0)/..
...
+for f in packages/*; do
+    if [[ -f ${f}/package.yaml ]]; then
+        split_crds=$(cat ${f}/package.yaml | yq r - generateCRDChart.enabled)
+		if [[ "${split_crds}" == "true" && -d ${f}/charts-crd ]]; then
+			./scripts/clean-crds ${f}
+		fi
+        providesGVR="$(cat ${f}/package.yaml | yq r - providesGVR)"
+		if [[ -n ${providesGVR} && -f ${f}/charts/Chart.yaml ]]; then
+			if [[ "$(yq r ${f}/charts/Chart.yaml 'annotations[catalog.cattle.io/provides-gvr]')" == "${providesGVR}" ]]; then
+				yq d -i ${f}/charts/Chart.yaml "annotations[catalog.cattle.io/provides-gvr]"
```
aiyengar2, line 18: @mrajashree I pulled out your local changes into my fork and tested this out and it seems to be passing validate on your script as expected... with one exception. I had to make a tiny change to your chart in this commit: https://github.com/aiyengar2/charts/commit/e152e5214dfbb31a6499e22f970da3c558bd03fb.

The technical reason why this change is required is because `yq` doesn't support preserving the formatting with in place writing (based on https://github.com/mikefarah/yq/issues/465), so changes like `yq d -i` or `yq w -i` sprinkled throughout our codebase essentially reformat your `Chart.yaml`'s formatting to a standard YAML format.

Specifically, this causes an issue in the `validate` script since it checks to see if your Git is dirty when performing the validate (i.e. you have changes that have not been committed to your patch or directly to your chart). Since you saved your `Chart.yaml` with a different format than what the output of `yq [d/w] -i ${f}/charts/Chart.yaml` would have generated, the validate will always fail since it will always assume your git is dirty.

TLDR, if you see an issue with the make validate, just make sure the `Chart.yaml`'s output matches the output provided by `yq r ${f}/charts/Chart.yaml`, where `f=packages/backup-restore-operator` or something similar, and run `make validate` once more.

mrajashree: yup I'd noticed that too, tested my commit with this PR and the ci works fine

dramich: @aiyengar2 If yq is going to be changing formatting of files then failing a build this needs to be more obvious, and we need to include formatting as part of the make steps a dev will run locally similar to go-fmt so anyone will be compliant.

scripts/clean-crds
```diff
@@ -0,0 +1,33 @@
...
+
+# Reverts changes done to a chart from the prepare-crd script to go back to using one chart
+if [[ -z $1 ]]; then
+    echo "No directory provided to revert charts-crd changes within"
+	exit 1
+fi
+
+f=$1
+
+if ! [[ -d ${f}/charts-crd ]]; then
```
cbron, line 12: This is redundant but thats ok for now.

scripts/validate
cbron: Easy solution is just to remove these from package.yaml altogether and just manually put in the patch file. The dual way was too confusing. Since its disconnected from the CRD generation and pushed onto parent it doesn't make sense in package.yaml anymore and will greatly simplify scripts.

scripts/clean
```diff
@@ -3,12 +3,21 @@
...
 	if [[ -f ${f}/package.yaml ]]; then
 		if [[ -z $CHART || $CHART == $(basename -- ${f}) ]]; then
-			if [[ $(cat ${f}/package.yaml | yq r - url) ]]; then
+			# Remove special changes from optional flags before checking if Git is dirty
+			providesGVR="$(yq r ${f}/package.yaml providesGVR)"
+			if [[ -n ${providesGVR} && -f ${f}/charts/Chart.yaml ]]; then
+				if [[ "$(yq r ${f}/charts/Chart.yaml 'annotations[catalog.cattle.io/provides-gvr]')" == "${providesGVR}" ]]; then
+					yq d -i ${f}/charts/Chart.yaml "annotations[catalog.cattle.io/provides-gvr]" 
+				fi
+			fi
```
cbron, line 12: As mentioned elsewhere, let's just remove all this `providesGVR` stuff because its disconnected from CRD generation now.

cbron (commented):
Even if approved don't merge this until after 10am PST.

scripts/validate
aiyengar2: @dramich the new last commit on the PR https://github.com/rancher/charts/pull/620/commits/bf9a6228754c99af14d405566484cf5ac44cae67 ensures that yq won't break the build. @mrajashree with the last commit, you should no longer need to make the `Chart.yaml` change to get your chart to pass CI.

I tried adding formatting as part of the make validate instead, but it seems like there could be discrepancies between the output `yq r` gives on the machine that runs our tests vs. the local dev environment.

re: @cbron's comments, I agree that `providesGVR` should be removed from the package.yaml to simplify the scripts, but I think we still need to modify the Chart.yaml for `auto-install` annotations for CRD charts.

scripts/clean
aiyengar2: Not sure if we should just remove this providesGVR stuff since we'll have to modify the patches of all of the `Chart.yaml`s for each chart. With the changes added in the last commit (discussed in https://github.com/rancher/charts/pull/620#discussion_r485799851), there shouldn't be a need for removing this stuff at the moment since it will no longer break the build.

scripts/validate
cbron: We talked about this offline and to simplify things we are just removing all the providesGVR and `git update-index` stuff. Scripts has to stay small and simple until we rewrite. The downside is that validate might fail for original charts on chart.yaml formatting, we can make a separate make command for that later, but right now its a nice-to-have.

aiyengar2:
Chart changes for deprecated providesGVR are located in https://github.com/rancher/charts/pull/624

cbron (approved):
lgtm, merging this will break the provides->requires dependency but then 624 will fix them again, so we will merge in succession. 
The formatting issue will be a known issue with original charts, both current chart owners know and we'll have fix in next edition.

dramich (approved):

brendarearden (approved):

PR #622, opened 2020-09-09
Make monitoring gvr singular

cbron:
**Problem**
Id was `"id": "monitoring.coreos.com.prometheus"`, and not `"id": "monitoring.coreos.com.prometheuses",`

**Solution** 
Update provides/requires gvr to match id

Closes #621


brendarearden (approved):
Looks good!

PR #581, opened 2020-08-25
Istio version 1.7.0

brendarearden:
Update istio to version 1.7.0

Images mirrored: https://github.com/rancher/image-mirror/pull/23

PR #627, opened 2020-09-09
Add annotation changes to version Istio 1.6.8

brendarearden:
Rebuild Istio 1.6.8 with the newest set of annotation and chart changes

Updated to use cattle system registry

**Issue**
https://github.com/rancher/rancher/issues/28818


packages/rancher-istio/charts/configs/istio-base.yaml
```diff
@@ -4,6 +4,14 @@ spec:
...
     istiocoredns:
       enabled: {{ .Values.istiocoredns.enabled }}
+    prometheus:
+      enabled: false
+    grafana:
+      enabled: false
+    kiali:
+      enabled: false
+    tracing:
+      enabled: false
```
cbron, line 14: Why'd these come back ?

brendarearden: This pr shows them getting removed only when going to 1.7 https://github.com/rancher/charts/pull/581/files. We had them in as they still had default values in 1.6 so we wanted to ensure they were always false.

packages/rancher-istio/charts/values.yaml
```diff
@@ -54,13 +54,14 @@ sidecarInjectorWebhook:
...
   proxy:
     repository: rancher/istio-proxyv2
-    tag: 1.7.0
+    tag: 1.6.8
   proxy_init:
     repository: rancher/istio-proxyv2
-    tag: 1.7.0
+    tag: 1.6.8
   defaultPodDisruptionBudget:
     enabled: true
```
cbron, line 67: Kiali stays at same version too now ?

brendarearden: Yes, kiali version didn't change from 1.6.8 to 1.7.0. The new version was released later. I will be updating the kiali version in my next pr with some kiali chart changes as well (I still need to do the `global.cattle.systemDefaultRegistry` for kiali and a few other things)

paynejacob (approved):

packages/rancher-istio/charts/configs/istio-base.yaml
cbron: Guess it doesn't matter if we are deleting 1.6 anyway

packages/rancher-istio/charts/values.yaml
cbron: ok

cbron (approved):

rmweir (approved):
LGTM

PR #629, opened 2020-09-10
Documentation updates and minor fixes for monitoring and pushprox

aiyengar2:
This PR contains a couple of minor fixes such as:
- Updating the descriptions of charts
- Creating an app-README.md for rancher-monitoring
- Adding the hidden annotation to rancher-pushprox
- Moving all the exporter enabled logic into a helper template function
- Updating the CHANGELOG.md


packages/rancher-monitoring/overlay/app-README.md
```diff
@@ -0,0 +1,28 @@
+# Rancher Monitoring
+
+Modifies the upstream [Prometheus Operator](https://github.com/helm/charts/tree/master/stable/prometheus-operator) chart, which provides easy monitoring definitions for Kubernetes services and the deployment and management of Prometheus instances, and enables [Prometheus Adapter](https://github.com/helm/charts/tree/master/stable/prometheus-adapter) on a default Prometheus instance.
+
+You must install the Prometheus Operator CRDs first using the `rancher-monitoring-crd` chart before installing this chart. Once installed, this chart will contain the following components:
```
mrajashree, line 5: nit: is there a link to this chart we can give here?

mrajashree (approved):
LGTM

packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -2026,11 +2026,12 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-monitoring/charts-original/val
  
  ## Provide a k8s version to auto dashboard import script example: kubeTargetVersionOverride: 1.16.6
  ##
-@@ -76,8 +334,18 @@
+@@ -76,8 +334,19 @@
  
  ##
  global:
```
PennyScissors, line 2057: I think you are missing `global.cattle.clusterId` and `global.cattle.clusterName`

packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -2026,11 +2026,12 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-monitoring/charts-original/val
...
-+  systemDefaultRegistry: ""
++  cattle:
```
PennyScissors, line 2058: I think you are missing `global.cattle.clusterId` and `global.cattle.clusterName` (In all other values.yaml as well)

mrajashree: we only need to accept those for charts that utilize these values right? because backup-restore-operator chart doesn't have them either

aiyengar2: I agree with Rajashree; I think these fields could be added in within a future PR, but since the Charts don't support any changes based on these fields at the moment I think it should be safe to omit it.

packages/rancher-monitoring/overlay/app-README.md
aiyengar2: Discussed this offline with Rajashree; we'll provide the `helm install` command to install the crd chart first instead since there is no physical repository to point to for the CRD chart.

aiyengar2:
Modified the `app-README.md` to make it more concise in the latest commit and add in the changes discussed offline.

packages/rancher-monitoring/rancher-monitoring.patch
PennyScissors: @aiyengar2 @mrajashree I was going based on a comment in 2dotx-dev, but I wasn't part of the design meeting where this was discussed so if that's the case then LGTM.

PennyScissors (approved):

cbron (approved):

cbron:
Might squash your commits, otherwise lgtm

PR #637, opened 2020-09-11
Create Alertmanager secret in pre-install hook

aiyengar2:
This PR enables users to create the Alertmanager Config Secret as part of a pre-install hook instead of via the normal Helm deployment process. This ensures that the Alertmanager Config Secret is untouched on `helm upgrade` and `helm uninstall` to allow users to modify and persist their Alertmanager Config even after the Rancher Monitoring chart is uninstalled.

This is done by introducing the following elements into the deployment that are all tied to a pre-install hook:
- A Secret `alertmanager-rancher-monitoring-preinstall` that will be the exact copy of the Alertmanager secret that we want to use; however, this secret is managed by the Helm lifecycle, so we don't want to use it directly
- A Job that uses a ServiceAccount to patch the `alertmanager-rancher-monitoring-preinstall` with a new name `alertmanager-rancher-monitoring-alertmanager` (the naming convention used by the upstream chart) and remove the Helm 3 hook annotations
- A ServiceAccount, ClusterRoleBinding, and ClusterRole that gives our ServiceAccount permissions to create, get, and patch Secrets within the `cattle-monitoring-system` namespace.

If a user already has a secret named `alertmanager-rancher-monitoring-alertmanager` installed onto the cluster, we fail to deploy since the secret already exists and we don't want to override it.

Users will also be able to specify that they want the AlertManager Config Secret that was created to be deleted on an uninstall via `alertmanager.secret.cleanupOnUninstall`. This will be done by a separate Job with different RBAC that only gets created on a post-delete hook.

##### Why use `rancher/hyperkube` instead of `rancher/kubectl`?

Although all we perform in the Job is effectively a single `kubectl patch` command, the `kubectl patch` will always fail since we are trying to modify an immutable field (i.e. the name of the resource). Therefore, we need to pipe the output of `kubectl patch` into `kubectl apply` to get the new resource onto the cluster; however, multiple commands usually require spinning a shell on the image.

To my knowledge, `rancher/kubectl` doesn't have this ability (re: https://github.com/rancher/charts/pull/592), so I decided to use `rancher/hyperkube` and split the patch into a patch + annotate to make it easier to read.

Since we are using `rancher/hyperkube` for create, it would be easier to use the same exact field for create since there's no need to introduce a dependency on both `rancher/hyperkube` and `rancher/kubectl`.

Open to suggestions to resolve this if you know of a better way to get this secret onto the cluster in a pre-install hook!

Related Issue: https://github.com/rancher/rancher/issues/28779


aiyengar2 (commented):
Still needs README.md update for the newly exposed fields and CHANGELOG.md updates for this new process of creating a secret.

dramich (changes requested):
I think the approach is fine but I have an issue with using a 1.5GB image to just run a couple of kubectl commands. The agent image has kubectl and a shell and is much smaller and is also used in other places to do cleanup (see imported cluster cleanup job). I would also like @cbron to take a peek.

aiyengar2:
> I think the approach is fine but I have an issue with using a 1.5GB image to just run a couple of kubectl commands. The agent image has kubectl and a shell and is much smaller and is also used in other places to do cleanup (see imported cluster cleanup job). I would also like @cbron to take a peek.

Thanks for pointing this out! If we make this change here, we should also modify the image used by the CRD assume ownership chart [here](https://github.com/rancher/charts/blob/f4848c6c9e2331191c935fbef94f4976e9a2ed51/scripts/chart-templates/crd-assume-ownership/values.yaml#L10-L11) (in a separate PR) since it also just runs a couple of kubectl commands.

aiyengar2:
@dramich I updated the image to use `rancher/rancher-agent` in the latest commit and it works as expected. Kept it as a separate commit for now, but will squash the first three commits on a merge.

packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -706,6 +717,216 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-monitoring/charts-original/tem
...
++          imagePullPolicy: {{ .Values.alertmanager.secret.image.pullPolicy }}
++          command:
++          - /bin/sh
++          - -c
++          - >
++            if kubectl get secret -n {{ template "prometheus-operator.namespace" . }} alertmanager-{{ template "prometheus-operator.fullname" . }}-alertmanager > /dev/null 2>&1; then
++              kubectl delete secret -n {{ template "prometheus-operator.namespace" . }} alertmanager-{{ template "prometheus-operator.fullname" . }}-alertmanager
++            fi;
++          - -o
++          - yaml
```
dramich, line 757: Are these supposed to be here?

packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -706,6 +717,216 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-monitoring/charts-original/tem
...
++  annotations:
++    "helm.sh/hook": pre-install
++    "helm.sh/hook-delete-policy": hook-succeeded, hook-failed
++    "helm.sh/hook-weight": "3"
++rules:
++- apiGroups:
++  - ""
++  resources:
++  - secrets
++  verbs: ['create', 'get', 'patch']
```
dramich, line 897: Is `patch` required if we are doing dryrun with a yaml output?

packages/rancher-monitoring/rancher-monitoring.patch
aiyengar2: good catch, will remove this

dramich:
How does helm handle the job? Does it delete when it exits with a 0 status?

packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -706,6 +717,216 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-monitoring/charts-original/tem
...
++          - >
++            if kubectl get secret -n {{ template "prometheus-operator.namespace" . }} alertmanager-{{ template "prometheus-operator.fullname" . }}-alertmanager > /dev/null 2>&1; then
++              kubectl delete secret -n {{ template "prometheus-operator.namespace" . }} alertmanager-{{ template "prometheus-operator.fullname" . }}-alertmanager
++            fi;
++          - -o
++          - yaml
++      restartPolicy: OnFailure
++---
++apiVersion: rbac.authorization.k8s.io/v1
++kind: ClusterRole
```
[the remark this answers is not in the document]
aiyengar2: Yeah; the current way the code is assembled will not deploy the hooks at all in the case of an upgrade.

packages/rancher-monitoring/overlay/CHANGELOG.md
```diff
@@ -39,3 +39,4 @@ All notable changes from the upstream Prometheus Operator chart will be added to
 - Updated default Grafana `deploymentStrategy` to `Recreate` to prevent deployments from being stuck on upgrade if a PV is attached to Grafana
 - Modified the default `<serviceMonitor|podMonitor|rule>SelectorNilUsesHelmValues` to default to `false`. As a result, we look for all CRs with any labels in all namespaces by default rather than just the ones tagged with the label `release: rancher-monitoring`.
 - Modified the default images used by the `rancher-monitoring` chart to point to Rancher mirrors of the original images from upstream.
+- Modified the behavior of the chart to create the Alertmanager Config Secret via a pre-install hook instead of using the normal Helm lifecycle to manage the secret. The benefit of this approach is that all changes to the Config Secret done on a live cluster will never get overridden on a `helm upgrade` since the secret only gets created on a `helm install`. If you would like the secret to be cleaned up on an `helm uninstall`, enable `alertmanager.cleanupOnUninstall`; however, this is disabled by default to prevent the loss of alerting configuration on an uninstall. This secret will never be modified on a `helm upgrade`.
```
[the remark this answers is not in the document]
aiyengar2: Yeah, that would be ideal.

From my experience, any upgrade to the chart modifies the secret's contents to match the values provided on chart deployment (which is not the behavior we want, since we just want to set the default and let the user take care of it from there).

packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -706,6 +717,216 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-monitoring/charts-original/tem
...
++          command:
++          - /bin/sh
++          - -c
++          - >
++            if kubectl get secret -n {{ template "prometheus-operator.namespace" . }} alertmanager-{{ template "prometheus-operator.fullname" . }}-alertmanager > /dev/null 2>&1; then
++              echo "Secret already exists"
++              exit 1
++            fi;
++            kubectl patch secret -n {{ template "prometheus-operator.namespace" . }} --dry-run -o yaml
++            alertmanager-{{ template "prometheus-operator.fullname" . }}-pre-install
```
cbron, line 872: Can you explain this line quick. I'm confused about why we are using `hook-succeeded, hook-failed` on the secret above.

aiyengar2:
> How does helm handle the job? Does it delete when it exits with a 0 status?

Currently, the Jobs have been set with `"helm.sh/hook-delete-policy": hook-succeeded, hook-failed`, so regardless of whether the hook succeeds or fails, the resource is always deleted. 

However, the Jobs themselves will restart if they get a nonzero exit code because of `restartPolicy: OnFailure`.

packages/rancher-monitoring/rancher-monitoring.patch
aiyengar2: Explained here https://github.com/rancher/charts/pull/637#issuecomment-692366050, we want the resources to be deleted regardless of whether the hook is successful or fails since leaving the resources behind on either one of those scenarios will require a manual deletion to retry installing the chart iirc since the resources still exist in the cluster.

cbron: But that is on the secret, are we immediately deleting the `-pre-install` secret ?

packages/rancher-monitoring/rancher-monitoring.patch
dramich: `patch` is required for the `annotate` call

packages/rancher-monitoring/rancher-monitoring.patch
aiyengar2: Yes, we delete the pre-install secret the moment that it has been copied over into `alertmanager-rancher-monitoring-alertmanager`.

dramich (approved):

packages/rancher-monitoring/rancher-monitoring.patch
cbron: Ah, making more sense. So we template the `pre-install` secret, patch/rename on the fly, and kubectl apply that into the `alertmanager` secret, and then delete the former.

cbron (approved):

PR #638, opened 2020-09-11
renamed rke1 to rke

paynejacob:


brendarearden:
Delete the existing tgz, otherwise it looks good!

brendarearden (approved):

cbron (changes requested):
Logging is already moved up to 3.6, so this won't affect 3.4. Either we can leave those affected and not remove tgz's, or update those and remove tgz's. @brendarearden is familiar with backporting changes.

cbron (approved):
lgtm

PR #641, opened 2020-09-13
Rebase to new upstream at kube-prometheus-stack

aiyengar2:
The old upstream at https://github.com/helm/charts/tree/master/stable/prometheus-operator has been deprecated in favor of the chart at https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack.

This commit modifies the rancher-monitoring patch in order to rebase to this new upstream, so it contains a ton of different changes to the patch file.

- [x] Needs image changes from https://github.com/rancher/image-mirror/pull/35

Related Issue: https://github.com/rancher/rancher/issues/28822


packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -1,82 +1,60 @@
-diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-monitoring/charts-original/Chart.yaml packages/rancher-monitoring/charts/Chart.yaml
+diff -x '*.tgz' -x '*.lock' -x '*.DS_Store' -uNr packages/rancher-monitoring/charts-original/Chart.yaml packages/rancher-monitoring/charts/Chart.yaml
```
aiyengar2, line 1: `-x ‘.DS_Store’` is a relic of the script used to generate this patch, should be removed from the patch.

aiyengar2:
CI Build is broken due to a legitimate bug in the scripts for using subdirectories in git repositories fixed in https://github.com/rancher/charts/pull/642/commits/9971a14071b8fe32a148f98365d8d15ee3cde10a#diff-b54e67dc80bf2c540012ec37a9b81020R42.

Will introduce that change as part of this PR to fix CI issues.

scripts/generate-patch
```diff
@@ -29,18 +29,20 @@ for f in packages/*; do
...
 				git fetch origin $commit
 				git checkout $commit
 				cd $pwd
-				cp -r /tmp/tmp-charts/${subdirectory} ${f}/charts-original
+				cp -r /tmp/tmp-charts/${subdirectory}/* ${f}/charts-original
 				rm -rf /tmp/tmp-charts
 			else
-				mkdir -p ${f}/charts-original
 				curl -sLf ${url} | tar xvzf - -C ${f}/charts-original --strip ${fields} ${subdirectory} > /dev/null 2>&1
 			fi
```
aiyengar2, line 47: IMO it's worth thinking about making a commit like https://github.com/rancher/charts/pull/642/commits/9971a14071b8fe32a148f98365d8d15ee3cde10a that just pulls the `prepare-upstream` logic from `prepare` and `generate-patch` into the one script that gets called by both in order to avoid maintaining the exact same logic twice in our codebase.

Not sure if that's something we want to add into the scripts pre-2.5 (in a separate PR), but I've tested that the fix works as expected when I wrote the stuff to help me generate the patch for this PR as part of https://github.com/rancher/charts/pull/642.

cc: @cbron @StrongMonkey

packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -1,82 +1,60 @@
...
 +annotations:
 +  catalog.cattle.io/certified: rancher
 +  catalog.cattle.io/namespace: cattle-monitoring-system
 +  catalog.cattle.io/release-name: rancher-monitoring
 +  catalog.cattle.io/ui-component: monitoring
 +  catalog.cattle.io/provides-gvr: monitoring.coreos.com.prometheus/v1
-+description: Modifies the upstream Prometheus Operator chart, which provides easy monitoring definitions for Kubernetes services and the deployment and management of Prometheus instances, and enables Prometheus Adapter on a default Prometheus instance.
- engine: gotpl
- home: https://github.com/coreos/prometheus-operator
++description: Modifies the upstream Prometheus Operator chart, which collects Kubernetes manifests, Grafana dashboards, and Prometheus rules combined with documentation and scripts to provide easy to operate end-to-end Kubernetes cluster monitoring with Prometheus using the Prometheus Operator, and enables Prometheus Adapter on a default Prometheus instance.
```
mrajashree, line 13: nit: not sure what this part means and if we need it?
> which collects Kubernetes manifests

packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -1,82 +1,60 @@
...
-+  - kube-controller-manager
-+  - kube-proxy
-+  - kube-etcd (only rke and kubeAdm)
- - service monitors to scrape internal kubernetes components
-   - kube-apiserver
-   - kube-scheduler
-@@ -136,6 +144,30 @@
- 
- The following tables list the configurable parameters of the prometheus-operator chart and their default values.
+ You may also `helm show values` on this chart's [dependencies](#dependencies) for additional options.
```
mrajashree, line 42:
```suggestion
 You may also run `helm show values` on this chart's [dependencies](#dependencies) for additional options.
```

packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -1,82 +1,60 @@
...
  
 +### Rancher Monitoring Configuration
++
++The following table shows values exposed by Rancher Monitoring's additions to the chart:
++
 +| Parameter | Description | Default |
 +| ----- | ----------- | ------ |
++| `nameOverride` | Provide a name in place of `prometheus-operator` |`"rancher-monitoring"`|
```
mrajashree, line 50: not sure what prometheus-operator here is referring, a deployment or helm release? can we include that?

scripts/generate-patch
mrajashree: I don't know enough about generate-patch script but just looking at the scripts I agree the common part can be pulled into a single one and referenced in both places

packages/rancher-monitoring/rancher-monitoring.patch
aiyengar2: This part comes straight from the Chart description on the new upstream in https://github.com/prometheus-community/helm-charts/blob/97e66d826c887b8cb1007d58c2e8cc7f0f7d645a/charts/kube-prometheus-stack/Chart.yaml#L2.

I think the intention of this is to say that Prometheus Operator also packages certain independent charts / manifests (i.e. `kube-state-metrics`, `node-exporter`, and `grafana` for upstream, `rancher-pushprox` and `prometheus-adapter` in addition for us) to deploy a full monitoring stack onto the cluster. These are declared as subcharts, but aren't technically required to deploy Prometheus Operator itself, so they are like independent manifests deployed alongside the chart.

packages/rancher-monitoring/rancher-monitoring.patch
aiyengar2: good catch, will be updated in the next commit!

packages/rancher-monitoring/rancher-monitoring.patch
aiyengar2: This refers to the naming convention that is added to all resources deployed by this chart; i.e. normally the prometheus stack would deploy a resource like `alertmanager-<release-name>-alertmanager` onto the cluster but with the nameOverride that resource would be `alertmanager-rancher-monitoring-alertmanager`.

Will modify this description to make it a bit more clear in the next commit!

aiyengar2: Actually, going to update this description entirely to:
`Provide a name that should be used instead of the chart name when naming all resources deployed by this chart`

scripts/generate-patch
```diff
@@ -29,18 +29,20 @@ for f in packages/*; do
...
 				fields='1'
 			fi
+			rm -rf ${f}/charts-original
+			mkdir -p ${f}/charts-original
 			if [[ $type == 'git' ]]; then
 				mkdir -p /tmp/tmp-charts
 				git clone --depth=1 --no-tags $url /tmp/tmp-charts
 				pwd=$(pwd)
 				cd /tmp/tmp-charts
+				git fetch --tags
```
PennyScissors, line 39: I thought pulling tags was removed to improve performance.

scripts/prepare
```diff
@@ -19,15 +19,16 @@ for f in packages/*; do
 			rm -rf ${f}/charts
 			mkdir -p ${f}/charts
 			if [[ $type == 'git' ]]; then
+				rm -rf /tmp/tmp-charts
 				mkdir -p /tmp/tmp-charts
 				git clone --depth=1 --no-tags $url /tmp/tmp-charts
 				pwd=$(pwd)
 				cd /tmp/tmp-charts
+				git fetch --tags
```
PennyScissors, line 27: I thought pulling tags was removed to improve performance.

aiyengar2: Hmm I added this is because it was related to the CI build issues I found [here](https://github.com/rancher/charts/pull/641#issuecomment-691610987); CI fails because the tag was not found locally.

Since this is probably just an issue with adding support for tags, I've removed this line from this PR and modified my `package.yaml` to use a literal commit rather than a tag for this. Thinking more about this, I also think this is the right decision since the base that a tag is referring to could be modified on a later push, so the patch might fail as a result.

cc: @StrongMonkey for this issue with specifying tags for Git-based `package.yaml`s

aiyengar2:
Updated to the latest commit on the new upstream in the latest push. The only thing that changed was the `Chart.yaml` file; I relocated the annotations we add from Rancher's end since upstream added their own artifact-hub annotations to the chat as well.

mrajashree (approved):

packages/rancher-monitoring/package.yaml
```diff
@@ -1,4 +1,7 @@
-url: https://kubernetes-charts.storage.googleapis.com/prometheus-operator-8.16.1.tgz
+url: https://github.com/prometheus-community/helm-charts.git
+subdirectory: charts/kube-prometheus-stack
+type: git
+commit: 5c8e31ae6cb6e355900e2cbbd5cbceaec0c85821
```
cbron, line 4: There a reason we don't want to use their index.yaml ? https://prometheus-community.github.io/helm-charts/index.yaml
I've seen some repos that do text manipulation between the tree and publishing to the index.

packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -1,82 +1,55 @@
...
 +++ packages/rancher-monitoring/charts/Chart.yaml
-@@ -1,22 +1,28 @@
+@@ -1,5 +1,5 @@
  apiVersion: v1
- appVersion: 0.38.1
--description: Provides easy monitoring definitions for Kubernetes services, and deployment
--  and management of Prometheus instances.
-+annotations:
+-description: kube-prometheus-stack collects Kubernetes manifests, Grafana dashboards, and Prometheus rules combined with documentation and scripts to provide easy to operate end-to-end Kubernetes cluster monitoring with Prometheus using the Prometheus Operator.
++description: Modifies the upstream Prometheus Operator chart, which collects Kubernetes manifests, Grafana dashboards, and Prometheus rules combined with documentation and scripts to provide easy to operate end-to-end Kubernetes cluster monitoring with Prometheus using the Prometheus Operator, and enables Prometheus Adapter on a default Prometheus instance.
```
cbron, line 7: Let's nix the "modifies" part, and just talk about what chart does.

packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -1592,35 +1439,49 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-monitoring/charts-original/tem
...
 +++ packages/rancher-monitoring/charts/templates/prometheus-operator/deployment.yaml
-@@ -32,7 +32,7 @@
-     {{- end }}
+@@ -33,9 +33,9 @@
        containers:
-         - name: {{ template "prometheus-operator.name" . }}
+         - name: {{ template "kube-prometheus-stack.name" . }}
+           {{- if .Values.prometheusOperator.image.sha }}
+-          image: "{{ .Values.prometheusOperator.image.repository }}:{{ .Values.prometheusOperator.image.tag }}@sha256:{{ .Values.prometheusOperator.image.sha }}"
++          image: "{g{ template "system_default_registry" . }}{{ .Values.prometheusOperator.image.repository }}:{{ .Values.prometheusOperator.image.tag }}@sha256:{{ .Values.prometheusOperator.image.sha }}"
```
cbron, line 1447: typo in the `{g`

cbron (changes requested):
Didn't review most of the middle section

packages/rancher-monitoring/rancher-monitoring.patch
aiyengar2: good catch, will update in next commit

packages/rancher-monitoring/rancher-monitoring.patch
aiyengar2: Based on this and Rajashree's comment [here](https://github.com/rancher/charts/pull/641#discussion_r487563132) about this description, I'm changing this to:

`Collects several related Helm charts, Grafana dashboards, and Prometheus rules combined with documentation and scripts to provide easy to operate end-to-end Kubernetes cluster monitoring with Prometheus using the Prometheus Operator.`

packages/rancher-monitoring/package.yaml
aiyengar2: No particular reason. We can switch to using the index in the next commit.

scripts/prepare
cbron, line 27: This isn't necessary anymore right ? Since we are doing tgz and not git.

packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -94,248 +101,111 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-monitoring/charts-original/REA
...
-@@ -49,8 +49,8 @@
+@@ -49,7 +49,7 @@
  # schedulerName: "default-scheduler"
  
  image:
 -  repository: grafana/grafana
--  tag: 7.0.5
 +  repository: rancher/grafana-grafana
-+  tag: 7.0.6
+   tag: 7.1.5
```
cbron, line 205: assuming these new tags have been mirrored already

cbron (commented):
conflicts

packages/rancher-monitoring/rancher-monitoring.patch
aiyengar2: Being tracked in https://github.com/rancher/image-mirror/pull/35

aiyengar2:
@PennyScissors let me know when https://github.com/rancher/image-mirror/pull/35 is merged so we can unblock this merge after the new images have been created.

scripts/prepare
aiyengar2: good catch. removed!

cbron (approved):
Didn't do deep dive on middle section, but overall seems fine.

dramich (changes requested):
I think this needs to be split into two PRs. Making changes to the scripts and charts changes. The issue linked also doesn't cover the changes being made to the scripts.

aiyengar2:
> I think this needs to be split into two PRs. Making changes to the scripts and charts changes. The issue linked also doesn't cover the changes being made to the scripts.

@dramich good catch, this PR actually doesn't need the script changes anymore since we switched to using the url based approach as part of [Caleb's review](https://github.com/rancher/charts/pull/641#discussion_r488303507). Dropped the script changes from this PR.

dramich (approved):

PR #642, opened 2020-09-13
[No Merge] Add script to perform interactive rebase

aiyengar2:
This PR introduces a useful utility script that makes it super easy in order to rebase patch files on new upstreams. In order to facilitate this script, it makes the following commits:
1. A couple of misc fixes to the charts that fix some minor broken functionality and do a little bit of re-ordering
2. Pulls out the logic for preparing an upstream chart and putting it in ${f}/charts-original into a `prepare-upstream` script. The code here is double-maintained by `prepare` and `generate-patch`.
3. The actual script changes + Makefile changes to allow running the script interactively or non-interactively.

#### TLDR: How does this actually work?

This script takes in two values:
- CHART: the chart that you want to rebase the patch for
- NEW_PACKAGE_YAML: a new package.yaml that represents the new upstream you are using

For example, I ran `make CHART=rancher-monitoring NEW_PACKAGE_YAML=./new-package.yaml rebase-patch` with the following contents in `./new-package.yaml` for https://github.com/rancher/charts/pull/641.

```yaml
type: git
commit: kube-prometheus-stack-9.4.1
packageVersion: 00
generateCRDChart:
  enabled: true
  assumeOwnershipOfCRDs: true
```

After resolving any conflicts, it modifies the `package.yaml` file in `${f}/packages/rancher-monitoring` with the contents in `./new-package.yaml` and modifies the patch file with the newly rebased patch file. These are shown in the changes within https://github.com/rancher/charts/pull/641.

#### Walkthrough: How does this actually work?

In simple terms, this script executes the following operations with those inputs:

1. Pulls in charts

- Prepares the local chart (based on ${f}/package.yaml + applying the patches we have) without prepare CRD logic and places it into ${f}/charts-local.
- Prepares the original upstream chart (based on ${f}/package.yaml) without prepare CRD logic and places it into ${f}/charts-old-upstream.
- Prepares the new upstream chart (based on $NEW_PACKAGE_YAML) without prepare CRD logic and places it into ${f}/charts-new-upstream.

2. Creates two git branches

- The first git branch creates 3 commits: 
  1. Contains the three charts described above.
  2. Contains the result of copying the original upstream chart into ${f}/charts.
  3. Contains the result of copying the local chart into ${f}/charts. 

Therefore, the difference between commit 2 and commit 3 effectively represents our current patch file.

- The second git branch creates 3 commits:
  1. Same as first branch.
  2. Same as second branch.
  3. Contains the result of copying the new upstream chart into ${f}/charts. 
  
Therefore, the difference between commit 2 and commit 3 effectively represents the patch file between the old upstream and the new upstream.

3. Rebase branch 1 based on branch 2

Checks into branch 1 and runs `git rebase <branch-2>`. If the rebase is successful continue. Otherwise, it spawns a shell with the prompt `<interactive-rebase-shell>` that prevents exit till the rebase conflicts have been resolved (in a non-interactive setting, it would just do a `git rebase --abort` and exit out)

At this point, a dev can resolve any conflicts using existing Git tools (i.e. I use VSCodium with the Git plugin to be able to easily review and modify any conflicts). While checking against conflicts, devs may also want to make other chanegs (i.e. I noticed that image version changes showed up in my conflicts, which notifies me to make a commit to the `image-mirror` repo to address the change).

Once conflicts are resolved, they simply need to exit the shell to continue the script.

4. Generate the new patch file based on the new package.yaml

Replaces the existing package.yaml with the new package.yaml, disables the prepare CRD logic, and generates the new patch based on the new package.yaml as a base. This generates the new patch!

5. Stash and cleanup

Stash the new `package.yaml` and patch file and checkout of the branch to the original one that we were on when calling this script. Delete the branches created as part of this script.

#### What if I abort?

It automatically cleans up everything it created and doesn't make any changes.


aiyengar2:
Closing this pull request as it's not actually intended to be merged.

PR #645, opened 2020-09-14
Use rancher-agent instead of kubectl image

aiyengar2:
The `rancher-agent` image is significantly smaller than `kubectl` and achieves the same goals.

Based on comments on https://github.com/rancher/charts/pull/637#pullrequestreview-487381744.


scripts/chart-templates/crd-assume-ownership/values.yaml
```diff
@@ -7,5 +7,5 @@ global:
     systemDefaultRegistry: ""
 
 image:
-  repository: rancher/kubectl
-  tag: v1.18.6
+  repository: rancher/rancher-agent
+  tag: v2.4-head
```
dramich, line 11: Update to use `-latest`, `-head` could potentially be broken at any point since it is not an official release.

aiyengar2:
Closing out this PR; didn't realize that the `rancher/kubectl` image was way smaller than the `rancher/rancher-agent` image.

https://github.com/rancher/charts/pull/637 needs the change since we're switching from `rancher/hyperkube` to `rancher/rancher-agent`, which makes sense.

PR #648, opened 2020-09-15
Use separate namespace only for dashboards

aiyengar2:
This commit renames `grafana-dashboards` to `cattle-dashboards` and deprecates the `grafana-datasources` namespace in favor of the normal release namespace.

It also fixes the RBAC stuff noted in https://github.com/rancher/rancher/issues/29013 and moves the RBAC stuff out to the overlay/ folder since it is self-contained.

Related Issue: https://github.com/rancher/rancher/issues/28887, https://github.com/rancher/rancher/issues/29013

Pending merge:
- [x] Rebase to new upstream at Kube-prometheus (https://github.com/rancher/charts/pull/641)

#### Note to Reviewers

The first three commits of this PR come from https://github.com/rancher/charts/pull/641. Since that PR represents a major rebase to a new upstream branch, I didn't want to have to rebase this PR after the fact, so this PR can only be merged after the rebase is complete.


aiyengar2:
Rebase has been merged, so this PR should be unblocked now

packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -2359,15 +2350,15 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-monitoring/charts-original/val
      dashboards:
        enabled: true
        label: grafana_dashboard
```
brendarearden, line 2352: So the labels will stay the same and does not need to match the rename?

aiyengar2: Yes, the label stays the same since it's given by `grafana.sidecar.dashboards.label`.

https://github.com/prometheus-community/helm-charts/blob/d3184e900b4120c85c782a1bfac2b62efe23fc0e/charts/kube-prometheus-stack/values.yaml#L532

brendarearden (approved):
LGTM

packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -2359,15 +2350,15 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-monitoring/charts-original/val
...
++      searchNamespace: cattle-dashboards
  
        ## Annotations for Grafana dashboard configmaps
        ##
 @@ -548,6 +909,7 @@
        ## ref: https://git.io/fjaBS
        createPrometheusReplicasDatasources: false
        label: grafana_datasource
-+      searchNamespace: grafana-datasources
++      # searchNamespace: "cattle-monitoring-system"
```
dramich, line 2361: Why is this updated and commented out?

aiyengar2: Based on https://github.com/rancher/charts/pull/648/files#diff-ed5b5b87cf149adc6a6408bf81993efdR848, we'll either take the value provided here or default to `cattle-monitoring-system`.

I could omit the comment here, but I thought it would just be helpful for a user to glance at to know where they should expect their datasources to be.

packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -1843,7 +1834,7 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-monitoring/charts-original/tem
 +apiVersion: rbac.authorization.k8s.io/v1
 +kind: ClusterRole
```
dramich, line 1835: This should be a Role to scope this properly. As a clusterRole it would grant perms to ALL resources listed in the cluster if someone used a clusterRoleBinding with it.

packages/rancher-monitoring/rancher-monitoring.patch
dramich: If you want to leave the comment can you expand with what you just said? Otherwise I think it's missing context

aiyengar2: on second thought, I think it's better to omit this; a user can always find this information out by looking at the relevant README on the Grafana helm chart, so it doesn't make sense for us to expose this here as a commented out option.

packages/rancher-monitoring/rancher-monitoring.patch
aiyengar2: @dramich based on the fact that we also want to remove ConfigMap / Secret permissions from the `monitoring-<admin/view/edit>` cluster roles based on https://github.com/rancher/rancher/issues/29013, I refactored the logic for the Roles to instead introduce two different sets of roles:

`monitoring-<admin/view/edit>`: only allows access to Prometheus CRs; previously, these permissions also allowed users to modify ConfigMaps / Secrets, but this has since been removed. I've also documented in the `CHANGELOG.md` that the expectation is for cluster admins to bind this role to users using a `ClusterRoleBinding` in order to allow users full access to view / edit Prometheus CRs across the entire cluster.

`monitoring-config-<admin/view/edit>`: this only allows access to ConfigMaps and Secrets. I've also documented in the `CHANGELOG.md` that the expectation is for cluster admins to bind this role to users using a `RoleBinding` in order to allow users limited access to view / edit ConfigMaps / Secrets only in specific namespaces.

Both sets of roles will be aggregated up to the expected default Kubernetes Roles if that option is provided.

packages/rancher-monitoring/overlay/templates/rancher-monitoring/clusterrole.yaml
```diff
@@ -0,0 +1,148 @@
+{{- if and .Values.global.rbac.create .Values.global.rbac.userRoles.create }}
```
aiyengar2, line 1: Confirmed that moving this to `overlay/` doesn't break anything:
```shell
$ helm template -n cattle-monitoring-system rancher-monitoring ./assets/rancher-monitoring/rancher-monitoring-9.4.200.tgz | yq r -d'*' - 'metadata.name' | awk '/monitoring(-config)?-(admin|edit|view)/{ print }' 
monitoring-admin
monitoring-edit
monitoring-view
monitoring-config-admin
monitoring-config-edit
monitoring-config-view
```

packages/rancher-monitoring/overlay/CHANGELOG.md
```diff
@@ -12,10 +12,8 @@ All notable changes from the upstream Prometheus Operator chart will be added to
...
+- Added `monitoring-config-edit` and `monitoring-config-view` default `ClusterRoles` to allow admins to assign roles to users to be able to edit / view `Secrets` and `ConfigMaps`. These can be enabled by setting `.Values.global.rbac.userRoles.create` (default: `true`). In a typical RBAC setup, you might want to use a `RoleBinding` to bind `monitoring-config-edit` or `monitoring-config-view` to a Subject within a specific namespace. A typical use case for this would be to bind `monitoring-config-edit` permissions to a Subject in the `cattle-monitoring-system` namespace to allow them to modify Secrets / ConfigMaps tied to the deployment, such as your Alertmanager Config Secret. Another typical use case for this would be to bind `monitoring-config-edit` permissions to a Subject in the `grafana-dashboards` namespace to allow them to create / delete ConfigMaps that allow the user to persist Grafana Dashboards on the cluster.
```
aiyengar2: Should this information be documented anywhere else?

packages/rancher-monitoring/overlay/templates/rancher-monitoring/clusterrole.yaml
```diff
@@ -105,21 +90,32 @@ rules:
...
   - 'watch'
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: monitoring-config-admin
+  labels: {{ include "kube-prometheus-stack.labels" . | nindent 4 }}
+  {{- if .Values.global.rbac.userRoles.aggregateToDefaultRoles }}
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+  {{- end }}
```
aiyengar2, line 101: does it make sense for these to also be aggregated into the default k8s roles?

packages/rancher-monitoring/overlay/CHANGELOG.md
maggieliu: Can you send this to Catherine?

packages/rancher-monitoring/rancher-monitoring.patch
dramich: I don't agree with creating the 2nd sets of roles as ClusterRoles. It feels like we are handing out a foot gun for a role called monitoring-config-edit but in reality it gives you access to secrets. Docs are a suggestion, not what people are really going to do in the wild once this is deployed. 
The first set need to be ClusterRoles in order to roll up to their k8s parents. But the 2nd set does not need to role up as those k8s default roles already have access to secrets/configmaps. If you want to add those 2nd roles I think they should be Roles not ClusterRoles.

aiyengar2: Hmm I see your point. If we want to make them `Roles`, I think it will need to be split into `monitoring-config-<admin|edit|view>` which is a set of `Roles` in the `cattle-monitoring-system` namespace and `monitoring-dashboard-<admin|edit|view>`, which is a set of `Roles` in the `cattle-dashboards` namespace. Aggregation will also be removed, as you mentioned.

Will make the appropriate changes in the next commit.

packages/rancher-monitoring/overlay/templates/rancher-monitoring/clusterrole.yaml
aiyengar2: Based on https://github.com/rancher/charts/pull/648#discussion_r491640992, this will be removed in the next commit

packages/rancher-monitoring/overlay/templates/rancher-monitoring/dashboard-role.yaml
```diff
@@ -0,0 +1,47 @@
...
+kind: Role
+metadata:
+  name: monitoring-dashboard-admin
+  namespace: {{ .Values.grafana.sidecar.dashboards.searchNamespace }}
+  labels: {{ include "kube-prometheus-stack.labels" . | nindent 4 }}
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
```
aiyengar2, line 13: Persisted dashboards should be configmaps; there shouldn't be a need for secrets permissions in this namespace.

packages/rancher-monitoring/overlay/templates/rancher-monitoring/dashboard-role.yaml
```diff
@@ -0,0 +1,47 @@
...
+  name: monitoring-dashboard-edit
+  namespace: {{ .Values.grafana.sidecar.dashboards.searchNamespace }}
+  labels: {{ include "kube-prometheus-stack.labels" . | nindent 4 }}
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - '*'
```
dramich, line 29: Should this be star? That would allow them to delete and create. This applies to the other roles as well if it needs changed.

aiyengar2: Yes, a user with edit permissions should have the ability to delete and create as well; for example, this would give someone with `monitoring-dashboard-edit` permissions to persist their Grafana Dashboards (create) and also remove some dashboards that they had added that no longer need to be persisted (delete).

I believe we discussed this in a previous design meeting, but there's really no difference between a user with admin vs. edit permissions at the moment.

dramich (approved):

PR #649, opened 2020-09-15
[proposal] Replace yq in place commands with awk replaces

aiyengar2:
Until yq has a resolution that allows us to cleanly deal with in-place insertion, we need to provide a solution in the scripts that makes it possible for developers to be able to run the scripts without encountering random changes associated with `Chart.yaml` changes. This commit replaces all our inplace modifications using yq with awk-based (or should I say awk-ward?) functions that effectively do the same YAML manipulation.

Tested it out on several test cases (missing annotations, random indentations, etc.) and it seems to be working as expected.

### DIFF
--- a/scripts/awk/delete_annotation.awk
+++ b/scripts/awk/delete_annotation.awk
@@ -1,0 +1,26 @@
+BEGIN {
+    # By default, we are not within the annotations section yet
+    within_annotations=0;
+}
+
+{
+    if (within_annotations==1 && ($0~/^ /) == 0) {
+        # Since the first line here is not a space (i.e. indented)
+        # we are no longer in the annotation section
+        within_annotations=0
+    }
+}
+
+{
+    if (within_annotations==1 && $0~field) {
+        # If we are within the annotation section and we see our
+        # annotation, just skip printing that line
+        next
+    }
+}
+
+# Check to see if we have entered the annotations part of the YAML file
+/^annotations:/{within_annotations=1}
+
+# print any values that haven't been skipped
+{print}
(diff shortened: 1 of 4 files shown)


cbron (commented):
Is there any benefit to doing this now besides lowering patch changes ? We solved the immediate problem around dirty git status issues. Seems like something to do in the rewrite.

aiyengar2:
> Is there any benefit to doing this now besides lowering patch changes ? We solved the immediate problem around dirty git status issues. Seems like something to do in the rewrite.

I think this still affects @mrajashree and @prachidamle with `backup-restore-operator` and `rancher-cis-benchmark`, since each run of `make prepare` and `make clean` locally still leaves behind a modified `Chart.yaml` in the latest upstream. There's no way to commit those changes since it will break based on the indentation expected during the CI.

cbron:
But thats not breaking anything right ? Its just an oddity occuring in the CI that gets thrown away.

aiyengar2:
> But thats not breaking anything right ? Its just an oddity occuring in the CI that gets thrown away.

Yeah, nothing is specifically broken; however, iirc it also requires all the charts to maintain unnecessary patch changes like [this](https://github.com/rancher/charts/blob/be719fc40583ac7d901a5a4a8260f3f9d8a6abf1/packages/rancher-kiali-server/rancher-kiali-server.patch#L12-L36) or [this](https://github.com/rancher/charts/blob/be719fc40583ac7d901a5a4a8260f3f9d8a6abf1/packages/rancher-monitoring/rancher-monitoring.patch#L19-L42) that are simply due to the weird formatting requirements that our `Chart.yaml` is forced to have w.r.t. the indentation issues.

cbron:
Thats fine for 2.5.0, we need to bump this to 2.5.x.

PR #655, opened 2020-09-16
Add push to assets branch scripts

StrongMonkey:


StrongMonkey:
What this PR does:
1. set sha256sum of each chart folder so that changes are tracked. If sha changes, scripts will only run for that specific chart.
2. Push assets folder into dev-v2.5-assets branch.

This PR doesn't modify or delete the existing assets folder on dev-v2.5 branch. Once it is fully migrated it can be deleted.

scripts/ci
```diff
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
 set -e
 
+if [ ! -z ${1} ]; then
```
StrongMonkey, line 4: It is added some that `./scripts/ci` can take arguments and only run for specific chart

scripts/ci-wrapper
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+set -e
+set -x
+
+sudo ./scripts/bootstrap
+git fetch origin ${1}
+git checkout origin/${1} ./assets/index.yaml || true
```
StrongMonkey, line 7: fetch `./assets/index.yaml` and `./sha256sum` from asset branch. This is because dev-v2.5 branch will have no knowledge against what previous index.yaml we have and we will lose all the info on old helm chart. `./sha256sum` on asset branch is the last-run sha256sum of each folder in order to compare what's change in this new commit.

scripts/ci-wrapper
```diff
@@ -0,0 +1,14 @@
...
+git checkout origin/${1} ./sha256sum || true
+
+charts=$(./scripts/produce-sha256)
```
StrongMonkey, line 10: find all the diff charts, and only run `make ci` on diffed charts.

scripts/generate-charts
```diff
@@ -15,18 +15,26 @@ for f in packages/*; do
...
 		helm package ${f}/charts --destination assets/$(basename -- ${f})
-		git checkout assets/$(basename -- ${f})/"$(basename -- ${f})-$(yq r ${f}/charts/Chart.yaml version).tgz" 2>/dev/null || true
 		if [[ -d ${f}/charts-crd ]]; then
 			packageVersion=$(yq r ${f}/package.yaml packageVersion)
 			yq w -i ${f}/charts-crd/Chart.yaml 'version' "${version}${packageVersion}"
 
 			helm package ${f}/charts-crd --destination assets/$(basename -- ${f})
-			git checkout assets/$(basename -- ${f})/"$(basename -- ${f})-crd-$(yq r ${f}/charts-crd/Chart.yaml version).tgz" || true
 		fi
+		helm repo index --merge ./assets/index.yaml --url assets assets
```
StrongMonkey, line 24: We no longer checkout this in order to prevent user overriding existing chart. This can't done in the new scripts. Developer has to be very carefully in overrding charts as any changed introduced in the chart will get written into the current version of helm tarball.

scripts/push-to-assets-branch
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+set -e
+set -x
+
+rm -rf /tmp/tmp-assets
+mkdir -p /tmp/tmp-assets
+cp -r assets/ index.yaml index.html sha256sum/ /tmp/tmp-assets
```
StrongMonkey, line 8: copy what's has generated from dev-v2.5 branch and paste into tmp dir and eventually asset branch. Then commit and push.

StrongMonkey:
@cbron This is ready to be reviewed. Charts folder is also pushed. To give you a better context this is what assets branch looks like https://github.com/StrongMonkey/charts-1/tree/dev-v2.5-assets.

scripts/generate-charts
cbron: Not following this, we can chat offline.

scripts/generate-charts
```diff
@@ -14,19 +13,35 @@ for f in packages/*; do
 			packageVersion=$(yq r ${f}/package.yaml packageVersion)
 			yq w -i ${f}/charts/Chart.yaml 'version' "${version}${packageVersion}"
 		fi
+
 		helm package ${f}/charts --destination assets/$(basename -- ${f})
-		git checkout assets/$(basename -- ${f})/"$(basename -- ${f})-$(yq r ${f}/charts/Chart.yaml version).tgz" 2>/dev/null || true
+
+		mkdir -p charts/$(basename -- ${f})
+		cp -R ${f}/charts/* charts/$(basename -- ${f})/
```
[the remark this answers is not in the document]
cbron: Answer is that we are pushing the charts out to a folder so people can look at them.

scripts/produce-sha256
```diff
@@ -0,0 +1,28 @@
...
+# This script compares the sha256sum of each chart folder, find the sha256sum of that chart that has changed
+
+rm -rf ./sha256sum-new
+mkdir -p ./sha256sum-new
+for f in packages/*; do
+    rm -rf ./sha256sum-new/$(basename -- ${f})
+    mkdir -p ./sha256sum-new/$(basename -- ${f})
+    find ${f} -type f | xargs sha256sum > ./sha256sum-new/$(basename -- ${f})/$(basename -- ${f}).sum
+done
+
```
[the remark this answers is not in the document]
cbron: Its a starter block, only runs on first time.

cbron (commented):
Chatted offline, approach looks good. We need to swap branches it uses to use source branch, and also start with an orphaned branch.

scripts/push-to-assets-branch
```diff
@@ -0,0 +1,19 @@
...
+if [[ -d charts ]]; then
+  cp -r charts/ /tmp/tmp-assets
+fi
```
aiyengar2, line 11: We should check for the existence of all of these directories at the top of the file:
```shell
if ! [[ -d assets && -d sha256sum && -d charts ]]; then
    echo "Cannot find required assets to push to assets branch: assets/ sha256sum/ charts/"
    exit 1
fi
```

Is there any reason why the `charts/` directory should be treated specially? iirc we always want `assets/` and `charts/` in sync, so you shouldn't be able to push assets without modifying both.

scripts/push-to-assets-branch
```diff
@@ -0,0 +1,19 @@
...
+git checkout .
+git clean -fd
+git fetch ${1} ${2}
```
aiyengar2, line 14: We should add a check on the top of this file to ensure that these were passed in before making any changes and store these into readable variables, e.g.

```
if [[ -z ${1} || -z ${2} ]]; then
    echo "Usage: ./scripts/push-to-assets-branch <package-branch> <assets-branch>"
    exit 1
fi
package_branch=$1
asset_branch=$2
```

scripts/push-to-assets-branch
```diff
@@ -0,0 +1,19 @@
...
+git checkout ${2}
+cp -r /tmp/tmp-assets/* ./
+git add .
+git -c user.name="actions" -c user.email="actions@github.com" commit -m "Generated changes" || true
+git push ${1} ${2}
```
aiyengar2, line 19: I'm not a huge fan of including a `git push` directly as part of the scripts; I think this script should be changed from `push-to-assets-branch` to `generate-commit-in-assets-branch` with everything included in the current version of this script except for this line.

This would prevent a dev who accidentally ran this script from accidentally potentially pushing to the wrong branch.

This specific line could then be moved to the `run` field of the `.github/workflows/push.yml` e.g.
```
./scripts/generate-commit-in-assets-branch origin dev-v2.5-assets && git push origin dev-v2.5-assets
```

aiyengar2 (changes requested):

scripts/ci-wrapper
StrongMonkey: @aiyengar2  It can't be clean up as `assets/` and `sha256sum/` has to be copied over to asset branch. Also it doesn't really need to be cleaned up since this script should never be run through a developer.

scripts/ci
```diff
@@ -1,6 +1,10 @@
...
+    export CHART=${1}
+fi
+
 ./scripts/prepare
 ./scripts/generate-charts
 ./scripts/clean
```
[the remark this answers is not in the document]
StrongMonkey: yeah, I think this is out of the scope of this PR. we can do clean up in the future. For now I would like to keep `scripts/ci` untouched so that it would not impact developer's normal workflow for testing. Beside `scripts/ci-wrapper` is only a wrapper for CI and doesn't intended to be run by developers.

scripts/ci-wrapper
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+set -e
+set -x
```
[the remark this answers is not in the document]
StrongMonkey: :+1: agree, but out of scrope for this PR.

scripts/ci-wrapper
```diff
@@ -0,0 +1,14 @@
...
+
+sudo ./scripts/bootstrap
+git fetch origin ${1}
+git checkout origin/${1} ./assets/index.yaml || true
+git checkout origin/${1} ./sha256sum || true
```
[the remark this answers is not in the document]
StrongMonkey: good catch, I will change it.

scripts/produce-sha256
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+set -e
+
+# This script compares the sha256sum of each chart folder, find the sha256sum of that chart that has changed
```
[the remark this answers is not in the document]
StrongMonkey: :+1:

scripts/produce-sha256
```diff
@@ -0,0 +1,28 @@
...
+rm -rf ./sha256sum-new
+mkdir -p ./sha256sum-new
+for f in packages/*; do
+    rm -rf ./sha256sum-new/$(basename -- ${f})
+    mkdir -p ./sha256sum-new/$(basename -- ${f})
+    find ${f} -type f | xargs sha256sum > ./sha256sum-new/$(basename -- ${f})/$(basename -- ${f}).sum
+done
+
+if [ ! -d sha256sum ]; then
+    cp -r sha256sum-new sha256sum
```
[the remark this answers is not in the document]
StrongMonkey: :+1:

scripts/ci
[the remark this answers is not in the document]
StrongMonkey: this is for better layout

scripts/ci-wrapper
```diff
@@ -0,0 +1,14 @@
...
+
+charts=$(./scripts/produce-sha256)
+
+if [[ ! -z $charts ]]; then
+  echo $charts | xargs sh -c ./scripts/ci
```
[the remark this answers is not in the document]
StrongMonkey: sh -c has to used otherwise it will not take arguments

scripts/generate-charts
```diff
@@ -14,19 +13,35 @@ for f in packages/*; do
...
 		if [[ -d ${f}/charts-crd ]]; then
 			packageVersion=$(yq r ${f}/package.yaml packageVersion)
 			yq w -i ${f}/charts-crd/Chart.yaml 'version' "${version}${packageVersion}"
 
 			helm package ${f}/charts-crd --destination assets/$(basename -- ${f})
-			git checkout assets/$(basename -- ${f})/"$(basename -- ${f})-crd-$(yq r ${f}/charts-crd/Chart.yaml version).tgz" || true
+
+			mkdir -p charts/$(basename -- ${f})/charts-crd
+			cp -R ${f}/charts-crd/* charts/$(basename -- ${f})/charts-crd
 		fi
```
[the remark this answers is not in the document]
StrongMonkey: looks like it is the same then...

scripts/generate-charts
```diff
@@ -14,19 +13,35 @@ for f in packages/*; do
...
+<!doctype html>
+<html>
+  <head>
+    <title>This is the index page for charts</title>
+  </head>
+  <body>
+    <p>If you are looking for index.yaml, go to https://charts.rancher.io/index.yaml</p>
+  </body>
+</html>
+EOF
```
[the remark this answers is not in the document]
StrongMonkey: yeah this can be removed

scripts/generate-charts
```diff
@@ -14,19 +13,35 @@ for f in packages/*; do
...
 			packageVersion=$(yq r ${f}/package.yaml packageVersion)
 			yq w -i ${f}/charts-crd/Chart.yaml 'version' "${version}${packageVersion}"
 
 			helm package ${f}/charts-crd --destination assets/$(basename -- ${f})
-			git checkout assets/$(basename -- ${f})/"$(basename -- ${f})-crd-$(yq r ${f}/charts-crd/Chart.yaml version).tgz" || true
+
+			mkdir -p charts/$(basename -- ${f})/charts-crd
+			cp -R ${f}/charts-crd/* charts/$(basename -- ${f})/charts-crd
 		fi
+		helm repo index --merge ./assets/index.yaml --url assets assets
```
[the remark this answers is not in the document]
StrongMonkey: if no change is made, we do not call this in order to avoid modifying index.yaml even no chart is changed.

scripts/produce-sha256
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+set -e
+
+# This script compares the sha256sum of each chart folder, find the sha256sum of that chart that has changed
+
+rm -rf ./sha256sum-new
+mkdir -p ./sha256sum-new
+for f in packages/*; do
+    rm -rf ./sha256sum-new/$(basename -- ${f})
```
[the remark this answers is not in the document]
StrongMonkey: yes, can be removed :+1:

scripts/produce-sha256
```diff
@@ -0,0 +1,28 @@
...
+    mkdir -p ./sha256sum-new/$(basename -- ${f})
+    find ${f} -type f | xargs sha256sum > ./sha256sum-new/$(basename -- ${f})/$(basename -- ${f}).sum
```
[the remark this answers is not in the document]
StrongMonkey: actually this script is never intended to be run locally. But yes the check can be added. :+1:

scripts/push-to-assets-branch
StrongMonkey: added.

scripts/push-to-assets-branch
StrongMonkey: good catch, added

scripts/push-to-assets-branch
StrongMonkey: :+1:

scripts/ci-wrapper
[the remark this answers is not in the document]
StrongMonkey: actually.... I am thinking about that this might not need to be changed. We don't expect this to be run locally and if it is run by github action then there should only be one remote call origin. So leave it unchanged.

scripts/generate-charts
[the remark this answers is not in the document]
aiyengar2: Yeah, the logic for CRD chart packaging is identical to the normal chart packaging except for the path used.

However, on further thought I think that we can remove this as its out of scope for the PR.

scripts/push-to-assets-branch
```diff
@@ -0,0 +1,26 @@
...
+for i in assets/ index.yaml index.html sha256sum/ charts; do
+    if [[ -d $i ]]; then
+        cp -R $i /tmp/tmp-assets
+    fi
+done
+git checkout .
+git clean -fd
+git fetch $remote $branch
+git checkout $branch
+cp -r /tmp/tmp-assets/* ./
```
aiyengar2, line 24: nit:
```suggestion
cp -R /tmp/tmp-assets/* ./
```

.github/workflows/push.yml
```diff
@@ -13,8 +13,8 @@ jobs:
 
     - name: Run CI scripts
       run: |-
-        sudo make ci
+        ./scripts/ci-wrapper dev-v2.5
```
aiyengar2, line 16:
```suggestion
        ./scripts/ci-wrapper dev-v2.5-source
```

.github/workflows/push.yml
```diff
@@ -13,8 +13,8 @@ jobs:
...
 
-    - uses: stefanzweifel/git-auto-commit-action@480e111bf8274f3bbd429aef51504f71c808e3e4
-      with:
-        commit_message: Generated changes
+    - name: Push charts and assets to the dev-v2.5 branch
+      run: |-
+        ./scripts/push-to-assets-branch origin dev-v2.5 && git push origin dev-v2.5
```
aiyengar2, line 20:
```suggestion
        ./scripts/push-to-assets-branch origin dev-v2.5-source && git push origin dev-v2.5-source
```

scripts/ci
aiyengar2, line 4: nit:
```suggestion
if [ -n ${1} ]; then
```

scripts/ci-wrapper
```diff
@@ -0,0 +1,16 @@
...
+git fetch origin ${1}
+
+# checkout index.yaml and sha256sum folder from asset branch in order to compare sha and merge index
+git checkout origin/${1} ./assets/index.yaml || true
+git checkout origin/${1} ./sha256sum || true
+
+charts=$(./scripts/produce-sha256)
+
+if [[ ! -z $charts ]]; then
+  echo $charts | xargs sh -c ./scripts/ci
```
aiyengar2, line 15: nit: for consistency we should use brackets, and `-n` instead of `! -z`
```suggestion
if [[ -n ${charts} ]]; then
  echo ${charts} | xargs sh -c ./scripts/ci
```

scripts/push-to-assets-branch
```diff
@@ -0,0 +1,26 @@
...
+if [[ -z ${1} || -z ${2} ]]; then
+    echo "Usage: ./scripts/push-to-assets-branch <package-branch> <assets-branch>"
+    exit 1
+fi
+remote=${1}
+branch=${2}
+
+rm -rf /tmp/tmp-assets
+mkdir -p /tmp/tmp-assets
+for i in assets/ index.yaml index.html sha256sum/ charts; do
```
aiyengar2, line 15: Do we need the slashes?
```suggestion
mkdir -p /tmp/tmp-assets
for i in assets index.yaml index.html sha256sum charts; do
```

scripts/push-to-assets-branch
```diff
@@ -0,0 +1,26 @@
...
+    if [[ -d $i ]]; then
+        cp -R $i /tmp/tmp-assets
+    fi
```
aiyengar2, line 18: `index.yaml` and `index.html` will never be copied based on this logic since we only copy directories in.

Also, should we fail out of `push-to-assets-branch` if any of these directories doesn't exist? I'm concerned about a situation where the `charts/` directory fails to get created in an earlier script but the `assets/` succeeds and we end up out of sync.

aiyengar2 (changes requested):

.github/workflows/push.yml
StrongMonkey: This actually needs to be dev-v2.5. dev-v2.5 is going to the asset branch.

.github/workflows/push.yml
StrongMonkey: same, source is the branch that developer are working on. Target(asset) is dev-v2.5

.github/workflows/push.yml
aiyengar2: whoops, you're right. Nvm

scripts/push-to-assets-branch
StrongMonkey: whoops, will fix it. If it doesn't exist that means we are not generating anything and we shouldn't fail. If something failed on the previous step, this script should not be executed.

scripts/push-to-assets-branch
StrongMonkey: Fixed

aiyengar2 (approved):
LGTM! Thanks for all the changes @StrongMonkey 😄

cbron:
Handled via: https://github.com/rancher/charts/pull/683/commits/09d57f73d6d5c8ae055e2acd6b62fd386ed92b38

cbron:
Handled via https://github.com/rancher/charts/pull/683

PR #657, opened 2020-09-17
Add changes for links behind proxy

brendarearden:
**Problem**
Prometheus and Grafana are behind a kube-proxy and are not configured to be accessed at the redirect path

**Solution**
Utilize grafana's root_url and prometheus' externalUrl to define what the subpath/relative path is for each service

The url is currently set to (for example):
`/api/v1/namespaces/cattle-monitoring-system/services/http:rancher-monitoring-grafana:80/proxy`
but should be: 
`/k8s/clusters/local/api/v1/namespaces/cattle-monitoring-system/services/http:rancher-monitoring-grafana:80/proxy`

This solution works, but need to have discussion about why including the `/k8s/clusters/local` in path is not working with the ui 

**Issue**
https://github.com/rancher/rancher/issues/28495


packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -2268,12 +2322,13 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-monitoring/charts-original/val
...
  ## Provide a k8s version to auto dashboard import script example: kubeTargetVersionOverride: 1.16.6
  ##
-@@ -76,8 +334,19 @@
+@@ -76,8 +334,20 @@
  
  ##
  global:
 +  cattle:
 +    systemDefaultRegistry: ""
++    clusterId: ""
```
aiyengar2, line 2331: As you pointed out in the description, the external URLs we are using are `api/v1/namespaces/cattle-monitoring-system/services/http:rancher-monitoring-grafana:80/proxy`, so there should be no need to add this value into the PR as it is right now.

aiyengar2 (changes requested):
@brendarearden @cbron will this PR be blocked on resolving the issue mentioned in the description or do we plan to merge this in with the external URL being `/api/v1/namespaces/cattle-monitoring-system/services/http:rancher-monitoring-grafana:80/proxy` first and then followup with one that includes the cluster ID?

I'm assuming this only works because `/api/v1` points the the same thing as `/k8s/clusters/local/api/v1` using the path defined by ECM but not by using Dashboard's proxy. So what happens if we have two clusters being managed by Dashboard and click on the Grafana link; does it point to only one of the clusters' (i.e. local) Grafana endpoints?

If this PR will be blocked on trying to figure out how to get it to work with `/k8s/clusters/<clusterId>/api/v1` based external URLs, we might wanna make it based off of https://github.com/rancher/charts/pull/641 since the rebase to a new upstream would / should likely get merged in first and will require some changes to this PR (such as refactoring the `prometheus-operator.*` template names to `kube-prometheus-stack.*`)

aiyengar2:
Also @vincent99 do you know why it might be that this proxy mechanism works via `/api/v1` but not via `/k8s/clusters/local/api/v1`?

When Brenda and I were checking it out, it seemed to have to do with the fact that Grafana was getting redirected when making a request like this:

`/k8s/clusters/local/api/v1/namespaces/cattle-monitoring-system/services/http:rancher-monitoring-grafana:80/proxy/<some-path>`
=>
`/k8s/clusters/local/api/v1/namespaces/cattle-monitoring-system/services/http:rancher-monitoring-grafana:80/proxy/k8s/clusters/local/api/v1/namespaces/cattle-monitoring-system/services/http:rancher-monitoring-grafana:80/proxy/<some-path>`

vincent99:
No I don't know, but from it being exactly the same URL twice I would guess something like the leading slash is missing in the config so it thinks it doesn't match and sends a redirect (also without a leading slash).  That would then gets interpreted as relative to where you were, which was already that path, so now you  end up at double path.

aiyengar2:
Closed in favor of https://github.com/rancher/charts/pull/677

PR #663, opened 2020-09-18
Update resourceset to include configmaps for clusterrepos

mrajashree:
* The current default resourceset contains all new CRDs related to catalog.cattle.io, but the clusterrepo index information is stored in a configmap for all clusterrepos in the cattle-system namespace. So this commit includes a resourceSelector for that
https://github.com/rancher/backup-restore-operator/issues/32
* Tag v0.0.1-rc13 contains a fix to convert an error log to a warning log
 https://github.com/rancher/backup-restore-operator/issues/33


prachidamle (approved):

packages/backup-restore-operator/charts/templates/rancher-resourceset.yaml
```diff
@@ -11,6 +11,10 @@ resourceSelectors:
   - apiVersion: "v1"
     kindsRegexp: "^Secret$|^serviceaccounts$"
     namespaceRegexp: "^cattle-|^p-|^c-|^local$|^user-|^u-"
+  - apiVersion: "v1"
+    kindsRegexp: "^configmaps$"
+    namespaces:
+      - "cattle-system"
```
cbron, line 17: This would grab all configmaps in cattle-system ? We want to do that right.

mrajashree: @cbron yes looks like it. We already backup secrets and serviceaccounts from this namespace too. Plus all index info for clusterrepos will be in these configmaps

cbron (approved):

PR #670, opened 2020-09-19
Update gatekeeper to version 3.1.1

PennyScissors:
This updates gatekeeper to the v3.1.1 upstream which now has Helm 3 support. This cleaned up the patch significantly as we no longer need to move crds to a separate directory in the patch. 🎉


packages/rancher-gatekeeper/package.yaml
```diff
@@ -1,4 +1,4 @@
-url: https://open-policy-agent.github.io/gatekeeper/charts/gatekeeper-v3.1.0.tgz
+url: https://open-policy-agent.github.io/gatekeeper/charts/gatekeeper-3.1.1.tgz
```
PennyScissors, line 1: Don't be alarmed, they dropped the v intentionally 🤷‍♂️

packages/rancher-gatekeeper/rancher-gatekeeper.patch
```diff
@@ -16,7 +16,7 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-gatekeeper/charts-original/Cha
  sources:
 -- https://github.com/open-policy-agent/gatekeeper.git
 +  - https://github.com/open-policy-agent/gatekeeper.git
- version: v3.1.0
+ version: 3.1.1
```
PennyScissors, line 19: Don't be alarmed, they dropped the v intentionally 🤷‍♂️

packages/rancher-gatekeeper/rancher-gatekeeper.patch
```diff
@@ -28,401 +28,57 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-gatekeeper/charts-original/REA
...
-+  versions:
-+  - name: v1alpha1
-+    served: true
-+    storage: true
+@@ -3,7 +3,6 @@
+ metadata:
+   annotations:
+     controller-gen.kubebuilder.io/version: v0.3.0
+-    helm.sh/hook: crd-install
+     helm.sh/hook-delete-policy: before-hook-creation
```
aiyengar2, line 46: If this is no longer a Helm hook, we should remove the `hook-delete-policy` annotation as well.

Applies to all the CRDs on this chart

aiyengar2 (approved):
LGTM, had a minor comment about removing an annotation from CRDs

packages/rancher-gatekeeper/rancher-gatekeeper.patch
cbron: Very weird thing anyway. This would attempt to delete the CRD on re-install or something ? Yea let's remove them to be safe.

cbron: Wait if the CRD's were moved to the CRD folder, what file is this modifying ?

PennyScissors: @cbron This modifies their crds in the `crds` directory of their upstream. Their crds used to be in `template` because it was a helm 2 chart and we had to do this change and also move them to a `crds` directory.

aiyengar2: re: https://github.com/rancher/charts/pull/670#discussion_r491739018 I don't think this would do anything since it's not defined as a Helm hook resource. Most likely would just get ignored afaik unless Helm does something super weird

PennyScissors: @aiyengar2 I know that `crd-install` hook gets ignored by Helm 3 (not sure about others), but it doesn't hurt to remove them for peace of mind.

aiyengar2: @PennyScissors we definitely should remove them, but what I meant with https://github.com/rancher/charts/pull/670#discussion_r491752108 is that because we are removing the `helm.sh/hook` annotation from the pod, `helm.sh/hook-delete-policy` likely won't be evaluated either. So we might as well remove both.

PennyScissors: I will remove them in the next commit, also remove old tgz.

packages/rancher-gatekeeper/rancher-gatekeeper.patch
```diff
@@ -992,16 +272,18 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-gatekeeper/charts-original/val
  emitAuditEvents: false
  image:
 -  repository: openpolicyagent/gatekeeper
--  release: v3.1.0
+-  release: v3.1.1
 +  repository: rancher/openpolicyagent-gatekeeper
-+  tag: v3.1.0
++  tag: v3.1.1
```
aiyengar2, line 277: Should this be `tag: 3.1.1`?

aiyengar2 (changes requested):

packages/rancher-gatekeeper/rancher-gatekeeper.patch
PennyScissors: Nope, they only removed the v from the chart version and not the image or app

aiyengar2 (approved):

prachidamle (approved):
LGTM

packages/rancher-gatekeeper/package.yaml
cbron, line 1: You don't actually need to remove the old tgz's because you are creating a new version. You can leave for upgrade testing, or remove them to clean up in general. Either way works.

cbron (commented):
lgtm, one more question

packages/rancher-gatekeeper/package.yaml
PennyScissors: I removed them for cleanup since I would consider this version the best starting point for being a GA and cleaner.

cbron (approved):

PR #671, opened 2020-09-20
fixed eks and k3s logging in airgap

paynejacob:


cbron (changes requested):
If this is everything then looks good. But needs tgz removal.

cbron (approved):
lgtm

PR #673, opened 2020-09-21
allow tolerations to be passed to fluentbit containers

paynejacob:
PROBLEM:

fluentbit pods are not installed on controlplane and etcd nodes

SOLUTION:

allow tolerations to be passed to all fluentbit pods created by rancher.


rmweir (approved):
LGTM

packages/rancher-logging/overlay/templates/loggings/eks/logging.yaml
```diff
@@ -16,6 +16,12 @@ spec:
       Tag: "eks"
       Path: "/var/log/messages"
       Parser: "syslog"
+    {{ if .Values.fluentbit_tolerations }}
```
mrajashree, line 19: I think this `if` is redundant? same for the other places

paynejacob: good catch.

mrajashree (approved):

cbron (changes requested):
remove tgz's

cbron (approved):

PR #677, opened 2020-09-22
Fix broken magic links for Grafana and Prometheus

aiyengar2:
This PR introduces nginx proxy containers on Grafana and Prometheus to resolve any issues associated with broken magic links.

Related Issue: https://github.com/rancher/rancher/issues/28495


packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -193,6 +193,15 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-monitoring/charts-original/cha
      {{- end }}
      imagePullPolicy: {{ .Values.image.pullPolicy }}
    {{- if .Values.command }}
+@@ -285,7 +284,7 @@
+     {{- end }}
+     ports:
+       - name: {{ .Values.service.portName }}
+-        containerPort: {{ .Values.service.port }}
++        containerPort: {{ .Values.service.targetPort }}
```
aiyengar2, line 201: This is a small bug in the grafana chart; `containerPort` defines the port on which app can be reached out inside the container and `targetPort` defines the port on which container has exposed the application, so they are effectively the same thing. 

On the other hand, `port` is the port that we want to expose on the service that should lead to `targetPort`. Therefore, for Grafana we expose a port of `80`, but the `targetPort` / `containerPort` is `8080`, which is where the nginx proxy will be listening to requests to forward to the actual Grafana container running in port `3000` (unexposed).

packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -2216,27 +2225,69 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-monitoring/charts-original/val
...
-     portName: service
+-    portName: service
++    portName: nginx-http
 +    ## Port for Grafana Service to listen on
 +    ##
 +    port: 80
 +    ## To be used with a proxy extraContainer port
 +    ##
-+    targetPort: 3000
++    targetPort: 8080
```
aiyengar2, line 2248: Modified to point to the nginx proxy port instead of the Grafana container port.

packages/rancher-monitoring/overlay/charts/grafana/templates/nginx-config.yaml
```diff
@@ -0,0 +1,75 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: grafana-nginx-proxy-config
+  namespace: {{ template "grafana.namespace" . }}
+  labels:
+    {{- include "grafana.labels" . | nindent 4 }}
+data:
+  nginx.conf: |-
```
aiyengar2, line 9: Copied straight from [Monitoring V1](https://github.com/rancher/system-charts/blob/d6f4e8f7056917ed013ddd14c9a33f750dc3e95d/charts/rancher-monitoring/v0.1.3/charts/grafana/templates/nginx-configmap.yaml#L45-L111). We don't need the scripts from the Monitoring V1 ConfigMap since we aren't trying to validate the requests coming into Prometheus / Grafana based on service account credentials (unlike Monitoring V1, which enforces RBAC on Prometheus via [`rancher/prometheus-auth`](https://github.com/rancher/prometheus-auth)

packages/rancher-monitoring/overlay/templates/prometheus/nginx-config.yaml
```diff
@@ -0,0 +1,66 @@
...
+        location / {
+          proxy_cache         my_zone;
+          proxy_cache_valid   200 302 1d;
+          proxy_cache_valid   301 30d;
+          proxy_cache_valid   any 5m;
+          proxy_cache_bypass  $http_cache_control;
+          add_header          X-Proxy-Cache $upstream_cache_status;
+          add_header          Cache-Control "public";
+          
+          proxy_pass     http://localhost:9090/;
```
aiyengar2, line 55: Copied straight from [Monitoring V1](https://github.com/rancher/system-charts/blob/d6f4e8f7056917ed013ddd14c9a33f750dc3e95d/charts/rancher-monitoring/v0.1.3/charts/prometheus/templates/nginx-configmap.yaml#L29-L83) with one exception:

Since we aren't trying to validate the requests coming into Prometheus based on service account credentials (unlike Monitoring V1, which enforces RBAC on Prometheus via [`rancher/prometheus-auth`](https://github.com/rancher/prometheus-auth), we can query the Prometheus container via localhost and avoid sending in `Authorization` headers with the service account creds.


We don't need the script from the Monitoring V1 ConfigMap either for the same reason.

packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -1423,6 +1423,15 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-monitoring/charts-original/tem
  {{- if .Values.prometheus.prometheusSpec.nodeSelector }}
    nodeSelector:
  {{ toYaml .Values.prometheus.prometheusSpec.nodeSelector | indent 4 }}
+@@ -226,7 +229,7 @@
+ {{- end }}
+ {{- if .Values.prometheus.prometheusSpec.containers }}
+   containers:
+-{{ toYaml .Values.prometheus.prometheusSpec.containers | indent 4 }}
++{{ tpl .Values.prometheus.prometheusSpec.containers $ | indent 4 }}
```
aiyengar2, line 1431: Updated since we pass in a template that needs to be rendered by Helm in order to support private registry stuff with the containers we are inserting.

packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -2449,6 +2458,15 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-monitoring/charts-original/val
  ## Deploy a Prometheus instance
  ##
  prometheus:
+@@ -1403,7 +1834,7 @@
+     port: 9090
+ 
+     ## To be used with a proxy extraContainer port
+-    targetPort: 9090
++    targetPort: 8080
```
aiyengar2, line 2466: Modified to listen to the nginx proxy instead of the Prometheus pod directly.

aiyengar2 (commented):
Left a couple of notes to make it easier to review!

brendarearden (approved):
Looks good!

packages/rancher-monitoring/rancher-monitoring.patch
cbron: Aren't we exposing port 3000 right below this ?

packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -2273,27 +2291,69 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-monitoring/charts-original/val
...
 +    ## Only used if service.type is 'NodePort'
 +    ##
 +    nodePort: 30950
 +    ## Service type
 +    ##
 +    type: ClusterIP
++
++  proxy:
++    image:
++      repository: rancher/library-nginx
```
cbron, line 2316: Need to do airgap testing on this PR

packages/rancher-monitoring/rancher-monitoring.patch
aiyengar2: Yes, Grafana by default exposes both ports on https://github.com/grafana/helm-charts/blob/13382de5d9fd15f2ed28f24182c1d1fa4f4c4cba/charts/grafana/templates/_pod.tpl#L286-L292 as well.

I think this makes sense, since leaving port 3000 exposed on the deployment allows the cluster owner to use `kubectl port-forward` in order to directly access Grafana if the proxy is broken for some reason.

packages/rancher-monitoring/rancher-monitoring.patch
aiyengar2: left a comment on the issue for QA to track this https://github.com/rancher/rancher/issues/28495#issuecomment-696994508.

packages/rancher-monitoring/rancher-monitoring.patch
cbron: So the service doesn't expose it at all but deployment does, ok.

cbron (approved):
Seems fine to me, assuming we are ok removing the prom-auth stuff.

PR #679, opened 2020-09-22
Change encryption config field name in restore crd

mrajashree:
This commit renames field `encryptionConfigName` on Restore CR spec to `encryptionConfigSecretName`. Also updated tag to v0.1.0-rc1 

https://github.com/rancher/dashboard/issues/1125


cbron (approved):
lgtm, version should be in line now with others

PR #680, opened 2020-09-22
[2.5.1] Add rbac.pspAnnotations field to Grafana

aiyengar2:
This commit patches the Grafana subchart in order to use `rbac.pspAnnotations` to allow a user to opt-in to what PSP annotations they wanted added on Grafana's PSP on a chart deploy.

Related Issue: https://github.com/rancher/rancher/issues/28536


mrajashree (dismissed):

packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -193,19 +193,49 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-monitoring/charts-original/cha
...
++  pspAnnotations: {}
++  ## Specify pod annotations
++  ## Ref: https://kubernetes.io/docs/concepts/policy/pod-security-policy/#apparmor
++  ## Ref: https://kubernetes.io/docs/concepts/policy/pod-security-policy/#seccomp
++  ## Ref: https://kubernetes.io/docs/concepts/policy/pod-security-policy/#sysctl
++  ##
++  # seccomp.security.alpha.kubernetes.io/allowedProfileNames: 'docker/default'
++  # seccomp.security.alpha.kubernetes.io/defaultProfileName:  'docker/default'
++  # apparmor.security.beta.kubernetes.io/allowedProfileNames: 'runtime/default'
++  # apparmor.security.beta.kubernetes.io/defaultProfileName:  'runtime/default'
```
dramich, line 233: I don't follow the purpose of this section?

prachidamle (dismissed):

packages/rancher-monitoring/rancher-monitoring.patch
aiyengar2: Based on the issue logged by @jiaqiluo in https://github.com/rancher/rancher/issues/28536#issuecomment-696876105, it seems like these annotations aren't necessarily supported by all hardened clusters. So instead of using the original Grafana chart approach of [introducing flags like `rbac.pspUseAppArmor`](https://github.com/rancher/charts/pull/680/files#diff-ed5b5b87cf149adc6a6408bf81993efdR203-R209) to opt-out of certain annotations, I used the `kube-prometheus-stack` approach of allowing the user to [opt into annotations via `rbac.pspAnnotations`](https://github.com/prometheus-community/helm-charts/blob/a2b924bf7c439f0531634983dd941eeec2917725/charts/kube-prometheus-stack/values.yaml#L92-L100).

These commented out annotations are just the default ones that the upstream Grafana chart normally adds onto the PSP annotations.

cbron:
This was bumped to 2.5.1, so not high priority anymore.

mrajashree (approved):
LGTM

packages/rancher-monitoring/rancher-monitoring.patch
dramich: Can we drop them if they aren't needed? The patch file is already big without things that we don't need/use

aiyengar2: We could, but it’s possible / likely a user might want to uncomment these annotations if they have seccomp or apparmor in their cluster. Also it wouldn’t make much of a difference in the patch file since we’ll have to apply a patch anyways to replace pspUseAppArmor with pspAnnotations.

dramich (approved):

cbron:
This issue is in 2.5.1, so don't merge yet, not sure if QA has time to test in 2.5.0 cycle.

aiyengar2:
Closing this PR out in favor of https://github.com/rancher/charts/pull/745

PR #682, opened 2020-09-22
Remove rancher-monitoring 8.16.100

aiyengar2:
This version of the package no longer exists and does not need to be in the assets/ directory since it corresponds to an older version of the chart based on the old upstream before https://github.com/rancher/charts/pull/641 was merged.

Related Issue: https://github.com/rancher/rancher/issues/29098


aiyengar2:
@StrongMonkey is this the best way to approach this type of change?

aiyengar2:
This PR stays on the dev-v2.5 branch since it affects the `assets/index.yaml` directly

cbron (approved):
Makes sense to me, but get Daishans approval before merging

StrongMonkey (approved):

PR #685, opened 2020-09-22
Remove assets and index.yaml

StrongMonkey:


cbron (dismissed):

aiyengar2 (changes requested):
Can we drop CNAME too?

aiyengar2 (approved):

PR #688, opened 2020-09-23
Support semver-style dependency verisons

aiyengar2:


StrongMonkey (approved):

scripts/prepare-subcharts
```diff
@@ -117,25 +117,22 @@ while ! [[ -z $(yq r ${f}/charts/requirements.yaml "dependencies[${i}]") ]]; do
...
 		rm ${f}/charts/charts/${subchart_tgz}
 	fi
 	((i=i+1))
 done
 
 # Undo the patch on the requirements.yaml
-if [[ -f ${f}/requirements.patch ]] && [[ -s ${f}/requirements.patch ]]; then
-	patch -R -E -p3 -d ${f}/charts < ${f}/requirements.patch
+if [[ -f ${f}/requirements.patch ]]; then
+	[[ -s ${f}/requirements.patch ]] && patch -R -E -p3 -d ${f}/charts < ${f}/requirements.patch
```
cbron, line 136: feel like this got more confusing, why'd it move ?

scripts/generate-patch
```diff
@@ -51,7 +51,7 @@ for f in packages/*; do
 					helm_repo=$(yq r ${f}/charts/requirements.yaml "dependencies[${i}].repository" | head -n 1)
 					# Remove trailing slash
 					[[ ${helm_repo} == */ ]] && helm_repo=${helm_repo%/}
-					if [[ ${helm_repo} == file://* ]] || [[ ${version} == *'*'* ]]; then
+					if [[ ${helm_repo} == file://* ]] || [[ ! ${version} =~ ^[0-9\.]+$ ]]; then
```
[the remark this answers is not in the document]
aiyengar2: Yes, the previous version of the code was just checking to see if a wildcard character in the version. This version of the code checks to see that the version provided only consists of digits (`0-9`) or periods (`\.`), since it is possible / valid to specify a version string that has a non-fixed version.

It's not the best solution, since there are cases like [datadog](https://github.com/helm/charts/blob/ef5b8687622cce8d6b0c7ee1df80329603738c2a/stable/datadog/requirements.yaml#L3) where you are specifying a fixed version number but this script won't pick it up as expected, but instead of introducing a more complicated expression I kept it simple to unblock the failing CI. Properly validating the semver and checking can be done in https://github.com/rancher/rancher/issues/28730.

scripts/prepare-subcharts
```diff
@@ -117,25 +117,22 @@ while ! [[ -z $(yq r ${f}/charts/requirements.yaml "dependencies[${i}]") ]]; do
 		((i=i+1))
 		continue
 	fi
-	subchart_path=$(ls ${f}/charts/charts/${subchart_tgz})
-	if [[ ${version} == *'*'* ]]; then
+	if [[ ! ${version} =~ ^[0-9\.]+$ ]]; then
 		# Subcharts that are tracked remotely must have a fixed version number
-		if [[ -f ${subchart_path} ]]; then
-			rm -rf ${charts_path}
-		fi
```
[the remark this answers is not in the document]
aiyengar2: Yeah, `subchart_path` should have been `subchart_tgz_path` and chart_path should have been `subchart_dir_path` technically. I'll do a quick commit to refactor that

scripts/prepare-subcharts
aiyengar2: We want to delete the `requirements.patch` if it's an empty file but only patch it if it actually has contents.

Without this, an empty `requirements.patch` would be left behind.

I'll update it in the next commit to make that more clear.

cbron: Yes that reads cleaner now, thanks.

cbron (approved):

StrongMonkey (approved):

PR #689, opened 2020-09-23
Remove default values so they're not set while launching app

mrajashree:
The configuration fields in UI are populated with the default values, so removing these default values and setting them to empty values. Related to this bug https://github.com/rancher/dashboard/issues/1019#issuecomment-697098476

### DIFF
--- a/packages/backup-restore-operator/charts/templates/s3-secret.yaml
+++ b/packages/backup-restore-operator/charts/templates/s3-secret.yaml
@@ -11,18 +11,16 @@ stringData:
   {{- with .Values.s3 }}
   {{- if .credentialSecretName }}
   credentialSecretName: {{ .credentialSecretName }} 
-  {{- end }}
-  {{- if .credentialSecretNamespace }}
-  credentialSecretNamespace: {{ .credentialSecretNamespace }} 
+  credentialSecretNamespace: {{ required "When providing a Secret containing S3 credentials, a valid .Values.credentialSecretNamespace must be provided" .credentialSecretNamespace }}
   {{- end }}
   {{- if .region }}
   region: {{ .region }}
   {{- end }}
-  bucketName: {{ .bucketName }}
+  bucketName: {{ required "A valid .Values.bucketName is required for configuring S3 compatible storage as the default backup storage location" .bucketName }}
   {{- if .folder }}
   folder: {{ .folder }} 
   {{- end }}
-  endpoint: {{ .endpoint }}
+  endpoint: {{ required "A valid .Values.endpoint is required for configuring S3 compatible storage as the default backup storage location" .endpoint }}
   {{- if .endpointCA }} 
   endpointCA: {{ .endpointCA }} 
   {{- end }}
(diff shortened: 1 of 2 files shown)


packages/backup-restore-operator/charts/values.yaml
```diff
@@ -7,14 +7,14 @@ s3:
...
-  region: us-west-2
-  bucketName: rancherbackups
-  folder: base folder
-  endpoint: s3.us-west-2.amazonaws.com
-  endpointCA: base64 encoded CA cert
-  # insecureTLSSkipVerify: optional
+  region: ""
+  bucketName: ""
+  folder: ""
+  endpoint: ""
```
aiyengar2, line 15: If I deploy this as-is with just `s3.enabled=true` as a Helm chart user, what would happen if the endpoint was not changed here? Seems like we would end up with a secret whose stringData has just `endpoint: ""` in it, but no actual failures on chart deploy based on:

https://github.com/rancher/charts/blob/1a5c676c23b49d509eb94f39f3143eb7ed0fd2c0/packages/backup-restore-operator/charts/templates/s3-secret.yaml#L25

Is this a valid configuration to have? If not, we should modify the above line to use the Helm `required` function so that users are forced to pick a value for it if they set `s3.enabled=true`.

aiyengar2 (changes requested):

packages/backup-restore-operator/charts/values.yaml
```diff
@@ -7,14 +7,14 @@ s3:
...
+  endpointCA: ""
+  insecureTLSSkipVerify: false
```
prachidamle, line 17: What if I install this chart via helm cli? Is it ok to let these values be empty by the user?

mrajashree: Yes, if s3.enabled can be left as false. S3 can be configured to either use S3 or MinIO buckets. The only common required fields are Region and Endpoint so I've changed them to use the "required" function

aiyengar2 (approved):

prachidamle (approved):
LGTM

packages/backup-restore-operator/charts/templates/s3-secret.yaml
```diff
@@ -11,18 +11,16 @@ stringData:
   {{- with .Values.s3 }}
   {{- if .credentialSecretName }}
   credentialSecretName: {{ .credentialSecretName }} 
-  {{- end }}
-  {{- if .credentialSecretNamespace }}
-  credentialSecretNamespace: {{ .credentialSecretNamespace }} 
+  credentialSecretNamespace: {{ required "When providing a Secret containing S3 credentials, a valid .Values.credentialSecretNamespace must be provided" .credentialSecretNamespace }}
```
cbron, line 14: `credentialSecretNamespace` hasn't actually changed in values.yaml, did we mean to change `credentialSecretName ` ? Or is that not required.

mrajashree: @cbron yes in values.yaml `credentialSecretNamespace` was already empty, and I wanted to set `credentialSecretName` to an empty string too. Neither of the fields are required. But if the credentialSecretName is set then the namespace should also be set. That's why I made this change

cbron (approved):

PR #691, opened 2020-09-23
fixed airgap images

paynejacob:
PROBLEMS:
- airgap install fails
- k3s installs require a pvc

SOLUTION:
- added registry prefix to operator image
- override the config reloading images
- ensured the pvc toggle is passed to all loggings

### DIFF
--- a/packages/rancher-logging/overlay/templates/loggings/eks/logging.yaml
+++ b/packages/rancher-logging/overlay/templates/loggings/eks/logging.yaml
@@ -24,4 +24,8 @@ spec:
     image:
       repository: {{ template "system_default_registry" . }}{{ .Values.images.fluentd.repository }}
       tag: {{ .Values.images.fluentd.tag }}
+    configReloaderImage:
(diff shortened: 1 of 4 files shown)


packages/rancher-logging/overlay/templates/loggings/eks/logging.yaml
```diff
@@ -24,4 +24,8 @@ spec:
     image:
       repository: {{ template "system_default_registry" . }}{{ .Values.images.fluentd.repository }}
       tag: {{ .Values.images.fluentd.tag }}
+    configReloaderImage:
```
brendarearden, line 27: I don't think it needs to happen for this pr, but it seems like the fluentd section is the same in all of these logging yamls and might be a good candidate for a helper template

brendarearden (dismissed):
Looks good!

cbron (approved):
lgtm, after this goes out we need to download new chart and ensure its all updated before we do any testing

cbron (approved):

brendarearden (approved):

PR #692, opened 2020-09-24
Longhorn: Add upstream Longhorn chart

shuo-wu:
This PR is from #526
Issue: rancher/rancher#27725

### DIFF
--- a/packages/longhorn/longhorn.patch
+++ b/packages/longhorn/longhorn.patch
@@ -1,0 +1,758 @@
+diff -x '*.tgz' -x '*.lock' -uNr packages/longhorn/charts-original/Chart.yaml packages/longhorn/charts/Chart.yaml
+--- packages/longhorn/charts-original/Chart.yaml
++++ packages/longhorn/charts/Chart.yaml
+@@ -5,23 +5,29 @@
+ kubeVersion: ">=v1.14.0-r0"
+ description: Longhorn is a distributed block storage system for Kubernetes.
+ keywords:
+-- longhorn
+-- storage
+-- distributed
+-- block
+-- device
+-- iscsi
++  - longhorn
++  - storage
++  - distributed
++  - block
++  - device
++  - iscsi
+ home: https://github.com/longhorn/longhorn
+ sources:
+-- https://github.com/longhorn/longhorn
+-- https://github.com/longhorn/longhorn-engine
+-- https://github.com/longhorn/longhorn-instance-manager
+-- https://github.com/longhorn/longhorn-manager
+-- https://github.com/longhorn/longhorn-ui
+-- https://github.com/longhorn/longhorn-tests
++  - https://github.com/longhorn/longhorn
++  - https://github.com/longhorn/longhorn-engine
++  - https://github.com/longhorn/longhorn-instance-manager
++  - https://github.com/longhorn/longhorn-manager
++  - https://github.com/longhorn/longhorn-ui
++  - https://github.com/longhorn/longhorn-tests
+ maintainers:
+-- name: Longhorn maintainers
+-  email: maintainers@longhorn.io
+-- name: Sheng Yang
+-  email: sheng@yasker.org
+-icon: https://raw.githubusercontent.com/cncf/artwork/master/projects/longhorn/icon/color/longhorn-icon-color.svg?sanitize=true
++  - name: Longhorn maintainers
++    email: maintainers@longhorn.io
++  - name: Sheng Yang
++    email: sheng@yasker.org
++icon: https://raw.githubusercontent.com/cncf/artwork/master/projects/longhorn/icon/color/longhorn-icon-color.svg
++annotations:
++  catalog.cattle.io/certified: rancher
++  catalog.cattle.io/namespace: longhorn-system
++  catalog.cattle.io/release-name: longhorn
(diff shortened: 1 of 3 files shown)


packages/longhorn/longhorn.patch
```diff
@@ -0,0 +1,774 @@
...
++{{- end -}}
+diff -x '*.tgz' -x '*.lock' -uNr packages/longhorn/charts-original/templates/clusterrole.yaml packages/longhorn/charts/templates/clusterrole.yaml
+--- packages/longhorn/charts-original/templates/clusterrole.yaml
++++ packages/longhorn/charts/templates/clusterrole.yaml
+@@ -30,11 +30,8 @@
+   verbs: ["*"]
+ - apiGroups: ["coordination.k8s.io"]
+   resources: ["leases"]
+-  verbs: ["get", "watch", "list", "delete", "update", "create"]
++  verbs: ["*"]
```
cbron, line 385: I don't think this is necessary anymore ?

packages/longhorn/longhorn.patch
```diff
@@ -0,0 +1,774 @@
...
+ - apiGroups: ["longhorn.io"]
+   resources: ["volumes", "volumes/status", "engines", "engines/status", "replicas", "replicas/status", "settings",
+               "engineimages", "engineimages/status", "nodes", "nodes/status", "instancemanagers", "instancemanagers/status"]
+   verbs: ["*"]
+-- apiGroups: ["coordination.k8s.io"]
+-  resources: ["leases"]
+-  verbs: ["*"]
```
cbron, line 392: this either, I'm assuming the operator still needs this in clusterrole to function

packages/longhorn/longhorn.patch
```diff
@@ -0,0 +1,774 @@
...
++++ packages/longhorn/charts/templates/_helpers.tpl
+@@ -36,3 +36,12 @@
+ app.kubernetes.io/instance: {{ .Release.Name }}
+ app.kubernetes.io/version: {{ .Chart.AppVersion }}
+ {{- end -}}
++
++
++{{- define "system_default_registry" -}}
++{{- if .Values.global.systemDefaultRegistry -}}
++{{- printf "%s/" .Values.global.systemDefaultRegistry -}}
```
cbron, line 372: This changed to be global.cattle, I updated the doc with better links: https://github.com/rancher/charts/blob/b55e6ec01907964723942fdbd9f63b9afef7a1ed/packages/rancher-cis-benchmark/charts/templates/_helpers.tpl#L6-L12

Need to update values.yaml also

packages/longhorn/package.yaml
```diff
@@ -0,0 +1,8 @@
+url: https://github.com/longhorn/charts.git
+subdirectory: charts/longhorn
+type: git
+commit: 0a88eaae4d7bdce762af942ea490d72aee338bb1
+packageVersion: 00
+generateCRDChart:
+  enabled: true
+  providesGVR: longhorn.io/v1beta1
```
cbron, line 8: Can you remove this, it doesn't have an effect anymore. The annotation on chart is the correct way and it already is there.

cbron (changes requested):
Few last minor things. After this is done can you run `make CHART=longhorn charts` and ensure that it installs from the generated tgz's assets. You might get a git error but you can ignore that.

packages/longhorn/longhorn.patch
```diff
@@ -0,0 +1,774 @@
...
+         - longhorn-manager
+diff -x '*.tgz' -x '*.lock' -uNr packages/longhorn/charts-original/values.yaml packages/longhorn/charts/values.yaml
+--- packages/longhorn/charts-original/values.yaml
++++ packages/longhorn/charts/values.yaml
+@@ -1,16 +1,36 @@
+ # Default values for longhorn.
+ # This is a YAML-formatted file.
+ # Declare variables to be passed into your templates.
++global:
++  systemDefaultRegistry: ""
```
cbron, line 720: this is the part that also needs cattle added

packages/longhorn/longhorn.patch
shuo-wu: This is redundancy in Longhorn role. It's not related to this chart PR. I will apply the change in the Longhorn chart later.

cbron (commented):
This looks good but waiting on uninstaller bug first.

yasker:
I noticed that once I installed the chart, the ingress is `xip.io` instead of using kubeproxy/magic proxy. Is there any change compared to the chart we use in v2.4?

yasker:
It seems if Longhorn is installed using the Rancher app, the ingress will be selected by default. Previously the service provided the UI will have a label `kubernetes.io/cluster-service: "true"`, allow Rancher to discover it. I guess now we need a specific UI to do that.

yasker:
The issue of selecting `ingress` by default seems caused by a pair of unnecessary quotes. See the bug for details. https://github.com/rancher/dashboard/issues/1402

We should be able to fix it using a patch for now.

yasker:
@cbron The PR is good to merge. The uninstaller issue won't happen in every cluster and it's not a new issue. We're tracking separately for it. We will also provide a workaround if it happened on the customer side before our next release.

cbron (approved):

PR #695, opened 2020-09-24
Fix ci

StrongMonkey:
After new fleet release, fleet-agent and fleet-crd charts are not updated correctly
https://github.com/rancher/fleet/issues/94

### DIFF
--- a/packages/fleet-agent/package.yaml
+++ b/packages/fleet-agent/package.yaml
@@ -1,2 +1,2 @@
-url: https://github.com/rancher/fleet/releases/download/v0.3.0-beta5/fleet-agent-0.3.0-beta5.tgz
+url:  https://github.com/rancher/fleet/releases/download/v0.3.0-beta5/fleet-agent-0.3.0-beta5.tgz
 packageVersion: 00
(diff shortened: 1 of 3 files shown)


StrongMonkey:
@cbron @aiyengar2 The xargs doesn't seem to work if multiple charts are changed, switch to for loop which seems to be more reliable. If you guys have any clue on xargs please let me know. This is to fix https://github.com/rancher/charts/runs/1162981323 where three charts are changed but only one chart is uploaded

aiyengar2 (approved):

aiyengar2:
For future reference, this change was required since `./scripts/ci` only supports one argument based on: https://github.com/rancher/charts/blob/286df602a2af2b82ea9e52d1ca4d013f40f04718/scripts/ci#L5-L7.

xargs would pass in the charts as `$1`, `$2`, `$3`, etc. so using a for loop is necessary instead.

prachidamle (approved):
LGTM

StrongMonkey:
The space change in the second commit is to change content so that chart can be reuploaded.

cbron (approved):
Spaces in package.yaml are to force a push. LGTM.

PR #699, opened 2020-09-25
fixed logging roles

paynejacob:
Removed edit from cluster members and allowed all users to view cluster level crds.

### DIFF
--- a/packages/rancher-logging/overlay/templates/userroles.yaml
+++ b/packages/rancher-logging/overlay/templates/userroles.yaml
@@ -16,8 +16,10 @@ rules:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: "logging-edit"
+  name: "logging-view"
   labels:
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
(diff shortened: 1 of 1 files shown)


brendarearden (approved):
LGTM

cbron (commented):
Can you make a ticket in 2.5.0 that explains the new goals to check for, and assign it to you and @izaac and then move to test after this merges.

packages/rancher-logging/overlay/templates/userroles.yaml
```diff
@@ -15,31 +15,20 @@ rules:
...
     rbac.authorization.k8s.io/aggregate-to-view: "true"
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
 rules:
   - apiGroups:
       - "logging.banzaicloud.io"
     resources:
       - flows
       - outputs
+      - clusterflows
+      - clusteroutputs
```
cbron, line 31: If we are giving view to all users, is there any sensitive info they could see on any 4 of these CRD's ? Or are they all in secrets.

izaac: It is possible to store credentials/certificates as string values in the (Cluster)Outputs YAMLs. Although we're mounting from secrets using the UI.

cbron (approved):
We should document this caveat, and if we confirm any secrets are there we should add those to doc. Otherwise lgtm.

PR #696, opened 2020-09-25
Add README and rename chart to rancher-backup

mrajashree:
This commit:

- [x] Adds Readme for the chart. Addressing feedback from this previous PR's review comments https://github.com/rancher/charts/pull/628
https://github.com/rancher/backup-restore-operator/issues/36

Also includes following changes for https://github.com/rancher/backup-restore-operator/issues/37
- [x] Renames chart to rancher-backup

PR #703, opened 2020-09-25
Remove backup-restore-operator chart since its renamed

mrajashree:
The backup-restore-operator chart was renamed to rancher-backup as a part of this PR https://github.com/rancher/charts/pull/696 for this issue https://github.com/rancher/backup-restore-operator/issues/37
This commit removes the entries for backup-restore-operator chart (entries created for the previous name)


cbron (changes requested):
Need to do assets/index.yaml also

assets/index.yaml
```diff
@@ -1,1447 +0,0 @@
-apiVersion: v1
```
cbron, line 1: Just remove the chart, lines 1-39

mrajashree: ah, my bad

cbron (approved):

PR #704, opened 2020-09-25
Correct Kiali Version, Enable Telemetry V2

brendarearden:
**Problem:**
Telemetry V2 was is not getting enabled. Current chart is still using Telemetry v1. 
Kiail version for the chart.yaml auto-install annotation is 1.22.x, not 1.23.x

**Solution:**
Change auto-install annotation to correct kiali version
Disable mixer telemetry and policy. Add configuration for tTelemetry v2. 

**Issue:** 
https://github.com/rancher/rancher/issues/29169
https://github.com/rancher/rancher/issues/28738

### DIFF
--- a/packages/rancher-istio/charts/Chart.yaml
+++ b/packages/rancher-istio/charts/Chart.yaml
@@ -14,4 +14,4 @@ annotations:
   catalog.cattle.io/ui-component: istio
   catalog.cattle.io/requires-gvr: monitoring.coreos.com.prometheus/v1
   catalog.cattle.io/provides-gvr: networking.istio.io.virtualservice/v1beta1
-  catalog.cattle.io/auto-install: rancher-kiali-server-crd=1.22.001
+  catalog.cattle.io/auto-install: rancher-kiali-server-crd=1.23.001
--- a/packages/rancher-istio/charts/configs/istio-base.yaml
+++ b/packages/rancher-istio/charts/configs/istio-base.yaml
@@ -43,7 +43,7 @@ spec:
     policy:
       enabled: {{ .Values.policy.enabled }}
     telemetry:
-      enabled: {{ .Values.telemetry.enabled }}
+      enabled: {{ .Values.telemetry.v1.enabled }}
   hub: {{ .Values.systemDefaultRegistry | default "docker.io" }}
   profile: default
   tag: {{ .Values.tag }}
@@ -63,7 +63,7 @@ spec:
       proxy:
         image: {{ template "system_default_registry" . }}{{ .Values.global.proxy.repository }}:{{ .Values.global.proxy.tag }}
       proxy_init:
-        image: {{ template "system_default_registry" . }}{{ .Values.global.proxy_init.repository}}:{{ .Values.global.proxy_init.tag }}
+        image: {{ template "system_default_registry" . }}{{ .Values.global.proxy_init.repository }}:{{ .Values.global.proxy_init.tag }}
       {{- if .Values.global.defaultPodDisruptionBudget.enabled }}
       defaultPodDisruptionBudget:
         enabled: {{ .Values.global.defaultPodDisruptionBudget.enabled }}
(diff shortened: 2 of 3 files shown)


paynejacob (dismissed):

packages/rancher-istio/charts/Chart.yaml
```diff
@@ -14,4 +14,4 @@ annotations:
   catalog.cattle.io/ui-component: istio
   catalog.cattle.io/requires-gvr: monitoring.coreos.com.prometheus/v1
   catalog.cattle.io/provides-gvr: networking.istio.io.virtualservice/v1beta1
-  catalog.cattle.io/auto-install: rancher-kiali-server-crd=1.22.001
+  catalog.cattle.io/auto-install: rancher-kiali-server-crd=1.23.001
```
aiyengar2, line 17: Hmm, maybe specifying which CRD charts you chart is dependent on should be something addressed by https://github.com/rancher/rancher/issues/28730.

Will keep track of it

packages/rancher-istio/charts/configs/istio-base.yaml
```diff
@@ -73,18 +73,26 @@ spec:
...
+      {{- if .Values.mixer.policy.enabled }}
       policy:
-        image: {{ template "system_default_registry" . }}{{ .Values.policy.repository}}:{{ .Values.policy.tag }}
+        image: {{ template "system_default_registry" . }}{{ .Values.mixer.policy.repository}}:{{ .Values.mixer.policy.tag }}
       {{- end }}
-      {{- if .Values.telemetry.enabled }}
+      {{- if .Values.mixer.telemetry.enabled }}
       telemetry:
-        image: {{ template "system_default_registry" . }}{{ .Values.telemetry.repository}}:{{ .Values.telemetry.tag }}
+        image: {{ template "system_default_registry" . }}{{ .Values.mixer.telemetry.repository}}:{{ .Values.mixer.telemetry.tag }}
```
aiyengar2, line 82: nit:
```suggestion
        image: {{ template "system_default_registry" . }}{{ .Values.mixer.telemetry.repository }}:{{ .Values.mixer.telemetry.tag }}
```

packages/rancher-istio/charts/values.yaml
```diff
@@ -34,20 +34,27 @@ ingressGateways:
...
 
-policy:
+pilot:
     enabled: true
-    repository: rancher/istio-mixer
+    repository: rancher/istio-pilot
     tag: 1.7.1
 
 telemetry:
+  enabled: true
```
aiyengar2, line 53: I think we should drop this field, see above comment

packages/rancher-istio/charts/configs/istio-base.yaml
```diff
@@ -73,18 +73,26 @@ spec:
...
       image: {{ template "system_default_registry" . }}{{ .Values.pilot.repository}}:{{ .Values.pilot.tag }}
     {{- end }}
+    {{- if .Values.telemetry.enabled }}
+    telemetry:
+      enabled: {{ .Values.telemetry.enabled }}
+      v1:
+        enabled: {{ .Values.telemetry.v1.enabled }}
+      v2:
+        enabled: {{ .Values.telemetry.v2.enabled }}
+    {{- end}}
```
aiyengar2, line 95: I'm not a huge fan of the idea that you need to specify enabled twice. Can we modify this to:
```suggestion
    {{- if or .Values.telemetry.v1.enabled .Values.telemetry.v2.enabled }}
    telemetry:
      enabled: true # is this necessary anymore?
      v1:
        enabled: {{ .Values.telemetry.v1.enabled }}
      v2:
        enabled: {{ .Values.telemetry.v2.enabled }}
    {{- end}}
```

aiyengar2 (changes requested):

brendarearden:
I removed mixer from the istio-base.yaml all together since it is deprecated in 1.7.1 - if someone wants to enable it for some reason, they can utilize the overlay.yaml option

packages/rancher-istio/charts/configs/istio-base.yaml
brendarearden: I used this idea, but I also do need to still define the field `telemetry.enabled` so I went ahead and used the same if statement to set it

packages/rancher-istio/charts/configs/istio-base.yaml
```diff
@@ -41,9 +41,9 @@ spec:
     pilot:
       enabled: {{ .Values.pilot.enabled }}
     policy:
-      enabled: {{ .Values.policy.enabled }}
+      enabled: {{ .Values.mixer.policy.enabled }}
     telemetry:
-      enabled: {{ .Values.telemetry.enabled }}
+      enabled: {{ .Values.mixer.telemetry.enabled }}
```
cbron, line 46: If this is `spec.components.telemetry`, doesn't that correspond with values.telemetry.v1 or v2 ? I'm confused on both why this would be set via mixer and how v1 and v2 in values.yaml correspond to this. Same question for policy.

packages/rancher-istio/charts/configs/istio-base.yaml
```diff
@@ -73,18 +73,30 @@ spec:
...
       telemetry:
-        image: {{ template "system_default_registry" . }}{{ .Values.telemetry.repository}}:{{ .Values.telemetry.tag }}
+        image: {{ template "system_default_registry" . }}{{ .Values.mixer.telemetry.repository }}:{{ .Values.mixer.telemetry.tag }}
       {{- end }}
     {{- if .Values.pilot.enabled }}
     pilot:
-      image: {{ template "system_default_registry" . }}{{ .Values.pilot.repository}}:{{ .Values.pilot.tag }}
+      image: {{ template "system_default_registry" . }}{{ .Values.pilot.repository }}:{{ .Values.pilot.tag }}
     {{- end }}
+    {{- if or .Values.telemetry.v1.enabled .Values.telemetry.v2.enabled }}
```
cbron, line 88: Is this line necessary ? Even if they are both false they would just both be false below.

packages/rancher-istio/charts/configs/istio-base.yaml
brendarearden: Since `The functionality provided by Mixer is being moved into the Envoy proxies`, then the separate component installation responds to the deprecated mixer components. Envoy is managed via istiod so enable or disable occurs via the `.Values.telemetry.enabled` option.

cbron (dismissed):
Lgtm. Should probably lightly document how to turn on telemetry v1 vs v2, since that is somewhat confusing.

packages/rancher-istio/charts/configs/istio-base.yaml
```diff
@@ -75,16 +75,26 @@ spec:
     mixer:
```
cbron, line 75: Should prob wrap this, as if both policy and telemetry is off it'd be empty

brendarearden: added!

paynejacob (dismissed):

aiyengar2 (dismissed):

packages/rancher-istio/charts/values.yaml
```diff
@@ -7,47 +7,52 @@ installer:
...
 policy:
-    enabled: true
-    repository: rancher/istio-mixer
-    tag: 1.7.1
+  enabled: false
+  repository: rancher/istio-mixer
+  tag: 1.7.1
 
 telemetry:
-    enabled: true
```
cbron, line 48: Can you see in istio 1.8 what this is going to be ? My bet is they are removing v1 and v2 and leaving this

brendarearden: In the current master, I see that they removed the v1 check in the configmap, but kept the v2 check in the enoyfilter creation. They are tracking mixer removal here https://github.com/istio/istio/issues/25333

cbron (approved):

brendarearden:
@ aiyengar2 @cbron since we aren't sure how they are going to resolve the v1 v2 removal in istio 1.8 yet, I went ahead and addded `.Values.telemetry.enabled` back into the values.yaml. This will ensure future compatibility if they do decide to go that route.

cbron (dismissed):
Added telemetry.enabled back for future-proofing.

aiyengar2 (dismissed):

cbron (approved):

PR #710, opened 2020-09-26
Updating chart tags for cis-operator and security-scan images and set version to v0.2.0

prachidamle:
Updating chart tags for cis-operator and security-scan images and setting app version to v0.2.0

https://github.com/rancher/cis-operator/issues/37

### DIFF
--- a/packages/rancher-cis-benchmark/charts/Chart.yaml
+++ b/packages/rancher-cis-benchmark/charts/Chart.yaml
@@ -1,9 +1,10 @@
 apiVersion: v1
-appVersion: v0.0.1
+appVersion: v1.0.0
 description: The cis-operator enables running CIS benchmark security scans on a kubernetes
   cluster
 name: rancher-cis-benchmark
-version: 0.0.1
+version: 1.0.000
+icon: https://charts.rancher.io/assets/logos/cis-kube-bench.svg
 keywords:
 - security
 annotations:
--- a/packages/rancher-cis-benchmark/charts/values.yaml
+++ b/packages/rancher-cis-benchmark/charts/values.yaml
@@ -5,10 +5,10 @@
 image:
   cisoperator:
     repository: rancher/cis-operator
-    tag: v0.0.6
+    tag: v1.0.0
   securityScan:
     repository: rancher/security-scan
-    tag: v0.2.0
+    tag: v0.2.1
   sonobuoy:
     repository: rancher/sonobuoy-sonobuoy
     tag: v0.16.3


packages/rancher-cis-benchmark/charts/values.yaml
```diff
@@ -5,10 +5,10 @@
 image:
   cisoperator:
     repository: rancher/cis-operator
-    tag: v0.0.6
+    tag: v0.0.7
   securityScan:
     repository: rancher/security-scan
-    tag: v0.2.0
+    tag: v0.2.1
```
cbron, line 11: are these images already pushed/mirrored in dockerhub ?

packages/rancher-cis-benchmark/charts/Chart.yaml
```diff
@@ -1,9 +1,9 @@
 apiVersion: v1
-appVersion: v0.0.1
+appVersion: v0.2.0
 description: The cis-operator enables running CIS benchmark security scans on a kubernetes
   cluster
 name: rancher-cis-benchmark
-version: 0.0.1
+version: 0.2.0
```
cbron, line 6: Are we removing old version of chart ?

prachidamle: yes we want to start the app version start from v0.2.0 - I hope we need to change both version and appVersion fields?

packages/rancher-cis-benchmark/charts/values.yaml
prachidamle: yes they have been tagged and built- here https://github.com/rancher/security-scan/releases/tag/v0.2.1

packages/rancher-cis-benchmark/charts/Chart.yaml
cbron: The version corresponds to the version of the chart, the appVersion is the version of the operator. Thats why we often use the double-zero method and keep them in line. IE: operator version 1.2.1 gets `appVersion` of 1.2.1 and `version` of 1.2.100.

If we are removing the old version of the chart, you'll need to remove from the dev-v2.5 branch the entry in `assets/index.yaml`, and the assets.

aiyengar2 (dismissed):

packages/rancher-cis-benchmark/charts/Chart.yaml
prachidamle: Ok, Added this new PR for assets changes for dev-v2.5

https://github.com/rancher/charts/pull/716

packages/rancher-cis-benchmark/charts/Chart.yaml
```diff
@@ -1,9 +1,9 @@
 apiVersion: v1
-appVersion: v0.0.1
+appVersion: v0.2.0
 description: The cis-operator enables running CIS benchmark security scans on a kubernetes
   cluster
 name: rancher-cis-benchmark
-version: 0.0.1
+version: 0.2.000
```
cbron, line 6: Swapping this for 1.0.000 now

prachidamle: Swapped to 1.0.000  and also added a logo link and the logo image is added to dev-2.5 PR - is that correct @cbron

packages/rancher-cis-benchmark/charts/Chart.yaml
```diff
@@ -1,9 +1,10 @@
 apiVersion: v1
-appVersion: v0.0.1
+appVersion: v1.0.0
 description: The cis-operator enables running CIS benchmark security scans on a kubernetes
   cluster
 name: rancher-cis-benchmark
-version: 0.0.1
+version: 1.0.000
+icon: https://charts.rancher.io/assets/logos/cis-kube-bench.svg
```
cbron, line 7: This doesn't actually exist

cbron (approved):
looks good beside missing logo

packages/rancher-cis-benchmark/charts/Chart.yaml
prachidamle: Seems like all the logo images are in dev-2.5 branch - so I added it to dev-2.5 PR

prachidamle: Is this the way we need to follow? or there is some other method to get this added? @cbron

cbron: Whoops, missed your message above. Yea that works, so this PR is good.

PR #711, opened 2020-09-26
Remove webhook from index.yaml

StrongMonkey:


maggieliu:
related to: https://github.com/rancher/rancher/issues/29189

PR #715, opened 2020-09-27
Add ability to create new PVC on upgrade

mrajashree:
This commit modifies the helper for creating PVC name to address this bug
https://github.com/rancher/backup-restore-operator/issues/39

The chart name was used as PVC name till now, because of which if you upgrade the chart to use a different Persistent Volume, it would use the same PVC since the chart name stays constant. Now PVC is named with a combination of release name and revision number so it changes each time on upgrade. The PVC will be recreated and the annotation in the deployment will launch a new pod to use the new PV

Also removed the nfs pv helper which is no longer used

### DIFF
--- a/packages/rancher-backup/charts/templates/deployment.yaml
+++ b/packages/rancher-backup/charts/templates/deployment.yaml
@@ -17,7 +17,8 @@ spec:
       labels:
         {{- include "backupRestore.selectorLabels" . | nindent 8 }}
       annotations:
-        checksum/secret: {{ include (print $.Template.BasePath "/s3-secret.yaml") . | sha256sum }}
+        checksum/s3: {{ include (print $.Template.BasePath "/s3-secret.yaml") . | sha256sum }}
+        checksum/pvc: {{ include (print $.Template.BasePath "/pvc.yaml") . | sha256sum }}
     spec:
       serviceAccountName: {{ include "backupRestore.serviceAccountName" . }}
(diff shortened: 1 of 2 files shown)


packages/rancher-backup/charts/templates/deployment.yaml
```diff
@@ -17,7 +17,8 @@ spec:
       labels:
         {{- include "backupRestore.selectorLabels" . | nindent 8 }}
       annotations:
-        checksum/secret: {{ include (print $.Template.BasePath "/s3-secret.yaml") . | sha256sum }}
+        checksum/s3: {{ include (print $.Template.BasePath "/s3-secret.yaml") . | sha256sum }}
+        checksum/pvc: {{ include (print $.Template.BasePath "/pvc.yaml") . | sha256sum }}
```
prachidamle, line 21: Where are these annotations used?

aiyengar2 (approved):

packages/rancher-backup/charts/templates/deployment.yaml
mrajashree: They are just a way to make the deployment create new pods since nothing in the pod spec is changing https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments

prachidamle (approved):

dramich (approved):

PR #716, opened 2020-09-27
Removing assets, since new chart version is being started from v1.0.0

prachidamle:
Removing assets, since new chart version is being started from v0.2.0, as per the discussion here:
https://github.com/rancher/charts/pull/710#discussion_r495613240

https://github.com/rancher/cis-operator/issues/37


aiyengar2 (changes requested):
Need to remove contents of `charts/rancher-cis-benchmark`, `sha256sum/rancher-cis-benchmark`, and `assets/index.yaml` as well.

cbron (commented):
Looks fine to me, lets get other PR in first to get new version out

cbron (approved):

cbron:
If we are just removing a version and not renaming the chart, we only need to remove the assets/index.yaml and index.yaml. If renaming the chart we need to do all of them.

aiyengar2 (approved):
Needs a rebase but LGTM!

cbron (approved):
Sha removal isn't necessary but thats ok, it will just repopulate.

aiyengar2 (approved):

Issue #712, opened 2020-09-26
EFK Kibana mixed content


impsik:
I deployd EFK (ver. 7.3.0) but when i open Kibana, then i get error `Kibana did not load properly. Check the server output for more information.`
When i look browser console, then i see that content is served over http, not https. For example:
location: http://example.com/api/v1/namespaces/efk/services/kibana-http:80/proxy/built_assets/css/plugins/maps/index.light.css
`Mixed Content: The page at '<URL>' was loaded over HTTPS, but requested an insecure stylesheet '<URL>'. This request has been blocked; the content must be served over HTTPS.`

gueux:
I have faced the same issue. If enable mixed content in Chrome, next error appear:

```
Translations request failed with status code: 401
Version: 7.3.0
Build: 25250
TypeError: Failed to fetch
```

And in console:
```
Access to fetch at 'https://rancher.example.io/api/v1/namespaces/logging/services/kibana-http:80/proxy/translations/en.json' (redirected from 'https://rancher.example.io/api/v1/namespaces/logging/services/kibana-http:80/proxy/app//translations/en.json') from origin 'https://rancher.example.io' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
```

deniseschannon:
This repo was originally providing the Helm 2 charts in the "Library" catalog in Rancher in our original UI in the `master` branch. This UI was deprecated in Rancher 2.6 and the charts in them are no longer maintained. 

This repo is now serving the Rancher charts in the new Apps and Marketplace in our new UI that was introduced in Rancher 2.5. These new Helm 3 charts are in the `dev-*` and `release-*` are the ones that are being actively maintained by the Rancher team. 

Partner charts are now maintained in a different repo: rancher/partner-charts.

Due to these changes, this issue will be closed as it is referencing our old charts that aren't being updated.

PR #718, opened 2020-09-28
Add x_forwarded_proto as scheme for 301 redirect for kibana-proxy nginx-config

gueux:
Workaround for issue caused by nginx proxy between rancher and kibana #712


PennyScissors:
Closing this as we have migrated to use [rancher/partner-charts](https://github.com/rancher/partner-charts) for all partner charts moving forward.

PR #732, opened 2020-09-30
Bump webhook

dramich:
https://github.com/rancher/rancher/issues/29263


cbron (approved):
This is fine, but we also need to hide it: https://github.com/rancher/rancher/issues/29263

dramich:
beta5 adds the hidden annotation upstream

cbron (approved):

PR #731, opened 2020-09-30
Update chart version and opertor tag for rancher-backup

mrajashree:
This commit adds the new tag v1.0.0 for rancher-backup operator, and changes chart version to 1.0.000

PR #733, opened 2020-09-30
Remove older version of rancher-backup chart

mrajashree:
The chart and operator version are now updated https://github.com/rancher/charts/pull/731
This PR removes the older chart version


cbron (changes requested):
Can you leave the backup.sum as-is, that is actually correct right now

cbron (approved):

PR #738, opened 2020-10-01
Change version to avoid semver errors

mrajashree:


cbron (approved):
lgtm, need to remove 100 from dev-v2.5 and main branches

PR #740, opened 2020-10-01
Remove 1.0.000 version from chart

mrajashree:


cbron (approved):
lgtm, after other change merges to main-source we need to remove from main also

PR #743, opened 2020-10-01
Start operator version from 1.0.1 to avoid semver errors

prachidamle:


cbron (approved):
lgtm, merge this right after removal one merges or it will conflict, for a short time none will show up in dashboard

PR #746, opened 2020-10-01
added tolerations for rke controlplane/etcd nodes

paynejacob:
https://github.com/rancher/rancher/issues/29297

### DIFF
--- a/packages/rancher-logging/rancher-logging.patch
+++ b/packages/rancher-logging/rancher-logging.patch
@@ -68,7 +68,7 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-logging/charts-original/values
  
  resources: {}
    # We usually recommend not to specify default resources and to leave this as a conscious
-@@ -76,4 +76,37 @@
+@@ -76,4 +76,43 @@
  monitoring:
    # Create a Prometheus Operator ServiceMonitor object
    serviceMonitor:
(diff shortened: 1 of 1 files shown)


packages/rancher-logging/rancher-logging.patch
```diff
@@ -107,4 +107,10 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-logging/charts-original/values
 +  cattle:
 +    systemDefaultRegistry: ""
```
cbron, line 108: What value was that checkbox setting ?

packages/rancher-logging/rancher-logging.patch
```diff
@@ -107,4 +107,10 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-logging/charts-original/values
...
 +
-+fluentbit_tolerations: []
++fluentbit_tolerations:
++  - key: node-role.kubernetes.io/controlplane
++    value: "true"
++    effect: NoSchedule
++  - key: node-role.kubernetes.io/etcd
++    value: "true"
++    effect: NoExecute
```
cbron, line 116: This is specifically the settings for rke ?

paynejacob: yes, this counters the taints set on rke nodes

rmweir (approved):
LGTM

packages/rancher-logging/rancher-logging.patch
cbron: It was actually setting the fluentbit_tolerations array, so thats fine.

cbron (approved):

PR #748, opened 2020-10-01
Add default dashboard to Grafana home

aiyengar2:
Also added a small documentation update related to https://github.com/rancher/rancher/issues/29165

Related Issue: https://github.com/rancher/rancher/issues/29308

The final homepage looks like this:
Screen Shot 2020-10-01 at 4 59 56 PM

PR #756, opened 2020-10-02
Fix Grafana home dashboard layout when "Getting started" panel is injected

bashofmann:
Fix Grafana home dashboard layout when "Getting started" panel is injected

On first login Grafana automatically injects a "Getting started" panel into the home dashboard. This can not be deactivated (https://github.com/grafana/grafana/issues/8402).
To ensure that this additional panel does not break the layout, y axis line 3 has to be kept empty in the dashboard grid.

I totally missed this when I created the dashboard because I had already dismissed this magic panel.

Screenshot before the fix after logging in for the first time:

Bildschirmfoto 2020-10-02 um 20 47 19

Screenshot with the fix:

Bildschirmfoto 2020-10-02 um 21 00 15


bashofmann:
For reference: This is the injection logic in Grafana https://github.com/grafana/grafana/blob/8a5fc00330fadb0fa2cba9e21e4ca1367963418c/pkg/api/dashboard.go#L367-L383

aiyengar2 (approved):

aiyengar2:
Followup to https://github.com/rancher/charts/pull/748.

Related Issue: https://github.com/rancher/rancher/issues/29308

dramich (approved):

PR #757, opened 2020-10-02
Update rancher-backup chart version to use v1.0.2 tag

mrajashree:
v1.0.2 includes the fix for https://github.com/rancher/backup-restore-operator/issues/49

### DIFF
--- a/packages/rancher-backup/charts/Chart.yaml
+++ b/packages/rancher-backup/charts/Chart.yaml
@@ -1,11 +1,11 @@
 apiVersion: v1
-appVersion: v1.0.1
+appVersion: v1.0.2
 description: Provides ability to back up and restore the Rancher application running on any Kubernetes cluster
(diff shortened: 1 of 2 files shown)


packages/rancher-backup/charts/Chart.yaml
```diff
@@ -1,11 +1,11 @@
 apiVersion: v1
-appVersion: v1.0.1
+appVersion: v1.0.2
 description: Provides ability to back up and restore the Rancher application running on any Kubernetes cluster
 name: rancher-backup
 keywords:
 - applications
 - infrastructure
-version: 1.0.100
+version: 1.0.101
```
cbron, line 8: This should stay in line with image tag, so 1.0.200

cbron (changes requested):

cbron (approved):

PR #759, opened 2020-10-02
Update rancher-operator to v0.1.0

StrongMonkey:


cbron (approved):
Lgtm. We were bumping everything to 1.0.0, but since this is hidden it shouldn't matter. 
Can you submit a PR to remove old version from main, then we should be good.

PR #773, opened 2020-10-09
Update Istio and Kiali Versions

brendarearden:
**Problem**
Kiali released version 1.24.0
Istio released version 1.7.3 (this is a security release)
Older versions of installer use verify-install which does not work as expected in all scenarios

**Solution**
Update versions for kiali and istio to match the latest releases 
With the istio version update, the new installer without verify-install is used

Also added script to make updating the versions a lot easier. On occasion I would miss a value, the script ensures all values are getting updated as expected. 

**Issues**
rancher/rancher#29383
rancher/rancher#29249
rancher/rancher#29408

**Dependencies**
https://github.com/rancher/istio-installer/pull/12

### DIFF
--- a/packages/rancher-istio/charts/app-readme.md
+++ b/packages/rancher-istio/charts/app-readme.md
@@ -1,13 +1,3 @@
-# Rancher Istio Installers
-
-A Rancher created chart that packages the istioctl binary to install via a helm chart.
-
-# Installation Requirements 
-
-## Chart Dependencies
-- rancher-kiali-server-crd chart
-
-
 ## Kiali
 
 ###  Dependencies
@@ -27,8 +17,3 @@ To limit scraping to specific namespaces, set `prometheus.prometheusSpec.ignoreN
 
 1. Add a Service Monitor or Pod Monitor in the namespace with the targets you want to scrape.
 1. Add an additionalScrapeConfig to your rancher-monitoring instance to scrape all targets in all namespaces.
-
-# Installation
-```
-helm install rancher-istio . --create-namespace -n istio-system
-```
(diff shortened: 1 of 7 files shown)


packages/rancher-istio/scripts/version_update.go
```diff
@@ -0,0 +1,60 @@
+package main
```
nickgerace, line 1: Is bringing in source code to the charts repo intentional?

paynejacob (dismissed):

packages/rancher-istio/scripts/version_update.go
brendarearden: yes, this is a script for this specific chart update all the versions for istio and kiali easily - its just a nice to have script

packages/rancher-istio/charts/app-readme.md
```diff
@@ -0,0 +1,19 @@
+## Kiali
+
+###  Dependencies
+- rancher-monitoring chart or other monitoring installation
```
cbron, line 4:
```suggestion
- rancher-monitoring chart or other Prometheus installation
```

packages/rancher-istio/scripts/version_update.go
cbron: This won't end up in helm package right ? Want to confirm that.

brendarearden: it wont since its at the same level as the charts directory, and the rancher-istio/charts is what gets packaged

nickgerace (approved):

cbron (approved):

PR #775, opened 2020-10-09
WIP: Ensure logging pods only run on Linux nodes

nickgerace:
Ensure logging pods only run on Linux nodes by adding tolerations and
node selectors for fluentbit and fluentd. The tolerations have
"NoSchedule" for Windows nodes, and the node selectors look for nodes
with a Linux-based OS.

Solves issue: [28720](https://github.com/rancher/rancher/issues/28720)

**Please note: this is a WIP.**

PR #778, opened 2020-10-12
Ensure logging pods only run on Linux nodes

nickgerace:
Ensure logging pods only run on Linux nodes by adding tolerations and
node selectors for fluentbit and fluentd. The tolerations have
"NoSchedule" for Windows nodes, and the node selectors look for nodes
with a Linux-based OS.

Solves issue: [28720](https://github.com/rancher/rancher/issues/28720)

Please note: this is a re-opened version of PR #775 due to force-push error

### DIFF
--- a/packages/rancher-logging/overlay/templates/loggings/eks/logging.yaml
+++ b/packages/rancher-logging/overlay/templates/loggings/eks/logging.yaml
@@ -16,10 +16,15 @@ spec:
       Tag: "eks"
       Path: "/var/log/messages"
       Parser: "syslog"
-    {{ with .Values.fluentbit_tolerations }}
+  {{- $total_tolerations := concat (.Values.tolerations) (.Values.fluentbit_tolerations) }}
+  {{- with $total_tolerations }}
     tolerations:
-      {{ toYaml . | nindent 6 }}
-    {{ end }}
+      {{- toYaml . | nindent 6 }}
+  {{- end }}
+  {{- with .Values.nodeSelector }}
+    nodeSelector:
+      {{- toYaml . | nindent 6 }}
+  {{- end }}
   fluentd:
     image:
       repository: {{ template "system_default_registry" . }}{{ .Values.images.fluentd.repository }}
@@ -28,4 +33,12 @@ spec:
       repository: {{ template "system_default_registry" . }}{{ .Values.images.config_reloader.repository }}
       tag: {{ .Values.images.config_reloader.tag }}
     disablePvc: {{ .Values.disablePvc }}
+    {{- with .Values.tolerations }}
+    tolerations:
+      {{- toYaml . | nindent 6 }}
+    {{- end }}
+    {{- with .Values.nodeSelector }}
+    nodeSelector:
+      {{- toYaml . | nindent 6 }}
+    {{- end }}
 {{- end }}
--- a/packages/rancher-logging/rancher-logging.patch
+++ b/packages/rancher-logging/rancher-logging.patch
@@ -20,18 +20,6 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-logging/charts-original/Chart.
 +  catalog.cattle.io/release-name: rancher-logging
 +  catalog.cattle.io/ui-component: logging
 +  catalog.cattle.io/provides-gvr: logging.banzaicloud.io.clusterflow/v1beta1
-diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-logging/charts-original/templates/deployment.yaml packages/rancher-logging/charts/templates/deployment.yaml
---- packages/rancher-logging/charts-original/templates/deployment.yaml
-+++ packages/rancher-logging/charts/templates/deployment.yaml
-@@ -30,7 +30,7 @@
-     {{- end }}
-       containers:
-         - name: {{ .Chart.Name }}
--          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
-+          image: "{{ template "system_default_registry" . }}{{ .Values.image.repository }}:{{ .Values.image.tag }}"
-           imagePullPolicy: {{ .Values.image.pullPolicy }}
-           resources:
-             {{- toYaml .Values.resources | nindent 12 }}
 diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-logging/charts-original/templates/_helpers.tpl packages/rancher-logging/charts/templates/_helpers.tpl
 --- packages/rancher-logging/charts-original/templates/_helpers.tpl
 +++ packages/rancher-logging/charts/templates/_helpers.tpl
@@ -47,6 +35,31 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-logging/charts-original/templa
 +{{- "" -}}
 +{{- end -}}
 +{{- end -}}
+diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-logging/charts-original/templates/deployment.yaml packages/rancher-logging/charts/templates/deployment.yaml
+--- packages/rancher-logging/charts-original/templates/deployment.yaml
++++ packages/rancher-logging/charts/templates/deployment.yaml
+@@ -30,7 +30,7 @@
+     {{- end }}
+       containers:
+         - name: {{ .Chart.Name }}
+-          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
++          image: "{{ template "system_default_registry" . }}{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+           imagePullPolicy: {{ .Values.image.pullPolicy }}
+           resources:
+             {{- toYaml .Values.resources | nindent 12 }}
+@@ -45,10 +45,10 @@
+       securityContext: {{ toYaml .Values.podSecurityContext | nindent 8 }}
+     {{- end }}
+ 
+-      {{- with .Values.nodeSelector }}
++    {{- with .Values.nodeSelector }}
+       nodeSelector:
+         {{- toYaml . | nindent 8 }}
+-      {{- end }}
++    {{- end }}
+     {{- with .Values.affinity }}
+       affinity:
+         {{- toYaml . | nindent 8 }}
 diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-logging/charts-original/values.yaml packages/rancher-logging/charts/values.yaml
 --- packages/rancher-logging/charts-original/values.yaml
 +++ packages/rancher-logging/charts/values.yaml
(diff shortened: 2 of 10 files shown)


packages/rancher-logging/overlay/templates/loggings/eks/logging.yaml
```diff
@@ -20,6 +20,12 @@ spec:
     tolerations:
       {{ toYaml . | nindent 6 }}
     {{ end }}
+      - key: "cattle.io/os"
```
cbron, line 23: If `Values.fluentbit_tolerations` is empty, does the `tolerations` node still exist above for this to fall under ?

nickgerace: Good catch. It was actually the issue I ran into since I assumed we always specify a tolerance in ```values.yaml```. Fixed!

nickgerace:
This is no longer a WIP.

nickgerace:
While this is a full fix to issue [28720](https://github.com/rancher/rancher/issues/28720) from Rancher's perspective, we've uncovered an upstream bug. L204 of [this file](https://github.com/banzaicloud/logging-operator/blob/73030d200f6f02c6ce9c174fb34eb149db1312e0/pkg/resources/fluentd/appconfigmap.go#L204) shows that the ```fluentd-configcheck``` container does not inherit the tolerances and nodeselector settings from the logging-operator.

Regardless of the upstream bug, this PR is still valid, even though the next step is to address the upstream bug.

packages/rancher-logging/overlay/templates/loggings/eks/logging.yaml
```diff
@@ -16,10 +16,16 @@ spec:
       Tag: "eks"
       Path: "/var/log/messages"
       Parser: "syslog"
-    {{ with .Values.fluentbit_tolerations }}
     tolerations:
+      - key: "cattle.io/os"
```
paynejacob, line 20: can we put this with the other tolerations? The more configurable the better for users.

paynejacob (changes requested):

packages/rancher-logging/overlay/templates/loggings/eks/logging.yaml
nickgerace: In ```values.yaml``` rather than hardcoded here? I'd be fine with that, but I'm curious: if we do not want these pods to run on Windows, should we even expose that customization?

paynejacob: We don't want it scheduled on windows nodes but I think it is still important to give the users the option t. I don't have a use case for not having this toleration but putting it in values saves us the pr later if one does come up.

nickgerace: Hmmm, ok will do then!

nickgerace:
Addressed @paynejacob's comments and have moved all hardcoded values here to ```values.yaml``` instead.

paynejacob (approved):

nickgerace:
For the curious: we double-checked that this does not affect all Linux node clusters.
```bash
[nickgerace at rancherbook in ~/git/nickgerace/charts] (0) (dev-v2.5-source)
% kgpa
NAMESPACE               NAME                                           READY   STATUS      RESTARTS   AGE
cattle-logging-system   rancher-logging-d87d68c9c-kgsx7                1/1     Running     0          55s
cattle-logging-system   rancher-logging-fluentbit-68cbv                1/1     Running     0          41s
cattle-logging-system   rancher-logging-fluentbit-745st                1/1     Running     0          41s
cattle-logging-system   rancher-logging-fluentbit-7c6d7                1/1     Running     0          41s
cattle-logging-system   rancher-logging-fluentbit-bkt4p                1/1     Running     0          41s
cattle-logging-system   rancher-logging-fluentbit-dll4f                1/1     Running     0          41s
cattle-logging-system   rancher-logging-fluentbit-hl85n                1/1     Running     0          41s
cattle-logging-system   rancher-logging-fluentbit-sxwzr                1/1     Running     0          41s
cattle-logging-system   rancher-logging-fluentd-0                      2/2     Running     0          42s
cattle-logging-system   rancher-logging-fluentd-configcheck-c3054702   0/1     Completed   0          50s
```

packages/rancher-logging/rancher-logging.patch
```diff
@@ -47,6 +35,18 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-logging/charts-original/templa
...
++++ packages/rancher-logging/charts/templates/deployment.yaml
+@@ -30,7 +30,7 @@
+     {{- end }}
+       containers:
+         - name: {{ .Chart.Name }}
+-          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
++          image: "{{ template "system_default_registry" . }}{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+           imagePullPolicy: {{ .Values.image.pullPolicy }}
+           resources:
+             {{- toYaml .Values.resources | nindent 12 }}
```
cbron, line 49: Was this all just from the patch file moving things around ?

packages/rancher-logging/rancher-logging.patch
```diff
@@ -96,21 +113,33 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-logging/charts-original/values
...
++    nodeSelector:
++      kubernetes.io/os: linux
++    tolerations:
++      - key: cattle.io/os
++        operator: "Equal"
++        value: "linux"
++        effect: NoSchedule
 +  syslog_forwarder:
 +    repository: rancher/fluent-bit-out-syslog
 +    tag: 0.1.0
```
cbron, line 141: Putting it under images feels weird. The standard would be to put things above in the `tolerations` node at the main level, but since we are splitting across fluentd/fluent-bit, we could at least make a node next to that one that splits them. Like: 
```

tolerations:
   - key: cattle.io/os

fluentbit_tolerations:
    ....
fluentD_tolerations:
    ....
```
What do you think @paynejacob

cbron: We should probably also keep those standard `tolerations` ones working across the board, looks like they would be ignored in most places right now. It should use both.

packages/rancher-logging/rancher-logging.patch
nickgerace: Yes, but I asked the same question here. I ran ```git rest --hard``` to make sure, and got the same results in the patch file.

packages/rancher-logging/rancher-logging.patch
nickgerace: That makes sense. I'll adjust the PR.

nickgerace:
CI checks are failing. Checked in with @paynejacob, and it seems to be unrelated to this PR in particular.

packages/rancher-logging/overlay/templates/loggings/eks/logging.yaml
```diff
@@ -20,6 +20,10 @@ spec:
     tolerations:
       {{ toYaml . | nindent 6 }}
     {{ end }}
```
cbron, line 22: Node selector looks right now, but with tolerations you can't pass in the plain `.Values.tolerations`. Looks like you actually added that below in values.yaml but then don't use it. We should be using that one and then add-on the `fluentbit_tolerations` and `fluentd_tolerations` after

nickgerace: That's a good point. I had wanted to separate them, but they both need to be included in the templating.

paynejacob (commented):
Everything looks good outside of Caleb's comments.

packages/rancher-logging/overlay/templates/loggings/eks/logging.yaml
nickgerace: Done

nickgerace:
Fixed Caleb's comments.

mrajashree (dismissed):
lgtm

packages/rancher-logging/overlay/templates/loggings/eks/logging.yaml
```diff
@@ -16,10 +16,15 @@ spec:
       Tag: "eks"
       Path: "/var/log/messages"
       Parser: "syslog"
-    {{ with .Values.fluentbit_tolerations }}
+    {{ $total_tolerations := concat (.Values.tolerations) (.Values.fluentbit_tolerations) }}
+    {{ with $total_tolerations }}
```
cbron, line 20: Does the yaml turn out normally and spaced correctly when using concat ? I had assumed we would do a if statement that surround `tolerations:` and then two with statements. But this can work too.

cbron (commented):
Looks good

packages/rancher-logging/overlay/templates/loggings/eks/logging.yaml
nickgerace: I had started the ```if``` route, but opted to use ```concat``` after hitting a few errors and asking folks in the Kubernetes slack.

The indentation and spacing is as expected, but there are extra newlines. This does not seem to be related to using ```concat``` though. Here is some YAML from a ```--dry-run --debug``` install...

```yaml
  fluentbit:
    image:
      repository: rancher/fluent-fluent-bit
      tag: 1.5.4
    
    
    tolerations:
      
      - effect: NoSchedule
        key: cattle.io/os
        operator: Equal
        value: linux
      - effect: NoSchedule
        key: node-role.kubernetes.io/controlplane
        value: "true"
      - effect: NoExecute
        key: node-role.kubernetes.io/etcd
        value: "true"
    
    
    nodeSelector:
      
      kubernetes.io/os: linux
```

At first, I thought this was because I used ```concat```, but the ```fluentd``` section actually doesn't use it. The ```fluentd``` section is the same as the deployment below.

```yaml    
  fluentd:
    image:
      repository: rancher/banzaicloud-fluentd
      tag: v1.11.2-alpine-2
    configReloaderImage:
      repository: rancher/jimmidyson-configmap-reload
      tag: v0.2.2
    disablePvc: true
    
    tolerations:
      
      - effect: NoSchedule
        key: cattle.io/os
        operator: Equal
        value: linux
    
    
    nodeSelector:
      
      kubernetes.io/os: linux
```

Confusingly, the deployment YAML is fine.

```yaml
      containers:
        - name: rancher-logging
          image: "rancher/banzaicloud-logging-operator:3.6.0"
          imagePullPolicy: IfNotPresent
          resources:
            {}
          ports:
            - name: http
              containerPort: 8080
      nodeSelector:
        kubernetes.io/os: linux
      tolerations:
        - effect: NoSchedule
          key: cattle.io/os
          operator: Equal
          value: linux
```

I can dive deeper into this. It's functional, but may need tidying if deemed necessary.

nickgerace: Hm... I actually think I might've found the issue. The indentation of the templating lines might be creating the newlines. Testing locally... I'll update this thread.

cbron: Might just be the helm template braces not using a minus sign `{{-`

cbron (dismissed):

packages/rancher-logging/overlay/templates/loggings/eks/logging.yaml
nickgerace: Found it. Need to prepend templating conditionals with a dash character. The newlines do not appear now. Updating accordingly.

nickgerace: Fixed

nickgerace:
Fixed cosmetic issues from @cbron's comments.

nickgerace:
Removing Caleb until re-review from @paynejacob and @mrajashree. Only changes since previous approval have been cosmetic, however.

cbron (approved):
If only cosmetic we can move forward without it, but would be good to have @paynejacob confirm the new route real quick.

paynejacob (approved):

nickgerace:
Thanks to [@bacongobbler](https://github.com/bacongobbler) for the recommendation on using ```concat``` here.

PR #779, opened 2020-10-12
revert latest commit to bump versions first

cbron:
revert commit 62bf8e7 in order to bump versions first


StrongMonkey:
@cbron Did you manually revert? The commit should look something like `revert $sha...`

cbron:
Not manual, I did `git revert --no-commit 62bf8e7`. Didn't want to `reset --soft HEAD~1` because that could mess with people's history.

StrongMonkey:
Ok, ideally I think if commit contains information like which sha is reverted it should be useful

aiyengar2 (approved):

PR #780, opened 2020-10-12
Increment chart versions for new changes

cbron:
part of: https://github.com/rancher/rancher/issues/29219

### DIFF
--- a/packages/longhorn/package.yaml
+++ b/packages/longhorn/package.yaml
@@ -2,6 +2,6 @@ url: https://github.com/longhorn/charts.git
 subdirectory: charts/longhorn
 type: git
 commit: 0a88eaae4d7bdce762af942ea490d72aee338bb1
-packageVersion: 00
+packageVersion: 01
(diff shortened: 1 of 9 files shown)


aiyengar2 (approved):

StrongMonkey (commented):
Should we only bump chart that only has crd-chart?

aiyengar2:
> Should we only bump chart that only has crd-chart?

Yes, I think that all the CRD charts were modified due to the minor README.md update under `chart-templates/`, so all of the charts that use the generateCRDCharts flag should be bumped.

In general, this is something we'll also need to keep in mind for https://github.com/rancher/rancher/issues/28730#issuecomment-697596235 to enforce that CRD chart changes also necessitate a version bump

cbron:
Confirmed that bumping chart also bumps its CRD chart if we generate it. For fleet would have to do by hand.

StrongMonkey (approved):

PR #781, opened 2020-10-12
Unhide rio, update to v0.8.0

StrongMonkey:


deniseschannon (approved):

deniseschannon (approved):

packages/rio/charts/Chart.yaml
```diff
@@ -1,14 +1,13 @@
 apiVersion: v1
 name: rio
 version: 0.8.000
```
cbron, line 3: Do we not care about bumping version since it never went live ?

StrongMonkey: Yes, It is always hidden so overriding it should be fine.

PR #784, opened 2020-10-13
Fix Monitoring V2 issues in k8s 1.16 clusters

aiyengar2:
#### Summary

Root cause of the issue is that k8s 1.15 / 1.16 clusters seem to have known issues related to CRD fields that invoke `items` in it, such as the `volumes` we introduced as part of the nginx proxy that sits in front of Prometheus in Monitoring V2. This only seems to have been fixed in k8s 1.17+.

Related Issue: https://github.com/rancher/rancher/issues/29395

---

#### More context

https://github.com/kubernetes/kubernetes/issues/68466 is an issue that tracks a problem with CRDs that use fields named `items` introduced sometime before k8s 1.16.

In k8s 1.16's release notes, this issue seemed to have been resolved as it shows up under `Other Notable Changes` [here](https://v1-16.docs.kubernetes.io/docs/setup/release/notes/#other-notable-changes-7):
```
Fix CRD validation error on ‘items’ field. (#76124, @tossmilestone)
```

However, when looking into the actual [PR](https://github.com/kubernetes/kubernetes/pull/76124) that was executed to fix this issue, it seems like this issue had a followup in https://github.com/kubernetes/kubernetes/pull/85223 that was only resolved in k8s 1.17+.

This shows up in the k8s 1.17 release notes under `API Machinery` [here](https://v1-17.docs.kubernetes.io/docs/setup/release/notes/#api-machinery):
```
CRDs can have fields named type with value array and nested array with items fields without validation to fall over this. (#85223, @sttts)
```

Therefore, the quick fix was to just modify the contents provided to the `volumes` field of the `prometheusSpec`; once modified, Monitoring V2 successfully deployed in a k8s 1.16 cluster.


brendarearden (approved):
looks good!

mrajashree (dismissed):

aiyengar2:
Accidentally pushed a commit, no changes were made since the last review.

dramich (approved):

PR #790, opened 2020-10-16
Updating cis-operator tag and updating cis ClusterRoles to drop label and ns

prachidamle:
1) Updating the chart to point to new cis-operator tag that pulls the scheduled Scan functionality
- https://github.com/rancher/cis-operator/issues/27

2) Also removing the aggregator label from cis ClusterRoles since the aggregation does not apply to cluster scoped CIS CRDs
3) Also removing the namespace from ClusterRoles since namespace is not required for ClusterRoles that applies cluster wide.
- https://github.com/rancher/cis-operator/issues/41

### DIFF
--- a/packages/rancher-cis-benchmark/charts/Chart.yaml
+++ b/packages/rancher-cis-benchmark/charts/Chart.yaml
@@ -1,9 +1,9 @@
 apiVersion: v1
-appVersion: v1.0.1
+appVersion: v1.0.2
 description: The cis-operator enables running CIS benchmark security scans on a kubernetes
   cluster
 name: rancher-cis-benchmark
-version: 1.0.101
+version: 1.0.200
 icon: https://charts.rancher.io/assets/logos/cis-kube-bench.svg
 keywords:
 - security
--- a/packages/rancher-cis-benchmark/charts/templates/cis-roles.yaml
+++ b/packages/rancher-cis-benchmark/charts/templates/cis-roles.yaml
@@ -2,10 +2,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  labels:
-    rbac.authorization.k8s.io/aggregate-to-admin: "true"
   name: cis-admin
-  namespace: {{ template "cis.namespace" . }}
 rules:
   - apiGroups:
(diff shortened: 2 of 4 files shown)


packages/rancher-cis-benchmark/charts/Chart.yaml
```diff
@@ -1,9 +1,9 @@
 apiVersion: v1
-appVersion: v1.0.1
+appVersion: v1.0.2
 description: The cis-operator enables running CIS benchmark security scans on a kubernetes
   cluster
 name: rancher-cis-benchmark
-version: 1.0.101
+version: 1.0.201
```
mrajashree, line 6: not sure but I think this should be 1.0.200? Rest LGTM

prachidamle: yup, done

mrajashree (approved):

packages/rancher-cis-benchmark/charts/templates/cis-roles.yaml
```diff
@@ -19,26 +16,6 @@ rules:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  labels:
-    rbac.authorization.k8s.io/aggregate-to-edit: "true"
-  namespace: {{ template "cis.namespace" . }}
-  name: cis-edit
```
cbron, line 25: Are we getting rid of edit because it's the same as admin ? We'll need to update the docs.

prachidamle: I think you suggested that in the design meeting? hence I removed it

cbron: Ha, ok. Should be fine, only downside is if people already were using it but we can document that it changed to only admin.

cbron (approved):

packages/rancher-cis-benchmark/charts/templates/cis-roles.yaml
prachidamle: @cbron  added this docs PR as well: https://github.com/rancher/docs/pull/2785

PR #796, opened 2020-10-20
Revert the ScheduledScan CRD and cis-operator image tag

prachidamle:
There has been a Design chnage on https://github.com/rancher/cis-operator/issues/27

Hence removing the ScheduledScan CRD added and reverting to previous image tag for cis-operator


cbron (approved):
Assuming this was a literal `git revert` command it should be fine. This should merge after the other one to avoid conflicts.

prachidamle:
@cbron Actually the PR that is supposed to be reverted involves changes for 2 issues
- https://github.com/rancher/cis-operator/issues/41
- https://github.com/rancher/cis-operator/issues/27 

But only the changes for https://github.com/rancher/cis-operator/issues/27 need to be reverted. Hence I have manually deleted those files. 

I think we would still need to up the  chart version and so we need not revert the version change from dev-2.5 then. Correct?

PR #797, opened 2020-10-20
Add tidb-operator v1.1.6

AstroProfundis:
Signed-off-by: Allen Zhong <zhongbenli@pingcap.com>

Update tidb-operator to the latest stable release.


PennyScissors:
@AstroProfundis Could you please update the `questions.yaml`? This is used in the UI to enable users to configure the helm chart. It would be good to at least add a few question to cover common use cases so that users can install the chart with little effort.

AstroProfundis:
@PennyScissors Thanks for the remind, I've added some important fields to the questions list.

As Tidb Operator itself is very lightweight, there is no much to change for the operator to use, most configurable fields appear when the user want to use it to deploy a real TiDB cluster.

PennyScissors (approved):
lgtm

PR #799, opened 2020-10-21
Bump image version

brendarearden:
**Problem**
- Need to update to include grafana dashboard changes
- Kiali dependencies is still not very clear and inconsistency in capitalization 

**Solution**
- Need to bump installer version to add the grafana dashboard changes
- Updated readme to make some clarifications for the kiali dependencies, and added consistent capitalization. 

**Issues**
https://github.com/rancher/rancher/issues/28468
https://github.com/rancher/rancher/issues/29418

**Additional Info**
Since rancher-istio 1.7.3 chart is not in production, not bumping version - will just overwrite existing dev chart

### DIFF
--- a/packages/rancher-istio/charts/values.yaml
+++ b/packages/rancher-istio/charts/values.yaml
@@ -5,7 +5,7 @@ forceInstall: false
 
 installer:
   repository: rancher/istio-installer
-  tag: 1.7.3-rancher1
+  tag: 1.7.3-rancher2
 
 istiocoredns:
   enabled: false
--- a/packages/rancher-istio/charts/README.md
+++ b/packages/rancher-istio/charts/README.md
@@ -13,9 +13,9 @@ A Rancher created chart that packages the istioctl binary to install via a helm 
 ###  Dependencies
 - rancher-monitoring chart or other Prometheus installation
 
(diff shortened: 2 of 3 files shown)


nickgerace (dismissed):

paynejacob (approved):

paynejacob (dismissed):

brendarearden:
removed some wording from the >note section in the readme and app-readme

paynejacob (approved):

packages/rancher-istio/charts/values.yaml
```diff
@@ -5,7 +5,7 @@ forceInstall: false
 
 installer:
   repository: rancher/istio-installer
-  tag: 1.7.3-rancher1
+  tag: 1.7.3-rancher2
```
cbron, line 9: I thought we added a grafana flag for on or off, but I guess not. It just runs if its able. We might want to add that later.

cbron (approved):
Already new istio version in charts so no need to bump app version here

nickgerace (approved):

packages/rancher-istio/charts/values.yaml
brendarearden: Yeah, it seemed like grafana won't install unless rancher-monitoring is installed and you would set grafana enabled there....

PR #800, opened 2020-10-21
Add a pre-install job to remove existing crds

prachidamle:
https://github.com/rancher/rancher/issues/29188


PennyScissors:
Does the job need to be in the crd chart? This will create a new directory `crd/` in the parent chart and the job will be placed there. I tried renaming the directory to `crds/` which is what the script that splits them expects, but it seems to happen before placing overlays so it just ends up in the parent chart.

dramich:
This seems like a really big hammer, if any CRs live under those CRDs they will also be deleted. What are other charts doing to work around this?

prachidamle:
This is not going to work anyways with our current charts scripts. The scripts do not overlay anything in charts/crd folder and  so this job is not copied over to the chart and not run.

PR #803, opened 2020-10-22
Add generateCRDChart.assumeOwnershipOfCRDs for rancher-gatekeeper

prachidamle:
If gatekeeper is already installed on a rancher-2.4 cluster, then after upgrading to 2.5 if user wants to update gatekeeper app to its latest GA chart, there is no upgrade path. 
User has to uninstall the 2.4 gatekeeper via ember UI and then re-install the new rancher-gatekeeper app via Cluster-Explorer

Even if the previous gatekeeper is uninstalled, helm2 does not delete the crds. These crds are installed by another helm release and so while installing the 2.5 rancher-gatekeeper the crd installation errors out.

The CRD's have not changed between the 2 gatekeeper releases, hence we can just transfer the ownership to the 2.5 helm release by adding flag assumeOwnershipOfCRD to the chart

https://github.com/rancher/rancher/issues/29188

After applying this fix following tests have been performed and working fine:
A] Uninstalling OPA v1 on Rancher 2.4.8 and installing OPA v2 after upgrading to Rancher 2.5
1) On HA setup install Rancher v2.4.8 and then enable OPA Gatekeeper v1 via Dashboard
2) Create few constraints and make sure they work
3) Now Upgrade Rancher to v2.5 
4) OPA Gatekeeper v1 is still working
5) Now delete OPA Gatekeeper v1 by Removing the App from Ember UI or ClusterManager
6) After uninstall, go to ClusterExplorer and Install new rancher-gatekeeper App
7) Installation succeeds and you can check via kubectl that the 2 v1 CRDs are present and 2 new v2 CRDS are added.

8) Also gatekeeper v2 versions can be upgraded

### DIFF
--- a/packages/rancher-gatekeeper/package.yaml
+++ b/packages/rancher-gatekeeper/package.yaml
@@ -1,4 +1,5 @@
 url: https://open-policy-agent.github.io/gatekeeper/charts/gatekeeper-3.2.1.tgz
-packageVersion: 00
+packageVersion: 01
 generateCRDChart:
   enabled: true
+  assumeOwnershipOfCRDs: true


cbron (commented):
I'd really like Arvind's review here, this flag was originally meant for rancher-monitoring, and I'm not sure what the plans are for it in the scripts rewrite. I want to move towards any in-house chart not using automation, that would simplify scripts and take away the magic, as with the "magic" no one understands how things work or the current state of the chart.
Sadly we can't just generate this once and copy into the CRD chart, as that is also auto-generated.

prachidamle:
This monitoring PR has been referred for this change: https://github.com/rancher/charts/commit/9fe6ed19c14825b80f4618ced66d32a1e73d98fc#diff-93c5f8e80a04a2a5141b2c41f8f760a87649f1cd8a9b0112598103988b4f1317

StrongMonkey:
@cbron That's the legit concern, I think we should wait for @aiyengar2 to review then.

aiyengar2:
Sorry for the late review!

w.r.t. the charts script rewrite, I think that we probably still want to support automation when it comes to CRD charts in general because the broad majority of charts just need to:
1. Copy files from the CRD directory to the `templates/`
2. Fill out a basic `Chart.yaml` and `README.md` template for the CRD chart
3. Add a fixed set of annotations to the CRD chart
4. Add a fixed set of annotations to the main chart
5. Add a fixed validation file to the main chart to check if CRDs are installed before installing

While it does introduce some magic into the way things are built, I think that these basic steps are okay to automate for most charts so we don't need to introduce / maintain a second chart and keep them in sync.

However, w.r.t. special options like `assumeOwnershipOfCRDs`, I was planning to introduce the idea of a `crd-template/` folder (in the same directory as `overlay/`). It would basically follow the above steps, except it would place your CRDs in the `crd-manifest/` directory instead of the `template/` directory. Then it would overlay the contents of `crd-template/` into the CRD chart, with support for using environment variables to fill out some of the values.

This is effectively what the scripts are doing right now for `assumeOwnershipOfCRDs`. For the rewrite, I was just planning to do:
`mv -R ./scripts/chart-templates/crd-assume-ownership  ./packages/rancher-monitoring/crd-template`. We can perform this for CIS operator as well.

aiyengar2 (dismissed):
I think this LGTM.

w.r.t. [Caleb's comment](https://github.com/rancher/charts/pull/803#pullrequestreview-514904132):
> I want to move towards any in-house chart not using automation

I think the switch to not using automation should be really easy since we would just need to copy the generated CRD chart from the https://github.com/rancher/charts/tree/main/charts to create the new chart. If possible, I think we should stick to using the automation if it completely meets CIS's requirements and is already easily available.

That way problems like https://github.com/rancher/rancher/issues/29973 can be solved together.

prachidamle:
@cbron and @StrongMonkey can you confirm if this fix can be merged?

aiyengar2:
We should check if this exposes CIS to the problem in https://github.com/rancher/rancher/issues/29980, as described in https://github.com/rancher/rancher/issues/29980#issuecomment-724221496, before merging this PR. Should be a quick fix.

cbron:
This can't be merged until post-2.5.2 release. 

As far as the scripts rewrite goes, we could just do the manual copy now. Generate this once and then copy it out and handle manually from then on. The downside is that we can't use any CRD automation again, meaning if CRD charts need a new annotation or something that would have to be manually added going forward. But might not be a bad path since that is rare, and we could use the scripts to just regenerate it again manually. Otherwise the approach would be to use the current or new `crd-template` automation for now. Any opinions on either approach ?

aiyengar2:
@cbron if we want to split it out with a manual copy, that should be okay. I plan to add support for charts that have custom CRD charts in a separate repository (such as fleet) in the charts script rewrite as well, so that we could support CRD automation for those types of charts as well.

cbron:
The only reason is that we'd like to cut back on automation in general for charts. Long term goal is that all charts get moved to their own upstream repo and managed like fleet does. Maybe for solving those problems we could create a library of helpers or something to pull in. But I get that this one is rather large, not as simple as that. I guess I'm fine either way for now since we are in the middle of things. @StrongMonkey or @prachidamle do you guys have opinion ?

prachidamle:
@cbron I feel we should merge this for now and fix it using automation just as any other changes have been so far - and when we are making any chart scripts refactor we could take this up along with the move to separate chart repos?

cbron (dismissed):
Ok, I guess we can go that route for now as long as we attempt to move away from the automation in the future. Also we need to ensure that the upgrade process is smooth, and be aware of how removing this in the future could affect the chart/CRDs. 

Also make sure that gatekeeper has an issue for version change.

prachidamle:
Gatekeeper issue for version change: https://github.com/rancher/rancher/issues/30179

aiyengar2 (approved):
LGTM

PR #805, opened 2020-10-22
Add runAsNonRoot to false to override init container

StrongMonkey:
The initChown container needs to run as root user so that runAsNonRoot
has to be disabled to respect root user setting

### DIFF
--- a/packages/rancher-monitoring/rancher-monitoring.patch
+++ b/packages/rancher-monitoring/rancher-monitoring.patch
@@ -119,7 +119,7 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-monitoring/charts-original/cha
  {{- define "grafana.pod" -}}
  {{- if .Values.schedulerName }}
  schedulerName: "{{ .Values.schedulerName }}"
-@@ -21,9 +20,9 @@
+@@ -21,12 +20,13 @@
  {{- if ( and .Values.persistence.enabled .Values.initChownData.enabled ) }}
    - name: init-chown-data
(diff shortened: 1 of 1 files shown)


StrongMonkey:
https://github.com/rancher/rancher/issues/29639

StrongMonkey:
I also submitted an upstream PR. I think this can be fixed in upstream chart also. https://github.com/grafana/helm-charts/pull/63

packages/rancher-monitoring/rancher-monitoring.patch
```diff
@@ -131,7 +131,11 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-monitoring/charts-original/cha
      {{- end }}
      imagePullPolicy: {{ .Values.initChownData.image.pullPolicy }}
      securityContext:
-@@ -41,9 +40,9 @@
++      runAsNonRoot: false
```
cbron, line 134: Weird, wouldn't runAsUser handle this for you ? Or do you need both to enable root user.

StrongMonkey: `runAsNonRoot` will force you to run as non root. Due to the CIS run we made all the pod run with `runAsNonRoot=true`. This initChown container needs to run as root user so that `runAsNonRoot=true` can not be set. That's why we can set it on container level so it is overridden.

cbron: Mmk, so you need both then.

cbron (approved):

PR #811, opened 2020-10-27
Update storageos-operator to version 0.3.2

darkowlzz:


PennyScissors (approved):
I'm approving this because the chart itself is fine, and my understanding is that this PR is time sensitive, but it's in the wrong location. Please move the `storage-operator/` directory with all the charts out of `proposed/` into `charts/` in a separate PR. Thanks!

PR #813, opened 2020-10-28
Add updating chart with upstream workflow

nickgerace:
Add updating chart with upstream workflow. This includes a
troubleshooting section.

EDIT: you can see what the README looks like rendered [here](https://github.com/nickgerace/rancher-charts/blob/dev-v2.5-source-docs/README.md).

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ The step includes:
    
 2. Change the version in `package.yaml`. If upstream chart needs to be updated, update url to point the latest chart. `packageVersion` also needs to updated.
 
-3. Make modification to your charts. 
+3. Make modifications to your charts.
 
 4. Run `make CHART={CHART_NAME} patch`
  
@@ -97,6 +97,25 @@ The step includes:
 
 This repo provides a [workflow](./.github/workflows) that automatically uploads patch files and tarball of charts. Commit will only need to update `package/${chart-name}/charts` and make sure patches are 
 up-to-date with the latest chart. It also automatically build github pages to serve `index.yaml` and artifacts of charts.
+
+### Troubleshooting
+
+When updating your chart with the upstream version you may encounter conflicts, or have general questions about the process. Below are some troubleshooting tips.
+
+#### How do I find all the image tags from the upstream chart?
+
+Check the new image versions (and entirely new images) in the upstream chart, and update them in the local chart.
+There are multiple ways to find them, but the easiest may be to install the upstream chart on a sandboxed cluster and track all images used.
+
+#### How do I update the image tags or add new images to our mirror?
+
+Add or update images in the new chart by opening a PR in [image-mirror](https://github.com/rancher/image-mirror).
+
+#### How do I generate a CRD chart when there's conflicts?
+
+Encountering conflicts when running `prepare` can create a scenario where the `charts-crd` chart is never generated.
+While this workflow may not work for all conflicts, you can copy the files with conflicts to a temporary location, manually triage the conflicts, and then copy those files back.
+You may also want to remove the files with conflicts from the chart's `patch` file before running `prepare` again.
 
 ### Experimental: Splitting CRDs from an upstream package into a separate package
 


README.md
```diff
@@ -98,6 +98,50 @@ The step includes:
...
+
+# Fix the conflict files files manually and store them in a local directory.
+# Store the package.yaml file in the local directory as well/
+mkdir tmp/
+cp <modified-files> tmp/
+cp packages/<insert-chart>/package.yaml tmp/
+
+# Make the repository clean again. You may have to remove artifacts.
+make clean
+git reset --hard
```
aiyengar2, line 132: I'm not sure if the recommended Bash commands you added to the README to resolve patch conflicts would actually work for all (or most) charts at the moment due to special logic in our scripts to support things like CRD charts or patching subcharts; it would probably be something that works perfectly fine for charts that have simple patches (e.g. longhorn), so maybe we should leave a note about that.

For full coverage, we'd probably need something more heavyweight like https://github.com/aiyengar2/charts/blob/d3e90909c7f28d5c66eb1f287e877c61a58310e8/scripts/generate-rebased-patch, which is a script I used earlier to automate the process of rebasing Rancher Monitoring to a new upstream.

aiyengar2 (changes requested):
@nickgerace thanks for taking the initiative with this README.md update! Left some comments

README.md
aiyengar2: For reference, the stuff in https://github.com/aiyengar2/charts/blob/d3e90909c7f28d5c66eb1f287e877c61a58310e8/scripts/generate-rebased-patch#L77-L133 is what actually handled the special cases I mentioned above; this was a script I made for an earlier version of the charts scripts tho, so it probably needs to be rewritten now.

nickgerace: What do you suggest to do with this troubleshooting guide? I needed it for my workflow _because_ the `charts-crd` chart was not created. I couldn't patch my changes because the conflicts caused `charts-crd` to never be generated. The only way to avoid that was to remove the affected files from the patch file.

README.md
```diff
@@ -98,6 +98,50 @@ The step includes:
...
+To update an existing chart with upstream, run through the following steps.
+
+1. Change the URL in `package.yaml` to the upstream version.
+2. Run through the **Workflow** section above until the `patch` step.
+3. Check the new image versions (and entirely new images) in the upstream chart, and update them in the local chart. There are multiple ways to find them, but the easiest may be to install the upstream chart on a sandboxed cluster and track all images used.
+4. Add or update images in [image-mirror](https://github.com/rancher/image-mirror).
+5. Continue the **Workflow** section at the `patch` step, and complete it.
+6. Once the images have been added/updated, test the newly updated chart.
+
+#### Troubleshoting
```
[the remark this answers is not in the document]
nickgerace: Sure thing

README.md
```diff
@@ -98,6 +98,50 @@ The step includes:
...
+Here the recommended workflow in this scenario…
+
+```bash
+# Start with a clean repository. Change the URL to the upstream version.
+sed -i 's/<insert-old-url>/<insert-new-url>/g' packages/<insert-chart>/package.yaml
+
+# Now, run prepare. You'll encounter conflicts, but that is okay for now.
+make CHART=<insert-chart> prepare
+
+# Fix the conflict files files manually and store them in a local directory.
```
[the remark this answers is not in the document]
nickgerace: Fixed

README.md
```diff
@@ -98,6 +98,50 @@ The step includes:
...
+2. Run through the **Workflow** section above until the `patch` step.
+3. Check the new image versions (and entirely new images) in the upstream chart, and update them in the local chart. There are multiple ways to find them, but the easiest may be to install the upstream chart on a sandboxed cluster and track all images used.
+4. Add or update images in [image-mirror](https://github.com/rancher/image-mirror).
+5. Continue the **Workflow** section at the `patch` step, and complete it.
+6. Once the images have been added/updated, test the newly updated chart.
+
+#### Troubleshoting
+
+If upgrading the chart with the upstream version, you may encounter conflicts when running `prepare`.
+Here the recommended workflow in this scenario…
```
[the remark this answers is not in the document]
nickgerace: I'll do this

README.md
```diff
@@ -98,6 +98,50 @@ The step includes:
...
 
+### Updating Charts with Upstream
+
+To update an existing chart with upstream, run through the following steps.
+
+1. Change the URL in `package.yaml` to the upstream version.
+2. Run through the **Workflow** section above until the `patch` step.
+3. Check the new image versions (and entirely new images) in the upstream chart, and update them in the local chart. There are multiple ways to find them, but the easiest may be to install the upstream chart on a sandboxed cluster and track all images used.
+4. Add or update images in [image-mirror](https://github.com/rancher/image-mirror).
+5. Continue the **Workflow** section at the `patch` step, and complete it.
```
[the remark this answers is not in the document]
nickgerace: Okay that sounds good

nickgerace:
Addressed all your comments but one @aiyengar2. Thank you for taking a look!

Even if it's not the `bash` commands, I think that having *some* form of documentation on what to do in a conflict scenario is important though. I only knew to cache my changes in a `tmp/` directory and edit the patch file manually through word of mouth. However, that could also be because I am new to the patch workflow.

aiyengar2 (approved):

README.md
aiyengar2: Yeah the charts-crd fails to be created because we patch before we attempt to split out the CRD chart.

I think what you have now is okay since you're just recommending copying the specific files that do have patch conflicts to a temporary directory and removing the conflicting patches from the patch file (as opposed to copying the entire directory, which would probably fail); this should work in general, but probably (not 100% sure) will not work for charts that modify the `requirements.yaml` or try to patch subcharts like monitoring / Istio. It also won't be feasible probably for charts that have enormous patches (again like monitoring / istio).

But neither of those issues impact the majority of chart owners and this README will anyways be rewritten post https://github.com/rancher/rancher/issues/28730, I think this should be good for now 🙂

nickgerace: Fixed by removing the bash commands themselves. Replaced with generalized language that gets the same idea across.

nickgerace:
Addressed @aiyengar2's last review by removing the bash commands themselves. Replaced with generalized language that gets the same idea across.

README.md
```diff
@@ -98,6 +98,26 @@ The step includes:
 This repo provides a [workflow](./.github/workflows) that automatically uploads patch files and tarball of charts. Commit will only need to update `package/${chart-name}/charts` and make sure patches are 
 up-to-date with the latest chart. It also automatically build github pages to serve `index.yaml` and artifacts of charts.
 
+### Troubleshooting
+
+When updating your chart with the upstream version, you may encounter conflicts, or have general questions about the process.
+Below are some troubleshooting tips.
```
aiyengar2, line 104: nit:
```suggestion
When updating your chart with the upstream version, you may encounter conflicts or have general questions about the process. Below are some troubleshooting tips:
```

aiyengar2 (approved):
nit comment, LGTM otherwise

README.md
```diff
@@ -98,6 +98,26 @@ The step includes:
 This repo provides a [workflow](./.github/workflows) that automatically uploads patch files and tarball of charts. Commit will only need to update `package/${chart-name}/charts` and make sure patches are 
 up-to-date with the latest chart. It also automatically build github pages to serve `index.yaml` and artifacts of charts.
 
+### Troubleshooting
+
+When updating your chart with the upstream version, you may encounter conflicts, or have general questions about the process.
```
rmweir, line 103: nit: I'm not sure if the first comma is necessary

README.md
```diff
@@ -98,6 +98,26 @@ The step includes:
...
+Below are some troubleshooting tips.
+
+#### How do I find the all image tags from the upstream chart?
```
rmweir, line 106: Should this be
```suggestion
#### How do I find all the image tags from the upstream chart?
```

rmweir (changes requested):

aiyengar2 (approved):
LGTM

rmweir (approved):
LGTM

cbron (approved):

PR #814, opened 2020-10-29
Remove hostPath type from PushProx client

aiyengar2:
The issue in https://github.com/kubernetes/kubernetes/issues/83125 was encountered when trying to use /opt/rke/etc/kubernetes/ssl as the hostPath. Removing this resolves it.

Related Issue: https://github.com/rancher/rancher/issues/29815

### DIFF
--- a/packages/rancher-pushprox/charts/templates/pushprox-clients.yaml
+++ b/packages/rancher-pushprox/charts/templates/pushprox-clients.yaml
@@ -127,7 +127,6 @@ spec:
       - name: metrics-cert-dir-source
         hostPath:
           path: {{ required "Need access to volume on host with the SSL cert files to use HTTPs" .Values.clients.https.certDir }}
-          type: Directory
       - name: metrics-cert-dir
         emptyDir: {}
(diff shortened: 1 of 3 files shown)


packages/rancher-pushprox/charts/templates/pushprox-clients.yaml
```diff
@@ -127,7 +127,6 @@ spec:
       - name: metrics-cert-dir-source
         hostPath:
           path: {{ required "Need access to volume on host with the SSL cert files to use HTTPs" .Values.clients.https.certDir }}
-          type: Directory
```
mrajashree, line 130: do we need to replace this with anything?

aiyengar2: Nah, removing the type field makes it default to `type: ""`, which is the workaround described in https://github.com/kubernetes/kubernetes/issues/83125

mrajashree (approved):

dramich (approved):

PR #815, opened 2020-10-29
add rke2 as scan target

briandowns:
Add support for RKE2.

Reference issue: https://github.com/rancher/rke2/issues/1

Added a ticket under cis-operator  for tracking these changes: https://github.com/rancher/cis-operator/issues/59

### DIFF
--- a/packages/rancher-cis-benchmark/charts/templates/configmap.yaml
+++ b/packages/rancher-cis-benchmark/charts/templates/configmap.yaml
@@ -6,6 +6,7 @@ metadata:
 data:
   # Default ClusterScanProfiles per cluster provider type
   rke: "rke-profile-permissive"
+  rke2: "rke2-cis-1.5-profile-permissive"
   eks: "eks-profile"
   gke: "gke-profile"
   default: "cis-1.5-profile"
--- a/packages/rancher-cis-benchmark/charts/values.yaml
+++ b/packages/rancher-cis-benchmark/charts/values.yaml
@@ -5,10 +5,10 @@
 image:
   cisoperator:
     repository: rancher/cis-operator
-    tag: v1.0.1
+    tag: v1.0.3-rc2
   securityScan:
     repository: rancher/security-scan
-    tag: v0.2.1
+    tag: v0.2.2-rc1
   sonobuoy:
     repository: rancher/sonobuoy-sonobuoy
     tag: v0.16.3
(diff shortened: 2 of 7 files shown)


brandond (approved):

packages/rancher-cis-benchmark/charts/templates/configmap.yaml
```diff
@@ -6,6 +6,8 @@ metadata:
 data:
   # Default ClusterScanProfiles per cluster provider type
   rke: "rke-profile-permissive"
+  rke2: "rke2-profile-hardened"
+  rke2: "rke2-profile-permissive"
```
prachidamle, line 10: just add one profile to this configmap - its the profile to be run by default if user does not selects one

packages/rancher-cis-benchmark/charts/templates/configmap.yaml
```diff
@@ -6,6 +6,8 @@ metadata:
 data:
   # Default ClusterScanProfiles per cluster provider type
   rke: "rke-profile-permissive"
+  rke2: "rke2-cis-1.5-profile-hardened"
```
prachidamle, line 9: remove the hardened profile mapping from this file - here we maintain only one default profile per clusterProvider

prachidamle (commented):
We need to use the updated the image tags in Values.yaml for rancher/security-scan and rancher/cis-operator so that the new rke2 benchmark is available.

prachidamle (commented):
@briandowns Changes look good, but can you please squash all individual commits into one

prachidamle (approved):
LGTM

prachidamle:
@cbron Can you please review this CIS chart change PR - this will the first PR for 2.5.4 CIS changes and will increment the chart ad app version cc @mrajashree

packages/rancher-cis-benchmark/charts/values.yaml
```diff
@@ -5,10 +5,10 @@
 image:
   cisoperator:
     repository: rancher/cis-operator
-    tag: v1.0.1
+    tag: v1.0.3-rc2
```
cbron, line 8: You can use the other issue you created (that tracks the new chart version) to remember to tag the real image so that the RC2 image gets replaced.

cbron (approved):

packages/rancher-cis-benchmark/charts/values.yaml
prachidamle: @cbron if we update the image tag later, can we then keep the appVersion same as v1.0.3

cbron: yes

brandond (approved):

PR #819, opened 2020-10-29
Add app-README for rancher-backup

mrajashree:
Screenshot of what this looks like in the UI:
Screen Shot 2020-10-30 at 11 24 29 AM

### DIFF
--- a/packages/rancher-backup/charts/app-README.md
+++ b/packages/rancher-backup/charts/app-README.md
@@ -1,0 +1,15 @@
+# Rancher Backup
+
+The Rancher Backup chart provides ability to back up and restore the Rancher application running on any Kubernetes cluster. It can be used for migrating Rancher running on one cluster to another.
+-  It backs up all Kubernetes resources and CRDs that Rancher creates and manages from the local cluster. It gathers these resources by querying the Kubernetes API server, packages all resources to create a tarball file and saves it in the configured backup storage location.
+- The chart can be configured to store backups in S3-compatible object stores such as AWS S3 and MinIO, and in Persistent Volumes. The [storage configuration](https://rancher.com/docs/rancher/v2.x/en/backups/v2.5/configuration/storage-config/) guide has steps for configuring these storage locations. You will have the option to override this with each backup, but will be limited to using an S3-compatible object store.
+- It preserves the ownerReferences on all resources, hence maintaining dependencies between objects.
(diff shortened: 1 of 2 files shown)


packages/rancher-backup/charts/app-README.md
```diff
@@ -0,0 +1,34 @@
+# Rancher Backup
+
+This chart provides ability to back up and restore the Rancher application running on any Kubernetes cluster.
+Refer [this](https://github.com/rancher/backup-restore-operator) repository for implementation details.
+
+### Configuration
```
aiyengar2, line 6: Should the `Configuration` and `CRDs` section should be omitted here since they are already discussed in the `README.md`?

aiyengar2: Since the purpose of the `app-README.md` is to be the first page that a user sees when installing a chart via `Apps & Marketplace`, I think it should be more high-level about what the chart is (i.e. how you would summarize `rancher-backup` to someone in person) rather than how to configure it, which is already covered in the `README.md`.

aiyengar2: For example, for `rancher-backup` it might make sense to add something similar to the [User Flow](https://github.com/rancher/backup-restore-operator#user-flow) section in the repo's README.md. 

Something that explains how the chart actually provides the ability to backup and restore Rancher by 1. Using the default `ResourceSet` to identify which resources need to be backed up 2. creating a `Backup` (also should explain what storage locations are supported e.g. S3, PV, etc.) and 3. creating a `Restore` to actually perform the backup.

aiyengar2 (changes requested):

packages/rancher-backup/charts/app-README.md
mrajashree: yeah I was thinking of something similar, that's why I thought I'll mark the PR as draft and edit it more

deniseschannon: yeah. we could remove this so it looks nicer and is more high level descrition of the chart since the Helm README is available inside the app deployment.

mrajashree: @deniseschannon @aiyengar2 the PR is still Draft/WIP, I'll request review once it's ready

mrajashree:
As per [this](https://github.com/rancher/charts/pull/820#pullrequestreview-520961425) comment, adding sections for Workloads and CRDs, screenshot from UI:
Screen Shot 2020-11-02 at 1 35 46 PM

deniseschannon:
Closing in favor of https://github.com/rancher/charts/pull/827/files

PR #820, opened 2020-10-30
Adding app-readme for CIS chart

prachidamle:
Adding an app-readme for CIS chart

### DIFF
--- a/packages/rancher-cis-benchmark/charts/app-readme.md
+++ b/packages/rancher-cis-benchmark/charts/app-readme.md
@@ -1,0 +1,12 @@
+# Rancher CIS Benchmark
+
+This chart installs the following components to enable security scanning of your cluster using benchmarks published by CIS (Center for Internet Security):
+
+## Workloads
+* [cis-operator](https://github.com/rancher/cis-operator)- The cis-operator handles launching the [kube-bench](https://github.com/aquasecurity/kube-bench) tool that runs a suite of CIS tests on the nodes of your Kubernetes cluster. After scans finish, the cis-operator generates a compliance report that can be downloaded.
+
+## CRDs & CRs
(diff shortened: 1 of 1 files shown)


deniseschannon (changes requested):
Update text to this. I've left the CRD names and explained how we're installing CRDs and specific CRs within the chart. Let me know if this make sense. 
---

# Rancher CIS Benchmark

This chart installs the following components to enable security scanning of your cluster using benchmarks published by CIS (Center for Internet Security):

## Workloads
* [cis-operator](https://github.com/rancher/cis-operator)- The cis-operator handles launching the [kube-bench](https://github.com/aquasecurity/kube-bench) tool that runs a suite of CIS tests on the nodes of your Kubernetes cluster. After scans finish, the cis-operator generates a compliance report that can be downloaded.

## CRDs & CRs
* `ClusterScanBenchmark` - A `ClusterScanBenchmark` is a CRD that is used to define the CIS benchmark version to run using kube-bench as well as the valid configuration parameters for that benchmark. This chart installs a few default `ClusterScanBenchmark` custom resources.
* `ClusterScanProfile` - A `ClusterScanProfile` is a CRD that is used to define the configuration for the CIS scan, which is the benchmark versions to use and any specific tests to skip in that benchmark. This chart installs a few default `ClusterScanProfile` custom resources with no skipped tests, which can immediately be used to launch CIS scans. 
* `ClusterScan` - A `ClusterScan` is a CRD that is used to define when to trigger CIS scans on the cluster based on the defined profile. A report is created after the scan is completed.
* `ClusterScanReport` - A `ClusterScanReport` is a CRD that is used to display results of a CIS scan of the cluster.

prachidamle:
@deniseschannon yes this makes sense, I have added the changes

deniseschannon:
Closing in favor of https://github.com/rancher/charts/pull/827/files

PR #826, opened 2020-11-02
Adapt overridden kiali webroot to correct kiali service name

bashofmann:
The kiali service is not called `rancher-istio-kiali` anymore, but `kiali`. The overwritten `WEB_ROOT` url in https://github.com/rancher/charts/blob/dev-v2.5-source/packages/rancher-kiali-server/rancher-kiali-server.patch#L125 still references the old(?) service name. Also from my testing a `/kiali` suffix seems to be necessary now.

Because of the wrong `WEB_ROOT`, when you refresh the kiali page, you get a 404.

Addresses https://github.com/rancher/rancher/issues/29898.

### DIFF
--- a/packages/rancher-kiali-server/package.yaml
+++ b/packages/rancher-kiali-server/package.yaml
@@ -1,5 +1,5 @@
 url: https://kiali.org/helm-charts/kiali-server-1.24.0.tgz
-packageVersion: 01
+packageVersion: 02
 generateCRDChart:
   enabled: true
 
--- a/packages/rancher-istio/charts/Chart.yaml
+++ b/packages/rancher-istio/charts/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.7.3
 description: A basic Istio setup that installs with the istioctl. Refer to https://istio.io/latest/ for details.
 name: rancher-istio
-version: 1.7.300
+version: 1.7.301
 icon: https://charts.rancher.io/assets/logos/istio.svg
 keywords:
 - networking
@@ -13,4 +13,4 @@ annotations:
   catalog.cattle.io/release-name: rancher-istio
   catalog.cattle.io/ui-component: istio
   catalog.cattle.io/provides-gvr: networking.istio.io.virtualservice/v1beta1
-  catalog.cattle.io/auto-install: rancher-kiali-server-crd=1.24.001
+  catalog.cattle.io/auto-install: rancher-kiali-server-crd=1.24.002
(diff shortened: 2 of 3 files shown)


bashofmann:
I'm not sure if this is enough, or if I also need to bump some package or chart version, or if I need to merge this somewhere else.

brendarearden (approved):
LGTM

packages/rancher-kiali-server/package.yaml
```diff
@@ -1,5 +1,5 @@
 url: https://kiali.org/helm-charts/kiali-server-1.24.0.tgz
-packageVersion: 01
+packageVersion: 02
```
aiyengar2, line 2: @brendarearden I don't think it makes a difference w.r.t. the packaging of the chart since the `requirements.yaml` uses a local chart archive, but should we be updating https://github.com/rancher/charts/blob/70bcfcbee35caff374021bcc1fd0cb721f766640/packages/rancher-istio/charts/requirements.yaml#L6 with the correct `rancher-kiali-server` version (including the packageVersion)?

aiyengar2: Note for future: This is a weird corner case of our charts scripts that needs to be addressed in https://github.com/rancher/rancher/issues/28730.

aiyengar2 (approved):
LGTM, let's wait on merging based on the response to my comment below. We may need to make a small change to bump the `rancher-kiali-server` version as well, just for documentation purposes.

packages/rancher-kiali-server/package.yaml
brendarearden: I thought that the base version stayed the same, and we were appending the extra `000` or `001` on the end - is that not what the package version is for? Since we are just adjusting what we patch over the top of the 1.24.0 version, I think that requirements version should stay the same, but https://github.com/rancher/charts/blob/948c7b8f9a7a847eca3d680271ee0c2da11ddde5/packages/rancher-istio/charts/Chart.yaml#L16 would update... at least this was my understanding

aiyengar2: Yeah I think that makes sense. That's how we've done it so far so probably that shouldn't change.

For this PR, I don't think we need to necessarily make that change (or rather shouldn't, to keep it consistent with the prior release) but it's definitely weird that our Helm packages point to a non-existent chart in the `requirements.yaml` when it should also be pointing to the actual version that the annotation points to. 

e.g. the file we point to in the `requirements.yaml` doesn't even exist in the `main` branch 

https://github.com/rancher/charts/blob/828fdd507ae0d5be0b6f017f39aff6742356c362/charts/rancher-istio/requirements.yaml#L6-L7

Something to fix in the charts script rewrite, but not something that needs to be addressed here 😅

aiyengar2:
@bashofmann once you resolve conflicts, we can add the committer review and hopefully get this merged in!

packages/rancher-kiali-server/package.yaml
brendarearden: yeah that is very weird - I see what you are saying now.

brendarearden:
these changes were merged here https://github.com/rancher/charts/pull/848

PR #828, opened 2020-11-03
Instana agent chart v1.1.4

mmanciop:


mmanciop:
We recently introduced the usage of `lookup`, which breaks Helm 2. We'll consider whether to revert this addition or drop support for Helm 2.

PR #831, opened 2020-11-04
Adds display-name annotation

cbron:
https://github.com/rancher/rancher/issues/29934

### DIFF
--- a/packages/rancher-backup/charts/Chart.yaml
+++ b/packages/rancher-backup/charts/Chart.yaml
@@ -14,3 +14,4 @@ annotations:
   catalog.cattle.io/ui-component: rancher-backup
   catalog.cattle.io/provides-gvr: resources.cattle.io.resourceset/v1
   catalog.cattle.io/scope: management
+  catalog.cattle.io/display-name: "Rancher Backups"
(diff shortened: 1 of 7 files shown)


deniseschannon (approved):

vincent99:
The annotations look good, but are the patch directories supposed to be changing in some of them?

packages/rancher-backup/charts/Chart.yaml
```diff
@@ -14,3 +14,4 @@ annotations:
   catalog.cattle.io/ui-component: rancher-backup
   catalog.cattle.io/provides-gvr: resources.cattle.io.resourceset/v1
   catalog.cattle.io/scope: management
+  catalog.cattle.io/display-name: "Rancher Backups"
```
aiyengar2, line 17: nit: Should this just be `Backups`? Or should `Monitoring` be `Rancher Monitoring` as well?

cbron: We decided to just remove all the "Rancher" parts. Backup I will ask.

cbron: We'll do "backups"

PR #841, opened 2020-11-09
[DNM - 2.5.4] Split up CRDs based on ConfigMap 1mb limit for assumeOwnershipOfCRDs

aiyengar2:
This PR changes the chart template for `assumeOwnershipOfCRDs` to create a new ConfigMap if a single ConfigMap has over 1mb in it. It only affects charts that use `assumeOwnershipOfCRDs`.

The previous behavior was to use one ConfigMap for all CRDs, which does not work for Rancher Monitoring since it has exceeded that (see issue).

Related Issue: https://github.com/rancher/rancher/issues/29973

### DIFF
--- a/scripts/chart-templates/crd-assume-ownership/templates/_helpers.tpl
+++ b/scripts/chart-templates/crd-assume-ownership/templates/_helpers.tpl
@@ -5,3 +5,47 @@
 {{- printf "%s/" .Values.global.cattle.systemDefaultRegistry -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Creates a list of ConfigMaps that contain CRD files with the following format.
+It ensures that no ConfigMap is >1mb in size.
+name: $name-manifest-xx
+crds:
+  {{ path }}: {{ File }}
+size: {{ size in characters }}
+*/}}
+{{- define "crd_manifest_configmaps" -}}
+{{- \$currentScope := . -}}
+{{- \$counter := dict "i" 0 -}}
+{{/* Initialize the first configMap and add it to the list of ConfigMaps necessary */}}
+{{- \$currName := (printf "%s-manifest-%02d" \$currentScope.Chart.Name (get \$counter "i")) -}}
+{{- \$currConfigMap := dict "name" \$currName "size" 0 "crds" (dict) -}}
+{{- \$configMaps := dict \$currName \$currConfigMap -}}
+{{/* Iterate through the CRDs to add to the configMap or create a new one */}}
+{{- range \$filepath, \$_ :=  (.Files.Glob "crd-manifest/**.yaml") -}}
+{{- with \$currentScope -}}
+{{/* Get current values from the dictionary */}}
+{{- \$currConfigMap := get \$configMaps \$currName -}}
+{{- \$currSize := get \$currConfigMap "size" -}}
+{{- \$currCRDs := get \$currConfigMap "crds" -}}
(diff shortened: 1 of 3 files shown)


scripts/chart-templates/crd-assume-ownership/templates/_helpers.tpl
```diff
@@ -5,3 +5,47 @@
...
+{{/* Get current values from the dictionary */}}
+{{- \$currConfigMap := get \$configMaps \$currName -}}
+{{- \$currSize := get \$currConfigMap "size" -}}
+{{- \$currCRDs := get \$currConfigMap "crds" -}}
+{{/* Get the next CRD file that needs to be added to a ConfigMap */}}
+{{- \$path := base \$filepath -}}
+{{- \$file := .Files.Get \$filepath -}}
+{{/* Check if the file size of the CRD causes size to exceed 1048576 characters, which requires a new ConfigMap */}}
+{{- \$currSize := (add \$currSize (len \$file)) -}}
+{{- if ge \$currSize 1000000 -}}
```
aiyengar2, line 36: This new Helm template function is the crux of how we choose to split the files located in `crd-manifest/` (each file usually represents a CRD) into separate ConfigMaps.

Most of the code here is just verbose due to Helm's constraints on defining templates + the fact that we need to escape `$` characters used by Helm, but I'm hoping the Helm comments might make this a little easier to parse.

In general, this code seeks to:
- Initialize 1 ConfigMap that will be used by CRDs
- Iterate through each CRD to collect information on it (path, file contents, size)
- If there's space in the existing ConfigMap (e.g. file size + current size doesn't exceed `1,000,000`, which is a little less than 1mb), add the CRD to this ConfigMap and continue.
- Otherwise, add an entry for a new ConfigMap and add the CRD to it. Continue to iterate through CRDs with the new ConfigMap until that also fills up.

aiyengar2 (commented):
I'm not a huge fan of the approach of this PR since it's tough to read, but I wasn't able to figure out a better solution for it. 

Here are some things I tried and why I decided against it:
- `Create 1 ConfigMap per CRD + 1 Job per CRD`: too bulky. The helm upgrade takes a crazy long amount of time to just do the CRD chart part since it's processing all the Jobs in sequence.
- `Create 1 ConfigMap per CRD + 1 container within the single Job per CRD`: Seems like the process of loading 8 ConfigMaps as volumes onto the Job was the botteneck

The current approach is `Create 1 ConfigMap only when we reach the 1mb limit + 1 container per ConfigMap within the single Job`. It's still somewhat slower than the current install from what I can tell (probably because it needs to mount another volume / run 2 containers instead of 1?) but it's better than the alternatives I tried.

Would love any comments on how to make this a little bit more readable (or if there's a better approach to circumvent the 1mb limit)!

#### Note To Reviewers: `\$` in the `scripts/chart-templates` directory

The files in `scripts/chart-templates` represent a template that we run a `sed` operation on in order to place in certain package specific variables such as the parent chart name or version. These are encoded in the scripts at the moment.

For example, `name: $name-manifest-xx` would actually result in `name: rancher-monitoring-manifest-xx` in the actual CRD chart.

So when we need to use Helm variables (which idiomatically start with `$`), it needs to be escaped so that it doesn't show up as blank.

aiyengar2:
Dropped reviewers from the PR since it's being pushed to the backlog. Will address this when https://github.com/rancher/rancher/issues/29871 is prioritized.

PR #842, opened 2020-11-11
Add cleanup steps to make charts

aiyengar2:
Related Issue: https://github.com/rancher/rancher/issues/29064

### DIFF
--- a/scripts/generate-charts
+++ b/scripts/generate-charts
@@ -31,6 +31,14 @@ for f in packages/*; do
 			cp -R ${f}/charts-crd/* charts/$(basename -- ${f})/charts-crd
 		fi
 		helm repo index --merge ./assets/index.yaml --url assets assets
+
+		# Clean up changes
+		if [[ -f ${f}/package.yaml ]]; then
+			yq w -i ${f}/charts/Chart.yaml 'version' "${version}"
+		fi
+		for file in $(find ./${f}/overlay -type f); do
+			rm ${f}/charts/${file#./${f}/overlay/}
+		done
 	fi
 done
 


aiyengar2:
@cbron @StrongMonkey once this PR has been approved, I'll cut identical PRs to copy over our current scripts to:
- [`rancher/partner-charts`](https://github.com/rancher/partner-charts) @ `main-source`
- [`rancher/rke2-charts`](https://github.com/rancher/rke2-charts) @ `main-source`

Are there any other repo/branch pairs that I need to port this PR over to?

scripts/generate-charts
```diff
@@ -31,6 +31,14 @@ for f in packages/*; do
 			cp -R ${f}/charts-crd/* charts/$(basename -- ${f})/charts-crd
 		fi
 		helm repo index --merge ./assets/index.yaml --url assets assets
+		
+		# Clean up changes
+		if [[ -f ${f}/package.yaml ]]; then
+			yq w -i ${f}/charts/Chart.yaml 'version' "${version}"
```
cbron, line 37: Whats this doing ?

scripts/generate-charts
```diff
@@ -31,6 +31,14 @@ for f in packages/*; do
...
+		fi
+		for file in $(find ./${f}/overlay -type f); do
+			rm ${f}/charts/${file#./${f}/overlay/}
+		done
```
cbron, line 41: This is already cleaned up with make clean isn't it ? Are these just left behind on patch ?

scripts/generate-charts
aiyengar2: Resetting the version number on the chart by removing the `packageVersion`, e.g. the inverse of https://github.com/rancher/charts/blob/a5baa341654df0e333979f850c2277a95fb8b469/scripts/generate-charts#L11-L15

scripts/generate-charts
aiyengar2: This is the inverse of https://github.com/rancher/charts/blob/a5baa341654df0e333979f850c2277a95fb8b469/scripts/generate-charts#L7-L9

While it is cleaned up on `make clean`, [the original issue](https://github.com/rancher/rancher/issues/29064) was about making sure that a user who runs `make charts` followed by `make patch` doesn't end up leaving behind these overlaid resources (e.g. `make charts` should leave the contents of `packages/{chart}/*` identical to what it was before).

aiyengar2: It's a pretty common use case with the following developer workflow:
- Make changes to the chart and run `make patch`
- Run `make charts` (which also runs `make prepare`) and push your changes to your branch
- Add your branch to Rancher Dashboard and try installing your chart. See more changes that need to be made

At this step, currently you would have to run `make prepare` once more before trying to add the new changes and running `make patch`. With this fix, you could omit the extra `make prepare` step.

cbron: >  With this fix, you could omit the extra make prepare step.

I don't understand that part, prepare doesn't do anything with overlays ?

aiyengar2: Yeah, it’s one of the things I wanted to fix during the rewrite. Adding support to using overlays on prepare would also require changes to the patch logic, but that’s another option we could take here.

cbron: Ok, I guess this is fine, but we should limit changes to bash stuff.

cbron (approved):
Lets wait to merge for @StrongMonkey approval

StrongMonkey (dismissed):

scripts/generate-charts
```diff
@@ -31,6 +31,14 @@ for f in packages/*; do
 			cp -R ${f}/charts-crd/* charts/$(basename -- ${f})/charts-crd
 		fi
 		helm repo index --merge ./assets/index.yaml --url assets assets
+
```
aiyengar2, line 34: nit: Remove whitespace
```suggestion

```

aiyengar2:
@cbron @StrongMonkey I just removed some extra whitespace that was there are re0pushed. Since both of you have already approved (got dismissed as stale reviews), can I get one more approval to merge it?

StrongMonkey (dismissed):

cbron (approved):

PR #843, opened 2020-11-11
Change default webroot helper

brendarearden:
**Problem** 
The server.web_root has a helper method that is defaulting to `/kiali` so on refresh, the window.web_root is expected to have `/kiali` appended, but it doesn't so an error page occurs

**Solution**
Changed the helper method to be `/` by default if web_root is not set
Update window web_root to match the kiali server name

**2nd Solution**
https://github.com/rancher/charts/pull/826/files

**3rd Solution that was not implemented**
Set the nameOverride to `nameOverride: "rancher-istio-kiali"`
Set the fullNameOverride to `fullnameOverride: "rancher-istio-kiali"`
Changed the helper method to be `/` by default if web_root is not set
Do not change the window web_root
* UI change needed so that the kiali url dashboard will update based on if an older version name kiali or newer version named rancher-istio-kiali is deployed. 

**Issue**
https://github.com/rancher/rancher/issues/29898

### DIFF
--- a/packages/rancher-istio/charts/Chart.yaml
+++ b/packages/rancher-istio/charts/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.7.3
 description: A basic Istio setup that installs with the istioctl. Refer to https://istio.io/latest/ for details.
 name: rancher-istio
-version: 1.7.300
+version: 1.7.301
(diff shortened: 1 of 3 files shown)


brendarearden:
closing in favor of https://github.com/rancher/charts/pull/826/files

PR #849, opened 2020-11-17
Move storageos-operator to charts/

darkowlzz:
Moves storageos-operator from proposed to charts dir.
Refer: #811


PennyScissors:
Hello @darkowlzz, this PR looks good, but before I can approve and merge I need confirmation from you that this change won't break an existing app deployed for the end user.

darkowlzz:
@PennyScissors since the chart is the same, it's just moved from one directory to another, I don't think it'll break anything. The chart name is the same. I don't know enough about the internals of rancher charts repository to confidently confirm but I can only confirm with my limited understanding of it.

PennyScissors (approved):

PR #850, opened 2020-11-17
Add in jaeger tracing all-in-one

brendarearden:
**Problem**
Istio 1.7.x needs Jaeger support

**Solution**
No upstream helm chart of jaeger all-in-one exists, so added jaeger all in one templates to istio chart

This is not a solution that is for heavy production needs and is the same Jaeger installation type that was provided as part Istio V1. This is documented in the app-readme. 

**Issues**
https://github.com/rancher/rancher/issues/29903

**Dependent PRs**
https://github.com/rancher/image-mirror/pull/52

Related Issues:
https://github.com/rancher/dashboard/issues/1928
https://github.com/rancher/dashboard/issues/1863

Additional Information: 
Here are the additional Jaeger upstream charts https://github.com/jaegertracing/helm-charts/tree/master/charts
Using Jaeger with Istio: https://istio.io/latest/docs/tasks/observability/distributed-tracing/jaeger/

### DIFF
--- a/packages/rancher-istio/charts/values.yaml
+++ b/packages/rancher-istio/charts/values.yaml
@@ -93,4 +93,10 @@ kiali:
       custom_metrics_url: "http://rancher-monitoring-prometheus.cattle-monitoring-system.svc:9090"
       url: "http://rancher-monitoring-prometheus.cattle-monitoring-system.svc:9090"
     tracing:
-      enabled: false
+      in_cluster_url: "http://tracing.istio-system.svc:16686"
+
+tracing:
+  enabled: false
+  jaeger:
+    repository: rancher/jaegertracing-all-in-one
+    tag: 1.20.0
--- a/packages/rancher-tracing/charts/templates/_helpers.tpl
+++ b/packages/rancher-tracing/charts/templates/_helpers.tpl
@@ -1,0 +1,32 @@
+{{- define "system_default_registry" -}}
+{{- if .Values.global.cattle.systemDefaultRegistry -}}
+{{- printf "%s/" .Values.global.cattle.systemDefaultRegistry -}}
+{{- else -}}
+{{- "" -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "tracing.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
+*/}}
+{{- define "tracing.fullname" -}}
+{{- if .Values.fullnameOverride -}}
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
+{{- $name := default .Chart.Name .Values.nameOverride -}}
+{{- if contains $name .Release.Name -}}
+{{- .Release.Name | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
--- a/packages/rancher-tracing/charts/templates/deployment.yaml
+++ b/packages/rancher-tracing/charts/templates/deployment.yaml
@@ -1,0 +1,80 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "tracing.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ .Values.provider }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+spec:
+  selector:
+    matchLabels:
+      app: {{ .Values.provider }}
+  template:
+    metadata:
+      labels:
+        app: {{ .Values.provider }}
+        heritage: {{ .Release.Service }}
+        release: {{ .Release.Name }}
+      annotations:
+        sidecar.istio.io/inject: "false"
+        prometheus.io/scrape: "true"
+        prometheus.io/port: "14269"
+{{- if .Values.jaeger.podAnnotations }}
+{{ toYaml .Values.jaeger.podAnnotations | indent 8 }}
+{{- end }}
+    spec:
+      containers:
+        - name: jaeger
+          image: "{{ template "system_default_registry" . }}{{ .Values.jaeger.repository }}:{{ .Values.jaeger.tag }}"
+          imagePullPolicy: {{ .Values.global.imagePullPolicy }}
+          env:
+          {{- if eq .Values.jaeger.spanStorageType "badger" }}
+          - name: BADGER_EPHEMERAL
+            value: "false"
+          - name: SPAN_STORAGE_TYPE
+            value: "badger"
+          - name: BADGER_DIRECTORY_VALUE
+            value: "/badger/data"
+          - name: BADGER_DIRECTORY_KEY
+            value: "/badger/key"
+          {{- end }}
+          - name: COLLECTOR_ZIPKIN_HTTP_PORT
+            value: "9411"
+          - name: MEMORY_MAX_TRACES
+            value: "{{ .Values.jaeger.memory.max_traces }}"
+          - name: QUERY_BASE_PATH
+            value: {{ if .Values.contextPath }} {{ .Values.contextPath }} {{ else }} /{{ .Values.provider }} {{ end }}
+          livenessProbe:
+            httpGet:
+              path: /
+              port: 14269
+          readinessProbe:
+            httpGet:
+              path: /
+              port: 14269
+{{- if eq .Values.jaeger.spanStorageType "badger" }}
+          volumeMounts:
+            - name: data
+              mountPath: /badger
+{{- end }}
+          resources:
+{{- if .Values.jaeger.resources }}
+{{ toYaml .Values.jaeger.resources | indent 12 }}
+{{- else }}
+{{ toYaml .Values.global.defaultResources | indent 12 }}
+{{- end }}
+      affinity:
+      {{- include "nodeAffinity" . | indent 6 }}
+      {{- include "podAntiAffinity" . | indent 6 }}
+{{- if eq .Values.jaeger.spanStorageType "badger" }}
+      volumes:
+      - name: data
+{{- if .Values.jaeger.persistentVolumeClaim.enabled }}
+        persistentVolumeClaim:
+          claimName: istio-jaeger-pvc
+{{- else }}
+        emptyDir: {}
+{{- end }}
+{{- end }}
--- a/packages/rancher-tracing/charts/templates/service.yaml
+++ b/packages/rancher-tracing/charts/templates/service.yaml
@@ -1,0 +1,63 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: tracing
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    {{- range $key, $val := .Values.service.annotations }}
+    {{ $key }}: {{ $val | quote }}
+    {{- end }}
+  labels:
+    app: {{ .Values.provider }}
+    heritage: {{ .Release.Service }}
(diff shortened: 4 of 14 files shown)


brendarearden:
add docs issue https://github.com/rancher/docs/issues/2860 to ensure the production qualification issue is known to all users

nickgerace (dismissed):

packages/rancher-istio/charts/templates/_helpers.tpl
```diff
@@ -10,3 +10,23 @@
...
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
+*/}}
+{{- define "tracing.fullname" -}}
+{{- if .Values.tracing.fullnameOverride -}}
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
+{{- $name := default "rancher-tracing" .Values.nameOverride -}}
```
aiyengar2, line 30: Hmm since this is all part of the `rancher-istio` chart, should there be a separate `rancher-tracing` naming convention? Seems like it should be packaged as a separate chart (similar to rancher-pushprox) if we're going with that convention.

aiyengar2: were there any other changes you made to stuff that you pulled from upstream we need to take a closer look at?

packages/rancher-istio/charts/templates/deployment-tracing.yaml
```diff
@@ -0,0 +1,91 @@
...
+      - name: data
+{{- if .Values.tracing.jaeger.persist }}
+        persistentVolumeClaim:
+          claimName: istio-jaeger-pvc
+{{- else }}
+        emptyDir: {}
+{{- end }}
+{{- end }}
+{{- end }}
+{{- end }}
```
aiyengar2, line 91:
```suggestion
{{- end }}

```

packages/rancher-istio/charts/templates/service-tracing.yaml
```diff
@@ -0,0 +1,69 @@
...
+    port: 14268
+    targetPort: 14268
+    protocol: TCP
+  - name: jaeger-collector-grpc
+    port: 14250
+    targetPort: 14250
+    protocol: TCP
+  selector:
+    app: jaeger
+{{- end }}
```
aiyengar2, line 69:
```suggestion
{{- end }}

```

packages/rancher-istio/charts/templates/pvc-tracing.yaml
```diff
@@ -0,0 +1,20 @@
...
+spec:
+  storageClassName: {{ .Values.tracing.jaeger.storageClassName }}
+  accessModes:
+    - {{ .Values.tracing.jaeger.accessMode }}
+  resources:
+    requests:
+      storage: 5Gi
+{{- end }}
+{{- end }}
+{{- end }}
```
aiyengar2, line 20:
```suggestion
{{- end }}

```

packages/rancher-istio/charts/templates/_affinity.tpl
```diff
@@ -0,0 +1,93 @@
...
+            values:
+            {{- $vals := split "," $item.values }}
+            {{- range $i, $v := $vals }}
+            - {{ $v | quote }}
+            {{- end }}
+            {{- end }}
+        topologyKey: {{ $item.topologyKey }}
+      weight: 100
+    {{- end }}
+{{- end }}
```
aiyengar2, line 93:
```suggestion
{{- end }}

```

aiyengar2 (changes requested):
Here are some followup questions to make sure we don't break hardened clusters with these changes (check the box if you've confirmed it):

- [x] Were any new images introduced (requires the system default registry to be passed in and a new repo to be created)?
- [x] Were any new namespaces added (requires changes for hardened clusters)?
- [x] Were any CRDs added (requires additional testing + modification of ClusterRoles / Roles introduced by Rancher)
- [x] Were any new resources that create Pods, like Jobs or Deployments, introduced (requires PSPs to be added for hardened cluster support)?

packages/rancher-istio/charts/templates/_helpers.tpl
brendarearden: Yeah, I was kind of going back and forth on this, so definitely open to changes.

 We are not pulling from an upstream chart, so I would be making a chart similar to how I made the istio chart. For this reason it made sense to just bundle it in with istio as just templates or as a chart dependency. Since its only a few templates, I opted to not make it a nested chart inside the istio chart. I could see why it might be better to do this so that is does allow for that separate naming conventions. Do you think this would be a better approach, or do you feel like we should still go in the direction of rancher-pushprox. Either one would be simple to change to.

Essentially the changes I added are a mixture of our Istio V1 jaeger integration https://github.com/rancher/system-charts/tree/dev-v2.5/charts/rancher-istio/1.5.900/charts/tracing, and the Istio quick start example https://istio.io/latest/docs/ops/integrations/jaeger/#option-1-quick-start and there wasn't any significant deviation from these

aiyengar2: Discussed this offline, I think it makes sense for `rancher-tracing` to be a separate chart because:
1. it can / should be deployable without requiring rancher-istio
2. rancher-tracing is not production-grade, but rancher-istio is
3. we want to use two separate naming schemes, so a separate chart makes sense instead of one chart creating multiple resources with different naming schemes (e.g. `rancher-istio-*` v.s. `rancher-tracing-*`)

Since they are mostly independent, we should go with the new-chart-pulled-in-as-a-subchart approach rather than adding it into our existing chart.

cbron: We can separate it out, but it will still live as a sub-chart of Istio. For now it will be hidden so it can't be installed separately, and since its a non-production service it probably shouldn't be installable separately.

packages/rancher-istio/charts/templates/deployment-tracing.yaml
```diff
@@ -0,0 +1,91 @@
...
+{{ toYaml .Values.tracing.jaeger.resources | indent 12 }}
+{{- else }}
+{{ toYaml .Values.global.defaultResources | indent 12 }}
+{{- end }}
+      affinity:
+      {{- include "nodeaffinity" . | indent 6 }}
+      {{- include "podAntiAffinity" . | indent 6 }}
+{{- if .Values.tracing.tolerations }}
+      tolerations:
+{{ toYaml .Values.tracing.tolerations | indent 6 }}
```
cbron, line 75: We don't have windows tolerations yet, so this can be added then.

packages/rancher-istio/charts/values.yaml
```diff
@@ -74,6 +77,15 @@ global:
...
     enabled: true
+  defaultTolerations: []
+  defaultResources: {}
+  imagePullPolicy: IfNotPresent
+  imagePullSecrets: []
+  arch:
+    amd64: 2
+    s390x: 2
+    ppc64le: 2
+  defaultNodeSelector: {}
```
cbron, line 88: If we are splitting chart out, all of this should be done in sub-chart, and then edited below under the tracing node. Under that node, we should only be setting things we are explicitly overriding.

packages/rancher-istio/charts/templates/deployment-tracing.yaml
```diff
@@ -0,0 +1,91 @@
+{{- if .Values.tracing.enabled }}
+{{- if eq .Values.tracing.provider "jaeger" }}
+apiVersion: apps/v1
+kind: Deployment
```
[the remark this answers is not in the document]
brendarearden: Istio is not set up for hardened clusters yet, so I will add a note to our open issue for it to ensure jaeger also works 
https://github.com/rancher/rancher/issues/28622

packages/rancher-istio/charts/values.yaml
```diff
@@ -93,4 +93,10 @@ kiali:
       custom_metrics_url: "http://rancher-monitoring-prometheus.cattle-monitoring-system.svc:9090"
       url: "http://rancher-monitoring-prometheus.cattle-monitoring-system.svc:9090"
     tracing:
-      enabled: false
+      in_cluster_url: "http://tracing.istio-system.svc:16686"
+
+tracing:
+  enabled: true
```
cbron, line 99: If one doesn't exist we need to make a UI issue to allow configuring this on/off.

packages/rancher-tracing/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,91 @@
+{{- if .Values.enabled }}
+{{- if eq .Values.provider "jaeger" }}
```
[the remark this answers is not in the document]
brendarearden: Jaeger, Zipkin and Lightstep are all potential tracing providers

https://istio.io/latest/docs/tasks/observability/distributed-tracing/overview/

packages/rancher-tracing/charts/templates/service.yaml
```diff
@@ -0,0 +1,69 @@
...
+      protocol: TCP
+      {{ if eq .Values.provider "jaeger" }}
+      targetPort: 16686
+      {{ else }}
+      targetPort: {{ .Values.zipkin.queryPort }}
+      {{ end}}
+  selector:
+    app: {{ .Values.provider }}
+---
+# Jaeger implements the Zipkin API. To support swapping out the tracing backend, we use a Service named Zipkin.
```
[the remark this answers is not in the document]
brendarearden: These are the three services that Istio has for their quickstart install of Jaeger https://raw.githubusercontent.com/istio/istio/release-1.8/samples/addons/jaeger.yaml. 

From what I could gather, you need a ui, an api for queries to whats been collected, and a collector to gather the traces. The jaeger service is the ui, the zipkin service is the api and the collector is the jaeger collector. You can also add in additional services or database.

packages/rancher-tracing/charts/templates/deployment.yaml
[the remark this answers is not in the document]
brendarearden: but to your point, there is only one option currently in our chart, so I can remove this check and add it back in if / when we add support for the other options.

packages/rancher-tracing/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,82 @@
+{{- if .Values.enabled }}
```
cbron, line 1: Whats the point of values.enabled ? If you disable the sub-chart the whole thing should get skipped.

packages/rancher-tracing/charts/templates/_helpers.tpl
```diff
@@ -0,0 +1,39 @@
...
+{{- else -}}
+{{- "" -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "tracing.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
```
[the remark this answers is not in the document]
cbron: ^

packages/rancher-tracing/charts/templates/deployment.yaml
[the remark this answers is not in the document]
cbron: yea lets keep simple for now

packages/rancher-istio/charts/templates/pvc-tracing.yaml
```diff
@@ -0,0 +1,20 @@
...
+  namespace: {{ template "istio.namespace" . }}
+  labels:
+    app: jaeger
+spec:
+  storageClassName: {{ .Values.tracing.jaeger.storageClassName }}
+  accessModes:
+    - {{ .Values.tracing.jaeger.accessMode }}
+  resources:
+    requests:
+      storage: 5Gi
```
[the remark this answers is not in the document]
cbron: This doesn't _really_ matter, as the request is just to make sure you get a volume thats big enough for you. You can't lower this to 4gi to free up an extra Gi.

packages/rancher-istio/charts/values.yaml
cbron: we discussed this going to false by default now

cbron (approved):

nickgerace (approved):

cbron:
@aiyengar2 merging now, lmk if you have additional concerns.

PR #825, opened 2020-11-02
stable(openebs): update v2.2.0 release version 

prateekpandey14:
- Updated README
- Updated questions.yaml to point the NDM images instead of defaults to `quay.io`
- Updated values.yaml

PR #854, opened 2020-11-19
stable(openebs): update v2.2.0 release version 

prateekpandey14:
Update to v2.2.0 release, 
- Updated README
- Updated values.yaml
- Updated questions.yaml

Update to v2.1.0  and v2.0.0 release, 
- Updated questions.yaml to point the NDM images instead of defaults to quay.io
- Updated values.yaml

### DIFF
--- a/charts/openebs/v2.2.0/Chart.yaml
+++ b/charts/openebs/v2.2.0/Chart.yaml
@@ -1,0 +1,19 @@
+apiVersion: v1
+version: 2.1.0
+name: openebs
+appVersion: 2.1.0
+description: Containerized Storage for Containers
+icon: file://../openebs-icon-color.png
+home: http://www.openebs.io/
+keywords:
+  - cloud-native-storage
+  - block-storage
+  - iSCSI
+  - storage
+sources:
+  - https://github.com/openebs/openebs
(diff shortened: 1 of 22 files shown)


PennyScissors:
Just reviewed this, and have a few things to bring up:

1. Does this PR replaces PR https://github.com/rancher/charts/pull/825 and is it okay to close it?

2. All changes in version v2.2.0 are good to merge (including changes carried over from the v2.0.1 and v2.1.1 updates). However, old charts that are already live shouldn't be modified, and your first 2 commits will modify charts v2.0.0 and v2.1.0 in place and that could cause problems for users using any of those versions. If you need those charts updated too, please create separate PRs and follow the same steps of one commit to copy the previous version and another to update it. Otherwise, it should be okay to skip them if they are not needed in v2.2.0.

prateekpandey14:
ok got it @PennyScissors,  i have removed the changes of the older charts , will raise a different PR for them. Just kept the new version changes (2.2.0) which can we can be merge now, thanks .

> Does this PR replaces PR #825 and is it okay to close it?

Yes, i did closed the PR

PennyScissors (approved):
lgtm!

PR #862, opened 2020-12-01
[2.5.4] Quote insecure skip verify to avoid parse errors

mrajashree:
The s3 configuration details are passed to the backup-restore-operator through a Secret, setting the stringData field. InsecureSkipVerify cannot have boolean value as a part of stringData. So the operator will now parse this field as a string and then convert it to boolean while using it. This commit quotes the value to avoid errors related to stringData given in this bug
https://github.com/rancher/backup-restore-operator/issues/56

This only applies when installing/upgrading the chart.

https://github.com/rancher/charts/issues/864

### DIFF
--- a/packages/rancher-backup/charts/values.yaml
+++ b/packages/rancher-backup/charts/values.yaml
@@ -1,6 +1,6 @@
 image:
   repository: rancher/backup-restore-operator
-  tag: v1.0.2
+  tag: v1.0.3
 
 ## Default s3 bucket for storing all backup files created by the backup-restore-operator
 s3:
--- a/packages/rancher-backup/charts/Chart.yaml
+++ b/packages/rancher-backup/charts/Chart.yaml
@@ -1,11 +1,11 @@
 apiVersion: v1
-appVersion: v1.0.2
+appVersion: v1.0.3
 description: Provides ability to back up and restore the Rancher application running on any Kubernetes cluster
 name: rancher-backup
 keywords:
 - applications
 - infrastructure
-version: 1.0.201
+version: 1.0.300
 icon: https://charts.rancher.io/assets/logos/backup-restore.svg
 annotations:
   catalog.cattle.io/certified: rancher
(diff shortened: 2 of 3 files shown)


aiyengar2 (dismissed):
LGTM

Double checked that image exists in https://hub.docker.com/r/rancher/backup-restore-operator/tags.

cbron (commented):
Has this chart already had a change made to it ? If not it needs a version bump. It also needs an issue made to track new chart versions going into the milestone, a parent issue that tracks all child issues with changes.

packages/rancher-backup/charts/values.yaml
```diff
@@ -1,6 +1,6 @@
 image:
   repository: rancher/backup-restore-operator
-  tag: v1.0.2
+  tag: v1.0.3
```
prachidamle, line 3: We will also need to increment the chart/app version in chart.yaml? So that users who already have installed this can upgrade to the new

mrajashree: done

mrajashree:
@cbron I've created this issue for incrementing the chart version https://github.com/rancher/charts/issues/864

packages/rancher-backup/charts/values.yaml
cbron: This is the image tag, was there a change to the backup restore image ? If thats the case then the chart should probably bump to `1.0.300`. If no changes in image just bump the chart version like you have above.

mrajashree: @cbron hmm I don't follow this part
> This is the image tag, was there a change to the backup restore image ?

yes this is a new image, so I can bump it straight to 1.0.300 is it?

cbron: Yes thats fine. Wanted to make sure you weren't bumping the image just to bump the chart.

cbron (dismissed):

cbron (approved):

PR #863, opened 2020-12-01
CIS report changes and add CIS 1.6

mrajashree:
This PR adds the CRs for CIS 1.6 benchmark and profiles. https://github.com/rancher/rancher/issues/29649
It also adds following changes to the clusterScan CRD:
1. AdditionalPrinterColumn "Warn" to show number of tests with "Warn" output. https://github.com/rancher/security-scan/issues/32
2. Field called `scoreWarning`, with which users can decide whether a scan should pass or fail if there are tests with "warn" output. https://github.com/rancher/cis-operator/issues/54

### DIFF
--- a/packages/rancher-cis-benchmark/charts/crds/clusterscan.yaml
+++ b/packages/rancher-cis-benchmark/charts/crds/clusterscan.yaml
@@ -19,11 +19,17 @@ spec:
   - JSONPath: .status.summary.skip
     name: Skip
     type: string
+  - JSONPath: .status.summary.warn
+    name: Warn
+    type: string
   - JSONPath: .status.summary.notApplicable
     name: Not Applicable
     type: string
   - JSONPath: .status.lastRunTimestamp
     name: LastRunTimestamp
+    type: string
+  - JSONPath: .spec.cronSchedule
+    name: CronSchedule
     type: string
   group: cis.cattle.io
   names:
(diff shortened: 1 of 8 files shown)


packages/rancher-cis-benchmark/charts/values.yaml
```diff
@@ -5,10 +5,10 @@
 image:
   cisoperator:
     repository: rancher/cis-operator
-    tag: v1.0.1
+    tag: v1.0.3-rc1
```
mrajashree, line 8: This will be changed once the final image is out, might even be removed from this PR

packages/rancher-cis-benchmark/charts/crds/clusterscan.yaml
```diff
@@ -19,6 +19,9 @@ spec:
   - JSONPath: .status.summary.skip
     name: Skip
     type: string
+  - JSONPath: .status.summary.warn
```
mrajashree, line 22: clusterScan CRD is also changed in the other PR to include recurring scans. So I'm not sure how to merge the PRs in which order, or will there be one PR to include all these changes?
@prachidamle @cbron

packages/rancher-cis-benchmark/charts/values.yaml
mrajashree: removing the image changes, will be done through @briandowns 's PR

prachidamle (commented):
LGTM - let's merge after Brian's rke2 PR merges

prachidamle (approved):
LGTM

prachidamle (approved):

cbron (approved):

PR #866, opened 2020-12-02
Remove assets to force a rebuild

brendarearden:
Deleted Istio 1.7.302tgz froms assets
Deleted rancher-tracing-1.20.001 tgz from istio chart
Removed Istio 1.7.302 from index


cbron (approved):
Think this is right, but @aiyengar2 can double check

aiyengar2 (changes requested):
@StrongMonkey correct me if I'm wrong but I don't think we need this change.

https://github.com/rancher/charts/pull/867 will update the sha256sum for `rancher-istio`, which will automatically replace the tgz within `rancher-istio` (the index.yaml does not need to be modified though since the chart version is the same).

 `rancher-tracing` was already updated as part of https://github.com/rancher/charts/pull/865/files

StrongMonkey:
Yeah I don't think we need this change, It will override regardless.

brendarearden:
so I will go ahead merge https://github.com/rancher/charts/pull/867 @cbron based on the comment here, I will leave this pr oepn until we confirm the other merge fixed it

cbron:
Ok we can go that route

brendarearden:
confirmed other pr resolved the issue. closing this one.

PR #870, opened 2020-12-07
[2.5] Added rancher-externalip-webhook embargo cve chart

rawmind0:


packages/rancher-externalip-webhook/package.yaml
```diff
@@ -0,0 +1,2 @@
+url: https://github.com/rancher/externalip-webhook/releases/download/v0.0.1-rc2/rancher-externalip-webhook-0.0.1-rc2.tgz
+packageVersion: 00
```
cbron, line 2: * Lets start at 0.1.100. Technically .001 is invalid. 
* The `rancher/externalip-webhook` needs a tag to go with it. 
* The authProxy image needs to be mirrored to Rancher. All repository images must start with `rancher/`. 
* Need to add linux taints/tolerations as outlined in doc

rawmind0: Updated

cbron (approved):

PR #876, opened 2020-12-08
[2.5] Updated rancher-externalip-webhook chart, renamed to vue-ranche…

rawmind0:
…r-externalip-webhook at 2.5


cbron (approved):
Might not fully work until we remove the old ones. We'll see.

PR #886, opened 2020-12-11
Add catalog.cattle.io/os linux annotation to logging

nickgerace:
Relevant issue: https://github.com/rancher/rancher/issues/30450


cbron (approved):

PennyScissors:
Shouldn't the package version be increased with this change?

cbron:
@PennyScissors only if there are no other changes in this release. We should ideally do 1 bump per milestone.

prachidamle (approved):

PR #888, opened 2020-12-11
Add catalog.cattle.io/os linux annotation to rancher-externai-ip-webhook

rawmind0:


packages/rancher-external-ip-webhook/package.yaml
```diff
@@ -1,2 +1,2 @@
-url: https://github.com/rancher/externalip-webhook/releases/download/v0.1.5/rancher-externalip-webhook-0.1.5.tgz
-packageVersion: 00
+url: https://github.com/rancher/externalip-webhook/releases/download/v0.1.6/rancher-externalip-webhook-0.1.6.tgz
+packageVersion: 01
```
cbron, line 2: 00 is fine ?

packages/rancher-external-ip-webhook/rancher-external-ip-webhook.patch
```diff
@@ -1,40 +1,28 @@
...
    catalog.cattle.io/display-name: External IP Webhook
    catalog.cattle.io/namespace: cattle-externalip-system
+   catalog.cattle.io/os: linux
 -  catalog.cattle.io/release-name: rancher-externalip-webhook
 -  catalog.cattle.io/ui-component: rancher-externalip-webhook
 +  catalog.cattle.io/release-name: rancher-external-ip-webhook
 +  catalog.cattle.io/ui-component: rancher-external-ip-webhook
  apiVersion: v1
- appVersion: v0.1.5
+ appVersion: v0.1.6
```
cbron, line 13: I don't see why this is necessary if we are rebasing to 0.1.6  in the package.yaml

rawmind0: We don't set the appVersion??

packages/rancher-external-ip-webhook/package.yaml
rawmind0: updated

cbron (approved):

PR #889, opened 2020-12-11
Add annotation to hide backup chart from windows cluster

mrajashree:
https://github.com/rancher/rancher/issues/30450


cbron (approved):
Chart version already bumped

PR #890, opened 2020-12-12
Add linux annotation to istio packages

cbron:
https://github.com/rancher/rancher/issues/30450


packages/rancher-istio/charts/Chart.yaml
```diff
@@ -15,3 +15,4 @@ annotations:
   catalog.cattle.io/provides-gvr: networking.istio.io.virtualservice/v1beta1
   catalog.cattle.io/auto-install: rancher-kiali-server-crd=1.24.002
```
brendarearden, line 16: since you bumped the package version for kiali, it should change here as well

brendarearden (changes requested):
Looks good except for the kiali package update needs to be carried through to the istio chart.yaml for auto-install annotation.

brendarearden (approved):
lgtm

PR #894, opened 2020-12-14
Adds linux annotation to longhorn chart

cbron:
Needs a package bump because there is no new chart yet in this release: https://github.com/rancher/charts/blob/main-source/packages/longhorn/package.yaml#L5

Issue: rancher/rancher#30450

### DIFF
--- a/packages/longhorn/longhorn.patch
+++ b/packages/longhorn/longhorn.patch
@@ -1,7 +1,7 @@
 diff -x '*.tgz' -x '*.lock' -uNr packages/longhorn/charts-original/Chart.yaml packages/longhorn/charts/Chart.yaml
 --- packages/longhorn/charts-original/Chart.yaml
 +++ packages/longhorn/charts/Chart.yaml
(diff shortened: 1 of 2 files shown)


nickgerace (approved):

packages/longhorn/longhorn.patch
```diff
@@ -49,6 +49,7 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/longhorn/charts-original/Chart.yaml pa
 +  catalog.cattle.io/ui-component: longhorn
 +  catalog.cattle.io/provides-gvr: longhorn.io/v1beta1
```
brendarearden, line 50: is this being used by another chart?

rmweir (approved):
LGTM

packages/longhorn/longhorn.patch
cbron: Not that I'm aware of, we do it preemptively though.

PR #900, opened 2020-12-18
Add "catalog.cattle.io/os: linux" annotation

thedadams:
The annotation was added to all non-crd charts and all sub-charts that are also rancher charts (e.g. `rancher-kiali` chart inside of `rancher-istio`)

Issue:
https://github.com/rancher/rancher/issues/30542


cbron (changes requested):
Needs a re-index of index.yaml / assets/index.yaml. The former is just a copy of the latter.

StrongMonkey:
@cbron So we are only changing for main branch, correct? if qa want to verify this they have to point the branch on rc builds from dev-2.5 to main

StrongMonkey:
Did we test the upgrade scenario? For example I deployed istio 1.7.0 in rancher 2.5.3 then we pushed a new version that has the new label. Does this trigger upgrade/chaos since it is the same version but with different content? Want to make sure either nothing is happening on existing setup or if upgrade is triggered, helm won't proceed with an error.

cbron:
@StrongMonkey I was assuming QA could just point to the fork branch directly. Is there a reason to make sure it gets into dev-v2.5 and 2.5-head ? 
For your scenario, yes we need to test that. I don't think anything will happen on the tar change, but an edit to that existing app should still work.

StrongMonkey:
@cbron I don't think we need to change on dev-2.5 either. Just wondering what's the workflow. I think we can let qa test donnie's fork branch first before merging this.

cbron (commented):
K. I am going to say this is ready for QA then unless anyone disagree's. I confirmed that we don't need to do sub-or-crd charts

aiyengar2:
> Did we test the upgrade scenario? For example I deployed istio 1.7.0 in rancher 2.5.3 then we pushed a new version that has the new label. Does this trigger upgrade/chaos since it is the same version but with different content? Want to make sure either nothing is happening on existing setup or if upgrade is triggered, helm won't proceed with an error.

I’d assume it won’t make a difference since we’re just adding an annotation to the Chart that Rancher 2.5.3 doesn’t check at all.

The upgrade scenario I’m concerned about would be a user upgrading to Rancher 2.5.4 from 2.5.3 without doing a catalog refresh, since that means they won’t get the new annotation until they refresh. We should note somewhere (to the release notes?) that users might need to refresh their catalogs when upgrading to 2.5.4 from 2.5.3.

aiyengar2:
> @cbron I don't think we need to change on dev-2.5 either. Just wondering what's the workflow. I think we can let qa test donnie's fork branch first before merging this.

Agreed. We can merge straight to main; dev-v2.5 doesn’t store any information about previously released packageVersions either way, so this should be fine.

aiyengar2 (changes requested):
LGTM

aiyengar2 (approved):
LGTM

StrongMonkey (approved):

cbron (approved):

PR #901, opened 2020-12-18
Add "catalog.cattle.io/os: linux" annotations

thedadams:
Related PR:
https://github.com/rancher/charts/pull/900

Issue:
https://github.com/rancher/rancher/issues/30542


assets/index.yaml
```diff
@@ -12,9 +12,9 @@ entries:
       catalog.cattle.io/release-name: fleet
```
cbron, line 12: Want to confirm, a bunch of these tars already had the change in them, this would result in no change in the end right ? 
I saw that one of them (kiali-server 1.24.003) had a change in the binary but when I downloaded them, the diff said they were the same, so I'm assuming thats timestamp changes and such.

thedadams: Yes, the script I wrote didn't do anything to charts that already had the annotation. The weirdness with running the `tar` command in a script versus on the command line caused me to just decompress and compress everything.

I can checkout the charts versions that already had annotation, if that matters.

PR #906, opened 2020-12-22
Removing istio versions that are not needed for prod

brendarearden:
Need to remove version 1.7.302 since I am changed the version back to 1.7.301 to make it less confusing for users in production. Removing 1.7.301 to ensure asset is the newest set of changes. 

**Issue**
https://github.com/rancher/rancher/issues/30598


aiyengar2 (approved):
Neither charts have been released yet, so this sounds good to me.

brendarearden:
Regenerated using the following commands: 
```
helm repo index --url assets assets 
cp ./assets/index.yaml .
```

cbron (approved):
As long as we regenerate its good

PR #907, opened 2020-12-22
Change istio version so it is only one version up from current produc…

brendarearden:
**Problem**
Multiple changes were completed for Istio for the 2.5.4 release which caused the Istio version to get bumped twice. This will be confusing for users since 1.7.300 is the current version in production. 

**Solution**
Now that all changes are in, update the istio version to be one up from the current production version

**Issue**
https://github.com/rancher/rancher/issues/30598

**Dependent PR**
https://github.com/rancher/charts/pull/906


cbron (approved):
Think this is right, since Kiali is already upgraded to new version.

aiyengar2 (approved):
LGTM

PR #908, opened 2020-12-22
reset index.yaml

paynejacob:
Rancher builds are currently failing because the index is pointing to a non existent version of longhorn https://drone-pr.rancher.io/rancher/rancher/12838/1/2


aiyengar2 (changes requested):
From the root of the repository, run the following ([copied from `make charts`](https://github.com/rancher/charts/blob/42eece69bd68da9ada2172b42f056cbd8c9a4e86/scripts/generate-charts#L47-L51)):

```bash
helm repo index --merge ./assets/index.yaml --url assets assets
cp ./assets/index.yaml .` 
```

This will generate the new index so we can see the patch of changes instead of relying on a merge from dev-v2.5-source to regenerate the index.

Also @brendarearden could we just fold this in as a commit in https://github.com/rancher/charts/pull/906#event-4142088545 since you're modifying the assets anyways?

brendarearden:
@aiyengar2 Yes I could do that!

brendarearden:
I have regenerated the index.yaml as part of this pr, if you want to close this PR in favor of that one https://github.com/rancher/charts/pull/906#event-4142088545

brendarearden:
I think we can close the pr sine this has been merged https://github.com/rancher/charts/pull/909

cbron:
Yes, I did this manually already to unblock us, and Brenda's will clean it up again, so closing.

PR #912, opened 2020-12-29
Instana agent chart v1.2.1

mmanciop:


charts/instana-agent/v1.2.1/Chart.yaml
```diff
@@ -1,5 +1,5 @@
-apiVersion: v1
-appVersion: 1.1
+apiVersion: v2
```
PennyScissors, line 1: This repository is for Helm 2 charts, so this change will cause the installation to fail in Rancher. Your options would be to either keep the Helm 2 compatibility, or move the chart to [rancher/helm3-charts](https://github.com/rancher/helm3-charts/) and update the `app-readme.me` in the previous version to let users know new versions will be in the Helm 3 catalog. If you decide to move it, all you have to do is move the commit to that repo and submit a PR. Basically the same you did here, but in that other repo.

charts/instana-agent/v1.2.1/questions.yml
```diff
@@ -7,7 +7,7 @@ questions:
 # Basic agent configuration
```
PennyScissors, line 7: It might be good to add a question for `cluster.name` as a required value. Installation will fail if you don't specify it, but you can't set it unless you edit it as a YAML.
```
CreateContainerConfigError: couldn't find key cluster_name in ConfigMap instana-agent/instana-agent
```

PennyScissors (commented):
I created a custom catalog that uses Helm 3 to be able to test this chart before you decide whether you are going to move it or keep Helm 3 compatibility, and I'm running into the following error:
```
CreateContainerError: Error response from daemon: Minimum memory limit allowed is 4MB
```
I'm setting the agent key 
Is there anything in the configuration steps I'm missing?
These are my steps:
1. get agent key from my instana account
2. set agent key to my agent key
3. set zone name to "testzone"
4. set cluster name to "testcluster"
5. change the endpoint to `ingress-orange-saas.instana.io` as that's what my account shows in the agent installation instructions.
6. install chart

PennyScissors:
Hello @mmanciop, just wanted to follow up on the status of this PR as I noticed new commits were pushed, but the apiVersion still remains v2 which will cause the chart to fail installation.

PennyScissors (commented):
Hello @mmanciop, just wanted to follow up on the status of this PR as I noticed new commits were pushed, but it still uses apiVersion: v2 which will cause the chart to fail installation in Rancher.

mmanciop:
Moving to the Helm3 branch

PR #923, opened 2021-01-07
Fix premature end block in default Prometheus CR

aiyengar2:
When `prometheus.enabled=false`, the current chart fails to render the manifest since anything that comes after the premature end block is not excluded. As a result, we get the following error on install when trying to validate the created manifest:

```
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: [apiVersion not set, kind not set]
```

This error also exists and is reproducible in [the upstream chart](https://github.com/prometheus-community/helm-charts/blob/5dc19ca22e6038b98c03e77df4d9b313b642acdb/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml#L264-L265'). Issue filed in upstream here: https://github.com/prometheus-community/helm-charts/issues/550.

Related Issue: https://github.com/rancher/rancher/issues/30712


aiyengar2:
Closing this out in favor of rebased PR

PR #931, opened 2021-01-11
Add PodSecurityPolicies for rancher-logging

bashofmann:
Add PodSecurityPolicies for rancher-logging.

Addresses https://github.com/rancher/rancher/issues/30771

A branch with the compiled charts for testing is available at https://github.com/bashofmann/charts-1/tree/rancher-logging-psp-built3.

### DIFF
--- a/packages/rancher-logging/overlay/templates/loggings/aks/logging.yaml
+++ b/packages/rancher-logging/overlay/templates/loggings/aks/logging.yaml
@@ -15,6 +15,9 @@ spec:
     inputTail:
       Tag: "aks"
       Path: "/var/log/azure/kubelet-status.log"
+    security:
+      podSecurityPolicyCreate: true
+      roleBasedAccessControlCreate: true
   {{- $total_tolerations := concat (.Values.tolerations) (.Values.fluentbit_tolerations) }}
   {{- with $total_tolerations }}
     tolerations: {{- toYaml . | nindent 6 }}
@@ -30,6 +33,9 @@ spec:
       repository: {{ template "system_default_registry" . }}{{ .Values.images.config_reloader.repository }}
       tag: {{ .Values.images.config_reloader.tag }}
     disablePvc: {{ .Values.disablePvc }}
+    security:
+      podSecurityPolicyCreate: true
+      roleBasedAccessControlCreate: true
     {{- with .Values.tolerations }}
     tolerations: {{- toYaml . | nindent 6 }}
     {{- end }}
--- a/packages/rancher-logging/overlay/templates/loggings/rke/daemonset.yaml
+++ b/packages/rancher-logging/overlay/templates/loggings/rke/daemonset.yaml
@@ -40,6 +40,7 @@ spec:
         - name: config
           configMap:
             name: "{{ .Release.Name }}-rke"
+      serviceAccountName: "{{ .Release.Name }}-rke-aggregator"
       {{- $total_tolerations := concat (.Values.tolerations) (.Values.fluentbit_tolerations) }}
       {{- with $total_tolerations }}
       tolerations:
@@ -49,4 +50,65 @@ spec:
       nodeSelector:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: "{{ .Release.Name }}-rke-aggregator"
+  namespace: "{{ .Release.Namespace }}"
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: "{{ .Release.Name }}-rke-aggregator"
+rules:
+  - apiGroups:
+    - policy
+    resourceNames:
+    - "{{ .Release.Name }}-rke-aggregator"
+    resources:
+    - podsecuritypolicies
+    verbs:
+    - use
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: "{{ .Release.Name }}-rke-aggregator"
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: "{{ .Release.Name }}-rke-aggregator"
+subjects:
+  - kind: ServiceAccount
+    name: "{{ .Release.Name }}-rke-aggregator"
+---
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: "{{ .Release.Name }}-rke-aggregator"
+  namespace: "{{ .Release.Namespace }}"
+spec:
+  allowPrivilegeEscalation: false
+  allowedHostPaths:
+  - pathPrefix: /var/lib/docker/containers/
+    readOnly: false
+  - pathPrefix: /var/lib/rancher/rke/log/
+    readOnly: false
+  - pathPrefix: /var/lib/rancher/logging/
+    readOnly: false
+  fsGroup:
+    rule: RunAsAny
+  readOnlyRootFilesystem: true
+  runAsUser:
+    rule: RunAsAny
+  seLinux:
+    rule: RunAsAny
+  supplementalGroups:
+    rule: RunAsAny
+  volumes:
+  - configMap
+  - emptyDir
+  - secret
+  - hostPath
 {{- end }}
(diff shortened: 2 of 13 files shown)


packages/rancher-logging/rancher-logging.patch
```diff
@@ -100,6 +100,10 @@ diff -x '*.tgz' -x '*.lock' -uNr packages/rancher-logging/charts-original/values
  
  affinity: {}
  
+@@ -55,2 +60,2 @@
```
paynejacob, line 103: Do we have a policy for the charts repo on if PSP should be enabled by default? We may also want to consider making this option available on the chart install page.

packages/rancher-logging/overlay/templates/loggings/aks/logging.yaml
```diff
@@ -15,6 +15,9 @@ spec:
     inputTail:
       Tag: "aks"
       Path: "/var/log/azure/kubelet-status.log"
+    security:
+      podSecurityPolicyCreate: true
+      roleBasedAccessControlCreate: true
```
nickgerace, line 20: Is there value in exposing this with values.yaml?

```yaml
{{- with .Values.security }}
security: # insert values here
{{- end }}
```

packages/rancher-logging/rancher-logging.patch
bashofmann: In rancher-monitoring and longhorn we have PSPs enabled as default.

packages/rancher-logging/overlay/templates/loggings/aks/logging.yaml
bashofmann: I can't think of any reason why you would want to deactivate it. If a cluster does not use PodSecurityPolicies, it does not matter if the objects are there or not, they are just ignored.
Theoretically someone could want to use their own PodSecurityPolicy instead of the one provided by the operator, but I can't think of any value that would give you.
But maybe I'm missing something here.

nickgerace: @bashofmann I'm inclined to agree. We can always add it later then.

packages/rancher-logging/overlay/templates/loggings/rke/daemonset.yaml
```diff
@@ -49,4 +50,65 @@ spec:
...
+  name: "{{ .Release.Name }}-rke-aggregator"
+rules:
+  - apiGroups:
+    - policy
+    resourceNames:
+    - "{{ .Release.Name }}-rke-aggregator"
+    resources:
+    - podsecuritypolicies
+    verbs:
+     - use
```
nickgerace, line 72: nit: spacing

packages/rancher-logging/overlay/templates/loggings/rke/daemonset.yaml
```diff
@@ -49,4 +50,65 @@ spec:
...
+  namespace: "{{ .Release.Namespace }}"
+spec:
+  allowPrivilegeEscalation: false
+  allowedHostPaths:
+  - pathPrefix: /var/lib/docker/containers
+    readOnly: false
+  - pathPrefix: /var/lib/rancher/rke/log
+    readOnly: false
+  - pathPrefix: /var/lib/rancher/logging/
+    readOnly: false
```
nickgerace, line 99: nit: We might want all of these `pathPrefix` to end with either `/` or no path separator

packages/rancher-logging/overlay/templates/loggings/rke2/daemonset.yaml
```diff
@@ -31,11 +31,69 @@ spec:
...
+  name: "{{ .Release.Name }}-rke2-journald-aggregator"
+rules:
+  - apiGroups:
+    - policy
+    resourceNames:
+    - "{{ .Release.Name }}-rke2-journald-aggregator"
+    resources:
+    - podsecuritypolicies
+    verbs:
+     - use
```
nickgerace, line 61: nit: spacing, same as above

nickgerace (approved):
LGTM, but I'd like @paynejacob's approval as well :)

paynejacob (approved):

nickgerace (approved):

nickgerace:
@bashofmann: Thanks for the PR, by the way. Once @cbron approves, we should be ready to ship it.

packages/rancher-logging/overlay/templates/loggings/aks/logging.yaml
```diff
@@ -15,6 +15,9 @@ spec:
     inputTail:
```
cbron, line 15: One general downside here is that PSP's are likely going away soon, so this will likely need editing in the future.

packages/rancher-logging/overlay/templates/loggings/rke2/daemonset.yaml
```diff
@@ -31,11 +31,69 @@ spec:
...
+---
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: "{{ .Release.Name }}-rke2-journald-aggregator"
+  namespace: "{{ .Release.Namespace }}"
+spec:
+  allowPrivilegeEscalation: false
+  allowedHostPaths:
+  - pathPrefix: /etc/rancher/logging/logs
```
cbron, line 83: Is this really all thats read ? cc @nickgerace

packages/rancher-logging/overlay/templates/loggings/aks/logging.yaml
cbron: Reading upstream, these 2 just makes a pre-baked fluentD PSP ?
I think only reason for shutting it off would be if it breaks something, but also I could see down the line PSP's coming with a warning deprecation message.

packages/rancher-logging/overlay/templates/loggings/rke2/daemonset.yaml
bashofmann: At least, it's the only volume I could find in the Daemonset: https://github.com/rancher/charts/pull/931/files/e3c75a2e2930663762440b044d83fb965f4729bb#diff-725e07f25e638efd8f1864b278104ab6d92602490dd07e0ebaf3fe4d6381fd3fR35-R38

packages/rancher-logging/overlay/templates/loggings/aks/logging.yaml
bashofmann: True, when we know what the PSP replacement will be, this will need some work again. But I think, as long as we offer PSP's in the UI when creating/configuring a cluster and have it in our hardening guide, our helm charts also should work with it.

nickgerace:
I cannot reply to [this](https://github.com/rancher/charts/pull/931#discussion_r559274680) in the conversation, so I will here...
> Reading upstream, these 2 just makes a pre-baked fluentD PSP ?
> https://github.com/banzaicloud/logging-operator/blob/master/pkg/resources/fluentd/psp.go#L27
>
> I think only reason for shutting it off would be if it breaks something, but also I could see down the line PSP's coming with a warning deprecation message.

I agree that we should reformat this to be part of the Logging CRs wherever possible. Thoughts @bashofmann?

packages/rancher-logging/overlay/templates/loggings/rke2/daemonset.yaml
nickgerace: As far as RKE2 goes, I believe that is correct @cbron. (cc @paynejacob optionally if you have more context)

paynejacob: Yes, this is the link between our fluenbit pods and banzai's pods.  Their configuration is too restrictive to collect all the rke2 logs so we aggregate them and present them in a way banzai can consume in this file.

bashofmann:
@nickgerace I'm not sure, if I understand you correctly. But in the PR I adapted all Logging CRs to set the necessary field. Or do you mean to not use `podSecurityPolicyCreate` from the Logging CR and instead add our own PSP?

nickgerace:
@bashofmann: that's actually what I meant. Perhaps I had a cached page of the PR, because I see that now. Looks good to me!

bashofmann:
Closing in favor of https://github.com/rancher/charts/pull/951

PR #935, opened 2021-01-13
Update README.md to include s3 secret namespace details for backup operator

mrajashree:


aiyengar2 (changes requested):
Can you rebase this PR on the contents of `dev-v2.5-source-alt` or close it out?

#### Context

As part of a migration to new build scripts for our Rancher charts, developers should now be merging changes into `dev-v2.5-source-alt` (a copy of dev-v2.5-source using the newer scripts) instead of `dev-v2.5-source`. The new scripts introduced breaking changes to the charts directory structures, so existing PRs need to be restructured to match that.

Eventually / soon, `<dev-v2.5-source|dev-v2.5|main>` will be forwarded to match the contents of `<dev-v2.5-source|dev-v2.5|main>-alt`.

#### Update:

No longer necessary.

aiyengar2 (approved):
LGTM, but I'm not sure what the context for this change was. Do we need to specify that it can be any namespace?

@SheilaghM @dramich should we merge this in or close it out?

PR #951, opened 2021-01-21
Add PodSecurityPolicies for rancher-logging

bashofmann:
Add PodSecurityPolicies for rancher-logging.

Addresses https://github.com/rancher/rancher/issues/30771

Original PR before charts changeover: https://github.com/rancher/charts/pull/931

### DIFF
--- a/packages/rancher-logging/generated-changes/overlay/templates/loggings/aks/logging.yaml
+++ b/packages/rancher-logging/generated-changes/overlay/templates/loggings/aks/logging.yaml
@@ -15,6 +15,11 @@ spec:
     inputTail:
       Tag: "aks"
       Path: "/var/log/azure/kubelet-status.log"
+    {{- if .Values.rbac.psp.enabled }}
+    security:
+      podSecurityPolicyCreate: true
+      roleBasedAccessControlCreate: true
+    {{- end }}
   {{- if .Values.additionalLoggingSources.aks.fluentbit.mountPath }}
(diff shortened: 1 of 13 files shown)


nickgerace (approved):

nickgerace:
LGTM, adding @aiyengar2 to confirm the `packageVersion` vs. `releaseCandidateVersion`.

packages/rancher-logging/generated-changes/overlay/templates/loggings/aks/logging.yaml
```diff
@@ -15,6 +15,9 @@ spec:
     inputTail:
       Tag: "aks"
       Path: "/var/log/azure/kubelet-status.log"
+    security:
+      podSecurityPolicyCreate: true
+      roleBasedAccessControlCreate: true
```
cbron, line 20: Did we decide if we are going to wrap these with `if psp.enabled` conditionals, for users who don't want them. I would say yes because its little work on our side for a failsafe, even if its unlikely to be used.

nickgerace: +1, we should do that. Probably would look like...
```yaml
{{- if values.psp.enabled }}
# insert security...
{{- end }}
```

bashofmann: @nickgerace I added the ifs

nickgerace (approved):

nickgerace (approved):

cbron (approved):

PR #952, opened 2021-01-22
Use generated-changes/dependencies for Istio

aiyengar2:
One of the misses from the charts-build-scripts migration was that rancher-istio was still using .tgz files within its `{workingDir}/charts/` directory (which contains all the dependencies for the Helm chart) instead of dynamically pulling them in.

This commit modifies Istio to use generated-changes and bumps the charts-build-scripts to point to a newer version that adds support for this.


aiyengar2:
Staging version bump: https://github.com/rancher/charts/pull/953

Live version bump: https://github.com/rancher/charts/pull/954

brendarearden (approved):
LGTM

PR #958, opened 2021-01-26
Increase Istio and Kiali version

brendarearden:
**Issue**
Istio and Kiali release new versions (1.8.1 & 1.29.0)

**Solution**
Upgrade Istio and Kiali to 1.8.1 and 1.29.0

**Issue**
https://github.com/rancher/rancher/issues/30410
https://github.com/rancher/rancher/issues/30470
https://github.com/rancher/rancher/issues/29700

**Dependent PRs**
https://github.com/rancher/image-mirror/pull/66
https://github.com/rancher/image-mirror/pull/61
https://github.com/rancher/istio-installer/pull/16


packages/rancher-istio/package.yaml
```diff
@@ -1,3 +1,3 @@
 url: local
-packageVersion: 01
-releaseCandidateVersion: 02
+packageVersion: 00
+releaseCandidateVersion: 01
```
nickgerace, line 3: I think this has to be `00`
CC: @aiyengar2

nickgerace (changes requested):

packages/rancher-istio/package.yaml
brendarearden: I had an error on my initial pr, so bumped the rc version. I ave now put it back to 00. You are correct, it should be 00.

nickgerace (approved):

packages/rancher-istio/charts/configs/istio-base.yaml
```diff
@@ -72,25 +68,12 @@ spec:
...
-    mixer:
-      {{- if .Values.policy.enabled }}
-      policy:
-        image: {{ template "system_default_registry" . }}{{ .Values.policy.repository }}:{{ .Values.policy.tag }}
-      {{- end }}
-      {{- if .Values.telemetry.v1.enabled }}
-      telemetry:
-        image: {{ template "system_default_registry" . }}{{ .Values.telemetry.v1.repository }}:{{ .Values.telemetry.v1.tag }}
-      {{- end }}
-    {{- end }}
```
cbron, line 85: This is because mixer is gone ?

packages/rancher-istio/charts/Chart.yaml
```diff
@@ -13,6 +13,6 @@ annotations:
   catalog.cattle.io/release-name: rancher-istio
   catalog.cattle.io/ui-component: istio
   catalog.cattle.io/provides-gvr: networking.istio.io.virtualservice/v1beta1
-  catalog.cattle.io/auto-install: rancher-kiali-server-crd=1.24.003
+  catalog.cattle.io/auto-install: rancher-kiali-server-crd=1.29.000
```
cbron, line 16: Doesn't this have to change in requirements.txt too ? Does your script still work ?

cbron (changes requested):

packages/rancher-istio/charts/Chart.yaml
brendarearden: This is part of the changes that we had to make with the new scripts. The version for dependencies now comes from `generated-changes/dependencies/<kiali>/dependency.yaml`

packages/rancher-istio/charts/configs/istio-base.yaml
brendarearden: Yes, mixer is not supported in any form in 1.8. I did put a comment in the values.yaml during the 1.7 release that said it would be deprecated soon. This will be something that would need to go in release notes.

packages/rancher-istio/charts/Chart.yaml
cbron: Ok, can we remove requirements.txt ?

brendarearden: The requirements.txt turned into requirements.yaml and it needs to remain for the pulling in the correct generated dependencies https://github.com/aiyengar2/charts-build-scripts/blob/a26221f4b798e9df62439b96d2c0c9c2acc3dbe7/templates/template/scripts/regenerate-packages#L100

cbron: Ok, so just the script/version_update needs updated. We can do that a diff time but try to hit eventually.

cbron (approved):

packages/rancher-istio/charts/Chart.yaml
aiyengar2: @brendarearden if you do `make charts` and then pull it into rancher and install, is it able to find the CRD chart to auto-install?

Just realized that if the CRD chart has `-rc00` appended to it the scripts may not be able to find it on the UI to auto-install. Wondering what happens in this case.

At worst case, this would only impact QA being able to install the CRD chart stuff. Should still be fine when we actually release, but the rc version in this annotation will need to be modified in `make charts`.

brendarearden: It is not able to find the CRD chart for auto install so this will need to be adjusted for testing RCs

cbron: Since we need to get these to QA asap, we should just take out the RC stuff and do version bumps directly. If it fails QA and we need to update again we can do another version bump or overwrite the current version and make QA aware of that. cc @brendarearden / @aiyengar2

aiyengar2: @brendarearden once https://github.com/rancher/charts/pull/964, https://github.com/rancher/charts/pull/965, https://github.com/rancher/charts/pull/966 are in (which vendor in changes to scripts for https://github.com/rancher/charts-build-scripts/pull/10), you will need to slightly modify the PR to set this annotation to `rancher-kiali-server-crd=1.29.000-rc00`.

Since the releaseCandidateVersion of the CRD chart of rancher-kiali-server is not tied to the releaseCandidateVersion of the rancher-istio chart, the fix in the scripts is to just to ensure that the rcVersion gets dropped when we sync to main (tested this out locally and it seems to work fine).

You'll still need to make sure you update the Chart annotation manually any time a new release candidate version comes in for rancher-kiali-server.

aiyengar2 (approved):
LGTM, checked that the CRD chart installed as expected

PR #960, opened 2021-01-27
[Logging v2] Split psp settings for logging

nickgerace:
Relevant issue: https://github.com/rancher/rancher/issues/31006
Original issue: https://github.com/rancher/rancher/issues/30771

### DIFF
--- a/packages/rancher-logging/generated-changes/overlay/templates/loggings/aks/logging.yaml
+++ b/packages/rancher-logging/generated-changes/overlay/templates/loggings/aks/logging.yaml
@@ -15,7 +15,7 @@ spec:
     inputTail:
       Tag: "aks"
       Path: "/var/log/azure/kubelet-status.log"
-    {{- if .Values.rbac.psp.enabled }}
+    {{- if .Values.global.psp.enabled }}
     security:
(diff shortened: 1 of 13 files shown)


brendarearden:
so both the rbac psp and the global psp will be set to true?

nickgerace:
@brendarearden: No, the `rbac.psp` will be set to false, which is the upstream default. The `global.psp` will be true because that only affects the overlay.

aiyengar2 (approved):
LGTM, as discussed offline this will not have any implications on an upgrade because `global.psp.enabled` is set to true by default. 

A user who did not have `rbac.psp.enabled=False` prior to this upgrade would only see new resources be created, and a user with `rbac.psp.enabled=True` prior to this upgrade would see no changes in the cluster.

The only way that resources would be lost on an upgrade is if is a user specifically override `global.psp.enabled=False` during the upgrade, which is expected. 

Confirmed that Rancher never automatically overrides `global.psp.enabled=False` as part of the UI.

brendarearden (approved):
LGTM

cbron (approved):

PR #961, opened 2021-01-28
Remove rc1 from chart version

thedadams:


aiyengar2 (approved):
LGTM

cbron (approved):
Believe that makes sense. Make sure that when you run “make charts” it sets version and RC right, it needs to move to 300.
We can wait for Arvind to double check if you want.

PR #962, opened 2021-01-28
Change to not increase package version

brendarearden:
**Problem**
Package version was at 01, but needed to be at 00, since istio 1.7.6 is not in production yet. 

**Solution**
Change istio package version to 00

**Issue**
https://github.com/rancher/rancher/issues/29812


brendarearden:
We are skipping 1.7.6 at this point.

PR #963, opened 2021-01-30
[Logging v2] Default to the global mountPath for logging vendors

nickgerace:
Relevant issue: https://github.com/rancher/rancher/issues/30329

### DIFF
--- a/packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/logging-rke2-journald.yaml
+++ b/packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/logging-rke2-journald.yaml
@@ -24,6 +24,8 @@ spec:
     {{- end }}
   {{- if .Values.additionalLoggingSources.rke2.fluentbit.mountPath }}
     mountPath: {{ .Values.additionalLoggingSources.rke2.fluentbit.mountPath }}
+  {{- else if .Values.fluentbit.mountPath }}
+    mountPath: {{ .Values.fluentbit.mountPath }}
   {{- end }}
   {{- $total_tolerations := concat (.Values.tolerations) (.Values.fluentbit.tolerations) }}
   {{- with $total_tolerations }}
--- a/packages/rancher-logging/generated-changes/patch/values.yaml.patch
+++ b/packages/rancher-logging/generated-changes/patch/values.yaml.patch
@@ -35,7 +35,7 @@
  rbac:
    enabled: true
    psp:
-@@ -80,3 +87,73 @@
+@@ -80,3 +87,70 @@
      additionalLabels: {}
      metricRelabelings: []
      relabelings: []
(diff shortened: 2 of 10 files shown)


rmweir:
I see our motivation for this is to help debug an issue but beyond that is this a necessary and desired change? Do you think no one will want separate mount paths?

nickgerace:
@rmweir that's a good question. I believe that we do want to combine them because the Logging CRs share the same nodes and ultimately deploy fluent pods in daemonsets. We will most likely run into more scenarioes where the right `mountPath` settings were not set correctly moreso than users will want to separate `mountPath` values.

If the user does want to separate `mountPath` values, they can edit a value after installation or creating their own Logging CR (which isn't uncommon).

nickgerace:
@rmweir: another thing: you are right to question the motivation behind this PR. For further context: this was a back-and-forth design question before the _original_ PR. We decided to go with the split `mountPath` fields, but were willing to change dependent on future UX observations or issues.

rmweir (approved):
LGTM. Asked some questions about upgrades elsewhere.

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/logging-rke2-journald.yaml
```diff
@@ -22,8 +22,8 @@ spec:
       podSecurityPolicyCreate: true
       roleBasedAccessControlCreate: true
     {{- end }}
-  {{- if .Values.additionalLoggingSources.rke2.fluentbit.mountPath }}
-    mountPath: {{ .Values.additionalLoggingSources.rke2.fluentbit.mountPath }}
+  {{- if .Values.fluentbit.mountPath }}
+    mountPath: {{ .Values.fluentbit.mountPath }}
```
prachidamle, line 26: Why don't we support both the cases:
1. Have support for separate mounthPaths in Values.yaml
2. If user does not set a separate mountPath, use the ".Values.fluentbit.mountPath"

nickgerace: I weighed this idea as well, but did not want to expose the option if it would never realistically get used. My concern was that users may be confused by the option existing in `values.yaml`. Perhaps, we can support both cases and add a comment clarifying that `fluentbit.mountPath` will affect every Logging CR unless a value is detected for `additionalLoggingSources.${SOURCE}.fluentbit.mountPath`.

prachidamle: yes comment, or readme can carry the chart values and explanations

nickgerace:
Scope of this PR has changed due to discussion: https://github.com/rancher/charts/pull/963#discussion_r568184476. Now, we will attempt to use the global `mountPath` if the vendor is unavailable.

packages/rancher-logging/generated-changes/patch/values.yaml.patch
```diff
@@ -85,14 +85,11 @@
 +    repository: rancher/banzaicloud-fluentd
 +    tag: v1.11.5-alpine-9
 +
++# These "fluentbit" settings apply to every Logging CR, including vendor Logging CRs enabled in
++# "additionalLoggingSources". Changing these affects every Logging CR installed. However, you can
++# override the "mountPath" setting in the corresponding Logging CR(s) for a given vendor (i.e. k3s).
```
nickgerace, line 90: @prachidamle I think this comment addresses the overriding use case.

prachidamle (approved):

PR #971, opened 2021-02-02
Adding cpu, memory chart annotations

brendarearden:
**Issue**
V1 version of Istio chart had a minimum cpu and memory requirement for cluster - V2 needs the same thing

**Solution**
Chart annotations will be added for rancher to use to check space and added ui warning

**Issue**
https://github.com/rancher/rancher/issues/30144

### DIFF
--- a/packages/rancher-istio/charts/Chart.yaml
+++ b/packages/rancher-istio/charts/Chart.yaml
@@ -16,3 +16,5 @@ annotations:
   catalog.cattle.io/auto-install: rancher-kiali-server-crd=1.29.000-rc00
   catalog.cattle.io/display-name: "Istio"
   catalog.cattle.io/os: linux
+  catalog.cattle.io/requests-cpu: "710m"
+  catalog.cattle.io/requests-memory: "2314Mi"
--- a/packages/rancher-istio/package.yaml
+++ b/packages/rancher-istio/package.yaml
@@ -1,3 +1,3 @@
 url: local
 packageVersion: 00
-releaseCandidateVersion: 00
+releaseCandidateVersion: 01


paynejacob (approved):

packages/rancher-istio/charts/Chart.yaml
```diff
@@ -16,3 +16,5 @@ annotations:
   catalog.cattle.io/auto-install: rancher-kiali-server-crd=1.29.000-rc00
   catalog.cattle.io/display-name: "Istio"
   catalog.cattle.io/os: linux
+  catalog.cattle.io/requests-cpu: "4500m"
+  catalog.cattle.io/requests-memory: "5620Mi"
```
cbron, line 20: Curious how you came to these numbers ?

brendarearden: I went into each chart in the manifest files and looked at their values.yaml default resource settings, for example this is from the ingress gateway chart
```
    resources:
      requests:
        cpu: 100m
        memory: 128Mi
      limits:
        cpu: 2000m
        memory: 1024Mi
```
And I added them together to get a minimum default setting - I did this when we initially switch to V2 https://rancher.com/docs/rancher/v2.x/en/istio/v2.5/resources/ - but now that I am looking at this side by side - something is wack... somehow...columns got switched? I will quickly do this again and get the PR updated as well as updates to the Docs

brendarearden: TLDR; I mixed up values for limits vs requests during calculations so they were incorrect

When making original calculations, some of the resources had switched the order in which they listed the resources requests and limits so I also did this in many places. 

I have redone the calculations, this no longer includes removed resources like mixer etc. These values were pulled from the default values for the resource https://github.com/rancher/docs/issues/2999

cbron (approved):
Totals look more correct now.

nickgerace (approved):
LGTM, apologies for the wait!

PR #967, opened 2021-02-01
Update rancher-operator to v0.1.3-rc3

thedadams:
Original PR:
https://github.com/rancher/rancher-operator/pull/19

PR #972, opened 2021-02-03
update rancher-operator to v0.1.3-rc4

kinarashah:
https://github.com/rancher/rancher/issues/29993

### DIFF
--- a/packages/rancher-operator-crd/package.yaml
+++ b/packages/rancher-operator-crd/package.yaml
@@ -1,3 +1,3 @@
-url: https://github.com/rancher/rancher-operator/releases/download/v0.1.3-rc3/rancher-operator-crd-0.1.3-rc3.tgz
+url: https://github.com/rancher/rancher-operator/releases/download/v0.1.3-rc4/rancher-operator-crd-0.1.3-rc4.tgz
 packageVersion: 00
-releaseCandidateVersion: 03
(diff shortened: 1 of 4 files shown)


StrongMonkey:
We shouldn't need to generate patch file..

packages/rancher-operator-crd/package.yaml
```diff
@@ -1,3 +1,3 @@
-url: https://github.com/rancher/rancher-operator/releases/download/v0.1.3-rc3/rancher-operator-crd-0.1.3-rc3.tgz
+url: https://github.com/rancher/rancher-operator/releases/download/v0.1.3-rc4/rancher-operator-crd-0.1.3-rc4.tgz
 packageVersion: 00
-releaseCandidateVersion: 03
```
StrongMonkey, line 3: We should remove this field `releaseCandidateVersion`

kinarashah: remove the field or not update it? I'm using https://github.com/rancher/charts/pull/967/files as an example

StrongMonkey (approved):

packages/rancher-operator-crd/package.yaml
StrongMonkey: nvm

PR #988, opened 2021-02-10
[2.5.8] Add global seLinux toggle to rancher-logging

cmurphy:
Add a new value to the rancher-logging chart to enable setting the
seLinuxOptions securityContext attribute for Logging resources and the
rke-aggregator DaemonSet. This is applied to any resource that needs to
read from /var/log/containers or /var/lib/docker/containers.

Without this patch, components in the rancher-logging chart mount and
attempt to read and sometimes write files on several hostPaths, and when
SELinux is enabled on the host, will fail because the containers are
unauthorized to do so. The added seLinuxOptions attribute in this patch
causes the fluentbit and rke-aggregator DaemonSets to run with the
SELinux context type `rke_logreader_t`, which is provided by the
rke2-linux package and enables the container to read files with the
label `container_log_t`, which includes files like
/var/log/containers/*.log, and files with type `rke_container_var_lib_t`
which the docker log files are relabeled to.

The logging aggregator will still fail to read and write log data from
/var/lib/rancher/logging, but this will be addressed in a different way
(https://github.com/rancher/rancher/issues/31309).

https://github.com/rancher/rancher/issues/30949

Depends on https://github.com/rancher/rke2-selinux/pull/13

### DIFF
--- a/packages/rancher-logging/generated-changes/patch/README.md.patch
+++ b/packages/rancher-logging/generated-changes/patch/README.md.patch
@@ -1,0 +1,10 @@
+--- charts-original/README.md
++++ charts/README.md
+@@ -67,6 +67,7 @@
+ | `securityContext`                                   | Container SecurityContext for Logging operator. [More info](https://kubernetes.io/docs/concepts/policy/security-context/) | `{"allowPrivilegeEscalation": false, "readOnlyRootFilesystem": true}` |
+ | `createCustomResource`                              | Create CRDs. | `true` |
+ | `monitoring.serviceMonitor.enabled`                 | Create Prometheus Operator servicemonitor. | `false` |
++| `global.seLinux.enabled`                            | Add seLinuxOptions to Logging resources, requires the [rke2-selinux RPM](https://github.com/rancher/rke2-selinux/releases) | `false` |
+ 
+ Alternatively, a YAML file that specifies the values for the parameters can be provided while installing the chart. For example:
+ 
--- a/packages/rancher-logging/package.yaml
+++ b/packages/rancher-logging/package.yaml
@@ -1,6 +1,6 @@
 url: https://kubernetes-charts.banzaicloud.com/charts/logging-operator-3.9.0.tgz
-packageVersion: 01
-releaseCandidateVersion: 08
+packageVersion: 02
+releaseCandidateVersion: 01
 additionalCharts:
   - workingDir: charts-crd
     crdOptions:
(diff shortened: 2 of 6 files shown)


nickgerace (dismissed):
LGTM, awesome

dweomer (approved):

packages/rancher-logging/package.yaml
```diff
@@ -1,6 +1,6 @@
 url: https://kubernetes-charts.banzaicloud.com/charts/logging-operator-3.9.0.tgz
 packageVersion: 00
-releaseCandidateVersion: 08
+releaseCandidateVersion: 09
```
cbron, line 3: Once rest of PR is confirmed you'll need to pull in upstream again and bump packageVersion.

packages/rancher-logging/generated-changes/patch/README.md.patch
```diff
@@ -0,0 +1,10 @@
+--- charts-original/README.md
++++ charts/README.md
+@@ -67,6 +67,7 @@
+ | `securityContext`                                   | Container SecurityContext for Logging operator. [More info](https://kubernetes.io/docs/concepts/policy/security-context/) | `{"allowPrivilegeEscalation": false, "readOnlyRootFilesystem": true}` |
+ | `createCustomResource`                              | Create CRDs. | `true` |
+ | `monitoring.serviceMonitor.enabled`                 | Create Prometheus Operator servicemonitor. | `false` |
++| `global.seLinux.enabled`                            | Add seLinuxOptions to Logging resources | `false` |
```
cbron, line 7: We should add that it requires the RPM. Could also link out to docs once they are ready.

paynejacob (commented):
We should add a note about the rpm dependency and then I think it is good.

packages/rancher-logging/generated-changes/patch/README.md.patch
cmurphy: Added a link to the rke2-selinux github release page

packages/rancher-logging/package.yaml
cmurphy: Bumped the packageVersion, does the releaseCandidateVersion stay the same or get reset?

cbron: Needs to get reset to 01

packages/rancher-logging/generated-changes/patch/README.md.patch
cbron: That works for now, we should try to remember to swap once it moves, ideally to the official docs.

packages/rancher-logging/package.yaml
cmurphy: done

cbron (approved):

PR #990, opened 2021-02-11
Adding Istio 1.8.3

brendarearden:
**Problem**
1. Istio release new Istio 1.8.3 version
2. Uninstall error occurs when chart dependencies are removed prior to rancher-istio gets removed

**Solution**
1.Add a chart to support Istio 1.8.3
2.Add documentation about the uninstall process to chart readme. 

**Issue**
https://github.com/rancher/rancher/issues/31187
https://github.com/rancher/rancher/issues/31046

**Dependent PRs**
https://github.com/rancher/image-mirror/pull/73
https://github.com/rancher/istio-installer/pull/17

**Additional Notes**
Kiali and Jaeger version did not change from 1.8.1 to 1.8.3

### DIFF
--- a/packages/rancher-backup/package.yaml
+++ b/packages/rancher-backup/package.yaml
@@ -1,6 +1,6 @@
 url: local
 packageVersion: 01
-releaseCandidateVersion: 00
+releaseCandidateVersion: 01
 additionalCharts:
   - workingDir: charts-crd
     crdOptions:
--- a/packages/rancher-istio/charts/README.md
+++ b/packages/rancher-istio/charts/README.md
@@ -7,6 +7,12 @@ A Rancher created chart that packages the istioctl binary to install via a helm 
 ## Chart Dependencies
 - rancher-kiali-server-crd chart
 
+# Uninstallation Requirements 
+To ensure rancher-istio uninstalls correctly, you must uninstall rancher-istio prior to uninstalling chart dependencies (see installation requirements for chart dependencies). This is because all definitions need to be available in order to properly build the rancher-istio objects for removal.
+
+If you remove dependent CRD charts prior to removing rancher-istio, you may encounter the following error::
+
+`Error: uninstallation completed with 1 error(s): unable to build kubernetes objects for delete: unable to recognize "": no matches for kind "MonitoringDashboard" in version "monitoring.kiali.io/v1alpha1"`
 
 # Addons
 
--- a/packages/rancher-istio/package.yaml
+++ b/packages/rancher-istio/package.yaml
@@ -1,3 +1,3 @@
 url: local
 packageVersion: 00
-releaseCandidateVersion: 01
+releaseCandidateVersion: 00
--- a/packages/rancher-istio/charts/Chart.yaml
+++ b/packages/rancher-istio/charts/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: 1.8.1
+appVersion: 1.8.3
 description: A basic Istio setup that installs with the istioctl. Refer to https://istio.io/latest/ for details.
 name: rancher-istio
-version: 1.8.1
+version: 1.8.3
 icon: https://charts.rancher.io/assets/logos/istio.svg
 keywords:
 - networking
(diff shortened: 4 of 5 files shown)


paynejacob (dismissed):

brendarearden:
It looks like we merged the rancher-backup chart without bumping the rc version so its causing the CI to break. I bumped the rc version in this pr and it seems to have fixed the issue.

packages/rancher-istio/charts/README.md
```diff
@@ -7,6 +7,12 @@ A Rancher created chart that packages the istioctl binary to install via a helm
 ## Chart Dependencies
 - rancher-kiali-server-crd chart
 
+# Uninstallation Requirements 
+To ensure rancher-istio uninstalls correctly, you must uninstall rancher-istio prior to uninstalling chart dependencies (see installation requirements for chart dependencies). This is because all definitions need to be available in order to properly build the rancher-istio objects for removal.
+
+Example uninstallation error you might encounter if you remove dependent CRD charts prior to removing rancher-istio:
```
nickgerace, line 13: nit: If you remove dependent CRD charts prior to removing rancher-istio, you may encounter the following error:

packages/rancher-istio/package.yaml
```diff
@@ -1,3 +1,3 @@
 url: local
 packageVersion: 00
-releaseCandidateVersion: 01
+releaseCandidateVersion: 00
```
nickgerace, line 3: RC is getting decremented?

nickgerace (dismissed):
LGTM barring nits

packages/rancher-istio/package.yaml
nickgerace: Oh, this is the cause? https://github.com/rancher/charts/pull/990#issuecomment-777633336

brendarearden: Its a new minor version, so package and rc version start back at 00.

nickgerace: Doesn't package version get incremented then? Or is that not the case since it's the same release? (CC: @aiyengar2)

brendarearden: If 1.8.100-rc00 is released in the release branch, and we need to make a change, we would increment package version >> 1.8.101-rc00
If 1.8.100-rc00 is only in dev branch, and we are making changes, we would increment rc version >> 1.8.100-rc01
If minor version does not exist in dev branch or release branch, we create a pr for 1.8.300-rc00 with no package and version increments

aiyengar2: Add guidance for this in https://github.com/rancher/charts/tree/dev-v2.5-source#versioning-packages

nickgerace: Got it. All good then

nickgerace (approved):

packages/rancher-backup/package.yaml
```diff
@@ -1,6 +1,6 @@
 url: local
 packageVersion: 01
-releaseCandidateVersion: 00
+releaseCandidateVersion: 01
```
cbron, line 3: Assuming this is just to fix CI, if its still a problem we need to look into it. 
I don't see backup on our list of new charts in 2.5.6, can you add it there.

cbron (approved):

packages/rancher-backup/package.yaml
brendarearden: https://github.com/rancher/charts/pull/935 << yeah it looks like this pr was merged without an rc change.

aiyengar2: Yup, this looks right. The sync workflow (applied on a push to dev-v2.5-source) also seems to have failed accordingly, so the changes from that PR never made it into dev-v2.5: https://github.com/rancher/charts/runs/1875711331

Seems like the push from this PR took care of that though, as seen here:
https://github.com/rancher/charts/commit/fc035102ae98090412de5b33615f6d71c866a741

nickgerace: I rebased with my current PR and the CI passed. Think that did the trick
https://github.com/rancher/charts/pull/991

PR #992, opened 2021-02-12
Update default kubelet path for k8s 1.20

aiyengar2:
Related Issue: https://github.com/rancher/rancher/issues/31167, https://github.com/rancher/rancher/issues/31305

### DIFF
--- a/packages/rancher-monitoring/generated-changes/patch/templates/_helpers.tpl.patch
+++ b/packages/rancher-monitoring/generated-changes/patch/templates/_helpers.tpl.patch
@@ -1,6 +1,6 @@
 --- charts-original/templates/_helpers.tpl
 +++ charts/templates/_helpers.tpl
-@@ -1,3 +1,61 @@
+@@ -1,3 +1,70 @@
 +# Rancher
 +{{- define "system_default_registry" -}}
 +{{- if .Values.global.cattle.systemDefaultRegistry -}}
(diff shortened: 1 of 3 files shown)


PennyScissors (approved):

packages/rancher-monitoring/generated-changes/patch/templates/_helpers.tpl.patch
```diff
@@ -57,6 +57,15 @@
 +{{- end -}}
 +{{- end }}
 +
++{{- define "kubelet.serviceMonitor.resourcePath" -}}
++{{- $kubeTargetVersion := default .Capabilities.KubeVersion.GitVersion .Values.kubeTargetVersionOverride }}
++{{- if and (semverCompare ">=1.20.0-0" $kubeTargetVersion) (eq .Values.kubelet.serviceMonitor.resourcePath "/metrics/resource/v1alpha1") -}}
```
cbron, line 62: This feels kinda weird, users can never set it to this path in future. Perhaps thats ok if it's really gone from there.

aiyengar2: This path was deprecated in an earlier k8s version and removed entirely in k8s 1.20, so it should be gone for good.

cbron: Ok. Well then its just making sure its inline with semver compare convo from other PR.

cbron (dismissed):

cbron (approved):

PR #995, opened 2021-02-13
[Logging v2] Add extraVolumeMounts to dockerRootDir

nickgerace:
Relevant issue: https://github.com/rancher/rancher/issues/30329

### DIFF
--- a/packages/rancher-logging/generated-changes/overlay/templates/loggings/rke/logging-rke.yaml
+++ b/packages/rancher-logging/generated-changes/overlay/templates/loggings/rke/logging-rke.yaml
@@ -1,4 +1,5 @@
 {{- if .Values.additionalLoggingSources.rke.enabled }}
+{{- $containers := printf "%s/containers/" (default "/var/lib/docker" .Values.global.dockerRootDirectory) }}
 apiVersion: logging.banzaicloud.io/v1beta1
 kind: Logging
 metadata:
@@ -26,7 +27,7 @@ spec:
       roleBasedAccessControlCreate: true
     {{- end }}
   {{- if .Values.global.dockerRootDirectory }}
-    mountPath: {{ .Values.global.dockerRootDirectory }}
+    mountPath: {{ $containers }}
   {{- end }}
   {{- $total_tolerations := concat (.Values.tolerations) (.Values.fluentbit.tolerations) }}
   {{- with $total_tolerations }}
--- a/packages/rancher-logging/generated-changes/overlay/templates/loggings/root/logging.yaml
+++ b/packages/rancher-logging/generated-changes/overlay/templates/loggings/root/logging.yaml
@@ -1,3 +1,4 @@
+{{- $containers := printf "%s/containers/" (default "/var/lib/docker" .Values.global.dockerRootDirectory) }}
 apiVersion: logging.banzaicloud.io/v1beta1
 kind: Logging
 metadata:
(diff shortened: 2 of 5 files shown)


packages/rancher-logging/generated-changes/overlay/templates/loggings/rke/logging-rke.yaml
```diff
@@ -1,4 +1,5 @@
 {{- if .Values.additionalLoggingSources.rke.enabled }}
+{{- $containers := printf "%s/%s" .Values.global.dockerRootDirectory "containers" -}}
```
rmweir, line 2: Does this not need a default in case .Values.global.dockerRootDirectory is empty similar to above usages?

rmweir (changes requested):
I feel like the commit message is not describing anything that would cause the problem. Can you clarify or explain how the change fixes the problem?

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke/logging-rke.yaml
nickgerace: No because `$containers` is only used within protected blocks within this file.
```yaml
{{- if .Values.global.dockerRootDirectory }}
# foo
{{- end }}
```
However, it might be a good idea to set a default anyway to prevent mishaps in the future. Thoughts?

nickgerace:
RE: https://github.com/rancher/charts/pull/995#pullrequestreview-592753990
> I feel like the commit message is not describing anything that would cause the problem. Can you clarify or explain how the change fixes the problem?
Sure thing. Admittedly, this was a multi-faceted issue since numbers small problems were fixed rather one big one, but I can amend the commit title and message.

paynejacob (approved):

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke/logging-rke.yaml
nickgerace: Thought about this some more and asked @paynejacob. Going to resolve. I think pre-optimizing isn't necessary here.

cbron, line 2: The differing ways of doing this is odd. Why not same as above
```suggestion
{{- $containers := printf "%s/containers" .Values.global.dockerRootDirectory -}}
```
Also why not trailing slash here ? It might not make a difference but consistency would be good. I see same thing below also.

packages/rancher-logging/generated-changes/overlay/templates/loggings/root/logging.yaml
```diff
@@ -17,7 +18,11 @@ spec:
       roleBasedAccessControlCreate: true
     {{- end }}
   {{- if .Values.global.dockerRootDirectory }}
-    mountPath: {{ .Values.global.dockerRootDirectory }}
+    mountPath: {{ $containers }}
+    extraVolumeMounts:
+      - source: {{ $containers }}
+        destination: {{ $containers }}
+        readOnly: true
```
cbron, line 25: Looks like `extraVolumeMounts` is a banzai chart thing, just another way of specifying mounts ?

cbron (commented):
In comment you have
>  by minimizing inputTail usage
What is that referring to ?

nickgerace:
RE: https://github.com/rancher/charts/pull/995#pullrequestreview-593683767
This is an artifact from the previous commit message. It's meant to imply that by _not_ adding an inputTail, we are using the default. The "minimization" piece refers to removing unnecessary values in the templates. Regardless, I will remove this.

packages/rancher-logging/generated-changes/overlay/templates/loggings/root/logging.yaml
nickgerace: Yes. This is required for the symlink to work though.

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke/logging-rke.yaml
nickgerace: There's a few ways to go about this (https://github.com/rancher/charts/pull/995#discussion_r578102738), but I will make it consistent across the repository.

cbron (approved):

rmweir (approved):
LGTM

PR #996, opened 2021-02-13
[DNM 2.5.7] Add rancher-wins-upgrader

aiyengar2:
`rancher-wins-upgrader` creates a quick and easy Helm chart that just upgrades the config based on the provided YAML and updates the wins binary. Every time you do a helm upgrade, it will restart the DaemonSet (if something has changed) to bump the wins server versions to the one you declared in the values.

Example manifest pumped out by the default on `helm template`:
```yaml
---
# Source: rancher-wins-upgrader/templates/rbac.yaml
apiVersion: policy/v1beta1
kind: PodSecurityPolicy
metadata:
  name: wins-upgrader
  namespace: default
  labels: 
    k8s-app: wins-upgrader
    release: RELEASE-NAME
    provider: kubernetes
spec:
  privileged: false
  hostNetwork: false
  hostIPC: false
  hostPID: false
  runAsUser:
    rule: 'RunAsAny'
  seLinux:
    rule: 'RunAsAny'
  supplementalGroups:
    rule: 'MustRunAs'
    ranges:
      - min: 0
        max: 65535
  fsGroup:
    rule: 'MustRunAs'
    ranges:
      - min: 0
        max: 65535
  readOnlyRootFilesystem: false
  volumes:
    - 'secret'
    - 'hostPath'
  allowedHostPaths:
  - pathPrefix: c:/etc/rancher/wins
---
# Source: rancher-wins-upgrader/templates/rbac.yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: wins-upgrader
  namespace: default
  labels: 
    k8s-app: wins-upgrader
    release: RELEASE-NAME
    provider: kubernetes
---
# Source: rancher-wins-upgrader/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: wins-upgrader-default
  namespace: default
  labels: 
    k8s-app: wins-upgrader
    release: RELEASE-NAME
    provider: kubernetes
data:
  config: |- 
    debug: false
    listen: rancher_wins
    proxy: rancher_wins_proxy
    white_list: 
      process_paths:
      - c:\\etc\wmi-exporter\windows-exporter.exe
      - c:\\etc\wmi-exporter\wmi-exporter.exe
      - c:\\etc\kubernetes\bin\kube-proxy.exe
      - c:\\etc\kubernetes\bin\kubelet.exe
      - c:\\etc\nginx\nginx.exe
      - c:\\opt\bin\flanneld.exe
      proxy_ports:
      - 9796
    upgrade: 
      mode: watching
      watching_path: c:\\etc\rancher\wins\wins.exe
    
  upgrade.ps1: |-
    $ErrorActionPreference = 'Stop'
    
    function Create-Directory
    {
        param (
            [parameter(Mandatory = $false, ValueFromPipeline = $true)] [string]$Path
        )
    
        if (Test-Path -Path $Path) {
            if (-not (Test-Path -Path $Path -PathType Container)) {
                # clean the same path file
                Remove-Item -Recurse -Force -Path $Path -ErrorAction Ignore | Out-Null
            }
    
            return
        }
    
        New-Item -Force -ItemType Directory -Path $Path | Out-Null
    }
    
    function Transfer-File
    {
        param (
            [parameter(Mandatory = $true)] [string]$Src,
            [parameter(Mandatory = $true)] [string]$Dst
        )
    
        if (Test-Path -PathType leaf -Path $Dst) {
            $dstHasher = Get-FileHash -Path $Dst
            $srcHasher = Get-FileHash -Path $Src
            if ($dstHasher.Hash -eq $srcHasher.Hash) {
                return
            }
        }
    
        $null = Copy-Item -Force -Path $Src -Destination $Dst
    }
    
    Create-Directory -Path "c:\host\etc\rancher\wins"
    Transfer-File -Src "c:\scripts\config" -Dst "c:\host\etc\rancher\wins\config"
    Transfer-File -Src "c:\Windows\wins.exe" -Dst "c:\host\etc\rancher\wins\wins.exe"
    
    # Sleep forever, since a DaemonSet's restartPolicy must be Always
    while(1) { Start-Sleep -s 3600 }
---
# Source: rancher-wins-upgrader/templates/rbac.yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: wins-upgrader
  namespace: default
  labels: 
    k8s-app: wins-upgrader
    release: RELEASE-NAME
    provider: kubernetes
rules:
- apiGroups: ['policy']
  resources: ['podsecuritypolicies']
  verbs:     ['use']
  resourceNames: ["wins-upgrader"]
---
# Source: rancher-wins-upgrader/templates/rbac.yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: wins-upgrader
  namespace: default
  labels: 
    k8s-app: wins-upgrader
    release: RELEASE-NAME
    provider: kubernetes
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: wins-upgrader
subjects:
- kind: ServiceAccount
  name: wins-upgrader
  namespace: default
---
# Source: rancher-wins-upgrader/templates/daemonset.yaml
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: wins-upgrader-default
  namespace: default
  labels: 
    k8s-app: wins-upgrader
    release: RELEASE-NAME
    provider: kubernetes
spec:
  selector:
    matchLabels: 
      k8s-app: wins-upgrader
      release: RELEASE-NAME
      provider: kubernetes
  template:
    metadata:
      labels: 
        k8s-app: wins-upgrader
        release: RELEASE-NAME
        provider: kubernetes
    spec:
      nodeSelector: 
        kubernetes.io/os: windows
      tolerations: 
        - operator: Exists
      serviceAccountName: wins-upgrader
      containers:
      - name: wins-upgrader
        image: arvindiyengar/windows_exporter-package:dev-windows-1809
        command: ["pwsh", "-f", "c:/scripts/upgrade.ps1"]
        volumeMounts:
        - name: wins
          mountPath: c:/host/etc/rancher/wins
        - name: upgrade-scripts
          mountPath: c:/scripts/
        env:
        - name: HELM_REVISION_NUMBER
          value: "1"
      volumes:
      - name: wins
        hostPath:
          path: c:/etc/rancher/wins
      - name: upgrade-scripts
        configMap:
          name: wins-upgrader-default
```

Related Issue: https://github.com/rancher/rancher/issues/28327


packages/rancher-wins-upgrader/charts/values.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+global:
+  cattle:
+    systemDefaultRegistry: ""
+
+prefixPath: c:\\
+
+## Contains one or more configurations for the wins server to be applied across all of the nodes based on the nodeSelector and tolerations provided
+##
+winsConfigs:
```
aiyengar2, line 15: takes in multiple wins configs, so you can limit the wins server whitelist and versions across sets of nodes.

This way you aren't forced to have one universal wins config or wins server version across all nodes, if you so wish to do so, and can configure everything based on nodeSelectors and tolerations.

packages/rancher-wins-upgrader/charts/values.yaml
```diff
@@ -0,0 +1,40 @@
...
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+# Configuration
+
+global:
+  cattle:
+    systemDefaultRegistry: ""
+
+prefixPath: c:\\
```
aiyengar2, line 11: mimics exactly what we do in bootstrapping

packages/rancher-wins-upgrader/charts/values.yaml
```diff
@@ -0,0 +1,40 @@
...
+    # TODO(aiyengar2): replace with an image that just contains wins
+    repository: rancher/wmi_exporter-package
+    tag: v0.0.4
+  config: |
+    debug: false
+    listen: rancher_wins
+    proxy: rancher_wins_proxy
+    white_list: 
+      process_paths:
+      - {{ .Values.prefixPath }}etc\wmi-exporter\wmi-exporter.exe
```
aiyengar2, line 27: takes in a go template so we can support prefixPath based configs

packages/rancher-wins-upgrader/charts/values.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+prefixPath: c:\\
+
+## Contains one or more configurations for the wins server to be applied across all of the nodes based on the nodeSelector and tolerations provided
+##
+winsConfigs:
+- name: default
+  image:
+    # TODO(aiyengar2): replace with an image that just contains wins
+    repository: rancher/wmi_exporter-package
```
aiyengar2, line 19: We don't currently have a Dockerfile that tracks different packaged versions of wins, but this needs to be created. 

This should pretty much be `https://github.com/rancher/wmi_exporter-package/blob/master/package/windows/Dockerfile` + `https://github.com/rancher/wmi_exporter-package/blob/master/package/windows/Dockerfile.20H2` maintained in the `rancher/wins` repository after dropping the `$BINARY_URL` + `COPY --from=download /wmi-exporter.exe /etc/wmi-exporter/` steps from it and the final command.

We'll also need to copy over https://github.com/rancher/wmi_exporter-package/blob/master/.drone.yml.

@luthermonson do you have any thoughts about this?

packages/rancher-wins-upgrader/charts/templates/daemonset.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - name: wins-upgrader
+        image: {{ template "system_default_registry" $ }}{{ .image.repository }}:{{ .image.tag }}
+        command: ["pwsh", "-f", "c:/scripts/upgrade.ps1"]
+        volumeMounts:
+        - name: wins
+          mountPath: c:/host/etc/rancher/wins
+        - name: upgrade-scripts
+          mountPath: c:/scripts/
+        env:
+        - name: HELM_REVISION_NUMBER
```
aiyengar2, line 35: This is a hack that forces helm to redeploy the DaemonSet on a helm upgrade since the revision number is always modified 😄

packages/rancher-wins-upgrader/charts/scripts/upgrade.ps1
```diff
@@ -0,0 +1,44 @@
...
+            return
+        }
+    }
+
+    $null = Copy-Item -Force -Path $Src -Destination $Dst
+}
+
+Create-Directory -Path "c:\host\etc\rancher\wins"
+Transfer-File -Src "c:\scripts\config" -Dst "c:\host\etc\rancher\wins\config"
+Transfer-File -Src "c:\Windows\wins.exe" -Dst "c:\host\etc\rancher\wins\wins.exe"
```
aiyengar2, line 41: `c:\Windows/wins.exe` will need to be modified based on the path within the Docker image that should be created for this based on discussion in https://github.com/rancher/charts/pull/996/files#r575614057.

Currently, this is set based on the location of wins.exe provided in https://github.com/rancher/wmi_exporter-package/blob/8e0c309b986cfbfb96ec1dba0a40df612f149f36/package/windows/Dockerfile#L33.

packages/rancher-wins-upgrader/charts/scripts/upgrade.ps1
```diff
@@ -0,0 +1,44 @@
...
+        [parameter(Mandatory = $true)] [string]$Dst
+    )
+
+    if (Test-Path -PathType leaf -Path $Dst) {
+        $dstHasher = Get-FileHash -Path $Dst
+        $srcHasher = Get-FileHash -Path $Src
+        if ($dstHasher.Hash -eq $srcHasher.Hash) {
+            return
+        }
+    }
```
nickgerace, line 34: What does this check ensure?

packages/rancher-wins-upgrader/charts/values.yaml
nickgerace: This requires one at minimum, right?

nickgerace (dismissed):

nickgerace:
The noop container is a good idea for the upgrader

packages/rancher-wins-upgrader/charts/values.yaml
aiyengar2: yup, will add the fix. Good call.

packages/rancher-wins-upgrader/charts/scripts/upgrade.ps1
aiyengar2: It's boilerplate code from our other windows stuff, but essentially it only performs the transfer if the hash is different between the files (e.g. if its the same file, no need to copy).

packages/rancher-wins-upgrader/charts/README.md
```diff
@@ -0,0 +1,41 @@
...
+```powershell
+(Get-CimInstance Win32_Service -Filter 'Name = "rancher-wins"').PathName
+```
+
+Once complete, restart the service:
+```powershell
+Restart-Service -Name "rancher-wins" | Stop-Service
+```
+
+### Masquerading (Use at your own risk. Here be dragons...)
```
aiyengar2, line 37: @luthermonson @nickgerace we should discuss whether this is something we want to expose in this chart. The idea behind it is to provide a way for existing windows clusters to masquerade the first payload for wins under another processes's name, which allows Windows clusters who temporarily disable a service like Monitoring to get the wins upgrader's foot through the door.

Worked on an existing Windows cluster with no changes to the bootstrap process when I tried it out.

aiyengar2:
Closing this PR out in favor of https://github.com/rancher/wins/pull/26

PR #997, opened 2021-02-13
Add rancher-windows-exporter

aiyengar2:
Forward ports a newer version of exporter-node-windows from Monitoring V1.

The charts are content-wise the same but structured very differently, so no base copy commit was added.

Blocked on creation of image:
- [x] Merge https://github.com/rancher/windows_exporter-package/pull/1
- [x] Cut release for wins v0.1.0
- [x] Ensure that rancher/windows_exporter-package:v0.0.1 exists
- [x] Merge https://github.com/rancher/windows_exporter-package/pull/2

Related Issue: https://github.com/rancher/rancher/issues/28327, https://github.com/rancher/rancher/issues/31497

### DIFF
--- a/packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
+++ b/packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
@@ -1,6 +1,6 @@
 --- charts-original/values.yaml
 +++ charts/values.yaml
-@@ -2,13 +2,283 @@
+@@ -2,13 +2,291 @@
  # This is a YAML-formatted file.
  # Declare variables to be passed into your templates.
  
@@ -24,7 +24,7 @@
 +    create: true
 +
 +## RKE PushProx Monitoring
-+## ref: https://github.com/rancher/charts/tree/master/packages/rancher-pushprox
++## ref: https://github.com/rancher/charts/tree/dev-v2.5-source/packages/rancher-pushprox
 +##
 +rkeControllerManager:
 +  enabled: false
@@ -89,8 +89,16 @@
 +    - effect: "NoSchedule"
 +      operator: "Exists"
 +
++## Windows Monitoring (note: currently RKE1 only)
++## ref: https://github.com/rancher/charts/tree/dev-v2.5-source/packages/rancher-windows-exporter
++## Runs https://github.com/prometheus-community/windows_exporter as a DaemonSet
++## Relies on the existence of a wins server of version v0.1.0+ on every Windows host to allow 
++## windows_exporter to run as a host process that can publish host metrics to a port on the Pod
++windowsExporter:
++  enabled: true
++
 +## k3s PushProx Monitoring
-+## ref: https://github.com/rancher/charts/tree/master/packages/rancher-pushprox
++## ref: https://github.com/rancher/charts/tree/dev-v2.5-source/packages/rancher-pushprox
 +##
 +k3sServer:
 +  enabled: false
@@ -106,7 +114,7 @@
 +      operator: "Exists"
 +
 +## KubeADM PushProx Monitoring
-+## ref: https://github.com/rancher/charts/tree/master/packages/rancher-pushprox
++## ref: https://github.com/rancher/charts/tree/dev-v2.5-source/packages/rancher-pushprox
 +##
 +kubeAdmControllerManager:
 +  enabled: false
@@ -175,7 +183,7 @@
 +      operator: "Exists"
 +
 +## rke2 PushProx Monitoring
-+## ref: https://github.com/rancher/charts/tree/master/packages/rancher-pushprox
++## ref: https://github.com/rancher/charts/tree/dev-v2.5-source/packages/rancher-pushprox
 +##
 +rke2ControllerManager:
 +  enabled: false
@@ -286,7 +294,7 @@
  
  ## Provide a k8s version to auto dashboard import script example: kubeTargetVersionOverride: 1.16.6
  ##
-@@ -76,8 +346,23 @@
+@@ -76,8 +354,23 @@
  
  ##
  global:
@@ -310,7 +318,7 @@
      pspEnabled: true
      pspAnnotations: {}
        ## Specify pod annotations
-@@ -130,6 +415,22 @@
+@@ -130,6 +423,22 @@
    ## ref: https://prometheus.io/docs/alerting/configuration/#configuration-file
    ##      https://prometheus.io/webtools/alerting/routing-tree-editor/
    ##
@@ -333,7 +341,7 @@
    config:
      global:
        resolve_timeout: 5m
-@@ -145,6 +446,8 @@
+@@ -145,6 +454,8 @@
          receiver: 'null'
      receivers:
      - name: 'null'
@@ -342,7 +350,7 @@
  
    ## Pass the Alertmanager configuration directives through Helm's templating
    ## engine. If the Alertmanager configuration contains Alertmanager templates,
-@@ -160,25 +463,76 @@
+@@ -160,25 +471,76 @@
    ## ref: https://prometheus.io/docs/alerting/notifications/
    ##      https://prometheus.io/docs/alerting/notification_examples/
    ##
@@ -438,7 +446,7 @@
  
    ingress:
      enabled: false
-@@ -208,6 +562,25 @@
+@@ -208,6 +570,25 @@
    ## Configuration for Alertmanager secret
    ##
    secret:
@@ -464,7 +472,7 @@
      annotations: {}
  
    ## Configuration for creating an Ingress that will map to each Alertmanager replica service
-@@ -334,7 +707,7 @@
+@@ -334,7 +715,7 @@
      ## Image of Alertmanager
      ##
      image:
@@ -473,7 +481,7 @@
        tag: v0.21.0
        sha: ""
  
-@@ -410,9 +783,13 @@
+@@ -410,9 +791,13 @@
      ## Define resources requests and limits for single Pods.
      ## ref: https://kubernetes.io/docs/user-guide/compute-resources/
      ##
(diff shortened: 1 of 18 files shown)


nickgerace (approved):

aiyengar2:
Dropping @luthermonson as a reviewer in favor of @brendarearden since this is very Helm chart specific. The wins related stuff can be reviewed by @nickgerace.

Also @nickgerace @brendarearden this PR will be blocked for a merge on the creation of the windows_exporter-package image, but it's ready to be reviewed.

nickgerace:
>  Cut release for wins v0.0.1
@aiyengar2 shouldn't this be v0.1.0?

nickgerace (approved):

aiyengar2:
> Cut release for wins v0.0.1
> @aiyengar2 shouldn't this be v0.1.0?

Good catch. Updated

packages/rancher-windows-exporter/charts/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,35 @@
...
+    matchLabels: {{ include "windowsExporter.labels" . | nindent 6 }}
+  namespaceSelector:
+    matchNames:
+    - {{ template "windowsExporter.namespace" . }}
+  endpoints:
+  - port: windows-metrics
+    metricRelabelings:
+    - sourceLabels: [__name__]
+      regex: 'wmi_(.*)'
+      replacement: 'windows_$1'
```
aiyengar2, line 19: FYI, this comes from https://github.com/prometheus-community/windows_exporter/issues/519#issuecomment-633249466

packages/rancher-windows-exporter/charts/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,35 @@
...
+      targetLabel: __name__
+    - sourceLabels: [__name__]
+      regex: windows_mssql_transactions_active_total
+      replacement: 'windows_mssql_transactions_active'
+      targetLabel: __name__
```
aiyengar2, line 24: FYI this comes from https://github.com/prometheus-community/windows_exporter/releases/tag/v0.15.0

packages/rancher-windows-exporter/charts/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,35 @@
...
+    - sourceLabels: [volume, nic]
+      regex: (.*);(.*)
+      separator: ''
+      targetLabel: device
+      action: replace
+      replacement: $1$2
+    - sourceLabels: [__name__]
+      regex: windows_cs_logical_processors
+      replacement: 'system'
+      targetLabel: mode
```
aiyengar2, line 34: Long-standing relabelings from V1 Monitoring https://github.com/rancher/system-charts/blob/3e8ab13429caede329976f7f556b3cac8b0421cd/charts/rancher-monitoring/v0.2.1/charts/exporter-node-windows/templates/servicemonitor.yaml#L24-L33

packages/rancher-windows-exporter/charts/scripts/proxy-entry.ps1
```diff
@@ -0,0 +1,11 @@
+# default
```
aiyengar2, line 1: This is a new file, unique to Monitoring V2

packages/rancher-windows-exporter/charts/scripts/copy-binary.ps1
```diff
@@ -0,0 +1,40 @@
+$ErrorActionPreference = 'Stop'
```
aiyengar2, line 1: Copied straight from https://github.com/rancher/system-charts/blob/3e8ab13429caede329976f7f556b3cac8b0421cd/charts/rancher-monitoring/v0.2.1/charts/exporter-node-windows/templates/configmap.yaml#L11, this gets placed in a ConfigMap just like before.

Just using helm templating here to make it easier to maintain

packages/rancher-windows-exporter/charts/templates/configmap.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: ConfigMap
+metadata:
+  name: {{ template "windowsExporter.name" . }}-scripts
+  namespace: {{ template "windowsExporter.namespace" . }}
+  labels: {{ include "windowsExporter.labels" . | nindent 4 }}
+data:
+  copy-binary.ps1: |-
+{{ .Files.Get "scripts/copy-binary.ps1" | indent 4 }}
+  proxy-entry.ps1: |-
+{{ .Files.Get "scripts/proxy-entry.ps1" | indent 4 }}
```
aiyengar2, line 11: This part of the file is new, the copy-binary script is a copy from v1 https://github.com/rancher/system-charts/blob/3e8ab13429caede329976f7f556b3cac8b0421cd/charts/rancher-monitoring/v0.2.1/charts/exporter-node-windows/templates/configmap.yaml#L11

aiyengar2 (commented):
Calling out some differences to keep a special eye out for on reviewing this PR.

packages/rancher-windows-exporter/charts/Chart.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+version: 0.1.0
+appVersion: 0.0.4
+annotations:
+  catalog.rancher.io/certified: rancher
+  catalog.rancher.io/namespace: cattle-monitoring-system
```
brendarearden, line 6: Should we be adding request cpu and request memory to all charts?

packages/rancher-windows-exporter/charts/templates/daemonset.yaml
```diff
@@ -0,0 +1,69 @@
...
+        volumeMounts:
+        - name: wins-pipe-proxy
+          mountPath: \\.\pipe\rancher_wins_proxy
+        - name: exporter-scripts
+          mountPath: c:/scripts/
+      - name: exporter-node
+        image: {{ template "system_default_registry" . }}{{ .Values.clients.image.repository }}:{{ .Values.clients.image.tag }}
+{{- if .Values.clients.args }}
+        args: {{ .Values.clients.args }}
+{{ end }}
```
brendarearden, line 38: should this end tag have a dash? it might not because the second if statement on line 1 doesn't have a dash and the closing ends on the last line both have dashes so that might be working out the dashes in the end.

packages/rancher-windows-exporter/charts/scripts/proxy-entry.ps1
```diff
@@ -0,0 +1,11 @@
...
+$listenPort = "9796"
```
brendarearden, line 2: https://github.com/rancher/rancher/issues/28327#issuecomment-754274987 << basedd on this comment I am wondering if users will still have to keep this port reserved for metrics? or how does this setup accomplish that?

brendarearden (commented):
LGTM, but had a few question - will approve after questions resolved.

packages/rancher-windows-exporter/charts/Chart.yaml
aiyengar2: Hmm I don't think we need this here since this chart is never expected to be installed without Monitoring V2 (it will fail since the CRDs aren't installed, similar to `rancher-pushprox`). The CPU / Request memory annotations in Monitoring V2 should be sufficient.

@cbron agree?

packages/rancher-windows-exporter/charts/templates/daemonset.yaml
aiyengar2: Oof good catch. Will update!

packages/rancher-windows-exporter/charts/scripts/proxy-entry.ps1
aiyengar2: Good question! The port on the host will still have to be reserved for `wmi_exporter`; we're only publishing the port on the host via the Pod. 

The difference though is that you will not have to open up the node port on the host to the external world (e.g. in Monitoring V1, --exposes would actually modify the firewall rules on the Windows host itself to allow traffic outside of `localhost` or `127.0.0.1` to send packets to that port). Additionally, you would not need to modify your Cloud Provider based firewall rules (e.g. security groups) to expose that port either.

brendarearden: gotcha!! thank you for the explanation! 😄

packages/rancher-windows-exporter/charts/Chart.yaml
cbron: Yes cpu/mem only at top level chart, unless chart can be deployed separately. It would be cool to define at all levels and aggregate that back up, but not practical.

aiyengar2:
@brendarearden @nickgerace all other relevant PRs are merged and I tested this out with the new wins image; needed to add two minor commits (did not modify any existing commits since you reviewed), but this should be good to go.

Will merge post 1 approval for the changes I introduced.

packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
```diff
@@ -95,7 +95,7 @@
 +## Relies on the existence of a wins server of version v0.1.0+ on every Windows host to allow 
 +## windows_exporter to run as a host process that can publish host metrics to a port on the Pod
 +windowsExporter:
-+  enabled: true
++  enabled: false
```
nickgerace, line 98: We want this to be false now? What's the reasoning?

aiyengar2: by default, it should be false since it is expected it will be deployed in a non-Windows cluster. I'll be creating a UI ticket for Windows to enable this by default automatically only if the cluster is detected to be a Windows cluster

nickgerace (dismissed):
LGTM: seems like we have all the pre-requisites now

brendarearden (approved):
LGTM

aiyengar2:
Force pushing since branch is out-of-date with base branch. No commits have been added to this PR.

nickgerace (approved):
Approving based on https://github.com/rancher/charts/pull/997#issuecomment-799617379

PR #1000, opened 2021-02-18
Fix RKE2 aggregator DaemonSet for rancher-logging

cmurphy:
Without this patch, the rancher-logging chart fails to install when RKE2
is included:

```
$ helm install rancher-logging rancher/rancher-logging --set additionalLoggingSources.rke2.enabled=true
Error: YAML parse error on rancher-logging/templates/loggings/rke2/daemonset.yaml: error converting YAML to JSON: yaml: line 25: did not find expected key
```

This is because the indentation being generated was invalid, which
itself was caused by the invalid use of `tolerations` and `nodeSelector`
under the `containers` object, when these are only Pod attributes. This
change fixes the indendation so that the end result is a valid pod spec
and the chart installs correctly.

https://github.com/rancher/rancher/issues/31372

### DIFF
--- a/packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
+++ b/packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
@@ -23,15 +23,15 @@ spec:
               name: logdir
             - mountPath: /fluent-bit/etc/
               name: config
-          {{- with .Values.tolerations }}
-          tolerations:
-            {{- toYaml . | nindent 6 }}
-          {{- end }}
-          {{- with .Values.nodeSelector }}
-          nodeSelector:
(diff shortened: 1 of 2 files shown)


nickgerace:
Great catch. We might need to change the RC version based on https://github.com/rancher/charts/pull/995 as well.

(by the way, congrats on #1000!)

nickgerace (approved):

cbron:
Looks good, thanks. I guess no one has used this in rke2 yet, which makes sense. @cmurphy can you create a rancher/rancher issue and put into 2.5.6 describing issue, that was QA can track and validate.

paynejacob (approved):

cmurphy:
@cbron sure, here: https://github.com/rancher/rancher/issues/31372

cbron (approved):

PR #1006, opened 2021-02-22
Add back in linux-only toleration

aiyengar2:
This commit addresses a regression introduced in the charts script rewrite as seen in charts-diff [here](https://github.com/aiyengar2/charts-diff/blob/a6535725f6295921bd2c7bc830dad9f28b9a291a/diff/rancher-monitoring.diff#L17).

The annotation needs to be added back in till we get Windows Monitoring out, which is now scheduled for 2.5.7.

Related Issue: https://github.com/rancher/rancher/issues/31421


brendarearden (approved):
LGTM

PR #1010, opened 2021-02-25
Support Grafana dashboards in multiple namespaces across cluster

aiyengar2:
Related Issue: https://github.com/rancher/rancher/issues/31412, https://github.com/rancher/rancher/issues/31769

### DIFF
--- a/packages/rancher-monitoring/generated-changes/overlay/templates/grafana/namespaces.yaml
+++ b/packages/rancher-monitoring/generated-changes/overlay/templates/grafana/namespaces.yaml
@@ -1,10 +1,9 @@
-{{- if and .Values.grafana.enabled }}
-{{- if or .Values.grafana.sidecar.dashboards.enabled .Values.grafana.defaultDashboardsEnabled }}
+{{- if and .Values.grafana.enabled .Values.grafana.defaultDashboardsEnabled (not .Values.grafana.defaultDashboards.useExistingNamespace) }}
 apiVersion: v1
 kind: Namespace
 metadata:
-  name: {{ .Values.grafana.sidecar.dashboards.searchNamespace }}
+  name: {{ .Values.grafana.defaultDashboards.namespace }}
   labels:
-    name: {{ .Values.grafana.sidecar.dashboards.searchNamespace }}
-{{- end }}
+    name: {{ .Values.grafana.defaultDashboards.namespace }}
+{{ include "kube-prometheus-stack.labels" . | indent 4 }}
 {{- end }}
--- a/packages/rancher-monitoring/generated-changes/overlay/templates/grafana/namespaces.yaml
+++ b/packages/rancher-monitoring/generated-changes/overlay/templates/grafana/namespaces.yaml
@@ -6,4 +6,8 @@ metadata:
   labels:
     name: {{ .Values.grafana.defaultDashboards.namespace }}
 {{ include "kube-prometheus-stack.labels" . | indent 4 }}
+  annotations:
+{{- if not .Values.grafana.defaultDashboards.cleanupOnUninstall }}
+    helm.sh/resource-policy: "keep"
+{{- end }}
 {{- end }}
--- a/packages/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/dashboard-role.yaml
+++ b/packages/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/dashboard-role.yaml
@@ -1,10 +1,10 @@
 {{- if and .Values.global.rbac.create .Values.global.rbac.userRoles.create .Values.grafana.enabled }}
-{{- if or .Values.grafana.sidecar.dashboards.enabled .Values.grafana.defaultDashboardsEnabled }}
+{{- if .Values.grafana.defaultDashboardsEnabled }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
   name: monitoring-dashboard-admin
-  namespace: {{ .Values.grafana.sidecar.dashboards.searchNamespace }}
+  namespace: {{ .Values.grafana.defaultDashboards.namespace }}
   labels: {{ include "kube-prometheus-stack.labels" . | nindent 4 }}
 rules:
 - apiGroups:
@@ -18,7 +18,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
   name: monitoring-dashboard-edit
-  namespace: {{ .Values.grafana.sidecar.dashboards.searchNamespace }}
+  namespace: {{ .Values.grafana.defaultDashboards.namespace }}
   labels: {{ include "kube-prometheus-stack.labels" . | nindent 4 }}
 rules:
 - apiGroups:
@@ -32,7 +32,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
   name: monitoring-dashboard-view
-  namespace: {{ .Values.grafana.sidecar.dashboards.searchNamespace }}
+  namespace: {{ .Values.grafana.defaultDashboards.namespace }}
   labels: {{ include "kube-prometheus-stack.labels" . | nindent 4 }}
 rules:
 - apiGroups:
--- a/packages/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/hardened.yaml
+++ b/packages/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/hardened.yaml
@@ -1,6 +1,6 @@
 {{- $namespaces := dict "_0" .Release.Namespace -}}
-{{- if and .Values.grafana.enabled (or .Values.grafana.sidecar.dashboards.enabled .Values.grafana.defaultDashboardsEnabled) -}}
-{{- $_ := set $namespaces "_1" .Values.grafana.sidecar.dashboards.searchNamespace -}}
+{{- if and .Values.grafana.enabled .Values.grafana.defaultDashboardsEnabled (not .Values.grafana.defaultDashboards.useExistingNamespace) -}}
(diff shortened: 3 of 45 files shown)


packages/rancher-monitoring/generated-changes/overlay/templates/grafana/namespaces.yaml
```diff
@@ -1,10 +1,12 @@
 {{- if and .Values.grafana.enabled }}
 {{- if or .Values.grafana.sidecar.dashboards.enabled .Values.grafana.defaultDashboardsEnabled }}
+{{- if or (not .Values.grafana.sidecar.dashboards.useExistingNamespace) (eq .Values.grafana.sidecar.dashboards.searchNamespace "ALL") }}
```
aiyengar2, line 3: There are 3 possible inputs that we expect:
1. searchNamespace = ALL: ignore useExistingNamespace and create the cattle-dashboards namespace by default
2. searchNamespace != ALL && useExistingNamespace=true: don't create a namespace, just create resources in an existing namespace
3. searchNamespace != ALL && useExistingNamespace=false: create the namespace and create resources within it.

Question: Should we also split case 1 based on useExistingNamespace?

The only case in which this would matter is if you already created the cattle-dashboard namespace from another chart / in a way that is not managed by Helm and you want to use that as your existing namespace. But I don't see that as a likely possibility.

vs. it's extremely likely a user would have left useExistingNamespace=true when switching searchNamespace = ALL, which would result in a helm error since the namespace we are creating resources within (default cattle-dashboard) no longer exists.

aiyengar2:
No longer blocked by release, so this PR is ready for review

brendarearden (dismissed):
LGTM

packages/rancher-monitoring/generated-changes/overlay/templates/grafana/namespaces.yaml
aiyengar2: Based on offline discussion with @brendarearden, I'll be making changes to this PR to logically separate searchNamespace from defaultDashboardsNamespace, which will alleviate the problem I brought up here. Will resolve this comment and re-request both reviews once that change is tested / in.

packages/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/hardened.yaml
```diff
@@ -1,6 +1,6 @@
 {{- $namespaces := dict "_0" .Release.Namespace -}}
-{{- if and .Values.grafana.enabled (or .Values.grafana.sidecar.dashboards.enabled .Values.grafana.defaultDashboardsEnabled) -}}
-{{- $_ := set $namespaces "_1" .Values.grafana.sidecar.dashboards.searchNamespace -}}
+{{- if and .Values.grafana.enabled .Values.grafana.defaultDashboards.enabled (not .Values.grafana.defaultDashboards.useExistingNamespace) -}}
```
aiyengar2: Dropping `.Values.grafana.sidecar.dashboards.enabled` was intentional since the Job that is deployed to patch the default service account in the namespace is tied to whether or not we introduce it.

Since we modified the logic for deciding whether or not to deploy the namespace, this one is modified in conjunction with that.

packages/rancher-monitoring/generated-changes/overlay/templates/grafana/namespaces.yaml
```diff
@@ -1,10 +1,9 @@
-{{- if and .Values.grafana.enabled }}
-{{- if or .Values.grafana.sidecar.dashboards.enabled .Values.grafana.defaultDashboardsEnabled }}
+{{- if and .Values.grafana.enabled .Values.grafana.defaultDashboardsEnabled (not .Values.grafana.defaultDashboards.useExistingNamespace) }}
```
aiyengar2, line 1: Don't need to check if the sidecar exists to create the namespace.

packages/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/dashboard-role.yaml
```diff
@@ -1,10 +1,10 @@
 {{- if and .Values.global.rbac.create .Values.global.rbac.userRoles.create .Values.grafana.enabled }}
-{{- if or .Values.grafana.sidecar.dashboards.enabled .Values.grafana.defaultDashboardsEnabled }}
+{{- if .Values.grafana.defaultDashboardsEnabled }}
```
aiyengar2, line 2: Don't need to check if the sidecar exists to create the roles.

packages/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/hardened.yaml
```diff
@@ -1,6 +1,6 @@
 {{- $namespaces := dict "_0" .Release.Namespace -}}
-{{- if and .Values.grafana.enabled (or .Values.grafana.sidecar.dashboards.enabled .Values.grafana.defaultDashboardsEnabled) -}}
-{{- $_ := set $namespaces "_1" .Values.grafana.sidecar.dashboards.searchNamespace -}}
+{{- if and .Values.grafana.enabled .Values.grafana.defaultDashboardsEnabled (not .Values.grafana.defaultDashboards.useExistingNamespace) -}}
```
aiyengar2, line 2: This if statement should be executed if and only if we create a namespace for the default dashboards.

As such, it follow the same logic as the if statement on top of `packages/rancher-monitoring/generated-changes/overlay/templates/grafana/namespaces.yaml`.

packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
```diff
@@ -526,15 +526,25 @@
    ## Deploy default dashboards.
    ##
    defaultDashboardsEnabled: true
```
aiyengar2, line 528: I considered modifying `defaultDashboardsEnabled` to `defaultDashboards.enabled` but I didn't want to introduce a breaking change if a user had previously set `defaultDashboardsEnabled=false`.

packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
```diff
@@ -526,15 +526,25 @@
...
-@@ -530,6 +936,7 @@
+ 
++  # Additional options for defaultDashboards
++  defaultDashboards:
++    namespace: cattle-dashboards
```
aiyengar2, line 532: This is somewhat of a breaking change.

A user who overrode `grafana.sidecar.searchNamespace` before to something non-default (e.g. `my-namespace`) would see the current Helm chart ignore that override on an upgrade (since they would need to modify `grafana.defaultDashboards.namespace` instead now).

However, I think it's highly unlikely anyone modified the name of the default namespace in which dashboards are deployed, so I don't think this is a risky move.

aiyengar2 (commented):
@brendarearden @PennyScissors this PR has been refactored and is ready for a re-review. Left some comments on things to look at.

packages/rancher-monitoring/generated-changes/overlay/templates/grafana/namespaces.yaml
aiyengar2: Changes have been made.

packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
brendarearden: Are there steps that could be documented that a user could follow to recover their dashboards in the event they did the very unlikely thing?

aiyengar2: If a user was just using the default dashboards, then there's nothing to recover since it will be recreated. If they modified default dashboards in place, then that's not supported since they are trying to manage a Helm managed resource outside of Helm.

However, I'm not sure what would happen if they created non-Helm managed resources in a Helm managed namespace resource that then got deleted. I'll try to look into what happens in that case and get back to you.

brendarearden (dismissed):
LGTM

packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
aiyengar2: @brendarearden just tried it out, I think that I need to refactor this PR to ensure the cattle-dashboards namespace never gets deleted unless you entirely uninstall the chart. We should also probably add something in the docs since we currently tell users to place dashboards in cattle-dashboards; as a result, those would get wiped out on a Helm uninstall / reinstall.

aiyengar2: @brendarearden I added the docs to https://github.com/rancher/docs/issues/3125 w.r.t. what you can do to prevent dashboards from being deleted in the current monitoring chart (workaround is adding a Helm resource annotation to the namespace manually) and added a flag to `cleanupOnUninstall` that defaults to false. 

When set, a helm resource policy is set to keep the namespace, which prevents the namespace from being destroyed on a helm uninstall.

aiyengar2:
Added a small annotation `helm.sh/resource-policy: "keep"` gated by `not .Values.grafana.defaultDashboards.cleanupOnUninstall` to the `cattle-dashboards` Namespace resource to ensure that
1) a user who changes the name of the `.Values.grafana.defaultDashboards.namespace` when `.Values.grafana.defaultDashboards.useExistingNamespace=false` does not nuke their non-Helm managed dashboards added to `cattle-dashboards`.
2) a user who uninstalls monitoring does not nuke their non-Helm managed dashboards added to `cattle-dashboards`.

Added docs in https://github.com/rancher/docs/issues/3125 to address this and created an issue in https://github.com/rancher/rancher/issues/31769 to reproduce and test this.

brendarearden (approved):

PennyScissors (approved):

PR #1014, opened 2021-03-01
Drop older rancher-tracing packageVersion and add a more granular script to move asset to released/

aiyengar2:
New script tracked in https://github.com/rancher/charts-build-scripts/pull/19.


scripts/release-asset
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
aiyengar2, line 1: if preferred, I can drop this commit from this PR in favor of getting it reviewed first in https://github.com/rancher/charts-build-scripts/pull/19 before pulling it in with `make docs`.

But since this is a script only utilized by release captains, imo it should be safe to merge. We can always revisit and pull in the latest post-review via `make docs`.

aiyengar2: FYI, we create a `to-be-released` intermediary directory due to the workaround suggested in https://github.com/helm/helm/issues/7363#issuecomment-572369872.

This avoids modifying every single chart's timestamp on `helm repo index --merge` so we only modify the fields of charts that are actually being moved into release.

cbron (approved):
Lgmt, this should remove already released tracing chart.

aiyengar2:
Made a slight change to the script to remove the charts/ directory for the already released chart as well.

cbron (approved):

PR #1017, opened 2021-03-02
Bump rancher-operator to release

thedadams:
Issue:
https://github.com/rancher/rancher/issues/30888


packages/rancher-operator-crd/package.yaml
```diff
@@ -1,3 +1,3 @@
-url: https://github.com/rancher/rancher-operator/releases/download/v0.1.3-rc7/rancher-operator-crd-0.1.3-rc7.tgz
-packageVersion: 00
-releaseCandidateVersion: 07
+url: https://github.com/rancher/rancher-operator/releases/download/v0.1.3/rancher-operator-crd-0.1.3.tgz
+packageVersion: 01
```
cbron, line 2: I would do packageVersion 00

packages/rancher-operator-crd/package.yaml
```diff
@@ -1,3 +1,3 @@
...
+releaseCandidateVersion: 00
```
cbron, line 3: We just discovered we have to do +1 on previous RC for this project. See: https://github.com/rancher/charts/pull/1016/files#diff-e92b26bfc5708ae53b7059830653831cad111e8a2b5e21df7126172b654a3af6R3

thedadams: Done

cbron (approved):
need to pull in upstream, then lg

cbron (approved):

PR #1018, opened 2021-03-03
Add vsphere-cpi chart from helm3-charts

aiyengar2:
Ports vsphere-cpi post merging https://github.com/rancher/helm3-charts/pull/62


packages/vsphere-cpi/charts/templates/vsphere-cpi-ds.yaml
```diff
@@ -0,0 +1,74 @@
...
+        runAsUser: 1001
+      tolerations:
+      - effect: NoExecute
+        operator: Exists
+      - effect: NoSchedule
+        operator: Exists
+      serviceAccountName: cloud-controller-manager
+      containers:
+        - name: vsphere-cloud-controller-manager
+          image: gcr.io/cloud-provider-vsphere/cpi/release/manager:v1.2.1
```
aiyengar2, line 42: @PennyScissors which image should be replaced here? Can you also reference the image-mirror PR in this comment?

packages/vsphere-cpi/charts/README.md
```diff
@@ -0,0 +1,56 @@
...
+More information on managing Secrets using kubectl [here](https://kubernetes.io/docs/tasks/configmap-secret/managing-secret-using-kubectl/).
+
+## Migration
+
+If using this chart to migrate volumes provisioned by the in-tree provider to the out-of-tree CPI + CSI, you need to taint all nodes with the following:
+```
+node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule
+```
+
+A script [taints.sh](taints.sh) has been added for your convenience, and it can be executed with the following command:
```
aiyengar2, line 51: This link is currently broken when viewed in the Cluster Explorer UI. Do we need to fix this link to point to some external URL? 

My suggestion is to point it to `https://github.com/rancher/charts/blob/dev-v2.5-source/packages/vsphere-cpi/charts/taints.sh`, but it might make sense to place this in some external repository instead

packages/vsphere-cpi/charts/templates/vsphere-cpi-ds.yaml
cbron: This has to be moved to values.yaml in repository:tag format.

packages/vsphere-cpi/charts/Chart.yaml
```diff
@@ -0,0 +1,16 @@
...
+description: vSphere CPI driver
+sources:
+  - https://github.com/kubernetes/cloud-provider-vsphere
+icon: file://logo.svg
+maintainers:
+  - name: Rancher
+    email: caleb@rancher.com
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/namespace: kube-system
```
cbron, line 14: kube-system ? is that right ?

packages/vsphere-cpi/charts/Chart.yaml
```diff
@@ -0,0 +1,16 @@
...
+  catalog.cattle.io/release-name: vsphere-cpi
+  catalog.cattle.io/os: linux
```
cbron, line 16: Need `display-name`. 
Also  `catalog.cattle.io/scope: management` if I understand this feature correctly. Shouldn't be able to deploy it downstream.

packages/vsphere-cpi/charts/questions.yaml
```diff
@@ -0,0 +1,44 @@
+rancher_min_version: 2.5.6-rc1
+rancher_max_version: 2.5.6-rc99
```
cbron, line 2: I'm not familiar with putting this inside questions.yml, what is goal for this ?

packages/vsphere-cpi/charts/templates/vsphere-cpi-ds.yaml
```diff
@@ -0,0 +1,74 @@
...
+  labels:
+    component: cloud-controller-manager
+    tier: control-plane
+    k8s-app: vsphere-cloud-controller-manager
+  name: vsphere-cloud-controller-manager
+  namespace: {{ .Release.Namespace }}
+spec:
+  selector:
+    matchLabels:
+      k8s-app: vsphere-cloud-controller-manager
```
cbron, line 16: Technically we should be limiting this to linux only with tolerations, but if it is local cluster only then we could skip that part.

packages/vsphere-cpi/charts/Chart.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+name: vsphere-cpi
+version: 1.0.0
+appVersion: 1.0.0
+description: vSphere CPI driver
+sources:
+  - https://github.com/kubernetes/cloud-provider-vsphere
+icon: file://logo.svg
+maintainers:
```
cbron, line 9: Ideally we would add a Keyword with 'Infrastructure'

packages/vsphere-cpi/charts/values.yaml
```diff
@@ -0,0 +1,11 @@
+vCenter:
+  host: ""
+  insecureflag: "1"
+  port: "443"
```
cbron, line 4: This is alarming, but looks like its only used in conf files.

packages/vsphere-cpi/charts/Chart.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
```
cbron, line 1: Do we need to define a minimum k8s version ? Is it not compatible with <1.20 ?

packages/vsphere-cpi/charts/templates/vsphere-cpi-rbac.yaml
```diff
@@ -0,0 +1,127 @@
...
+  verbs:
+  - create
+  - patch
+  - update
+- apiGroups:
+  - ""
+  resources:
+  - nodes
+  verbs:
+  - '*'
```
cbron, line 26: Does this come straight from upstream ? Seems heavy. And then nodes/status only gets patch.

packages/vsphere-cpi/charts/templates/vsphere-cpi-rbac.yaml
```diff
@@ -0,0 +1,127 @@
...
+- apiGroups:
+  - ""
+  resources:
+  - services/status
+  verbs:
+  - patch
+- apiGroups:
+  - ""
+  resources:
+  - serviceaccounts
```
cbron, line 51: Interesting that this SA can create other SA's

packages/vsphere-cpi/charts/Chart.yaml
cbron: apparently yes

packages/vsphere-cpi/charts/Chart.yaml
cbron: So this should change to downstream

packages/vsphere-cpi/charts/README.md
aiyengar2: Tracking in https://github.com/rancher/charts/pull/1019. We will use the charts.rancher.io released asset.

packages/vsphere-cpi/charts/questions.yaml
aiyengar2: @PennyScissors can chime in but I believe we added this to block releasing this Helm chart to users before the release was about to happen.

Do we need any limits on rancher_min_version and rancher_max_version for this chart as a feature chart? I think the answer is no, but I want to confirm.

packages/vsphere-cpi/charts/Chart.yaml
aiyengar2: @PennyScissors correct me if I'm wrong, but this should be deployable within the local cluster and downstream clusters right? Since both local cluster and downstream cluster can be vSphere clusters

cbron: If thats case then we don't add annotation at all.

packages/vsphere-cpi/charts/templates/vsphere-cpi-ds.yaml
aiyengar2: Addressed in latest, resolving this comment in favor of a new one that points to the right location that needs to be updated.

packages/vsphere-cpi/charts/values.yaml
```diff
@@ -0,0 +1,23 @@
...
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+global:
+  cattle:
+    systemDefaultRegistry: ""
+
+image:
+  repository: gcr.io/cloud-provider-vsphere/cpi/release/manager
+  tag: v1.2.1
```
aiyengar2, line 11: @PennyScissors which image should be replaced here? Can you also reference the image-mirror PR in this comment?

packages/vsphere-cpi/charts/templates/vsphere-cpi-rbac.yaml
dramich: Yes, see comment at the top of this file.

packages/vsphere-cpi/charts/Chart.yaml
aiyengar2: Addressed display name in PR, still need confirmation w.r.t. scope annotation

packages/vsphere-cpi/charts/templates/vsphere-cpi-ds.yaml
aiyengar2: We don't care about windows. Chart.yaml annotations specifically call this out as a linux only chart

packages/vsphere-cpi/charts/Chart.yaml
aiyengar2: added

packages/vsphere-cpi/charts/templates/vsphere-cpi-rbac.yaml
aiyengar2: Comes from upstream as well, so I don't think we need to address it

packages/vsphere-cpi/charts/questions.yaml
PennyScissors: This is to hide the charts because helm3-charts is an external facing repo and QA needs to validate them first. AFAIK, feature charts don't use this, so if that's the case we can just remove it. 

@deniseschannon You mentioned you wanted the vsphere charts to be available for v2.5.6+, so I figured you should be aware of this.

cbron: Apparently this doesn't work in vue at all, so lets just remove. We haven't added min/max to feature charts yet in general

packages/vsphere-cpi/charts/values.yaml
PennyScissors: Yes, this is only for the vsphere's cloud-config file.

packages/vsphere-cpi/charts/values.yaml
cbron: Yes it needs to be in format: `rancher/mirrored-`.....

packages/vsphere-cpi/charts/Chart.yaml
PennyScissors: Upstream's docs say they are compatible with kubernetes v1.14+

cbron: Below it has `Kubernetes v1.14+`. Not sure where that came from but if true then we are ok.

packages/vsphere-cpi/charts/Chart.yaml
PennyScissors: I'm not really sure about this one. @bmdepesa any ideas?

bmdepesa: It should be deployable on both local and downstream clusters.

packages/vsphere-cpi/charts/questions.yaml
aiyengar2: Updated in latest commit.

packages/vsphere-cpi/charts/values.yaml
aiyengar2: https://github.com/rancher/image-mirror/pull/78 will track creating this image. Updating this PR to use the newly created repo name and tag

aiyengar2:
#### Note for Reviewers

This PR will be entirely rebased after https://github.com/rancher/helm3-charts/pull/62 is merged, which will impact the first commit and every commit that results in a Git conflict after that.

Probably makes more sense to give this PR are second run-through after that has been merged. Will re-request for your reviews at that time.

packages/vsphere-cpi/charts/values.yaml
aiyengar2: https://hub.docker.com/r/rancher/mirrored-cloud-provider-vsphere-cpi-release-manager/tags?page=1&ordering=last_updated exists

aiyengar2:
This PR is ready for review.

This has been split up into a system-charts style PR where the first commit is a base copy (+ adding the package.yaml) and the following commits make the changes needed to get stuff added for the feature charts.

Probably easier to review commit by commit.

packages/vsphere-cpi/charts/Chart.yaml
```diff
@@ -0,0 +1,19 @@
...
+maintainers:
+  - name: Rancher
+    email: caleb@rancher.com
+keywords:
+- infrastructure
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/namespace: kube-system
+  catalog.cattle.io/release-name: vsphere-cpi
+  catalog.cattle.io/display-name: "vSphere Cloud Provider Interface"
```
deniseschannon, line 18: vSphere Cloud Provider Interface (CPI)

aiyengar2: Updated.

packages/vsphere-cpi/charts/values.yaml
```diff
@@ -0,0 +1,20 @@
+vCenter:
+  host: ""
+  port: 443
+  insecureFlag: "1"
```
cbron, line 4: I swear this was true a second ago.

packages/vsphere-cpi/package.yaml
```diff
@@ -0,0 +1,3 @@
+url: local
+packageVersion: 00
+releaseCandidateVersion: 00
```
cbron, line 3: Technically this falls to the whole 1.0.000 semver problem, but I think its ok functionality wise.

cbron (approved):

packages/vsphere-cpi/charts/values.yaml
PennyScissors: it was, we reverted it

aiyengar2: https://github.com/rancher/helm3-charts/pull/62#discussion_r586912945

PennyScissors (approved):

packages/vsphere-cpi/charts/values.yaml
cbron: I think it was other PR

packages/vsphere-cpi/package.yaml
aiyengar2: The semver problem here was that 1.0.000 would be interpreted as 1.0.0 right? I think that actually works to our favor here since, for local charts, Rancher shouldn't be using packageVersion at all (since we should just bump the patch version on each of our changes)

cbron: That or the compare would error out. I haven't seen that actually fail though.

PR #998, opened 2021-02-13
Add windows nodeSelector and tolerations to rancher-pushprox

aiyengar2:
Related Issue: https://github.com/rancher/rancher/issues/28327, https://github.com/rancher/rancher/issues/31498

PR #1028, opened 2021-03-04
Remove values.yaml.orig from released logging chart

aiyengar2:
Currently, dev-v2.5-source is blocked from merging since all workflows are failing to validate the rancher-logging chart due to this bug introduced in release (e.g. https://github.com/rancher/charts/pull/998).

This just removes a file that was never intended to be released in the first place and should have 0 impact on the actual functionality of the chart (since the .orig version of the file is never referenced by Helm and, as seen in https://github.com/aiyengar2/charts-diff/blob/release-v2.5.6/diff/rancher-logging.diff, the values.yaml that was released was what was expected).

Probably related to https://github.com/rancher/charts-build-scripts/issues/6, but more investigation is required to diagnose why this happened in the first place. This seems to be a bug in the release workflow from Staging to Live.


paynejacob (approved):

charts/rancher-logging/rancher-logging/3.9.000/values.yaml.orig
```diff
@@ -1,82 +0,0 @@
-# Default values for logging-operator.
```
aiyengar2, line 1: I'm not sure why the make sync workflow generated this file, but this could be related to https://github.com/rancher/charts-build-scripts/issues/6. We need to investigate this further.

In future, I think a quick sanity-check workflow would be to use the scripts in https://github.com/aiyengar2/charts-diff/tree/release-v2.5.6 (that I just built) to be able to double check that the assets being released directly match the contents developers pushed to source, except for known changes due to the scripts (e.g. dropping rcVersions).

Based on the contents I generated, it seems like this was the only miss from the release workflow.

cbron: Would be nice to include that as part of CI, although once it happens its hard for anyone outside the charts-build-scripts devs to fix it.

aiyengar2: Good idea. Created a blanket issue here to track this effort https://github.com/rancher/charts-build-scripts/issues/21

cbron (approved):
Looks like this is not included in the asset itself which is good. 
Would be nice to figure out why this happened, maybe that last round of changes we did where Ryan ctrl-c'd. But if not we'll just look out for it in future.

PR #1032, opened 2021-03-08
drop intermediate log file for rke and rke2

paynejacob:

PROBLEM:

Currently we write rke and rke2 logs through an intermediate file on the host.  This creates a headache for selinux deployments.  

SOLUTION:

Write directly to the fluentd service.  This alleviates the need for the intemediate file


nickgerace (dismissed):
LGTM, but are there any drawbacks? Why did we not take this approach in the first place? Good idea overall

paynejacob:
@nickgerace The only drawback is this solution is more fragile than what we had before.  If banzai changes their service name or how they route tags it could break rke  and rke2.  We will need to consider this when testing new versions.

cmurphy (dismissed):
lgtm

nickgerace (dismissed):

nickgerace (approved):

PR #1008, opened 2021-02-24
Create default ClusterRole for accessing Monitoring UIs

aiyengar2:
This commit adds the minimal permissions required for users to be able to access the various UIs for accessing Prometheus, Grafana, and Alertmanager via a kubectl proxy and access the Monitoring Pane.

Related Issue: https://github.com/rancher/rancher/issues/31411

PR #1036, opened 2021-03-15
Give monitoring-ui-view full permissions on service/proxy

aiyengar2:
Followup for https://github.com/rancher/charts/pull/1008.

Gives * permissions instead of get permissions so that a user can make POST and other requests as well.

Related Issue: https://github.com/rancher/rancher/issues/31411


brendarearden (approved):
makes me nervous, but LGTM

PR #1039, opened 2021-03-17
Add drone initial tests

brendarearden:
**Issue**
https://github.com/rancher/rancher/issues/28630


recena:
Closing since we are not using [Drone](https://www.drone.io/) anymore.

PR #1040, opened 2021-03-17
add rke2 cis 1.6 profile and benchmark

MonzElmasry:
This pr adds the necessary profiles and benchmark to rancher-cis-benchmark chart to support rke2 CIS-1.6
Issue: https://github.com/rancher/rke2/issues/350


briandowns (dismissed):
LGTM

packages/rancher-cis-benchmark/charts/templates/configmap.yaml
```diff
@@ -8,7 +8,9 @@ data:
   rke: |-
     <1.16.0: rke-profile-permissive-1.5
     >=1.16.0: rke-profile-permissive-1.6
-  rke2: "rke2-cis-1.5-profile-permissive"
+  rke2: |-
+    <1.20.0: rke2-cis-1.5-profile-permissive
+    >=1.20.0: rke2-cis-1.6-profile-permissive
```
prachidamle, line 13: The ClusterScanBenchmark above shows  minKubernetesVersion: "1.18.0"
But this change here restricts any version <1.20.0 to rke2-cis-1.5-profile-permissive - why? Should the `minKubernetesVersion` be "1.20.0" then?

prachidamle (commented):
Also, other changes like new image tag for rancher/security-scan that contains the new rke2 profile needs to be added to the chart. Otherwise if user creates a scan using the rke2-cis-1.6 profile, cis-operator will face errors.

briandowns (dismissed):
LGTM

prachidamle:
We need to update the security-scan image tag https://github.com/rancher/charts/blob/dev-v2.5-source/packages/rancher-cis-benchmark/charts/values.yaml#L9-L11

davidnuzik:
Are we all set @prachidamle @briandowns ? Just needs an approval again?

briandowns (approved):

galal-hussein (approved):
LGTM

cbron:
Also we need at least one rancher dev to approve PR's like this, looks like no one has. Without that there will be process missed that RKE2 devs are unfamiliar with.

packages/rancher-cis-benchmark/package.yaml
```diff
@@ -1,6 +1,6 @@
 url: local
-packageVersion: 00
-releaseCandidateVersion: 01
+packageVersion: 02
+releaseCandidateVersion: 00
```
prachidamle, line 3: @aiyengar2 @cbron We are changing rancher-cis-benchmark chart to release version 1.0.4 in this release 2.5.8 - Are the value for packageVersion and releaseCandidateVersion correct?

davidnuzik:
@cbron my bad I told @MonzElmasry she could merge with the 2 LGTMs. I will keep this in mind going forward

packages/rancher-cis-benchmark/package.yaml
prachidamle: there is no upstream chart for CIS v2, it is a local or rancher original chart.

prachidamle: There are multiple changes coming in for the rancher-cis-benchmark chart in this release, hence I wanted to check what should be the values for these. Will each PR need to change the packageVersion?

aiyengar2: We should only bump the packageVersion when absolutely necessary IMO. If it's possible to bump the patch version instead of the packageVersion, let's do that.

Some reasons why we cannot bump the patch version include:
- The chart is forked from an upstream, so we don't have control over the patch version
- We want to keep the chart version in line with the appVersion

For CIS, I think the latter case applies, so it may make sense to bump the packageVersion instead of the chart version's patch version.

prachidamle: ok so if multiple PRs are going in for the new appVersion version to be released in 2.5.8 (1.0.4 is to be released), should each PR change the packageVersion? or do we need one final packageVersion value?

aiyengar2: To be clear, we only need to bump the packageVersion once per release though. Once you bump it once in a release, you should always only bump the releaseCandidateVersion in any future changes.

aiyengar2: Just talked to @prachidamle offline about this. Since the latest version released today in https://github.com/rancher/charts/tree/release-v2.5/charts/rancher-cis-benchmark/rancher-cis-benchmark is 1.0.301 and we have already bumped the chart version to 1.0.4, the package.yaml should look like:
```
packageVersion: 00
releasedCandidateVersion: 0X
```

Where we always bump the releasedCandidateVersion only within the context of this release.

This would result in the final shipped package being 1.0.400, which is exactly what we want.

prachidamle: @MonzElmasry so we should change this to be:
```
packageVersion: 00
releasedCandidateVersion: 02
```

cbron: And then 1.0.400 is what we put in the table of that github issue

MonzElmasry: okay I'll create another pr that changes the versions.

prachidamle: @aiyengar2 Do we need to delete the pushed chart artifacts or will a new PR correcting the ```releasedCandidateVersion: 02``` regenerate them?

MonzElmasry: but should we manually delete the pushed chart ? @aiyengar2 @cbron

PR #1041, opened 2021-03-17
Bump kubectl and agent versions across feature charts

aiyengar2:
Bumps images to the latest available today.

Modifies rancher-monitoring, rancher-gatekeeper, and rancher-cis-benchmark.

Related Issue: https://github.com/rancher/rancher/issues/31729

### DIFF
--- a/packages/rancher-cis-benchmark/package.yaml
+++ b/packages/rancher-cis-benchmark/package.yaml
@@ -1,6 +1,6 @@
 url: local
-packageVersion: 01
-releaseCandidateVersion: 01
+packageVersion: 00
+releaseCandidateVersion: 00
 additionalCharts:
   - workingDir: charts-crd
     crdOptions:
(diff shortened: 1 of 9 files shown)


brendarearden (approved):
LGTM

packages/rancher-cis-benchmark/package.yaml
```diff
@@ -1,6 +1,6 @@
 url: local
-packageVersion: 01
-releaseCandidateVersion: 01
+packageVersion: 00
```
prachidamle, line 2: packageVersion also decreases to 00?

aiyengar2: I dropped the packageVersion to 00 in favor of just bumping the Chart version from 1.0.3 to 1.0.4.

packageVersion == rancherPatchVersion, so for local charts (where we have control over the patch number) it just makes more sense to bump the real patch version itself.

prachidamle (approved):

PR #1046, opened 2021-03-18
Add liveness probe to fluentd

### LINKED ISSUE #1043: [rancher-logging] Include liveness check for fluentd
As per https://github.com/banzaicloud/logging-operator/issues/579#issuecomment-726698590 the fluentd process can die while the container remains running, which causes log shipping to fail.

The [FluentdSpec](https://banzaicloud.com/docs/one-eye/logging-operator/configuration/crds/v1beta1/fluentd_types/) includes options to define a LivenessProbe, which we should include in all of the Logging objects installed by the chart.  The default liveness check looks at whether the buffers have changed in the last 5 minutes.  It might be cleaner to just check that the 24240 port is open with a socket check.

dbason:
Fluentd process can silently die, leaving the container running.  Have added a liveness probe to check that the Fluentd port is open and listening

Fixes #1043

### DIFF
--- a/packages/rancher-logging/generated-changes/overlay/templates/loggings/aks/logging.yaml
+++ b/packages/rancher-logging/generated-changes/overlay/templates/loggings/aks/logging.yaml
@@ -52,4 +52,7 @@ spec:
     {{- with .Values.fluentd.resources }}
     resources: {{- toYaml . | nindent 6 }}
     {{- end }}
+    {{- with .Values.fluentd.livenessProbe }}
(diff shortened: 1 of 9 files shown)


nickgerace (approved):
LGTM. Great change! :)

paynejacob (dismissed):

dbason:
Are we able to merge this (I don't have permission)?

nickgerace (approved):

true64gurus:
I am running into this issue . Any work around until fixed.

PR #1049, opened 2021-03-19
added missing mounts for journal logs for rke2

paynejacob:

PROBLEM:
we were not collecting any journal logs from rke2 nodes

SOLUTION:
mount the log location into fluentbit


packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
```diff
@@ -21,6 +21,8 @@ spec:
           volumeMounts:
             - mountPath: /fluent-bit/etc/
               name: config
+            - mountPath: /run/log/journal
+              name: journal
```
cmurphy, line 25:
```suggestion
              name: journal
              readOnly: true
```

nickgerace (approved):
LGTM, but with @cmurphy's suggested change

cmurphy (changes requested):
I proposed https://github.com/rancher/rke2-selinux/pull/14/commits/0a2d08b38c367b148281c57c585fd662dbffc79c but this change will also need to add seLinuxOptions to this daemonset in order to use it, similar to the containers aggregator daemonset https://github.com/rancher/charts/blob/184ab063a9cbdc4dedbc11de62d988c67f33f087/charts/rancher-logging/rancher-logging/3.9.002-rc01/templates/loggings/rke2/logging-rke2-containers.yaml#L27-L30

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
```diff
@@ -15,12 +15,20 @@ spec:
       labels:
         name: {{ .Release.Name }}-rke2-journald-aggregator
     spec:
+      {{- if .Values.global.seLinux.enabled }}
+      securityContext:
+        seLinuxOptions:
+          type: rke_logreader_t
+      {{- end }}
```
cmurphy, line 22: nit: this is fine since it's just one container but to be consistent with the rke aggregator it would be nice to put it at the container level rather than the pod level https://github.com/rancher/charts/blob/2420e452723d064580ff839105b25131c5954a86/charts/rancher-logging/rancher-logging/3.9.002-rc01/templates/loggings/rke/daemonset.yaml#L32-L36

paynejacob: I will make the change now, it will avoid a misunderstanding in the future.

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
```diff
@@ -18,9 +18,17 @@ spec:
...
+          securityContext:
+            seLinuxOptions:
+              type: rke_logreader_t
+          {{- end }}
           volumeMounts:
             - mountPath: /fluent-bit/etc/
               name: config
+            - mountPath: /run/log/journal
+              name: journal
+              readOnly: true
```
nickgerace, line 31: Should we wrap this with seLinux.enabled? I'm leaning towards "no", but wanted to ask.

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
```diff
@@ -34,6 +42,9 @@ spec:
         - name: config
           configMap:
             name: "{{ .Release.Name }}-rke2"
+        - name: journal
+          hostPath:
```
nickgerace, line 46: > Should we wrap this with seLinux.enabled? I'm leaning towards "no", but wanted to ask.
Same here

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
paynejacob: no we need to mount it for non se envs too

cmurphy (dismissed):

nickgerace (approved):

PR #1055, opened 2021-03-23
Add kubelet and node-exporter pushprox

### LINKED ISSUE #1051: [rancher-monitoring] Add pushprox for kubelet and node-exporter
Raising RFE for this as per the discussion in https://github.com/rancher/docs/issues/2493#issuecomment-790963731.

Node-exporter runs on the hostNetwork by default so we will need a pushprox to avoid opening ports to all nodes.  Kubelet is also monitored so we need a pushprox for that one.

dbason:
Fixes #1051 

Adds a pushprox for the kubelet and for node-exporter which runs on the host network.

Kubelet ServiceMonitor is still enabled by default, will disable it using values in the dashboard after this is merged.  The node-exporter ServiceMonitor is disabled, and the node-exporter pushprox is enabled by default.

### DIFF
--- a/packages/rancher-monitoring/generated-changes/dependencies/hardenedKubelet/dependency.yaml
+++ b/packages/rancher-monitoring/generated-changes/dependencies/hardenedKubelet/dependency.yaml
@@ -1,0 +1,2 @@
+workingDir: ""
+url: packages/rancher-pushprox
--- a/packages/rancher-monitoring/generated-changes/dependencies/hardenedNodeExporter/dependency.yaml
+++ b/packages/rancher-monitoring/generated-changes/dependencies/hardenedNodeExporter/dependency.yaml
@@ -1,0 +1,2 @@
+workingDir: ""
+url: packages/rancher-pushprox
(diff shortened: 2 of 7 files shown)


aiyengar2 (changes requested):
Do we need to create cluster-specific pushprox options here? Maybe this should just be exposed as hardenedKubelet and hardenedNodeExporter

dbason:
Sure that makes sense.  I wasn't sure so I followed the pattern that was already there (except for the node-exporter).  I can rename the node-exporter one, and move the kubelet to be more generic.

dbason:
@aiyengar2 I've squashed and pushed up the changes now.  I've also rebased in the latest changes from upstream.

abhi1693:
@aiyengar2 @dbason Will this be available in `v2.5.8`?

dbason:
> @aiyengar2 @dbason Will this be available in `v2.5.8`?

@abhi1693 it's missed the window for v2.5.8 which is being finalized at the moment.  Once that release is sorted I'll touch base to try and get this into the next release.

aiyengar2 (changes requested):

dbason:
Closing this in favour of #1221

PR #1064, opened 2021-03-30
Add istio version 1.7.8

brendarearden:
**Problem**
Upgrade for certain applications requires a version of istio >1.7.4 prior to being able to move to 1.8.x

**Solution**
Add istio version 1.7.8. This splits istio into separate versioned directories. 
Uses kiali 1.29.0

**Issue**
https://github.com/rancher/rancher/issues/31811


nickgerace (approved):

packages/rancher-istio-1.7/charts/Chart.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+appVersion: 1.7.6
+description: A basic Istio setup that installs with the istioctl. Refer to https://istio.io/latest/ for details.
+name: rancher-istio
+version: 1.7.6
+icon: https://charts.rancher.io/assets/logos/istio.svg
+keywords:
+- networking
```
aiyengar2, line 8: I'm relatively confused on the structure of this commit, so I think it would be easier to review if we split these up into separate commits.

From my understanding, what you are trying to achieve with this commit is:

#### Step 1: rename rancher-istio to rancher-istio-1.8
`mv packages/rancher-istio packages/rancher-istio-1.8`

#### Step 2: Copy packages/rancher-istio-1.7 from contents of the last commit for 1.7.6

Copy https://github.com/rancher/charts/tree/dd802e825ff1993a6c121874a9248467b20402c6/packages/rancher-istio directly into packages/rancher-istio

Ideally, the version of istio you referenced for this base copy should be contained within the commit message body so we can keep track of exactly what you ported in here.

#### Step 3: Modify rancher-istio-1.7 with some changes

Change the kiali server image version and other things

---

Is this accurate? If so, can you split these commits and drop the kiali-server stuff (that seems like it gets dropped anyways at the end) so that I can re-review just the last commit?

aiyengar2 (changes requested):

aiyengar2 (approved):
LGTM, need to finish up the cleanup step to make it easier for release captains, but that can be done independently of this PR.

packages/rancher-istio-1.7/charts/Chart.yaml
aiyengar2: Thanks @brendarearden ! Commits look great

nickgerace (approved):
Re-reviewed with the new commit structure to make sure I didn't miss anything. LGTM

PR #1067, opened 2021-03-31
Add workflows to keep dev-v2.6 in sync

aiyengar2:
Temporarily modifies our PR workflows to validate that we aren't corrupting changes in the dev-v2.6 branch and modifies our push workflow to also trigger a make sync on dev-v2.6, which will absorb changes from dev-v2.5-source pending https://github.com/rancher/charts/pull/1068.

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -15,7 +15,7 @@ jobs:
       run: sudo make pull-scripts
 
     - name: Pull in all relevant branches
-      run: git fetch origin dev-v2.5-source dev-v2.5 release-v2.5
+      run: git fetch origin dev-v2.5-source dev-v2.5 release-v2.5 dev-v2.6 release-v2.6
 
     - name: Checkout staging branch
       run: git checkout dev-v2.5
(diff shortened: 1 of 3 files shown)


aiyengar2:
Do not merge till https://github.com/rancher/charts/pull/1068 is merged.

.github/workflows/push.yaml
```diff
@@ -15,7 +15,7 @@ jobs:
       run: sudo make pull-scripts
 
     - name: Pull in all relevant branches
-      run: git fetch origin dev-v2.5-source dev-v2.5 release-v2.5
+      run: git fetch origin dev-v2.5-source dev-v2.5 release-v2.5 dev-v2.6 release-v2.6
```
cbron, line 18: Do we need release-v2.6 in there ?

cbron (approved):

.github/workflows/push.yaml
aiyengar2: We can drop it. I just added it to be consistent, but technically we should drop release-v2.5 as well since its not needed

PR #957, opened 2021-01-26
Add PodSecurityPolicies for rancher-istio,rancher-tracing and kiali

bashofmann:
Add PodSecurityPolicies for rancher-istio, rancher-tracing and rancher-kiali-server.

Addresses: rancher/rancher#30744
Rebase of https://github.com/rancher/charts/pull/930

PR #1071, opened 2021-04-01
Add PodSecurityPolicies for rancher-istio,rancher-tracing and kiali 

brendarearden:
https://github.com/rancher/charts/pull/957 << original PR

Opening this pr to complete the need for enablePSP checks on the psp resources per this comment: https://github.com/rancher/charts/pull/957#discussion_r597940236


packages/rancher-istio-1.7/charts/Chart.yaml
```diff
@@ -1,20 +0,0 @@
-apiVersion: v1
```
aiyengar2: Why did we delete these files?

aiyengar2 (changes requested):

packages/rancher-istio-1.7/charts/Chart.yaml
aiyengar2: Clarification offline: We are no longer splitting rancher-istio to 1.7 and 1.8, but we are continuing to rename rancher-istio => rancher-istio-1.8 to facilitate a future release of 1.9

packages/rancher-istio-1.7/charts/Chart.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
```
aiyengar2: @brendarearden let's drop the first commit and refactor the second commit to remove 1.7 changes since 1.7 is no longer being released

aiyengar2 (changes requested):
Need to modify the commits slightly but overall LGTM. Once changes have been made, I'll hit approve.

nickgerace (approved):

packages/rancher-kiali-server/package.yaml
```diff
@@ -1,6 +1,6 @@
 url: https://kiali.org/helm-charts/kiali-server-1.29.0.tgz
-packageVersion: 00
-releaseCandidateVersion: 01
+packageVersion: 01
+releaseCandidateVersion: 00
```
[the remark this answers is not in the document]
cbron: 000 and 001 are both invalid, so this specific change doesn't matter. 
Wouldn't helm complain whether is sub-chart or not ? And CRD chart is top-level chart so that is null point anyway.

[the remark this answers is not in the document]
cbron: If this is a problem we could literally change the kiali version getting installed since the user never see's it I don't think. And then once they release `.1` we just bump the package version.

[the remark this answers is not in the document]
aiyengar2: 000 is valid since it's not a trailing zero iirc. @mattfarina may be able to confirm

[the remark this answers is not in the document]
aiyengar2: Helm won't complain for a subchart, but the CRD chart will definitely cause complaints. It might make sense to bump kiali's patch version to 100 IMO.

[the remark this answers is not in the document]
brendarearden: does this also apply for jaeger? I will get kiali updated.

aiyengar2 (dismissed):
LGTM. I'll approve and leave the call about kiali-server's version in https://github.com/rancher/charts/pull/1071#discussion_r607201990 up to @cbron and @brendarearden

packages/rancher-kiali-server/package.yaml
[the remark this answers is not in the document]
aiyengar2: yes good catch @brendarearden, can we just bump the patch version since it's a local chart? https://github.com/rancher/charts/blob/54da22aa720ef05656768cd0d8767f0e5916a042/packages/rancher-tracing/charts/Chart.yaml#L5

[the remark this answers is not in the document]
cbron: I used Matt's tool to check: 
```
$ semver-isvalid 1.2.000
Illegal leading 0 found in "patch" part
Invalid Semantic Version: Version segment starts with 0. For more information see https://semver.org

$ semver-isvalid 1.2.001
Illegal leading 0 found in "patch" part
Invalid Semantic Version: Version segment starts with 0. For more information see https://semver.org
```

As far as what to do with kiali and jaeger for now. We will have to upgrade helm with a rancher release, and there will likely be a kiali and istio version in that rancher release. Jaeger may be less likely.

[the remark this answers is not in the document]
brendarearden: Okay, I added a commit just for the kiali and jaeger version updates so its ready for -re-review

aiyengar2 (dismissed):
LGTM. Good call keeping the packageVersion at `00` so that we can reserve `major.minor.101` when upstream pushes patch 1 😄

packages/rancher-istio-1.8/charts/templates/istio-cni-psp.yaml
```diff
@@ -0,0 +1,51 @@
...
+metadata:
+  name: psp-istio-cni
+  namespace: {{ template "istio.namespace" . }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: psp-istio-cni
+subjects:
+  - kind: ServiceAccount
+    name: istio-cni
```
cbron, line 35: Where is this service account ? Does the default istio install set that ?

packages/rancher-istio-1.8/charts/templates/istio-cni-psp.yaml
```diff
@@ -0,0 +1,51 @@
...
+      rule: RunAsAny
+  seLinux:
+      rule: RunAsAny
+  supplementalGroups:
+      rule: RunAsAny
+  volumes:
+  - secret
+  - configMap
+  - emptyDir
+  - hostPath
```
cbron, line 22: Going to trust that this is all minimum privilege requirements for various PSP's. Where did these all come from, made up by hand or does istio publish them ?

packages/rancher-istio-1.8/charts/templates/istio-psp.yaml
```diff
@@ -0,0 +1,81 @@
...
+  - kind: ServiceAccount
+    name: istio-operator-authproxy
+  - kind: ServiceAccount
+    name: istiod-service-account
+  - kind: ServiceAccount
+    name: istio-sidecar-injector-service-account
+  - kind: ServiceAccount
+    name: istiocoredns-service-account
+  - kind: ServiceAccount
+    name: default
```
cbron, line 27: Are these from somewhere ? How will we know when a new one is added or existing one updated ?

packages/rancher-istio-1.8/charts/templates/istio-psp.yaml
```diff
@@ -0,0 +1,81 @@
...
+metadata:
+  name: istio-psp
+  namespace: {{ template "istio.namespace" . }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: istio-psp
+subjects:
+  - kind: ServiceAccount
+    name: istio-egressgateway-service-account
```
cbron, line 13: Do all of these deployments from istio already run as non-root ? If not that would violate this policy.

packages/rancher-istio-1.8/charts/templates/istio-uninstall-job.yaml
```diff
@@ -39,4 +39,7 @@ spec:
             name: istio-installer-overlay
         {{ end }}
       serviceAccountName: istio-installer
```
cbron, line 41: So the uninstaller gets the installer PSP linked to it, we'll have to test both.

cbron:
We'll need to tell QA how to test this, with CNI on.

nickgerace (approved):
Reviewing again after Arvind's and Caleb's comments for grammar, etc.

packages/rancher-istio-1.8/charts/templates/istio-cni-psp.yaml
brendarearden: Yes, it is installed by istioctl, we do not need to create an additional one.

packages/rancher-istio-1.8/charts/templates/istio-cni-psp.yaml
brendarearden: Istio does not publish this, but https://banzaicloud.com/blog/istio-operator-psp/ this was used as a blueprint to determine which privileges were needed. CNI PSP has the most privileged permissions of the three PSPs being added.

packages/rancher-istio-1.8/charts/templates/istio-psp.yaml
brendarearden: Each release we will have to determine if a new one was added to the manifests or not. These were pulled from what is being installed in the newer istio versions.

packages/rancher-istio-1.8/charts/templates/istio-psp.yaml
brendarearden: All of the deployments currently run as non-root. There is a global setting in the istio operator to change `runAsNonRoot` but we do not currently set it to false.

packages/rancher-istio-1.8/charts/templates/istio-uninstall-job.yaml
brendarearden: Since its the same image, and the same service account then it makes sense to have one PSP for the installer and uninstaller. But I agree, we will need to test both install and uninstall to ensure all scripts run successfully.

aiyengar2 (approved):

packages/rancher-istio-1.8/charts/templates/istio-psp.yaml
cbron: Ok, we need to add this to a checklist or something to ensure it gets reviewed regularly, definitely with minor releases.

packages/rancher-istio-1.8/charts/templates/istio-cni-psp.yaml
cbron: Ok, we just need to be careful using Banzai's stuff as theirs is diff from upstream.

cbron (dismissed):
lgtm assuming this all works, need to help QA test this and have them try lots of options to try to break it.

brendarearden:
I changed the kiali rc to 01 and updated the auto-install for istio chart.yaml to use that version.

cbron (dismissed):

cbron (approved):

PR #1072, opened 2021-04-01
Rebase Monitoring to 14.5.0 and update all dependencies

aiyengar2:
First commit switches rancher-monitoring to a Git-based package for the next step.

Second commit creates a script that allows commit-by-commit interactive rebases on upstreams for Git-based packages.

Remaining commits were done via scripts in pairs:
- First commit per pair creates generated-changes based on the existing base
- Second commit per pair simply modifies the package.yaml to the new base and regenerates generated-changes on that base

Some fixups were added after the fact to each chart when I tested deploying them and found bugs from the rebase.

Related Issue: https://github.com/rancher/rancher/issues/29871


aiyengar2 (commented):
Since this PR is huge and not reviewable in its original form, I have added 3 commits to make this process way easier to review:

### Third To Last Commit

I ran:

a) `PACKAGE=rancher-monitoring make charts` without *any* changes

b) `mv charts/rancher-monitoring/rancher-monitoring/<VERSION> charts/rancher-monitoring/rancher-monitoring/charts` 

c) `mv charts/rancher-monitoring/rancher-monitoring-crd/<VERSION> charts/rancher-monitoring/rancher-monitoring-crd/charts` 

This represents the state of the chart today.

### Second To Last Commit

Same steps as above once my changes were made. 

This represents the chart that will be shipped in this commit. 

**Reviewers should review this commit for changes since this contains the final diff that will be shipped from this PR. Will update if modifications are requested**

### Last Commit

I removed the `charts/` directory since we do not want to commit that.

charts/rancher-monitoring/rancher-monitoring/charts/charts/grafana/templates/image-renderer-deployment.yaml
```diff
@@ -0,0 +1,115 @@
+{{ if .Values.imageRenderer.enabled }}
+apiVersion: apps/v1
+kind: Deployment
```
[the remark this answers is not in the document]
aiyengar2: Modified the logic for this deployment to use the service account tied to Grafana itself when not specified; therefore, we are using the same PSP for image-renderer as we are for the core Grafana service.

aiyengar2:
Tested deploying in a hardened cluster, a windows cluster, and a normal cluster and successfully deployed all components.

All images and repositories have been created as well.

aiyengar2:
Script has been moved to https://github.com/rancher/charts/pull/1079.

prachidamle:
In upgrade case, doe the deleted CRDs get cleaned up?

aiyengar2:
> @aiyengar2 In upgrade case, doe the deleted CRDs and the CRs get cleaned up? I see a couple of them deleted by the new version?

I believe those should be cleaned up by the CRD chart upgrade

aiyengar2:
Also, I don't think the CRDs are being deleted, but rather renamed into different files that also have modifications. That's why they appear to show up as deletions

prachidamle (dismissed):

charts/rancher-monitoring/rancher-monitoring/charts/Chart.yaml
```diff
@@ -26,6 +81,7 @@ keywords:
 - prometheus
 - kube-prometheus
 - monitoring
+kubeVersion: '>=1.16.0-0'
```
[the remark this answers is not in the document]
brendarearden: https://github.com/rancher/rancher/blob/cc293604c3f189ee26bce7cef659d2d916f0fd1f/pkg/catalogv2/content/content.go#L212 << if you need a rancher version restriction, you could use the rancher-version chart annotation.

charts/rancher-monitoring/rancher-monitoring/charts/Chart.yaml
```diff
@@ -43,4 +99,5 @@ name: rancher-monitoring
 sources:
 - https://github.com/prometheus-community/helm-charts
 - https://github.com/prometheus-operator/kube-prometheus
-version: 9.4.204-rc09
+type: application
+version: 14.5.001-rc00
```
brendarearden: Is this going to be a leading zeros issue?

charts/rancher-monitoring/rancher-monitoring/charts/README.md
```diff
@@ -230,9 +258,100 @@ For more in-depth documentation of configuration options meanings, please see
...
+- [PodMonitors](https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/user-guides/getting-started.md#include-podmonitors)
+- [Running Exporters](https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/user-guides/running-exporters.md)
 
-By default, Prometheus discovers ServiceMonitors within its namespace, that are labeled with the same release tag as the prometheus-operator release. Sometimes, you may need to discover custom ServiceMonitors, for example used to scrape data from third-party applications. An easy way of doing this, without compromising the default ServiceMonitors discovery, is allowing Prometheus to discover all ServiceMonitors within its namespace, without applying label filtering. To do so, you can set `prometheus.prometheusSpec.serviceMonitorSelectorNilUsesHelmValues` to `false`.
+By default, Prometheus discovers PodMonitors and ServiceMonitors within its namespace, that are labeled with the same release tag as the prometheus-operator release.
+Sometimes, you may need to discover custom PodMonitors/ServiceMonitors, for example used to scrape data from third-party applications.
```
brendarearden: nit: This sentence is kind of confusing - mainly the "for example used to scrape" part. maybe something like "Sometimes, you may need to discover custom PodMonitors/ServiceMonitors. An example of this would be using a custom monitor to scrape data from third-party applications."

charts/rancher-monitoring/rancher-monitoring/charts/Chart.yaml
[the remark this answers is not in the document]
aiyengar2: We don't need a rancher version restriction since installing Monitoring V2 is only supported for 2.5+, which supports k8s 1.16+.

charts/rancher-monitoring/rancher-monitoring/charts/README.md
aiyengar2: This wording comes from upstream, so while I agree it's not the best w.r.t. the wording, we aren't maintaining actual patches to the README.md we source from there (except additional fields we expose in Monitoring V2).

charts/rancher-monitoring/rancher-monitoring/charts/Chart.yaml
aiyengar2: Ah yes. I'll follow what you did and patch this to 14.5.100-rc00

charts/rancher-monitoring/rancher-monitoring/charts/README.md
```diff
@@ -230,9 +258,100 @@ For more in-depth documentation of configuration options meanings, please see
...
+    storageClassName: gp2
+    accessModes: ["ReadWriteOnce"]
+    resources:
+     requests:
+       storage: 50Gi
+```
+
+You have to specify matching `volumeClaimTemplate` with 50Gi storage and `ReadWriteOnce` access mode.
+
+Additionally, you should check the current AZ of your legacy installation's PV, and configure the fresh release to use the same AZ as the old one. If the pods are in a different AZ than the PV, the release will fail to bind the existing one, hence creating a new PV.
```
brendarearden: nit: I feel like PV was introduced in prior to this so it has context, but what is AZ? Auth Zone?

aiyengar2: Availability Zone. Also, this falls under the same comment as https://github.com/rancher/charts/pull/1072#discussion_r608872738 in that we're not patching the README.md

charts/rancher-monitoring/rancher-monitoring/charts/Chart.yaml
aiyengar2: addressed in latest commit

packages/rancher-grafana/generated-changes/exclude/ci/with-dashboard-json-values.yaml
```diff
@@ -0,0 +1,53 @@
...
+      # An empty but valid dashboard
+      json: |
+        {
+          "__inputs": [],
+          "__requires": [
+            {
+              "type": "grafana",
+              "id": "grafana",
+              "name": "Grafana",
+              "version": "6.3.5"
```
brendarearden, line 13: JW: The version in the chart for grafana is 6.6.4 - does this version have anything to do that the actual version of grafana?

aiyengar2: Yes, this is the version of Grafana that the dashboards were created for.

But there's no backwards compatibility with Grafana, so these dashboards can still be installed in 6.6.4 even if they are meant for 6.3.5. So I don't think this field has an actual impact on the dashboard

aiyengar2:
Had to push another commit to patch the CRD version to 14.5.1 (for 14.5.100 to be exposed) as well

brendarearden (approved):
this was a huge rebase, well done!!! as far as I could tell everything looks good. You answered all of my questions and it looks like images, versions, and other port or chart.yaml changes were consistent and expected!

PR #1083, opened 2021-04-06
Bump rancher-operator to v0.1.4-rc1

thedadams:


packages/rancher-operator-crd/package.yaml
```diff
@@ -1,3 +1,3 @@
-url: https://github.com/rancher/rancher-operator/releases/download/v0.1.3/rancher-operator-crd-0.1.3.tgz
+url: https://github.com/rancher/rancher-operator/releases/download/v0.1.4-rc1/rancher-operator-crd-0.1.4-rc1.tgz
 packageVersion: 00
-releaseCandidateVersion: 08
+releaseCandidateVersion: 00
```
cbron, line 3: Should be RC1 on both

packages/rancher-operator-crd/generated-changes/patch/Chart.yaml.patch
```diff
@@ -0,0 +1,12 @@
+--- charts-original/Chart.yaml
```
cbron, line 1: FYI @aiyengar2 this will go into 2.5 while 2.6 has the master version, so we will see if they conflict once we merge

aiyengar2: Sounds good. This looks okay to me since they have different major.minor.patch versions; it's identical (from a scripts perspective) to having 2 packages that track the same chart at different major.minor.patch versions.

packages/rancher-operator-crd/package.yaml
thedadams: Fixed

cbron (approved):

PR #1089, opened 2021-04-07
Add patch for rancher-operator 0.2.0-alpha02

thedadams:
Issue:
https://github.com/rancher/rancher/issues/31954

### DIFF
--- a/packages/rancher-operator/package.yaml
+++ b/packages/rancher-operator/package.yaml
@@ -1,3 +1,3 @@
 url: https://github.com/rancher/rancher-operator/releases/download/v0.2.0-alpha02/rancher-operator-0.2.0-alpha02.tgz
-packageVersion: 01
+packageVersion: 02
 releaseCandidateVersion: 01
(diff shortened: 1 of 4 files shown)


packages/rancher-operator/package.yaml
```diff
@@ -1,3 +1,3 @@
 url: https://github.com/rancher/rancher-operator/releases/download/v0.2.0-alpha02/rancher-operator-0.2.0-alpha02.tgz
 packageVersion: 01
-releaseCandidateVersion: 01
+releaseCandidateVersion: 02
```
cbron, line 3: if packageVersion is 01 and rc version is 02, won't that make v0.2.0-alpha01-rc02 ? Should it be alpha02-rc01 ?

cbron: You can run `make charts` to check

thedadams: Yes. Fixed.

cbron (approved):
lgtm, but wait to merge until assets are removed

PR #1075, opened 2021-04-04
[2.5] Drop all Monitoring 9.4.204 releaseCandidateVersions

aiyengar2:
Since we are releasing a new major.minor.patch version of Monitoring V2 in https://github.com/rancher/charts/pull/1072, I ran:

```bash
./scripts/release-asset $(find assets/rancher-monitoring -type f)
```

This moves all of these rcVersions into the `released/` directory, so that it is no longer tracked by CI / sync workflows (for release), but doesn't remove them from access by QA (since a pointer to these versions still exists in `released/assets/rancher-monitoring`)

Related Issue: https://github.com/rancher/rancher/issues/29871

PR #1090, opened 2021-04-07
Remove disfunctional versions of rancher-operator

thedadams:
Issue:
https://github.com/rancher/rancher/issues/31954


cbron (approved):
Looks right to me, we aren't removing 0.2.0-alpha01-rc01 from crd dir in either index.yaml or asset.

aiyengar2 (changes requested):
@thedadams is your goal here to rewrite `rancher-operator-crd-0.2.0-alpha01-rc01` or do you just want to release a `rancher-operator-crd-0.2.0-alpha02-rc00`? If you aren't trying to overwrite the version, you should use the release-asset script instead.

Reasoning behind using the script is to avoid losing the rcVersion, in case we ever want to refer to it in the future.

e.g. see the description in https://github.com/rancher/charts/pull/1075#issue-608481186.

thedadams:
@aiyengar2 There are a couple things I am trying to accomplish with this:
1. Remove the release that has a messed up version (rancher-operator-0.2.0-alpha0201-rc01.tgz). I don't really care to replace it, but it doesn't need to be there.
2. There is a rancher-operator-crd in the rancher-operator directory. This needs to go because it is causing issues.

I also have another PR to get a new version of rancher-operator(-crd) so QA can continue testing: https://github.com/rancher/charts/pull/1089

If I am going about this the wrong way and I should use the script to accomplish these things, please just let me know.

aiyengar2 (approved):
LGTM, make sure we do this in dev-v2.5 as well to keep them in sync.

PR #1092, opened 2021-04-07
Fix Github Push workflow

aiyengar2:
Fixes issues that require restarts due to unclean Git as a result of adding the 2.6 steps by just cleaning up Git entirely before running the 2.6 synchronization.

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -15,7 +15,7 @@ jobs:
       run: sudo make pull-scripts
 
     - name: Pull in all relevant branches
-      run: git fetch origin dev-v2.5-source dev-v2.5 release-v2.5 dev-v2.6 release-v2.6
+      run: git fetch origin dev-v2.5-source dev-v2.5
 
     - name: Checkout staging branch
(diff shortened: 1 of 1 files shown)


.github/workflows/push.yaml
```diff
@@ -15,7 +15,7 @@ jobs:
       run: sudo make pull-scripts
 
     - name: Pull in all relevant branches
-      run: git fetch origin dev-v2.5-source dev-v2.5 release-v2.5 dev-v2.6 release-v2.6
```
aiyengar2, line 18: We never needed release-v2.5, so I'm just removing it here.

.github/workflows/push.yaml
```diff
@@ -29,6 +29,15 @@ jobs:
     - name: Push assets
       run: git push origin dev-v2.5
 
+    - name: (temporary) Clean up Git
+      run: git checkout -- .; git clean -xdf
```
aiyengar2, line 33: Removes everything from staging and unstaged

cbron (approved):

PR #1096, opened 2021-04-07
Change adapter repo and remove from top-level values.yaml

aiyengar2:
Related Issue: https://github.com/rancher/rancher/issues/29871

### DIFF
--- a/packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
+++ b/packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
@@ -1,6 +1,6 @@
 --- charts-original/values.yaml
 +++ charts/values.yaml
-@@ -2,13 +2,291 @@
+@@ -2,13 +2,286 @@
  # This is a YAML-formatted file.
  # Declare variables to be passed into your templates.
  
@@ -15,11 +15,6 @@
 +    # Change this if you change the namespaceOverride or nameOverride of prometheus-operator
 +    url: http://rancher-monitoring-prometheus.cattle-monitoring-system.svc
 +    port: 9090
-+  image:
-+    repository: rancher/mirrored-directxman12-k8s-prometheus-adapter-amd64
-+    tag: v0.8.3
-+    pullPolicy: IfNotPresent
-+    pullSecrets: {}
 +  psp:
 +    create: true
 +
@@ -294,7 +289,7 @@
  
  ## Provide a k8s version to auto dashboard import script example: kubeTargetVersionOverride: 1.16.6
  ##
-@@ -89,8 +367,23 @@
+@@ -89,8 +362,23 @@
  
  ##
  global:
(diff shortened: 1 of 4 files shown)


packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
```diff
@@ -15,11 +15,6 @@
 +    # Change this if you change the namespaceOverride or nameOverride of prometheus-operator
 +    url: http://rancher-monitoring-prometheus.cattle-monitoring-system.svc
 +    port: 9090
-+  image:
-+    repository: rancher/mirrored-directxman12-k8s-prometheus-adapter-amd64
-+    tag: v0.8.3
-+    pullPolicy: IfNotPresent
-+    pullSecrets: {}
```
aiyengar2, line 22: removing since this is already tracked in the subchart.

packages/rancher-prometheus-adapter/generated-changes/patch/values.yaml.patch
```diff
@@ -10,7 +10,7 @@
  
  image:
 -  repository: directxman12/k8s-prometheus-adapter-amd64
-+  repository: rancher/mirrored-directxman12-k8s-prometheus-adapter-amd64
++  repository: rancher/mirrored-directxman12-k8s-prometheus-adapter
```
aiyengar2, line 13: https://github.com/rancher/image-mirror/pull/86

packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
brendarearden: will the image still get populated properly for the  image-list? I am not sure how we generate that still, but I remember at one point all the images needed to be listed in the main chart.yaml to ensure it was still picked up for that.

prachidamle (dismissed):

packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
aiyengar2: I believe @PennyScissors and I had a conversation about this and it does pull through all of the subcharts. It essentially lists every single file within the entire repo, checks whether its name is values.yaml, and lists every single YAML entry within those values.yaml files to find something that has the key image.

brendarearden (approved):

packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
brendarearden: is this something we should revisit with all of our charts? We patch versions and list them at both levels for all of our charts with subcharts.

brendarearden (dismissed):

packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
aiyengar2: I think we should. It's a better idea to have them tracked in just one place for reviews and future changes.

prachidamle (approved):

PR #1097, opened 2021-04-07
Update external URLs if cattle values are provided

aiyengar2:
Sets the externalURL to the rancher proxy if and only if `.Values.global.cattle.url` and `.Values.global.cattle.clusterId` are set.

While these will be set magically by default in Rancher 2.5.8 onwards by the dashboard UI, this solution doesn't prevent a user from manually setting these fields to be able to use this in Rancher 2.5.7 or before.


packages/rancher-monitoring/generated-changes/patch/templates/alertmanager/alertmanager.yaml.patch
```diff
@@ -9,7 +9,12 @@
    version: {{ .Values.alertmanager.alertmanagerSpec.image.tag }}
    {{- if .Values.alertmanager.alertmanagerSpec.image.sha }}
    sha: {{ .Values.alertmanager.alertmanagerSpec.image.sha }}
-@@ -25,8 +25,8 @@
+@@ -22,11 +22,13 @@
+   externalUrl: "{{ tpl .Values.alertmanager.alertmanagerSpec.externalUrl . }}"
+ {{- else if and .Values.alertmanager.ingress.enabled .Values.alertmanager.ingress.hosts }}
+   externalUrl: "http://{{ tpl (index .Values.alertmanager.ingress.hosts 0) . }}{{ .Values.alertmanager.alertmanagerSpec.routePrefix }}"
++{{- else if not (or (kindIs "invalid" .Values.global.cattle.url) (kindIs "invalid" .Values.global.cattle.clusterId)) }}
```
aiyengar2, line 16: `kindIs "invalid" {field}` is a nice little hack that allows us to not have to check `.Values.global` and `.Values.global.cattle` before looking at this field. If `.Values.global` or `.Values.global.cattle` do not exist, these statements will return true.

prachidamle (approved):

brendarearden (approved):
LGTM

PR #1098, opened 2021-04-07
Only remove ignored files on push

aiyengar2:


.github/workflows/push.yaml
```diff
@@ -30,10 +30,7 @@ jobs:
       run: git push origin dev-v2.5
 
     - name: (temporary) Clean up Git
-      run: git checkout -- .; git clean -xdf
```
aiyengar2, line 33: x is only needed to drop ignored files, but CI doesn't like that removed `bin/charts-build-scripts` based on the error in https://github.com/rancher/charts/runs/2292089289.

```bash
warning: failed to remove bin/charts-build-scripts: Permission denied
```

cbron (approved):

PR #1109, opened 2021-04-09
Use global.cattle.windows.enabled for windowsExporter and deprecate windowsExporter.enabled

aiyengar2:
Related Issue: https://github.com/rancher/rancher/issues/31942

### DIFF
--- a/packages/rancher-monitoring/generated-changes/patch/Chart.yaml.patch
+++ b/packages/rancher-monitoring/generated-changes/patch/Chart.yaml.patch
@@ -48,3 +48,12 @@
  kubeversion: '>=1.16.0-0'
  dependencies:
  - name: grafana
+@@ -185,7 +195,7 @@
+ - name: windowsExporter
+   version: ""
+   repository: file://./charts/windowsExporter
+-  condition: windowsExporter.enabled
++  condition: global.cattle.windows.enabled
+   tags: []
+   enabled: false
+   importvalues: []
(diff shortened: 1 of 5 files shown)


packages/rancher-monitoring/generated-changes/patch/Chart.yaml.patch
```diff
@@ -48,3 +48,12 @@
  kubeversion: '>=1.16.0-0'
  dependencies:
  - name: grafana
+@@ -185,7 +195,7 @@
+ - name: windowsExporter
+   version: ""
+   repository: file://./charts/windowsExporter
+-  condition: windowsExporter.enabled
++  condition: global.cattle.windows.enabled
```
aiyengar2, line 56: Based on [Helm docs](https://helm.sh/docs/topics/charts/#tags-and-condition-fields-in-dependencies):

```
Only the first valid path found in the list is evaluated and if no paths exist then the condition has no effect.
```

Due to this constraint of Helm, I entirely deprecated the `windowsExporter.enabled` field in favor of `global.cattle.windows.enabled` for the following reasons:
- A user in a non-Windows cluster today (e.g. `global.cattle.windows.enabled=false, windowsExporter.enabled=true/false`) would have no reason to deploy windowsExporter. So either of these cases are not necessary to consider
- A user who is in a Windows cluster today deploying Rancher via the UI (e.g. `global.cattle.windows.enabled=true, windowsExporter.enabled=true/false`) would not be able to disable windowsExporter from being deployed since `global.cattle.windows.enabled=true` would be set automatically. 
- A user who is in a Windows cluster today installing via a script like Terraform could choose to enable or disable `global.cattle.windows.enabled` since they control the field, so it is effectively the same as providing them the ability to toggle `windowsExporter.enabled=true/false` since nothing else except for windowsExporter gets deployed specifically for a Windows cluster.

brendarearden (approved):
LGTM

nickgerace (approved):

PR #1114, opened 2021-04-12
fixed journald aggregation for rke2

paynejacob:
PROBLEM:
we were not getting any of the rke2 systemd logs

SOLUTION:
fluentbit's journald plugin needs the machine id to work correctly

TESTING:
installed logging on an rke2 cluster and was able to get this log message

```
{"_MACHINE_ID":"59e806f5bba74a5d8addb46a60a7ce78","PRIORITY":"6","SYSLOG_FACILITY":"3","_UID":"0","_GID":"0","_SELINUX_CONTEXT":"unconfined\n","_SYSTEMD_SLICE":"system.slice","_TRANSPORT":"stdout","_CAP_EFFECTIVE":"3fffffffff","_HOSTNAME":"jpayne-rke2","SYSLOG_IDENTIFIER":"rke2","_COMM":"rke2","_EXE":"/usr/local/bin/rke2","_SYSTEMD_CGROUP":"/system.slice/rke2-server.service","_SYSTEMD_UNIT":"rke2-server.service","_CMDLINE":"/usr/local/bin/rke2 server","_BOOT_ID":"caa7ced9023945a482a7e0bb726c1dbf","_STREAM_ID":"837d3787b0f548fc9cc3f633c08b988f","_PID":"15517","_SYSTEMD_INVOCATION_ID":"0db756f3fcc44b13902c01badc67e7b8","MESSAGE":"I0412 19:10:13.724172   15517 controllermanager.go:238] Started \"cloud-node-lifecycle\""}
```

### DIFF
--- a/packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/configmap.yaml
+++ b/packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/configmap.yaml
@@ -10,7 +10,8 @@ data:
     [INPUT]
         Name              systemd
         Tag               rke2
-        Systemd_Filter    _SYSTEMD_UNIT=rke2.service
+        Path              {{ .Values.systemdLogPath | default "/var/log/journal" }}
+        Systemd_Filter    _SYSTEMD_UNIT=rke2-server.service
 
     [OUTPUT]
         Name              forward
--- a/packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
+++ b/packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
@@ -26,8 +26,11 @@ spec:
           volumeMounts:
             - mountPath: /fluent-bit/etc/
               name: config
-            - mountPath: /run/log/journal
+            - mountPath: {{ .Values.systemdLogPath | default "/var/log/journal" }}
               name: journal
+              readOnly: true
+            - mountPath: /etc/machine-id
+              name: machine-id
               readOnly: true
       {{- with .Values.tolerations }}
       tolerations:
@@ -44,7 +47,10 @@ spec:
             name: "{{ .Release.Name }}-rke2"
         - name: journal
           hostPath:
-            path: /run/log/journal
+            path: {{ .Values.systemdLogPath | default "/var/log/journal" }}
+        - name: machine-id
+          hostPath:
+            path: /etc/machine-id
 ---
 apiVersion: v1
 kind: ServiceAccount
--- a/packages/rancher-logging/generated-changes/patch/values.yaml.patch
+++ b/packages/rancher-logging/generated-changes/patch/values.yaml.patch
@@ -36,12 +36,14 @@
  rbac:
    enabled: true
    psp:
-@@ -85,3 +93,77 @@
+@@ -85,3 +93,79 @@
      additionalLabels: {}
      metricRelabelings: []
      relabelings: []
 +
 +disablePvc: true
++
++systemdLogPath: "/var/log/journal"
 +
 +additionalLoggingSources:
 +  rke:
--- a/packages/rancher-logging/package.yaml
+++ b/packages/rancher-logging/package.yaml
@@ -1,6 +1,6 @@
 url: https://kubernetes-charts.banzaicloud.com/charts/logging-operator-3.9.4.tgz
 packageVersion: 00
-releaseCandidateVersion: 03
+releaseCandidateVersion: 04
 additionalCharts:
   - workingDir: charts-crd
     crdOptions:


nickgerace (approved):
Barring SELinux implications, LGTM.

paynejacob:

packages/rancher-logging/package.yaml
```diff
@@ -1,6 +1,6 @@
 url: https://kubernetes-charts.banzaicloud.com/charts/logging-operator-3.9.4.tgz
 packageVersion: 00
-releaseCandidateVersion: 03
+releaseCandidateVersion: 04
```
cbron, line 3: Can you squash commits.

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
```diff
@@ -26,9 +26,12 @@ spec:
           volumeMounts:
             - mountPath: /fluent-bit/etc/
               name: config
-            - mountPath: /run/log/journal
+            - mountPath: /var/log/journal
               name: journal
               readOnly: true
+            - mountPath: /etc/machine-id
```
cbron, line 32: We have not added a policy for this in SELinux, I do not think it will pass. But we can decide if that is dealbreaker for 2.5.8, or if we can fix quickly. Looks like we didn't have one for `/run/log` either, but thats ok now.

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/configmap.yaml
```diff
@@ -10,7 +10,8 @@ data:
     [INPUT]
         Name              systemd
         Tag               rke2
-        Systemd_Filter    _SYSTEMD_UNIT=rke2.service
+        Path              /var/log/journal
+        Systemd_Filter    _SYSTEMD_UNIT=rke2-server.service
```
cbron, line 14: Are all types of rke2 server ? Would agent/node have this name too.

cbron (changes requested):

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/configmap.yaml
paynejacob: all the systemd logs would fall under that unit.  I did not see any others mentioned in the rke2 docs.

paynejacob: The rest are under containers

nickgerace: +1 to @paynejacob they use the same systemd unit

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
cmurphy: > We have not added a policy for this in SELinux, I do not think it will pass

Correct, /etc/machine-id has its own label that's not part of our policy yet:

```
$ sudo semanage fcontext -l | grep /etc/machine-id
/etc/machine-id                                    regular file       system_u:object_r:machineid_t:s0
```

> Looks like we didn't have one for /run/log either

We do have that:

```
$ sudo ls -Zd /run/log/journal
drwxr-sr-x. root systemd-journal system_u:object_r:syslogd_var_run_t:s0 /run/log/journal
```

https://github.com/rancher/rancher-selinux/blob/main/policy/centos7/rancher.te#L27-L28

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
```diff
@@ -26,9 +26,12 @@ spec:
           volumeMounts:
             - mountPath: /fluent-bit/etc/
               name: config
-            - mountPath: /run/log/journal
+            - mountPath: /var/log/journal
```
cmurphy, line 29: This path doesn't exist when I look on centos 7, centos 8, and opensuse. It only exists on ubuntu.

cbron (dismissed):

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
```diff
@@ -26,9 +26,12 @@ spec:
           volumeMounts:
             - mountPath: /fluent-bit/etc/
               name: config
-            - mountPath: /run/log/journal
+            - mountPath: {{if .Values.global.seLinux.enabled }}"/run/log/journal"{{ else }}"/var/log/journal"{{ end }}
```
cmurphy, line 29: It doesn't have anything to do with selinux, it's just the way systemd is configured by default for the distro.

On CentOS:

```
$ grep Storage /etc/systemd/journald.conf
#Storage=auto
```

On Ubuntu:

```
$ grep Storage /etc/systemd/journald.conf 
Storage=persistent
```

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
cbron: @cmurphy can you work on a PR to add the machine-id policy to rancher selinux, this may not get tested till next monday so its a little lower priority.

packages/rancher-logging/generated-changes/patch/values.yaml.patch
```diff
@@ -36,13 +36,15 @@
...
    psp:
-@@ -85,3 +93,77 @@
+@@ -85,3 +93,79 @@
      additionalLabels: {}
      metricRelabelings: []
      relabelings: []
 +
 +disablePvc: true
 +
++systemdLogPath: "/var/log/journal"
```
cbron, line 46: if we go this route we definitely need to document it, can you open/add-to an issue

paynejacob: I want to test this a bit before comitting, assuming all goes well I will open an issue by eob

nickgerace: This route is probably the best one anyway since the underlying distro has the choice of where this path can be. Documentation + this approach is what I would do

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
cmurphy: It actually looks like it doesn't need an additional policy, I was able to run a pod that mounts and reads /etc/machine-id on an selinux-enforcing host with no permission errors.

paynejacob:
Tested on ubuntu and centos.  I was able to get logs for both.  Centos requires the rancher selinux package to work

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
```diff
@@ -44,7 +47,10 @@ spec:
             name: "{{ .Release.Name }}-rke2"
         - name: journal
           hostPath:
-            path: /run/log/journal
+            path: {{ .Values.systemdLogPath }}
```
nickgerace, line 50: This should default to `/var/log/journal` if empty. The user could accidentally unset it

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/configmap.yaml
```diff
@@ -10,7 +10,8 @@ data:
     [INPUT]
         Name              systemd
         Tag               rke2
-        Systemd_Filter    _SYSTEMD_UNIT=rke2.service
+        Path              {{ .Values.systemdLogPath }}
```
nickgerace, line 13: See comment below; we may want a default that's not in `values.yaml`

nickgerace (changes requested):
LGTM barring my one note. I believe we would rather want to fallback on a path that DNE rather than `""` if it's accidentally unset by the user.

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
paynejacob: the default in values.yaml is "/var/log/journal"

nickgerace: Apologies, I think I need to reword what I am asking: I am wondering if we need to either...

1. Validate that we have a non-empty value
2. Use a variable with a default value if the value passed in is empty

I don't know if we want a scenario where the value is `""` is what I am saying.

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
cbron: Hmm, I wonder if that exemption is in vanilla SELinux or somewhere else ? I'm guessing SELinux if this was a basic setup.

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
```diff
@@ -26,9 +26,12 @@ spec:
           volumeMounts:
             - mountPath: /fluent-bit/etc/
               name: config
-            - mountPath: /run/log/journal
+            - mountPath: {{ .Values.systemdLogPath | default "/var/log/journal" }}
```
cbron, line 29: > Centos requires the rancher selinux package to work

Does that mean even without SELinux enabled ? If so we have to figure out why.

cmurphy: No, the rancher-selinux package only does anything if SELinux is enabled. https://github.com/rancher/rancher/issues/31309#issuecomment-819749224 is the background

paynejacob: If you set selinux to permissive on centos it works fine but that was not the default state for me with centos8 on do.

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
cmurphy: I think it's just default in plain selinux

```
$ sudo sesearch --allow --class file | grep machineid_t
   allow init_t machineid_t : file { ioctl read write create getattr setattr lock append unlink link rename mounton open } ; 
   allow domain machineid_t : file { ioctl read getattr lock open } ;
```

It's not a protected file, it's not secret information. The journald logs are different because those have sensitive data.

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
nickgerace: I think this line is fine, but we should document something along the lines of...

"If SELinux is set to `Enforcing`, which most Fedora and RHEL derivatives distributions do by default, then you must install the `rancher-selinux` RPM as well"

...although we probably want something cleaner than that :) 

I like the idea of having the default option be the Ubuntu path. It's safer than having an empty option IMO

nickgerace (approved):
LGTM. Though, this will need docs changes.

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
cbron: I'm a bit lost, did we test with selinux on and our new rpm installed ? https://github.com/rancher/rancher/issues/31309#issuecomment-819751532

paynejacob: Yes this was tested on centos and ubuntu and was able to collect systemd logs on both.

cbron (dismissed):

nickgerace (approved):

PR #1116, opened 2021-04-13
Use a different prefix for Rancher dashboards than the Chart name

aiyengar2:
Currently, we prefix dashboards based on logic similar to `{{ .Chart.Name }}-<dashboard>`.

However, when Grafana has a default dashboard with the same name as Rancher, the name becomes the same, which results in a conflict, re: https://github.com/rancher/rancher/issues/32024#issuecomment-818312832

While one solution for this would be to just ensure there is no overlap between the dashboard ConfigMap names chosen by Rancher and Grafana, an easier fix seems to be just to change the naming prefix of dashboards that are deployed by default by Rancher instead.

Therefore, we can have a `rancher-monitoring-etcd` deployed by default with Grafana as well as a `rancher-default-dashboards-etcd` to keep the name etcd but avoid conflict.

Related Issue: https://github.com/rancher/rancher/issues/32024

### DIFF
--- a/packages/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/dashboards/rancher/default-dashboard.yaml
+++ b/packages/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/dashboards/rancher/default-dashboard.yaml
@@ -3,7 +3,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   namespace: {{ .Values.grafana.defaultDashboards.namespace }}
-  name: {{ printf "%s-%s" (include "kube-prometheus-stack.fullname" $) "home" | trunc 63 | trimSuffix "-" }}
+  name: rancher-default-dashboards-home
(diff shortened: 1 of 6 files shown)


nickgerace:
Why "dashboards" over "dashboard"? This is probably a pattern I'm unaware of.

aiyengar2:
> Why "dashboards" over "dashboard"? This is probably a pattern I'm unaware of.

Each ConfigMap can contain one or more dashboards within it since they correspond to all JSON files in a folder based on a glob path.

e.g.
https://github.com/rancher/charts/blob/f7d8732721de264fb79713750ee65f5e0dc95d59/packages/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/dashboards/rancher/etcd-dashboards.yaml#L16

nickgerace (approved):

brendarearden (dismissed):

brendarearden (approved):

PR #1125, opened 2021-04-19
Change released/ branch to stage release-v2.5

aiyengar2:
Related Issue: https://github.com/rancher/rancher/issues/32146


aiyengar2:
Needs to be forward-ported to dev-v2.6 as well, but will do once this has been approved.

Makefile
```diff
@@ -1,8 +1,8 @@
 pull-scripts:
 	./scripts/pull-scripts
 
-release:
-	./scripts/release-assets
+stage-release:
+	./scripts/stage-release
```
thedadams, line 5: If I understand correctly, release is no longer needed because release is not a straightforward copy and PR?

aiyengar2: Yup. We always regenerate the released/ directory based on the current contents of dev-v2.5 and release-v2.5 on merges to a source branch that feeds into it

aiyengar2: The caveat here is that, if you made changes directly to dev-v2.5, they would not be updated until either 1) you make the next merge into dev-v2.5-source or 2) you make the merge for the changes into dev-v2.5 and then make a separate PR to regenerate the released/ directory only by running `make stage-release`.

aiyengar2: The reason for this is that the make sync command run on the release-v2.5 branch is based off of the upstream git repository for dev-v2.5, not based on the current contents in your dev-v2.5 branch. 

We don't support local sync / validation workflows in the scripts

thedadams:
Needs rebase, but looks good to me.

scripts/stage-release
```diff
@@ -0,0 +1,50 @@
...
+mkdir -p released
+
+echo "Pulling in ${RELEASE_REPO_URL}@${RELEASE_BRANCH} into ${RELEASE_BRANCH_DIR}"
+rm -rf ${RELEASE_BRANCH_DIR}
+
+git clone --depth 1 --branch ${RELEASE_BRANCH} ${RELEASE_REPO_URL} ${RELEASE_BRANCH_DIR} 2>/dev/null
+
+cd ${RELEASE_BRANCH_DIR}
+
+make sync
```
cbron: May need ulimit adjustment here ?

cbron (approved):

Makefile
cbron: That reminds me, we'll need to update release workflow.

scripts/stage-release
aiyengar2: don't think this is necessary atm but we can add it later if we encounter that issue

PR #1126, opened 2021-04-19
Add stage-release workflow

aiyengar2:
Adds a step to call stage-release after https://github.com/rancher/charts/pull/1126 is merged


aiyengar2:
Needs to be forward-ported to dev-v2.6-source as well, but will do once this has been approved.

When the original PR in https://github.com/rancher/charts/pull/1125 is forward ported to dev-v2.6, I'll open a separate PR to update dev-v2.5-source's push-2.6 workflow to also regenerate the released/ directory

.github/workflows/push.yaml
```diff
@@ -28,3 +28,12 @@ jobs:
 
     - name: Push assets
       run: git push origin dev-v2.5
```
thedadams, line 30: Should we only push once?

thedadams:
Needs rebase, but looks good to me.

brendarearden (approved):
lgtm!

PR #1129, opened 2021-04-20
[2.5]  run fluentbit as a privileged container

paynejacob:

PROBLEM:
fluentbit did not display any errors when deployed on a centos cluster.  The systemd logs would never be delivered to the output.

SOLUTION:
run the fluentbit container as a privileged container.

TESTING:
I modified the running qa instance that discovered the bug and systemd logs were immediately delivered after the change was applied.

### DIFF
--- a/packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
+++ b/packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
@@ -20,6 +20,7 @@ spec:
           image: "{{ template "system_default_registry" . }}{{ .Values.images.fluentbit.repository }}:{{ .Values.images.fluentbit.tag }}"
           {{- if .Values.global.seLinux.enabled }}
           securityContext:
+            privileged: true
             seLinuxOptions:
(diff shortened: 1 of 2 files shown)


packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
```diff
@@ -20,6 +20,7 @@ spec:
           image: "{{ template "system_default_registry" . }}{{ .Values.images.fluentbit.repository }}:{{ .Values.images.fluentbit.tag }}"
           {{- if .Values.global.seLinux.enabled }}
           securityContext:
+            privileged: true
```
cbron, line 23: As I understand it, we don't want to run any privileged containers. It is incompatible with SELinux controls.

nickgerace: Agreed with Caleb, what's the motivation behind this change?

paynejacob:
This pr was a result of a misunderstanding on my part.  There will be another pr later with a proper fix for the permission issue.

PR #1130, opened 2021-04-20
fix the issue where the configmap is missing when upgrading rancher-alerting-drives to enable a new driver

jiaqiluo:
Issue: https://github.com/rancher/rancher/issues/32249

Problem:
If we enable a new driver when doing a helm upgrade, the new configMap will not be created because configmaps are created only when a `helm install` happens. 

Fix:
Create the configmaps if it is missing when a `helm upgrade` happens.


cbron (approved):

aiyengar2 (approved):
LGTM

brendarearden (approved):
LGTM

PR #1131, opened 2021-04-21
Fix CRD chart regression and fix CRDs for upgrades

aiyengar2:
The first commit reverts a change introduced in a previous commit that deletes CRDs before re-installing them. This is not an option in a production environment since it would wipe out all of the CRs in the cluster, which would effectively remove any customization that a user added between Monitoring chart upgrades.

The second commit modifies the CRD chart to use rancher/agent instead of rancher/kubectl (since we need to perform some light scripting and `sh` is not offered in rancher/kubectl).

Once the image was modified, I changed the remove-finalizer container to avoid doing an entire kubectl apply in favor of just removing the finalizers (the original intention of the container) and added a new initContainer on install that would destroy the preserveUnknownField on an upgrade due to an issue with upgrading an apiextensions.k8s.io/v1beta1 CRD to one in apiextensions.k8s.io/v1, as noted in https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#field-pruning using the workaround described in https://github.com/kubernetes-sigs/controller-tools/issues/476#issuecomment-691519936.

Related Issue: https://github.com/rancher/rancher/issues/31991, https://github.com/rancher/rancher/issues/32225

### DIFF
--- a/packages/rancher-monitoring/templates/crd-template/templates/jobs.yaml
+++ b/packages/rancher-monitoring/templates/crd-template/templates/jobs.yaml
@@ -19,20 +19,6 @@ spec:
       securityContext:
         runAsNonRoot: true
         runAsUser: 1000
-      initContainers:
-        - name: delete-crds
-          image: {{ template "system_default_registry" . }}{{ .Values.image.repository }}:{{ .Values.image.tag }}
-          imagePullPolicy: IfNotPresent
-          command:
-          - /bin/kubectl
-          - delete
-          - --ignore-not-found=true
-          - -f
-          - /etc/config/crd-manifest.yaml
-          volumeMounts:
-            - name: crd-manifest
-              readOnly: true
-              mountPath: /etc/config
       containers:
         - name: create-crds
           image: {{ template "system_default_registry" . }}{{ .Values.image.repository }}:{{ .Values.image.tag }}
--- a/packages/rancher-monitoring/templates/crd-template/templates/jobs.yaml
+++ b/packages/rancher-monitoring/templates/crd-template/templates/jobs.yaml
@@ -19,15 +19,29 @@ spec:
       securityContext:
         runAsNonRoot: true
         runAsUser: 1000
+      initContainers:
+        - name: set-preserve-unknown-fields-false
+          image: {{ template "system_default_registry" . }}{{ .Values.image.repository }}:{{ .Values.image.tag }}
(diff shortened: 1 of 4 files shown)


packages/rancher-monitoring/templates/crd-template/templates/jobs.yaml
```diff
@@ -20,28 +20,28 @@ spec:
...
-          volumeMounts:
-            - name: crd-manifest
-              readOnly: true
-              mountPath: /etc/config
+          - /bin/sh
+          - -c
+          - >
+            {{- range $path, $_ :=  (.Files.Glob  "crd-manifest/**.yaml") }}
+            {{- $crd := get (get ($.Files.Get $path | fromYaml) "metadata") "name" }}
+            if [[ -n "$(kubectl get crd {{ $crd }} -o jsonpath='{.spec.preserveUnknownFields}')" ]]; then
```
aiyengar2, line 32: This is a best effort operation (hence the `|| true`) that should be done only if the CRD exists and has `.spec.preserveUnknownFields` set to some value. 

If the CRD does not exist or if `.spec.preserveUnknownFields` is unset, this bash script will skip since the `kubectl get` command will return nothing to the stdout, but it will still print the error in finding the resource since the kubectl get command prints to stderr.

packages/rancher-monitoring/templates/crd-template/templates/jobs.yaml
```diff
@@ -20,28 +20,28 @@ spec:
...
-          - /etc/config/crd-manifest.yaml
-          volumeMounts:
-            - name: crd-manifest
-              readOnly: true
-              mountPath: /etc/config
+          - /bin/sh
+          - -c
+          - >
+            {{- range $path, $_ :=  (.Files.Glob  "crd-manifest/**.yaml") }}
+            {{- $crd := get (get ($.Files.Get $path | fromYaml) "metadata") "name" }}
```
aiyengar2, line 31: Helm code that basically says "look at each file in crd-manifest that ends with .yaml and parse it as a YAML doc. Then pull out `metadata.name`'s value and put it in the variable $crd."

e.g. anywhere that says `{{ $crd }}` below would look like `prometheus.monitoring.coreos.com`.

dramich (approved):

brendarearden (approved):
LGTM

PR #1132, opened 2021-04-21
Change the tracing url for kiali

brendarearden:
**Problem**
When attempting to look at traces for services in Kiali, if you change the filter options, an error occurs `Cannot not parse minDuration: time: missing unit in duration "1703875"' and the display button does not work after that. Additional error is seen when Kiali attempts to make a call to Tracing and is unable to reach the service using the provided url. 

**Solution**
Use the full URL value for `in_cluster_url` and `url` to ensure the Kiali service is able to reach the tracing service. Added the same notes to the readme and the app-readme for both 1.8 and 1.9 to cover how to properly set this value. 

**Issue** 
https://github.com/rancher/rancher/issues/31979

**Notes**
From various notes I found on different issues, it would appear that even though tracing has the option for an `in_cluster_url`, it still needs to be an external facing url. If it is not accessible externally, it will result in an error. A different way to resolve this issue would've been to proxy the in_cluster_url so that it is accessible externally. Because the call is being made to an "external" service, the authorization has to be set to insecure_skip_verify.

### DIFF
--- a/packages/rancher-istio-1.8/charts/README.md
+++ b/packages/rancher-istio-1.8/charts/README.md
@@ -20,7 +20,7 @@ If you remove dependent CRD charts prior to removing rancher-istio, you may enco
 
 Kiali allows you to view and manage your istio-based service mesh through an easy to use dashboard.
 
-####  Dependencies
+###  Dependencies
 - rancher-monitoring chart or other Prometheus installation
 
 This dependecy installs the required CRDs for installing Kiali. Since Kiali is bundled in with Istio in this chart, if you do not have these dependencies installed, your Istio installation will fail. If you do not plan on using Kiali, set `kiali.enabled=false` when installing Istio for a succesful installation.
@@ -34,7 +34,7 @@ To limit scraping to specific namespaces, set `prometheus.prometheusSpec.ignoreN
 1. Add a Service Monitor or Pod Monitor in the namespace with the targets you want to scrape.
 1. Add an additionalScrapeConfig to your rancher-monitoring instance to scrape all targets in all namespaces.
 
-####  External Services
+###  External Kiali Services
 
 ##### Prometheus
 The `kiali.external_services.prometheus` url is set in the values.yaml:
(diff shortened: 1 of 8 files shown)


packages/rancher-istio-1.8/charts/README.md
```diff
@@ -51,12 +51,18 @@ http://{{ .Values.nameOverride }}-grafana.{{ .Values.namespaceOverride }}.svc:{{
 The url depends on the default values for `nameOverride`, `namespaceOverride`, and `grafana.service.port` being set in your rancher-monitoring or other monitoring instance.
 
 ##### Tracing
-The `kiali.external_services.tracing` url and `.Values.tracing.contextPath` is set in the rancher-istio values.yaml:
+The `kiali.external_services.tracing.url` url, `kiali.external_services.tracing.in_cluster_url` url and `.Values.tracing.contextPath` is set in the rancher-istio values.yaml. To ensure traces work for your services in the Kiali dashboard, you must set `kiali.external_services.tracing.url` and `kiali.external_services.tracing.in_cluster_url` key value using the complete external url to your tracing instance.
```
nickgerace, line 54: nit: Should we split sentences into newlines?  Also, what do you think about removing "url" after the url values since contextPath does not have a proceeding name?

packages/rancher-istio-1.8/charts/values.yaml
```diff
@@ -82,7 +82,10 @@ kiali:
       custom_metrics_url: "http://rancher-monitoring-prometheus.cattle-monitoring-system.svc:9090"
       url: "http://rancher-monitoring-prometheus.cattle-monitoring-system.svc:9090"
     tracing:
-      in_cluster_url: "http://tracing.istio-system.svc:16686/jaeger"
+      auth:
+        insecure_skip_verify: true
+      in_cluster_url: "https://173.255.241.30/k8s/clusters/c-zbpdp/api/v1/namespaces/istio-system/services/http:tracing:16686/proxy/jaeger"
+      url: "https://173.255.241.30/k8s/clusters/c-zbpdp/api/v1/namespaces/istio-system/services/http:tracing:16686/proxy/jaeger"
```
nickgerace, line 88: Are these IP's ok to be in the chart?

brendarearden: I did mean to change this to a generic url - I have updated that with my most recent commit

packages/rancher-istio-1.8/charts/README.md
brendarearden: These should be fixed in the most recent commit

packages/rancher-istio-1.8/charts/values.yaml
```diff
@@ -82,7 +82,11 @@ kiali:
       custom_metrics_url: "http://rancher-monitoring-prometheus.cattle-monitoring-system.svc:9090"
       url: "http://rancher-monitoring-prometheus.cattle-monitoring-system.svc:9090"
     tracing:
-      in_cluster_url: "http://tracing.istio-system.svc:16686/jaeger"
+      auth:
+        insecure_skip_verify: true
+      # this configuration needs changed to match your current installation for it to work
+      in_cluster_url: "https://tracing-service:8080/jaeger"
+      url: "https://tracing-service:8080/jaeger"
```
cbron, line 89: Is this off by default ? Or is that actually a functioning url.

packages/rancher-istio-1.8/charts/app-readme.md
```diff
@@ -35,11 +35,19 @@ http://{{ .Values.nameOverride }}-grafana.{{ .Values.namespaceOverride }}.svc:{{
...
+https://<your domain>/k8s/clusters/<cluster id>/api/v1/namespaces/{{ .Values.namespaceOverride }}/services/http:{{ .Values.namespaceOverride }}.svc:{{ .Values.service.externalPort }}/{{ .Values.tracing.contextPath }}
 ```
+An example with all the values filled in:
+```
+https://111.222.333.44/k8s/clusters/c-zbpdp/api/v1/namespaces/istio-system/services/http:tracing:16686/proxy/jaeger
+```
+The url depends on the default values for `namespaceOverride`, and `.Values.service.externalPort` being set in your rancher-tracing or other tracing instance.
+
+> **Note:** If you are having a hard time determining the complete url value, deploy Istio with Kiali and Tracing enabled.
+Navigate to the Jaeger dashboard and copy the url from the browser up to the {{ .Values.tracing.contextPath }} context path to use as your complete url value. The context path is `/jaeger` by default.
```
cbron, line 51: This is a lot to put into the app-readme. Ideally that page is quite short. I would put the example and the note in the readme, not in the app-readme.

packages/rancher-istio-1.8/charts/values.yaml
brendarearden: Tracing is disabled by default. With that in mind, I have updated this to both be empty strings since we cannot set them to a default value and they don't need set until tracing is enabled.

packages/rancher-istio-1.8/charts/app-readme.md
brendarearden: I have removed the example and notes section from the app-readme. It is now just in the readme.

nickgerace (dismissed):

cbron (approved):

cbron (approved):

nickgerace (approved):

brendarearden:
This is not needed at this time. It does not resolve the error as we were expecting.

PR #1133, opened 2021-04-21
Add option to disable Windows logging for Windows clusters

nickgerace:
Relevant issues:
- rancher/rancher#28721
- rancher/dashboard#2684
- rancher/rancher#32264

### DIFF
--- a/packages/rancher-logging/generated-changes/overlay/templates/loggings/root/logging.yaml
+++ b/packages/rancher-logging/generated-changes/overlay/templates/loggings/root/logging.yaml
@@ -8,6 +8,7 @@ metadata:
 {{ include "logging-operator.labels" . | indent 4 }}
 spec:
   controlNamespace: {{ .Release.Namespace }}
+  {{- if not (kindIs "invalid" .Values.global.cattle.windows.enabled) }}
   {{- if .Values.global.cattle.windows.enabled }}
   nodeAgents:
     - name: win-agent
@@ -44,6 +45,7 @@ spec:
         tls:
           enabled: {{ .Values.nodeAgents.tls.enabled | default false }}
     {{- end }}
+  {{- end }}
   {{- end }}
   fluentbit:
     image:
--- a/packages/rancher-logging/generated-changes/patch/values.yaml.patch
+++ b/packages/rancher-logging/generated-changes/patch/values.yaml.patch
@@ -36,7 +36,7 @@
  rbac:
    enabled: true
    psp:
-@@ -85,3 +93,79 @@
+@@ -85,3 +93,83 @@
      additionalLabels: {}
      metricRelabelings: []
      relabelings: []
@@ -105,8 +105,12 @@
 +global:
 +  cattle:
 +    systemDefaultRegistry: ""
-+    windows:
-+      enabled: false
++    # Uncomment the below two lines to either enable or disable Windows logging. If this chart is
++    # installed via the Rancher UI, it will set this value to "true" if the cluster is a Windows
++    # cluster. In that scenario, if you would like to disable Windows logging on Windows clusters,
++    # set the value below to "false".
++    # windows:
++    #   enabled: true
 +  # Change the "dockerRootDirectory" if the default Docker directory has changed.
 +  dockerRootDirectory: ""
 +  # This psp setting differs from the upstream "rbac.psp" by only enabling psp settings for the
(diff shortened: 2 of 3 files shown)


paynejacob (dismissed):

packages/rancher-logging/generated-changes/patch/values.yaml.patch
```diff
@@ -105,6 +110,7 @@
 +global:
 +  cattle:
 +    systemDefaultRegistry: ""
++    # This value is immutable if set by the UI.
```
nickgerace, line 113: **Add comment:** 

If not using the Rancher UI, then you must set `global.cattle.windows.enabled` to `true` for Windows cluster support.

packages/rancher-logging/generated-changes/overlay/templates/loggings/root/logging.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
 {{ include "logging-operator.labels" . | indent 4 }}
 spec:
   controlNamespace: {{ .Release.Namespace }}
-  {{- if .Values.global.cattle.windows.enabled }}
+  {{- if and .Values.global.cattle.windows.enabled .Values.nodeAgents.windows.enabled }}
```
aiyengar2, line 11:
```suggestion
  {{- if not (kindIs "invalid" .Values.global.cattle.windows.enabled) }}
  {{- if .Values.global.cattle.windows.enabled }}
```

+ an `{{- end }}` at the very end of the file

aiyengar2: The reason why you need 2 if statements is because `.Values.global.cattle.windows.enabled` should only be evaluated if the first `if` statement is true, and I believe Helm evaluates both statements in an `{{- if and () () }}` clause, which would cause it to always fail out if `.Values.global.cattle.windows.enabled` is not defined.

aiyengar2 (changes requested):
Based on conversation with @vincent99 offline, I think we should not introduce a brand new field but rather comment out `global.cattle.windows.enabled` in the values.yaml for this chart and introduce a slight change to the way we check whether it is enabled to allow it to be unset in the values.yaml.

See my comment in this review.

nickgerace:
Entirely agree with @aiyengar2's idea from the offline discussion. PR will now reflect that.

packages/rancher-logging/generated-changes/patch/values.yaml.patch
```diff
@@ -105,8 +105,6 @@
 +global:
 +  cattle:
 +    systemDefaultRegistry: ""
-+    windows:
-+      enabled: false
```
aiyengar2, line 109: Should we comment this out instead?

aiyengar2 (changes requested):

aiyengar2 (approved):
LGTM

paynejacob (approved):

packages/rancher-logging/generated-changes/patch/values.yaml.patch
```diff
@@ -105,8 +105,12 @@
...
++    # Uncomment the below two lines to either enable or disable Windows logging. If this chart is
++    # installed via the Rancher UI, it will set this value to "true" if the cluster is a Windows
++    # cluster. In that scenario, if you would like to disable Windows logging on Windows clusters,
++    # set the value below to "false".
++    # windows:
++    #   enabled: true
```
cbron, line 113: The reason we are doing this is: if a user sets to false then the UI will respect that ? Whereas if it defaults to false then the UI won't ?
Also we need to double check that the UI can indeed set a value that isn't in the values.yaml

vincent99: What happens is a combination of several things:
- The user enters global.cattle.windows.enabled=false
- When you click save, I diff that against the values yaml and save ONLY the things that differ from the chart.  But you've set that false is the default, so what the user entered is dropped since it's the same.
- Then after that I add all the global.* values, and those are set only-if-not-already set.  But it's not set in this case, so I set it to true.

So you commenting out the enabled means that the user's provided `false` is not the default, then it doesn't get dropped, and then it's set so it doesn't get overwritten with `true`.

You can set any arbitrary values in the yaml editor.

cbron: Explanation:

> spec.chart.values is the values from the chart definition.  spec.values is the user-configured and ui-configured values.  The UI combines the two for the edit screen so you see all the possibilities, then diff's the two and saves only the changes from the chart's definition.
But the chart contains global.cattle.windows.enabled=false, so if a user sets it to false, then it's no different than the default, so it's dropped from the values.  Then at the end the UI adds it's values for global.* if they're not set, which they now aren't because it was dropped.  Which makes the effective value true again.

cbron (dismissed):
lgtm

paynejacob (approved):

PR #1134, opened 2021-04-21
Removed istiocoredns from 1.9

brendarearden:
**Problem**
istiocoredns is not longer supported in the IstioOperator file
istio-installer is not catching errors for istioctl install and upgrade commands so helm is reporting a successful install even when not true. 

**Solution**
Remove istiocoredns from the IstioOperator base-config map. Updated the 1.8 values.yaml to indicate that it will be unsupported in 1.9. 
Use new istio-intaller image that catches errors for istioctl commands 

**Issue**
https://github.com/rancher/rancher/issues/32252

Dependent: https://github.com/rancher/istio-installer/pull/23

### DIFF
--- a/packages/rancher-istio-1.8/package.yaml
+++ b/packages/rancher-istio-1.8/package.yaml
@@ -1,3 +1,3 @@
 url: local
 packageVersion: 00
-releaseCandidateVersion: 01
+releaseCandidateVersion: 03
--- a/packages/rancher-istio-1.8/charts/values.yaml
+++ b/packages/rancher-istio-1.8/charts/values.yaml
@@ -7,7 +7,7 @@ installer:
   repository: rancher/istio-installer
   tag: 1.8.5-rancher1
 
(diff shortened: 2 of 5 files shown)


nickgerace (approved):

packages/rancher-istio-1.8/package.yaml
```diff
@@ -1,3 +1,3 @@
 url: local
 packageVersion: 00
-releaseCandidateVersion: 01
+releaseCandidateVersion: 03
```
cbron, line 3: Why not rc 02 on both of these ?

brendarearden: I was being optimistic that https://github.com/rancher/charts/pull/1132 would get merged first

aiyengar2 (dismissed):
LGTM; let’s make sure rc02 gets merged first

cbron (dismissed):

cbron (approved):

PR #1135, opened 2021-04-21
[2.6] Update vsphere CSI to v2.2.0

PennyScissors:
Issue: https://github.com/rancher/rancher/issues/31105

### DIFF
--- a/packages/rancher-vsphere-csi/charts/app-readme.md
+++ b/packages/rancher-vsphere-csi/charts/app-readme.md
@@ -1,0 +1,14 @@
+# vSphere Container Storage Interface (CSI)
+
+[vSphere Container Storage Interface (CSI)](https://github.com/kubernetes-sigs/vsphere-csi-driver/tree/release-2.1/manifests/v2.1.0/vsphere-7.0u1/) is a specification designed to enable persistent storage volume management on Container Orchestrators (COs) such as Kubernetes. The specification allows storage systems to integrate with containerized workloads running on Kubernetes. Using CSI, storage providers, such as VMware, can write and deploy plugins for storage systems in Kubernetes without a need to modify any core Kubernetes code.
+
+CSI allows volume plugins to be installed on Kubernetes clusters as extensions. Once a CSI compatible volume driver is deployed on a Kubernetes cluster, users can use the CSI to provision, attach, mount, and format the volumes exposed by the CSI driver.
+
+The CSI driver for vSphere is `csi.vsphere.vmware.com`.
+
+## Prerequisites
+
+- vSphere 6.7 U3+
+- Kubernetes v1.14+
+- Out-of-tree vSphere Cloud Provider Interface (CPI)
+- A Secret on your Kubernetes cluster that contains vSphere CSI configuration and credentials (Refer to `README` or `Detailed Descriptions`)
--- a/packages/rancher-vsphere-csi/charts/app-readme.md
+++ b/packages/rancher-vsphere-csi/charts/app-readme.md
@@ -1,6 +1,6 @@
 # vSphere Container Storage Interface (CSI)
 
-[vSphere Container Storage Interface (CSI)](https://github.com/kubernetes-sigs/vsphere-csi-driver/tree/release-2.1/manifests/v2.1.0/vsphere-7.0u1/) is a specification designed to enable persistent storage volume management on Container Orchestrators (COs) such as Kubernetes. The specification allows storage systems to integrate with containerized workloads running on Kubernetes. Using CSI, storage providers, such as VMware, can write and deploy plugins for storage systems in Kubernetes without a need to modify any core Kubernetes code.
+[vSphere Container Storage Interface (CSI)](https://github.com/kubernetes-sigs/vsphere-csi-driver) is a specification designed to enable persistent storage volume management on Container Orchestrators (COs) such as Kubernetes. The specification allows storage systems to integrate with containerized workloads running on Kubernetes. Using CSI, storage providers, such as VMware, can write and deploy plugins for storage systems in Kubernetes without a need to modify any core Kubernetes code.
 
 CSI allows volume plugins to be installed on Kubernetes clusters as extensions. Once a CSI compatible volume driver is deployed on a Kubernetes cluster, users can use the CSI to provision, attach, mount, and format the volumes exposed by the CSI driver.
 
--- a/packages/rancher-vsphere-csi/charts/questions.yaml
+++ b/packages/rancher-vsphere-csi/charts/questions.yaml
@@ -1,0 +1,91 @@
+questions:
+  - variable: vCenter.configSecret.generate
+    label: Generate CSI vSphere Config Secret
+    description: Generates a Secret that contains a CSI vSphere config and credentials (If the option to generate it is enabled, credentials will be visible in the API to authorized users)
+    type: boolean
+    default: true
+    required: true
+    group: Configuration
+    show_subquestion_if: true
+    subquestions:
+      - variable: vCenter.host
+        label: vCenter Host
+        description: IP address or FQDN of the vCenter
+        type: string
+        group: Configuration
+
+      - variable: vCenter.datacenters
+        description: Comma-separated list of paths to data centers. E.g "<dc1-path>, <dc2-path>, ..."
+        label: Data Centers
+        type: string
+        group: Configuration
+
+      - variable: vCenter.username
+        label: Username
+        description: Username for vCenter
+        type: string
+        group: Configuration
+
+      - variable: vCenter.password
+        label: Password
+        description: Password for vCenter
+        type: password
+        group: Configuration
+
+  - variable: vCenter.configSecret.name
+    label: CSI vSphere Config Secret Name
+    description: Name of the Secret that contains a CSI vSphere config and credentials (Will not be visible in the API. More info in the README)
+    type: string
+    group: Configuration
+    show_if: "vCenter.configSecret.generate=false"
+
+  - variable: csiMigration.enabled
+    label: Enable CSI Migration
+    description: Enable migration of volumes provisioned by in-tree vSphere provider to CSI (Available for vSphere 7.0 U1+ only)
+    type: boolean
+    default: false
+    group: Migration
+
+  - variable: csiController.csiResizer.enabled
+    label: Enable CSI Volume Resizer
+    description: This feature is available for vSphere 7.0 U1+ only
+    type: boolean
+    default: false
+    group: Storage
+
+  - variable: storageClass.enabled
+    default: true
+    label: Create Storage Class
+    description: Create a storageClass with the vSphere CSI provisioner
+    type: boolean
+    required: true
+    show_subquestion_if: true
+    group: Storage
+    subquestions:
+      - variable: storageClass.name
+        label: Storage Class Name
+        default: "vsphere-csi-sc"
+        type: string
+
+      - variable: storageClass.isDefault
+        label: Default Storage Class
+        description: Set the Storage Class as the default
+        default: true
+        type: boolean
+
+      - variable: storageClass.storagePolicyName
+        label: Storage Policy Name
+        description: Name of the Storage Policy created in vCenter
+        type: string
+
+      - variable: storageClass.datastoreURL
+        label: Data Store URL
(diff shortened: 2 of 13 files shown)


packages/rancher-vsphere-csi/charts/app-readme.md
```diff
@@ -1,6 +1,6 @@
 # vSphere Container Storage Interface (CSI)
 
-[vSphere Container Storage Interface (CSI)](https://github.com/kubernetes-sigs/vsphere-csi-driver/tree/release-2.1/manifests/v2.1.0/vsphere-7.0u1/) is a specification designed to enable persistent storage volume management on Container Orchestrators (COs) such as Kubernetes. The specification allows storage systems to integrate with containerized workloads running on Kubernetes. Using CSI, storage providers, such as VMware, can write and deploy plugins for storage systems in Kubernetes without a need to modify any core Kubernetes code.
+[vSphere Container Storage Interface (CSI)](https://github.com/kubernetes-sigs/vsphere-csi-driver) is a specification designed to enable persistent storage volume management on Container Orchestrators (COs) such as Kubernetes. The specification allows storage systems to integrate with containerized workloads running on Kubernetes. Using CSI, storage providers, such as VMware, can write and deploy plugins for storage systems in Kubernetes without a need to modify any core Kubernetes code.
```
PennyScissors, line 3: Removed the specific version from the link. I don't think it is necessary, and this way we don't have to update it with every chart update.

cbron:
This is actually for 2.6, do we want to backport to 2.5 ? Right now if we add to 2.5-source it will add to both 2.5 and 2.6. Or we can add to both. Either way we need to add to the "new charts" issue.

PennyScissors:
> This is actually for 2.6, do we want to backport to 2.5 ? Right now if we add to 2.5-source it will add to both 2.5 and 2.6. Or we can add to both. Either way we need to add to the "new charts" issue.

I'll rebase this to 2.6.

packages/rancher-vsphere-csi/charts/questions.yaml
```diff
@@ -44,7 +44,20 @@ questions:
     description: Enable migration of volumes provisioned by in-tree vSphere provider to CSI (Available for vSphere 7.0 U1+ only)
     type: boolean
     default: false
-    group: Migration
+    group: "Features"
```
aiyengar2, line 47: Was this supposed to be changed?

packages/rancher-vsphere-csi/charts/questions.yaml
```diff
@@ -44,7 +44,20 @@ questions:
...
+
+  - variable: csiAuthCheck.enabled
+    label: Enable authorization check on datastores and get datastores
```
aiyengar2, line 50: Not sure I understand why this label means. Do you think a better phrasing of this would be:

`Enable authorization checks on operations involving datastores`

packages/rancher-vsphere-csi/charts/templates/vsphere-csi-controller-deployment.yaml
```diff
@@ -25,10 +25,19 @@ spec:
         node-role.kubernetes.io/controlplane: "true"
       {{- end }}
       tolerations:
-        - operator: "Exists"
+        - key: node-role.kubernetes.io/master
+          operator: Exists
```
aiyengar2, line 29: What is the purpose for this change? Before, we were tolerating any NoSchedule or NoExecute, but now we seem to be lessening the scope to only NoSchedule with a specific key, so a user who was relying on this toleration for a custom taint would no longer be able to deploy the chart right?

I think, if we plan to make this change, we need to expose a way to override tolerations.

Also, we need to make sure that the default taints and tolerations for Rancher and RKE2 (e.g. CriticalAddonsOnly) are tolerated by default.

packages/rancher-vsphere-csi/charts/templates/vsphere-csi-controller-deployment.yaml
```diff
@@ -92,6 +106,9 @@ spec:
             - name: healthz
               containerPort: 9808
               protocol: TCP
+            - name: prometheus
```
aiyengar2, line 109: nit: was this a copy paste? If this was from upstream that's okay, but if it is from us I think we should modify this name to `metrics` instead of `prometheus`

packages/rancher-vsphere-csi/charts/templates/vsphere-csi-controller-rbac.yaml
```diff
@@ -36,7 +36,7 @@ rules:
     verbs: ["create", "get", "list", "watch", "update", "delete"]
   - apiGroups: ["apiextensions.k8s.io"]
     resources: ["customresourcedefinitions"]
-    verbs: ["get", "create"]
+    verbs: ["get", "create", "update"]
```
aiyengar2, line 39: Just want to clarify why this permission is needed; does the vsphere CSI controller automatically create CRDs, similar to what Rancher does?

aiyengar2 (changes requested):

packages/rancher-vsphere-csi/charts/questions.yaml
PennyScissors: I changed this to features because we initially had the migration option by itself and it made sense to name it Migration, but they added 2 other options and grouped them together as "internal features" so it made sense to me to group them together in the questions.yaml as well.

https://github.com/kubernetes-sigs/vsphere-csi-driver/blob/fa891a5f1a2290824c97be19195c3f3b5cd5bdc7/manifests/v2.2.0/deploy/vsphere-csi-controller-deployment.yaml#L175-L183

packages/rancher-vsphere-csi/charts/questions.yaml
PennyScissors: I couldn't find a good description in their docs stating what the csi-auth-check is exactly, so I used [this](https://github.com/kubernetes-sigs/vsphere-csi-driver/blob/73f89c21a72a8921a0458b2e34a6f5e0e3179c43/pkg/csi/service/common/authmanager.go#L39-L40) comment since this option enables that service. I like your suggestion better though, so I'll update it.

packages/rancher-vsphere-csi/charts/templates/vsphere-csi-controller-deployment.yaml
PennyScissors: This change came from upstream. Nearly everything that isn't templated is a change from upstream. Should we modifying stuff like this from what's upstream, or stick to what's upstream?

packages/rancher-vsphere-csi/charts/templates/vsphere-csi-controller-deployment.yaml
PennyScissors: Change from upstream as well

packages/rancher-vsphere-csi/charts/templates/vsphere-csi-controller-rbac.yaml
PennyScissors: This is a change from upstream as well, so I'm not why this permission is needed (I don't see any CRDs in the chart). I tried to stick to what's upstream to make the chart easier to update.

packages/rancher-vsphere-csi/charts/questions.yaml
PennyScissors: updated

packages/rancher-vsphere-csi/charts/templates/vsphere-csi-controller-rbac.yaml
aiyengar2: okay sounds good to me

packages/rancher-vsphere-csi/charts/templates/vsphere-csi-controller-deployment.yaml
PennyScissors: Reverted change from upstream and kept tolerations as is.

packages/rancher-vsphere-csi/charts/questions.yaml
```diff
@@ -44,7 +44,20 @@ questions:
...
+
+  - variable: csiAuthCheck.enabled
+    label: Enable authorization checks on operations involving datastores
+    type: boolean
+    default: false
+    group: Features
+
+  - variable: onlineVolumeExtend.enabled
+    label: Enable Online Volume Extend
+    description: Enable expansion of PVCs that are in use by a Pod or mounted in a Node (Available for vSphere 7.0 U2+ only)
```
aiyengar2, line 57: Should this be `mounted on a Node`?

aiyengar2 (dismissed):
LGTM, left a nit comment

packages/rancher-vsphere-csi/charts/questions.yaml
PennyScissors: updated

packages/rancher-vsphere-csi/package.yaml
```diff
@@ -1,3 +1,3 @@
 url: local
 packageVersion: 00
-releaseCandidateVersion: 01
+releaseCandidateVersion: 00
```
cbron, line 3: Nit: rc01

packages/rancher-vsphere-csi/package.yaml
```diff
@@ -1,3 +1,3 @@
 url: local
 packageVersion: 00
```
cbron, line 2: We will have to consider this as one of our `.000` charts

packages/rancher-vsphere-csi/charts/templates/vsphere-csi-controller-deployment.yaml
```diff
@@ -25,10 +25,24 @@ spec:
         node-role.kubernetes.io/controlplane: "true"
       {{- end }}
       tolerations:
+        # Rancher specific change: These tolerations are intentionally different from upstream to avoid lessening the scope to only NoSchedule with a specific key
+        # - key: node-role.kubernetes.io/master
+        #   operator: Exists
+        #   effect: NoSchedule
```
cbron, line 31: Why are we adding commented out tolerations ? Does this example come from somewhere else

packages/rancher-vsphere-csi/charts/templates/vsphere-csi-node-rbac.yaml
```diff
@@ -0,0 +1,29 @@
+kind: ServiceAccount
```
cbron, line 1: So the DS previously didn't use a service account and now it does ?

PennyScissors: This is a change from upstream

packages/rancher-vsphere-csi/charts/templates/vsphere-csi-controller-deployment.yaml
PennyScissors: The commented out toleration is a toleration that was set upstream for this update, but based on a conversation with @aiyengar2, we decided to reject that change from upstream and keep the tolerations as is because of [this](https://github.com/rancher/charts/pull/1135#discussion_r621552035). I kept the toleration commented out so that we know it is a Rancher specific change next time we update it because it is difficult to distinguish between what we and upstream removed/added.

packages/rancher-vsphere-csi/package.yaml
PennyScissors: I'm not sure what this means, is there anything that I need to change?

packages/rancher-vsphere-csi/package.yaml
PennyScissors: will update in next commit

packages/rancher-vsphere-csi/package.yaml
cbron: No, its fine. Just cc @aiyengar2

packages/rancher-vsphere-csi/charts/templates/vsphere-csi-node-rbac.yaml
PennyScissors: It looks like they were using the default service account and then created their own in a separate manifest. You can see the release notes [here](https://vsphere-csi-driver.sigs.k8s.io/releases/v2.2.0.html), but it doesn’t really say anything about it.

cbron (approved):

PR #1136, opened 2021-04-22
[2.6] Introduce stage-release script and drop released/ directory

aiyengar2:
Forward ports changes for released/ directory. Didn't recreate the released/ directory in this PR unlike https://github.com/rancher/charts/pull/1125 since it will get regenerated on pushing the workflow for dev-v2.5-source / dev-v2.6-source

Related Issue: https://github.com/rancher/rancher/issues/32146

PR #1137, opened 2021-04-22
Add stage-release workflow for 2.6

aiyengar2:
Blocked on https://github.com/rancher/charts/pull/1136.

Related Issue: https://github.com/rancher/rancher/issues/32146

### DIFF
--- a/.github/workflows/push-2.6.yaml
+++ b/.github/workflows/push-2.6.yaml
@@ -28,3 +28,12 @@ jobs:
 
     - name: Push assets
       run: git push origin dev-v2.6
+
+    - name: Stage released/ directory
+      run: FROM_WORKFLOW=1 RELEASE_REPO_URL=https://github.com/rancher/charts.git RELEASE_BRANCH=release-v2.6 make stage-release
+
+    - name: Add released/ to branch
+      run: git add . && git -c user.name="actions" -c user.email="actions@github.com" commit -m "Regenerate released directory" -m "$(git log -b dev-v2.5-source --oneline -n1 --pretty=format:'%B')" || true
+
+    - name: Push assets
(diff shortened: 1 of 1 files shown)


.github/workflows/push-2.6.yaml
```diff
@@ -28,3 +28,12 @@ jobs:
...
       run: git push origin dev-v2.6
+
+    - name: Stage released/ directory
+      run: FROM_WORKFLOW=1 RELEASE_REPO_URL=https://github.com/rancher/charts.git RELEASE_BRANCH=release-v2.6 make stage-release
+
+    - name: Add released/ to branch
+      run: git add . && git -c user.name="actions" -c user.email="actions@github.com" commit -m "Regenerate released directory" -m "$(git log -b dev-v2.5-source --oneline -n1 --pretty=format:'%B')" || true
+
+    - name: Push assets
+      run: git push origin dev-v2.6
```
cbron, line 39: Whats point of adding all this to 2.6 if we are going to make them fully separate branches ?

aiyengar2: At the current moment, they aren't separate, but when we separate them, we can drop this Github Workflow entirely

thedadams (approved):

.github/workflows/push-2.6.yaml
aiyengar2: We could repurpose this PR to drop the 2.6 workflows entirely in favor of manual forward-ports.

cbron: Talked offline, we should keep this working for now until we confirm its being repurposed.

cbron (approved):

brendarearden (approved):

PR #1139, opened 2021-04-22
changes in rancher-sachet: add the link to the upstream example

jiaqiluo:
issue https://github.com/rancher/rancher/issues/32296

Note:

Because the change is only in `rancher-sachet`, so bump the versions of `rancher-sachet` and `rancher-alerting-drivers`, NOT `rancher-prom2teams`.

### DIFF
--- a/packages/rancher-sachet/charts/templates/configmap-pre-install.yaml
+++ b/packages/rancher-sachet/charts/templates/configmap-pre-install.yaml
@@ -12,6 +12,19 @@ metadata:
     "helm.sh/resource-policy": keep
 data:
   config.yaml: |-
+    {{- if and (not .Values.sachet.providers) (not .Values.sachet.receivers) }}
+    # please refer to the upstream documentation for configuration options:
+    #   https://github.com/messagebird/sachet
+    #
+    # providers:
+    #   aliyun:
+    #     region_id:
+    #     ...
+    # receivers:
+    #   - name: 'team-sms'
+    #     provider: 'aliyu'
+    #     ...
+    {{- end }}
     {{- with .Values.sachet.providers }}
     providers: {{ toYaml . | nindent 6 }}
     {{- end }}
--- a/packages/rancher-alerting-drivers/package.yaml
+++ b/packages/rancher-alerting-drivers/package.yaml
@@ -1,3 +1,3 @@
 packageVersion: 00
-releaseCandidateVersion: 03
+releaseCandidateVersion: 04
 url: local
--- a/packages/rancher-sachet/package.yaml
+++ b/packages/rancher-sachet/package.yaml
@@ -1,3 +1,3 @@
 url: local
 packageVersion: 00
-releaseCandidateVersion: 03
+releaseCandidateVersion: 04


packages/rancher-sachet/charts/templates/configmap-pre-install.yaml
```diff
@@ -12,6 +12,17 @@ metadata:
     "helm.sh/resource-policy": keep
 data:
   config.yaml: |-
+    # please refer to the upstream documentation for configuration options:
+    #   https://github.com/messagebird/sachet/blob/master/examples/config.yaml
```
nickgerace, line 16: Should this be pointing to a link on the docs site?

jiaqiluo: Sure. 
I need to confirm with the Doc team to see if we will mention this in Rancher's documentation and  
get the proper URL.

nickgerace: Well, it doesn't have to be, I do not believe. I was moreso curious if this was intentional. If so, I think that's fine, but @cbron might have a different take.

cbron: In this case I'd prefer to leave it pointing at master branch of upstream. As it changes over time I don't want our docs team to have to keep up to date with it. Though I think we should just point at the repo itself instead of the file, in case the file moves.

packages/rancher-sachet/charts/templates/configmap-pre-install.yaml
```diff
@@ -12,6 +12,17 @@ metadata:
...
+    #
+    # providers:
+    #   aliyun:
+    #     region_id:
+    #     ...
+    # receivers:
+    #   - name: 'team-sms'
+    #     provider: 'aliyu'
+    #     ...
```
cbron, line 25: I'm a little confused at the below, if they supply the values with `Values.sachet.providers`, would this default config not still be set ? I think we should only add default config if file would be empty.

packages/rancher-sachet/charts/templates/configmap-pre-install.yaml
nickgerace: Good point. Agreed

packages/rancher-sachet/charts/templates/configmap-pre-install.yaml
jiaqiluo: Since rancher-alerting-drivers' chart does not expose `Values.sachet.providers` which is in the sub-chart rancher-sechat, users can not configure it when installing/upgrading rancher-alerting-drivers. They will see the above comments when they edit the configMap after the app is installed.

cbron: So the below `with` statement would only happen when used by open source users ? Technically you could hit this through the UI you would just have to manually input in the yaml editor.

cbron: If it's not hard I think we should remove the comments if the values that fill out this file are provided.

jiaqiluo: You are right, Rancher users can edit it as YAML to add those fields when installing the app in the Cluster Explorer.

Now the logic is that the comment will be added if neither `Values.sachet.providers` nor `Values.sachet.receivers` exits.

cbron (approved):

cbron (approved):

PR #1144, opened 2021-04-23
Longhorn v1.1.1 release for branch "dev-v2.5-source"

cclhsu:
Longhorn v1.1.1 dev 2.5

Signed-off-by: Clark Hsu <clark.hsu@suse.com>

### DIFF
--- a/packages/longhorn/generated-changes/overlay/crds/crds.yaml
+++ b/packages/longhorn/generated-changes/overlay/crds/crds.yaml
@@ -3,10 +3,10 @@ kind: CustomResourceDefinition
 metadata:
   labels:
     app.kubernetes.io/name: longhorn
-    helm.sh/chart: longhorn-1.1.0
+    helm.sh/chart: longhorn-1.1.1
     app.kubernetes.io/managed-by: Helm
     app.kubernetes.io/instance: longhorn
-    app.kubernetes.io/version: v1.1.0
+    app.kubernetes.io/version: v1.1.1
     longhorn-manager: Engine
   name: engines.longhorn.io
 spec:
@@ -59,10 +59,10 @@ kind: CustomResourceDefinition
 metadata:
   labels:
     app.kubernetes.io/name: longhorn
-    helm.sh/chart: longhorn-1.1.0
+    helm.sh/chart: longhorn-1.1.1
     app.kubernetes.io/managed-by: Helm
     app.kubernetes.io/instance: longhorn
-    app.kubernetes.io/version: v1.1.0
+    app.kubernetes.io/version: v1.1.1
     longhorn-manager: Replica
   name: replicas.longhorn.io
 spec:
@@ -119,10 +119,10 @@ kind: CustomResourceDefinition
 metadata:
   labels:
     app.kubernetes.io/name: longhorn
-    helm.sh/chart: longhorn-1.1.0
+    helm.sh/chart: longhorn-1.1.1
     app.kubernetes.io/managed-by: Helm
     app.kubernetes.io/instance: longhorn
-    app.kubernetes.io/version: v1.1.0
+    app.kubernetes.io/version: v1.1.1
     longhorn-manager: Setting
   name: settings.longhorn.io
 spec:
@@ -156,10 +156,10 @@ kind: CustomResourceDefinition
 metadata:
   labels:
     app.kubernetes.io/name: longhorn
-    helm.sh/chart: longhorn-1.1.0
+    helm.sh/chart: longhorn-1.1.1
     app.kubernetes.io/managed-by: Helm
     app.kubernetes.io/instance: longhorn
-    app.kubernetes.io/version: v1.1.0
+    app.kubernetes.io/version: v1.1.1
     longhorn-manager: Volume
   name: volumes.longhorn.io
 spec:
@@ -216,10 +216,10 @@ kind: CustomResourceDefinition
 metadata:
   labels:
     app.kubernetes.io/name: longhorn
-    helm.sh/chart: longhorn-1.1.0
+    helm.sh/chart: longhorn-1.1.1
     app.kubernetes.io/managed-by: Helm
     app.kubernetes.io/instance: longhorn
-    app.kubernetes.io/version: v1.1.0
+    app.kubernetes.io/version: v1.1.1
     longhorn-manager: EngineImage
   name: engineimages.longhorn.io
 spec:
@@ -272,10 +272,10 @@ kind: CustomResourceDefinition
 metadata:
   labels:
     app.kubernetes.io/name: longhorn
-    helm.sh/chart: longhorn-1.1.0
+    helm.sh/chart: longhorn-1.1.1
     app.kubernetes.io/managed-by: Helm
     app.kubernetes.io/instance: longhorn
-    app.kubernetes.io/version: v1.1.0
+    app.kubernetes.io/version: v1.1.1
     longhorn-manager: Node
   name: nodes.longhorn.io
 spec:
@@ -324,10 +324,10 @@ kind: CustomResourceDefinition
 metadata:
   labels:
     app.kubernetes.io/name: longhorn
-    helm.sh/chart: longhorn-1.1.0
+    helm.sh/chart: longhorn-1.1.1
     app.kubernetes.io/managed-by: Helm
     app.kubernetes.io/instance: longhorn
-    app.kubernetes.io/version: v1.1.0
+    app.kubernetes.io/version: v1.1.1
     longhorn-manager: InstanceManager
   name: instancemanagers.longhorn.io
 spec:
@@ -376,10 +376,10 @@ kind: CustomResourceDefinition
 metadata:
   labels:
     app.kubernetes.io/name: longhorn
-    helm.sh/chart: longhorn-1.1.0
+    helm.sh/chart: longhorn-1.1.1
     app.kubernetes.io/managed-by: Helm
     app.kubernetes.io/instance: longhorn
-    app.kubernetes.io/version: v1.1.0
+    app.kubernetes.io/version: v1.1.1
     longhorn-manager: ShareManager
   name: sharemanagers.longhorn.io
 spec:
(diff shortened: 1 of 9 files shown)


shuo-wu (changes requested):
1. Remove the modification for other packages ` packages/rancher-grafana`
2. Should not remove  `packages/longhorn/generated-changes/patch/Chart.yaml.patch` and ` packages/longhorn/generated-changes/patch/README.yaml.patch`. Instead, you need to update them.
3. New CRDs are missing in `packages/longhorn/generated-changes/overlay/crds/crds.yaml`
4. Please rename the commit title. We are updating for dev-v2.5 rather than v2.6. And this is the v1.1.1 release.


Here are the patches I just generate. You can check the difference:
https://github.com/rancher/charts/compare/dev-v2.5-source...shuo-wu:dev-v2.5-source

shuo-wu:
Notice that the image repos (`rancher/longhornio-xxxxxxx`) in `question.yml` are different from those (`rancher/mirrored-longhornio-xxxxxxx`) in `value.yml`. @cbron Should we make them consistent?

packages/longhorn/package.yaml
```diff
@@ -1,7 +1,7 @@
 url: https://github.com/longhorn/charts.git
 subdirectory: charts/longhorn
-commit: 299f4ab8ad01137c463574fe013da30efc6b05d0
-packageVersion: 01
+commit: 93f250e4a175044d34c6982ea6168b213caeee9f
```
jenting, line 3: Use `commit: v1.1.1` would be easier for future maintenance.

packages/longhorn/package.yaml
```diff
@@ -1,7 +1,7 @@
...
+packageVersion: 00
 releaseCandidateVersion: 01
```
shuo-wu, line 5: This version should be 00.

packages/longhorn/generated-changes/patch/questions.yml.patch
```diff
@@ -18,15 +18,6 @@
...
      label: Longhorn Engine Image Repository
-@@ -35,7 +35,7 @@
-     label: Longhorn Engine Image Tag
-     group: "Longhorn Images Settings"
-   - variable: image.longhorn.ui.repository
--    default:  longhornio/longhorn-ui
-+    default:  rancher/longhornio-longhorn-ui
-     description: "Specify Longhorn UI Image Repository"
-     type: string
-     label: Longhorn UI Image Repository
```
shuo-wu, line 29: Why removing this?

shuo-wu (changes requested):

packages/longhorn/generated-changes/patch/questions.yml.patch
cclhsu: We need to remove it?

nickgerace (approved):
This looks fine to me from a Helm syntax perspective. The CRDs are supposed to be in the overlay though? They do not come with the chart?

cbron:
@shuo-wu / @yasker is this aiming to get into 2.5.8 ? We are already in the late QA stages for that.

yasker:
@cbron Yes. Longhorn v1.1.1 has been released. The chart should be ready very soon. cc @innobead

packages/longhorn/package.yaml
cclhsu: Though unusually, tag can be re-tag to a different commit hash and content.  This might cause some confusion.

shuo-wu: Ideally, the tag is immutable from the perspective of project management. You don't need to worry about that part.

jenting (approved):
LGTM

shuo-wu (dismissed):
LGTM

shuo-wu:
> This looks fine to me from a Helm syntax perspective. The CRDs are supposed to be in the overlay though? They do not come with the chart?

Yes. The CRDs are in the overlay right now. Please take a final check.

packages/longhorn/package.yaml
cbron: Our tooling doesn't support tag yet, just sha.

packages/longhorn/generated-changes/patch/values.yaml.patch
```diff
@@ -1,28 +1,32 @@
...
        tag: v1_20201216
      shareManager:
 -      repository: longhornio/longhorn-share-manager
 +      repository: rancher/mirrored-longhornio-longhorn-share-manager
-       tag: v1_20201204
+       tag: v1_20210416
+     backingImageManager:
+-      repository: longhornio/backing-image-manager
++      repository: rancher/mirrored-longhornio-backing-image-manager
+       tag: v1_20210422
```
cbron, line 29: Can't merge until this image exists

cbron:
Issue: https://github.com/rancher/rancher/issues/32369

packages/longhorn/generated-changes/patch/questions.yml.patch
```diff
@@ -48,13 +48,22 @@
 @@ -71,7 +71,7 @@
      label: Longhorn Share Manager Image Tag
      group: "Longhorn Images Settings"
+   - variable: image.longhorn.backingImageManager.repository
+-    default: longhornio/backing-image-manager
++    default: rancher/longhornio-backing-image-manager
```
cbron, line 53: I think all the images in this file should default to the `mirrored-` ones. I'm curious if this default has been causing people to have problems with airgap. Whatever is in values.yaml is what gets added to airgap images.txt.

packages/longhorn/generated-changes/overlay/crds/crds.yaml
```diff
@@ -3,10 +3,10 @@ kind: CustomResourceDefinition
 metadata:
```
cbron, line 3: I didn't deep dive into the CRD's or the version to ensure they were correct, I am assuming QA will find those things.

innobead:
cc @khushboo-rancher

meldafrawi:
LGTM

packages/longhorn/generated-changes/patch/Chart.yaml.patch
```diff
@@ -1,9 +1,9 @@
 --- charts-original/Chart.yaml
 +++ charts/Chart.yaml
 @@ -26,3 +26,12 @@
- - name: Sheng Yang
-   email: sheng@yasker.org
- icon: https://raw.githubusercontent.com/cncf/artwork/master/projects/longhorn/icon/color/longhorn-icon-color.svg?sanitize=true
+ - name: Longhorn maintainers
+   email: maintainers@longhorn.io
+ icon: https://raw.githubusercontent.com/cncf/artwork/master/projects/longhorn/icon/color/longhorn-icon-color.png
 +annotations:
```
PhanLe1010, line 7: We want to support Longhorn installation in Rancher windows cluster. So we may want to remove this annotation: `catalog.cattle.io/os: linux`

cbron: Yes if it supports both windows and linux you would remove that annotation

packages/longhorn/generated-changes/patch/values.yaml.patch
cbron: exists now

packages/longhorn/generated-changes/patch/Chart.yaml.patch
innobead: Let's wait for @cclhsu update before merging.

yasker: Just to be clear. Longhorn doesn't support Windows. The support we're adding is to allow Longhorn to be installed in the cluster with mixed Linux and WIndows node.

So we should NOT remove this annotation. @cbron @PhanLe1010 please confirm the understanding is right here.

cbron: Windows here would imply that it works in some way with the windows nodes. Both logging and monitoring interact with the windows nodes. If longhorn is not interacting with the windows nodes and just tolerates them then keep the Linux annotation. 

Also wait to merge until I or another committer approve, there are a lot of requirements around charts in rancher releases and I need to ensure they are met.

packages/longhorn/package.yaml
jenting: In Git, the tag implies the SHA commit ID.
I tried to use the tag, it works the same as using SHA commit ID.

packages/longhorn/generated-changes/patch/Chart.yaml.patch
PhanLe1010: @yasker @cbron 

Yes, Longhorn doesn't run on Windows nodes. We want to allow the user to install Longhorn on Linux nodes in a cluster that have both Linux and Windows nodes.

If we don't the remove `catalog.cattle.io/os: linux` annotation, Longhorn chart will not be shown in the new Rancher UI when the cluster have both Windows nodes and Linux nodes. So, user cannot install Longhorn even if their cluster have some Linux nodes.

Before removing the `catalog.cattle.io/os: linux` annotation, new Rancher UI shows:
Screen Shot 2021-04-27 at 8 25 05 PM

After removing the `catalog.cattle.io/os: linux` annotation, new Rancher UI shows:
Screen Shot 2021-04-27 at 8 24 45 PM

innobead: > My understanding is that we still show Longhorn chart in Windows cluster. When users install Longhorn, they need to specify toleration and node selector so that Longhorn components is only deployed on the Linux nodes as described in the documentation at https://longhorn.io/docs/1.1.1/advanced-resources/deploy/rancher_windows_cluster/

This is what I expect. @yasker any further concerns?

innobead: BTW, this frontend implementation is from the below snippets, so obviously if you don't remove this annotation, there is no chance to install longhorn from the hybrid (windows+linux) cluster. (note: the OS annotation is exclusive or empty to allow in the hybrid cluster mode)

https://github.com/rancher/dashboard/blob/176ffcedfd57af42dab27b5c9c7723e8eab1590f/store/catalog.js#L527-L533 and https://github.com/rancher/dashboard/blob/176ffcedfd57af42dab27b5c9c7723e8eab1590f/store/catalog.js#L477-L491

cclhsu: @khushboo-rancher @meldafrawi I have updated the test branch  to remove the `linux` annotation.
```
repo: https://github.com/cclhsu/charts
branch: dev-v2.5
```
Please help validate.  Thanks!

yasker: @PhanLe1010 I see. It makes sense. Do we have documents on how users can deploy Longhorn in a mixed environment?

@cbron Can you confirm this is expected behavior? So adding the label means it's going to be installed in Linux only cluster. And if we want to support an environment mixed with Windows and Linux, then we need to remove the annotation and ask users to add node selectors to the chart.

khushboo-rancher: @cbron Is there a middle way to have option to install Longhorn on a Linux or mixed cluster and doesn't show this for a cluster having only windows nodes?

PhanLe1010: >  Do we have documents on how users can deploy Longhorn in a mixed environment?
Yeah, we do have the document at https://longhorn.io/docs/1.1.1/advanced-resources/deploy/rancher_windows_cluster/

vincent99: @khushboo-rancher There is no such thing as a cluster with only Windows nodes; you need Linux for the system services.  That's why the filtering is just by "is the cluster Windows or not".

cbron: This is technically ok, as long as users don't get confused. They might see it show up in a windows cluster and expect it to work on windows nodes. There will be no way for them to know which way it works.

yasker: TLDR: After talking with @cbron , we agree we should keep the annotation there to prevent Longhorn from showing up for the Windows/Linux mixed cluster.

There were some misunderstandings regarding the original request filed by @cbron at https://github.com/rancher/rancher/issues/29220 . The requirement is: Any Rancher featured chart shouldn't be immediately broken when installing in a Windows/Linux mixed cluster, e.g. trying to run a Linux container on top of Windows nodes.

However, Rancher also doesn't recommend running any featured charts for a system it's not designed for. In this case, obviously, Longhorn isn't designed for Windows, so we shouldn't remove the annotation.

But if Rancher doesn't recommend that, why is it required? It's because users can still deploy Helm chart without Rancher UI, e.g. via API/Terraform/Helm. Those ways are not blocked in any sense. There are users who want to do that. But since it's not a common case to deploy Linux specific workload in a Windows/Linux mixed cluster, as the current policy, we don't encourage the user to do it and we want to make sure user knows what they're doing. If they choose to deploy the chart via API/Terraform/Helm, we reasonably believe that they know what they're doing.

If a chart just shows up in user's Windows/Linux mixed cluster, the users might just click and deploy it, then error out - since they need to set related `NodeSelector` and `TaintToleration` manually. It's not a good experience anyway. So it's easier to just not showing it unless the users explicitly deploy it via other means.

Longhorn is not designed for Windows. So we think it's better if we keep the annotation `catalog.cattle.io/os: linux`.

@PhanLe1010 @cclhsu Let me know what do you think.

PhanLe1010: Thanks @yasker ! I agree with the conclusion

cbron (commented):
Lgtm. After annotation is back in we can merge.

packages/longhorn/package.yaml
```diff
@@ -1,8 +1,8 @@
 url: https://github.com/longhorn/charts.git
 subdirectory: charts/longhorn
-commit: 299f4ab8ad01137c463574fe013da30efc6b05d0
-packageVersion: 01
-releaseCandidateVersion: 01
+commit: 93f250e4a175044d34c6982ea6168b213caeee9f
+packageVersion: 00
+releaseCandidateVersion: 00
```
cbron, line 5: This should be 01

cbron (changes requested):
Actually 1 more minor thing then good.

packages/longhorn/package.yaml
yasker: Just to clarify @cbron , what does this `releaseCandidateVersion` mean? Also, does it apply to `packageVersion` too?

cbron: Package version of 00 will turn 1.1.1.1 into 1.1.1.100 which is correct. 
releaseCandidateVersion of 01 will turn 1.1.1.100 into 1.1.1.100-rc01, which is also correct

yasker: This is the formal release, so it shouldn't be `rc01`. so the original PR looks good?

cclhsu: changed.

cclhsu:
I have added back linux annotation, changed releaseCandidateVersion, and squashed the commits.

packages/longhorn/package.yaml
cbron: For right now everything in feature charts (this branch) has an rc until it goes live. That transition happens automatically on rancher release. That process will be going away in the future.

cbron (approved):

cbron (approved):

PR #1150, opened 2021-04-24
Small updates for Monitoring

aiyengar2:
Comments out global.cattle.windows.enabled so that users can set it to false in Rancher Dashboard UI.

Adds dummy `os` values to support marking images as Windows images for private registry script.

Reverts upstream change to set prometheusOperator.tls.internalPort to 10250 to avoid conflicts with kubelet in GKE clusters. The port has now gone back to the original 8443 and users of GKE private clusters are expected to open up this port in their firewall rules manually to support admission webhooks.

Related Issues: https://github.com/rancher/rancher/issues/32274, https://github.com/rancher/rancher/issues/32265, https://github.com/rancher/rancher/issues/32280, https://github.com/rancher/rancher/issues/32201


aiyengar2:
I'm unsure why, but CI for a tgz-based package seems to be failing for some reason when trying to patch rancher-wins-upgrader. https://github.com/rancher/charts/pull/1150/checks?check_run_id=2432884734

Switching to a GIt-based package seems to have done the trick, so the last commit pushed to this PR should resolve the CI issue.

nickgerace (approved):

brendarearden (approved):
LGTM

PR #1152, opened 2021-04-26
Remove rancher-operator

thedadams:
rancher-operator is no longer needed in Rancher 2.6+, so it should be
removed.

Issue:
https://github.com/rancher/rancher/issues/32179


cbron (approved):
lgtm, most likely will need to remove from assets branch afterwards unless charts changes do that for us

PR #1153, opened 2021-04-26
Remove rancher-operator

thedadams:
rancher-opeartor is no longer neded in Rancher 2.6+, so it should be
removed.


cbron (approved):
Interesting that we are deleting the old ones too. I suppose that makes sense.

PR #1155, opened 2021-04-27
support install rancher-alerting-drivers in hardened cluster

jiaqiluo:
Issues:
https://github.com/rancher/rancher/issues/32358
https://github.com/rancher/rancher/issues/32665

Notes about the PR:
- The chart is tested on a custom hardened cluster under the [CIS 1.6 Benchmark](https://rancher.com/docs/rancher/v2.x/en/security/rancher-2.5/1.6-hardening-2.5/). 
- The app is installed into the hardened cluster successfully.
- After the app is installed, pass the CIS scan with the profile `rke-profile-hardened-1.6`

### DIFF
--- a/packages/rancher-alerting-drivers/charts/templates/_helpers.tpl
+++ b/packages/rancher-alerting-drivers/charts/templates/_helpers.tpl
@@ -1,3 +1,29 @@
+{{- define "system_default_registry" -}}
+{{- if .Values.global.cattle.systemDefaultRegistry -}}
+{{- printf "%s/" .Values.global.cattle.systemDefaultRegistry -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Windows cluster will add default taint for linux nodes,
+add below linux tolerations to workloads could be scheduled to those linux nodes
+*/}}
+
+{{- define "linux-node-tolerations" -}}
+- key: "cattle.io/os"
+  value: "linux"
+  effect: "NoSchedule"
+  operator: "Equal"
+{{- end -}}
+
+{{- define "linux-node-selector" -}}
+{{- if semverCompare "<1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+beta.kubernetes.io/os: linux
+{{- else -}}
+kubernetes.io/os: linux
+{{- end -}}
+{{- end -}}
+
 {{/*
 Expand the name of the chart.
 */}}
--- a/packages/rancher-alerting-drivers/charts/templates/hardened.yaml
+++ b/packages/rancher-alerting-drivers/charts/templates/hardened.yaml
@@ -1,0 +1,116 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ include "drivers.fullname" . }}-patch-sa
+  namespace: {{ .Release.Namespace }}
+  labels: {{ include "drivers.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": post-install, post-upgrade
+    "helm.sh/hook-delete-policy": hook-succeeded, before-hook-creation
+spec:
+  backoffLimit: 1
+  template:
+    spec:
+      serviceAccountName: {{ include "drivers.fullname" . }}-patch-sa
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 1000
+      restartPolicy: Never
+      nodeSelector: {{ include "linux-node-selector" . | nindent 8 }}
+      tolerations: {{ include "linux-node-tolerations" . | nindent 8 }}
+      containers:
+        - name: {{ include "drivers.fullname" . }}-patch-sa
+          image: "{{ include "system_default_registry" . }}{{ .Values.global.kubectl.repository }}:{{ .Values.global.kubectl.tag }}"
+          imagePullPolicy: IfNotPresent
+          command: ["kubectl", "-n", {{ .Release.Namespace | quote }}, "patch", "serviceaccount", "default", "-p", "{\"automountServiceAccountToken\": false}"]
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "drivers.fullname" . }}-patch-sa
+  namespace: {{ .Release.Namespace }}
+  labels: {{ include "drivers.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": post-install, post-upgrade
+    "helm.sh/hook-delete-policy": hook-succeeded, before-hook-creation
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ include "drivers.fullname" . }}-patch-sa
+  labels: {{ include "drivers.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": post-install, post-upgrade
+    "helm.sh/hook-delete-policy": hook-succeeded, before-hook-creation
+rules:
+  - apiGroups: [""]
+    resources: ["serviceaccounts"]
+    verbs: ["get", "patch"]
+  - apiGroups: ["policy"]
+    resources: ["podsecuritypolicies"]
+    verbs:     ["use"]
+    resourceNames:
+      - {{ include "drivers.fullname" . }}-patch-sa
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ include "drivers.fullname" . }}-patch-sa
+  labels: {{ include "drivers.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": post-install, post-upgrade
+    "helm.sh/hook-delete-policy": hook-succeeded, before-hook-creation
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ include "drivers.fullname" . }}-patch-sa
+subjects:
+  - kind: ServiceAccount
+    name: {{ include "drivers.fullname" . }}-patch-sa
+    namespace: {{ .Release.Namespace }}
+---
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: {{ include "drivers.fullname" . }}-patch-sa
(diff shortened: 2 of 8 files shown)


nickgerace (approved):
Left a nit, but LGTM

packages/rancher-alerting-drivers/charts/templates/_helpers.tpl
```diff
@@ -1,3 +1,38 @@
...
+Windows cluster will add default taint for linux nodes,
+add below linux tolerations to workloads could be scheduled to those linux nodes
+*/}}
+
+{{- define "linux-node-tolerations" -}}
+- key: "cattle.io/os"
+  value: "linux"
+  effect: "NoSchedule"
+  operator: "Equal"
+{{- end -}}
```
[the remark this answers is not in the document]
jiaqiluo: We should be fine because pods will still be assigned to nodes if the target taint does not exist. 
Those charts are curated for Rancher's clusters, so we don't have to put them in values.yaml for now. If we decide to move then we need to do it for all charts :)

packages/rancher-alerting-drivers/charts/templates/network-policy-allow-all.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: networking.k8s.io/v1
```
aiyengar2, line 1: Instead of adding the network policy, serviceaccount, psp, and roles manually in these files, let's just copy what we have for Monitoring today here:

https://github.com/rancher/charts/blob/dev-v2.5-source/packages/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/hardened.yaml

That way we can standardize this across feature charts, similar to what we do today for helpers.tpl functions.

aiyengar2: You still need the serviceaccount, rolebinding, and PSP to allow your deployments to be deployed in a hardened cluster, but this ^^ takes care of the scan requirements.

aiyengar2 (changes requested):

packages/rancher-alerting-drivers/charts/templates/network-policy-allow-all.yaml
jiaqiluo: changes are made based on the above comments.

nickgerace (approved):
LGTM after Arvind's changes

packages/rancher-alerting-drivers/package.yaml
```diff
@@ -1,3 +1,3 @@
-packageVersion: 00
-releaseCandidateVersion: 04
+packageVersion: 01
+releaseCandidateVersion: 00
```
cbron, line 2: Nit: 01

packages/rancher-alerting-drivers/charts/templates/hardened.yaml
```diff
@@ -0,0 +1,117 @@
...
+        runAsNonRoot: true
+        runAsUser: 1000
+      restartPolicy: Never
+      nodeSelector: {{ include "linux-node-selector" . | nindent 8 }}
+      tolerations: {{ include "linux-node-tolerations" . | nindent 8 }}
+      containers:
+        - name: {{ include "drivers.fullname" . }}-patch-sa
+          image: "{{ include "system_default_registry" . }}{{ .Values.global.kubectl.repository }}:{{ .Values.global.kubectl.tag }}"
+          imagePullPolicy: IfNotPresent
+          command: ["kubectl", "patch", "serviceaccount", "default", "-p", "{\"automountServiceAccountToken\": false}"]
```
cbron, line 25: Is this the only way to make hardened cluster pass? By changing the default service account ? I haven’t seen this pattern before. cc @aiyengar2

packages/rancher-alerting-drivers/package.yaml
jiaqiluo: resoved.

packages/rancher-alerting-drivers/charts/templates/hardened.yaml
aiyengar2: @cbron you need to patch the default service account in any namespace that you create as part of the chart to be un-automounted. Declared as a req in https://rancher.com/docs/rancher/v2.x/en/security/rancher-2.5/1.6-hardening-2.5/#set-automountserviceaccounttoken-to-false-for-default-service-accounts.

aiyengar2:
@jiaqiluo the CI issue is due to an unrelated problem tracked inhttps://github.com/rancher/charts/pull/1190. Thanks for filing the issue!

aiyengar2 (dismissed):
LGTM

jiaqiluo:
@cbron  Need your final approval then this PR can be merged.

packages/rancher-alerting-drivers/charts/templates/hardened.yaml
cbron: Why is the namespace in the arg instead of the command ? Also is specifying the namespace required if the SA is from the same namespace ? I'm not sure on that one, depends how default SA and kubectl act.

packages/rancher-alerting-drivers/charts/templates/hardened.yaml
```diff
@@ -0,0 +1,117 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ include "drivers.fullname" . }}-patch-sa
+  namespace: {{ .Release.Namespace }}
+  labels: {{ include "drivers.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": post-install, post-upgrade
```
cbron, line 8: Should this be pre-install/upgrade, as the default SA already exists. That way it gets edited before the custom SA gets installed.

packages/rancher-alerting-drivers/charts/templates/hardened.yaml
```diff
@@ -0,0 +1,117 @@
...
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "drivers.fullname" . }}-patch-sa
+  namespace: {{ .Release.Namespace }}
+  labels: {{ include "drivers.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": post-install, post-upgrade
+    "helm.sh/hook-delete-policy": hook-succeeded, before-hook-creation
```
cbron, line 36: In same vein as above comment, why are these in hooks, can't they be installed at same time as everything else ?

cbron (commented):
Logic is fine, just some questions on timing.

packages/rancher-alerting-drivers/charts/templates/hardened.yaml
jiaqiluo: The edit on the default SA can happen at either pre or post of the install/upgrade. 
I can move it to pre-install/upgrade if that makes the logic more clear.

packages/rancher-alerting-drivers/charts/templates/hardened.yaml
jiaqiluo: We put them in the hooks because editing the default SA is a one-time operation so there is no need to keep those resources (job, ServiceAccount, ClusterRole, ClusterRoleBinding) in the cluster after the operation is done.

packages/rancher-alerting-drivers/charts/templates/hardened.yaml
jiaqiluo: The job template is from https://github.com/rancher/charts/blob/dev-v2.5-source/packages/rancher-cis-benchmark/charts/templates/patch_default_serviceaccount.yaml

I agree that it is not necessary to pass the ns as an arg. I will update the PR.

aiyengar2 (approved):

packages/rancher-alerting-drivers/charts/templates/hardened.yaml
aiyengar2: It's post-install since Monitoring V2 needed this job to run after the Namespace resource was created by Helm. It's not a requirement for Alerting Drivers tho, since Alerting Drivers doesn't create namespaced Helm resources.

We could change it, but IMO we should just keep it the same if possible.

packages/rancher-alerting-drivers/charts/templates/hardened.yaml
cbron: ok lets go with passing explicitly in the command

packages/rancher-alerting-drivers/charts/templates/hardened.yaml
cbron: Ahh, that makes sense. Then its only a question of pre vs post.

packages/rancher-alerting-drivers/charts/templates/hardened.yaml
cbron: I'm ok with post, on one stipulation. I don't want to run into a race with hardened cluster where the default SA is attempted to be used before it gets patched away.

jiaqiluo: The namespace is created either by Rancher UI or the user prior to installing the chart, so we do not know if the default SA is used by others or not.  What we can confirm is that the rancher-alerting-driver app does not use the default SA. 

As to the case where the default SA is attempted to be used during the period between starting installing the rancher-alerting-drivers app and executing the post hook, how is it different from the default SA is attempted to be used at any time prior to installing the app? IMO, we do not need to worry about it, or that is something the system, not this app, needs to handle.

packages/rancher-alerting-drivers/charts/templates/hardened.yaml
jiaqiluo: Fixed: the arg values were moved to the command.

packages/rancher-alerting-drivers/charts/templates/hardened.yaml
cbron: Oh, I was thinking about this wrong. The default SA is only applied if the app doesn't specify a service account. As long as it does the default SA won't be used.

cbron (approved):

aiyengar2 (changes requested):

jiaqiluo:
close in favor of https://github.com/rancher/charts/pull/1216

PR #1166, opened 2021-04-29
Updating the rancher-windows-exporter chart to use windows prefix path.

phillipsj:
This part of the fix for the bug found [here](https://github.com/rancher/dashboard/issues/2837) with the windows prefix path.

Related Issue: https://github.com/rancher/rancher/issues/32194


nickgerace (dismissed):

packages/rancher-windows-exporter/charts/templates/daemonset.yaml
```diff
@@ -45,7 +45,7 @@ spec:
         - name: wins-pipe
           mountPath: \\.\pipe\rancher_wins
         - name: binary-host-path
-          mountPath: c:/host/etc/windows-exporter
+          mountPath: {{ regexReplaceAll ".:/" .Values.global.cattle.rkeWindowsPathPrefix  "${1}C:/host/"}}etc/windows-exporter
```
aiyengar2, line 48: No need to do the `regexReplaceAll` here; `C:/host/etc/windows-exporter` can be hard-coded.

This is the same logic for why we didn't need to modify the [mountPath in rancher-wins-upgrader](https://github.com/rancher/wins/blob/4bf76f4ecbc0ea7b835a4080d36705737a41559a/charts/rancher-wins-upgrader/templates/daemonset.yaml#L40); this path is just what the **container** sees (and what the scripts reference).

All that needs to be changed is the directory that this containerPath points to; e.g. the hostPath modified below.

packages/rancher-windows-exporter/charts/templates/daemonset.yaml
```diff
@@ -66,7 +66,7 @@ spec:
           path: \\.\pipe\rancher_wins_proxy
       - name: binary-host-path
         hostPath:
-          path: c:/etc/windows-exporter
+          path: {{ .Values.global.cattle.rkeWindowsPathPrefix }}etc/windows-exporter
```
aiyengar2, line 69: Looks good, but can you add validation of the rkeWindowsPathPrefix value in this PR too?

e.g. add [`{{ include "winsUpgrader.validatePathPrefix" . }}`](https://github.com/rancher/wins/blob/4bf76f4ecbc0ea7b835a4080d36705737a41559a/charts/rancher-wins-upgrader/templates/configmap.yaml#L1) to the top of this file and add that function to the [`_helpers.tpl`](https://github.com/rancher/wins/blob/4bf76f4ecbc0ea7b835a4080d36705737a41559a/charts/rancher-wins-upgrader/templates/_helpers.tpl#L32-L39).

aiyengar2 (changes requested):

packages/rancher-windows-exporter/charts/templates/daemonset.yaml
phillipsj: Done

aiyengar2 (approved):
LGTM! 😄

aiyengar2:
@phillipsj CI is failing because we didn’t bump the rcVersion of Monitoring V2 as well.

Can you bump the releasedCandidateVersion in packages/rancher-monitoring/package.yaml as part of this change as well?

phillipsj:
@aiyengar2 I have bumped that version.

luthermonson:
https://github.com/rancher/rancher/issues/32194

aiyengar2 (approved):

nickgerace (approved):

PR #1167, opened 2021-04-29
Drop assets and charts that should not be released

aiyengar2:
Need sign-off from relevant devs to make sure these are the assets that need to be dropped from the release.

NOTE: These are assets that should *not* be released as part of Rancher 2.5.8. e.g. Longhorn 1.1.0 was never intended for release, we only wanted to release Longhorn 1.1.1, so all 1.1.0 rcs are being dropped.

Once these are dropped from assets/ and charts/, I will make a dummy PR to dev-v2.5-source triggering a `make stage-release`, which will automatically update the index.yaml and released/ directory entirely to remove these assets from dev-v2.5 as a whole.

Related Issue: https://github.com/rancher/rancher/issues/31594


aiyengar2:
Second commit was trashed via script:

```bash
trash $(git status --porcelain | cut -d'/' -f2-3 | sed 's/.tgz//' | sed 's/\(.*\)-\([[:digit:]].*\)/charts\/\1\/\2/' | xargs)
```

nickgerace (approved):

aiyengar2:
https://github.com/rancher/charts/pull/1168 is the dummy PR that will be merged to regenerate released/ and index.yaml. Will merge that in once this is given all approvals from feature chart owners.

cbron:
I can confirm longhorn one should be dropped. THat release exists: https://github.com/rancher/charts/tree/release-v2.5/assets/longhorn
It was only produced because at that point in time, no other new chart existed.

prachidamle (approved):

brendarearden:
assets/rancher-istio/rancher-istio-1.8.300-rc01.tgz  << this is just getting rid of the duplicate one, correct? not the released 1.8.300 version one that was release with v2.5.7?

aiyengar2:
@brendarearden right, even if it deletes something that is already released that’s not a problem. We’ll always pull released charts and then overlay the contents from assets on top of it

brendarearden (approved):
lgtm

StrongMonkey:
@aiyengar2 No need to touch fleet v0.3.5-rcxx, correct?

thedadams (approved):

StrongMonkey (approved):

PR #1175, opened 2021-05-03
Preserve structural schema during preserveUnknownFields step if possible

aiyengar2:
Related Issue: https://github.com/rancher/rancher/issues/32499

### DIFF
--- a/packages/rancher-monitoring/templates/crd-template/templates/jobs.yaml
+++ b/packages/rancher-monitoring/templates/crd-template/templates/jobs.yaml
@@ -30,7 +30,12 @@ spec:
             {{- range $path, $_ :=  (.Files.Glob  "crd-manifest/**.yaml") }}
             {{- $crd := get (get ($.Files.Get $path | fromYaml) "metadata") "name" }}
             if [[ -n "$(kubectl get crd {{ $crd }} -o jsonpath='{.spec.preserveUnknownFields}')" ]]; then
-              kubectl patch crd {{ $crd }} -p '{"spec": {"preserveUnknownFields": false, "versions": [{"name": "v1", "served": false, "storage": true}]}}' --type="merge" || true;
+                if [[ -n "$(kubectl get crd {{ $crd }} -o jsonpath='{.spec.versions[0].schema}' 2>&1)" ]]; then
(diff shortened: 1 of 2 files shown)


packages/rancher-monitoring/templates/crd-template/templates/jobs.yaml
```diff
@@ -30,7 +30,12 @@ spec:
             {{- range $path, $_ :=  (.Files.Glob  "crd-manifest/**.yaml") }}
             {{- $crd := get (get ($.Files.Get $path | fromYaml) "metadata") "name" }}
             if [[ -n "$(kubectl get crd {{ $crd }} -o jsonpath='{.spec.preserveUnknownFields}')" ]]; then
-              kubectl patch crd {{ $crd }} -p '{"spec": {"preserveUnknownFields": false, "versions": [{"name": "v1", "served": false, "storage": true}]}}' --type="merge" || true;
+                if [[ -n "$(kubectl get crd {{ $crd }} -o jsonpath='{.spec.versions[0].schema}' 2>&1)" ]]; then
+                  patch="{\"spec\": {\"preserveUnknownFields\": false}}"
+                else
+                  patch="{\"spec\": {\"preserveUnknownFields\": false, \"versions\": [{\"name\": \"v1\", \"served\": false, \"storage\": true}]}}"
```
aiyengar2, line 36: We only need this patch when the schema does not exist since `preserveUnknownFields=false` requires a schema. This is the case for when Monitoring V2 installs onto a clean Rancher cluster, since Rancher does not define a schema.

However, when a schema exists, all we need to do is update `preserveUnknownFields`. This is the case for clusters that have enabled Monitoring V1 and then disabled it before installing Monitoring V2.

nickgerace: Rather an if/else, should we set patch and then update it if the condition is met?

aiyengar2: Sure

brendarearden (approved):
LGTM

nickgerace (approved):

PR #1178, opened 2021-05-04
Put binary in prefixPath within container

aiyengar2:
The reason why this change is necessary is that wins performs a checksum on the exact same path that must exist in the container and must exist on the host. Therefore, even though there is no need for the container to have knowledge of the CATTLE_PREFIX_PATH, it must for wins to be able to checksum correctly.

Related Issue: https://github.com/rancher/rancher/issues/32194, https://github.com/rancher/rancher/issues/32530

Wins PR: https://github.com/rancher/wins/pull/39

### DIFF
--- a/packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
+++ b/packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
@@ -294,8 +294,8 @@
 +    ## Every Windows host must have a wins version of 0.1.0+ to use this chart (default as of Rancher 2.5.8).
 +    ## To upgrade wins versions on Windows hosts, see https://github.com/rancher/wins/tree/master/charts/rancher-wins-upgrader.
 +    ##
-+    # windows:
-+    #   enabled: false
++    windows:
++      enabled: false
 +  kubectl:
 +     repository: rancher/kubectl
 +     tag: v1.20.2
--- a/packages/rancher-windows-exporter/charts/scripts/run.ps1
+++ b/packages/rancher-windows-exporter/charts/scripts/run.ps1
@@ -36,9 +36,43 @@ function Transfer-File
     $null = Copy-Item -Force -Path $Src -Destination $Dst
 }
 
-# Copy binary
-Create-Directory -Path "c:\host\etc\windows-exporter"
-Transfer-File -Src "c:\etc\windows-exporter\windows-exporter.exe" -Dst "c:\host\etc\windows-exporter\windows-exporter.exe"
+# Copy binary into host
+Create-Directory -Path "C:\host\etc\windows-exporter"
+Transfer-File -Src "C:\etc\windows-exporter\windows-exporter.exe" -Dst "C:\host\etc\windows-exporter\windows-exporter.exe"
 
-# Run wins
-Invoke-Expression -Command c:\entry.ps1
+# Copy binary into prefix path, since wins expects the same path on the host and on the container
+$prefixPath = 'C:\'
+if ($env:CATTLE_PREFIX_PATH) {
+    $prefixPath = $env:CATTLE_PREFIX_PATH
+}
+$winsDirPath = $('{0}etc\windows-exporter' -f $prefixPath)
+$winsPath = $('{0}\windows-exporter.exe' -f $winsDirPath)
+
+Create-Directory -Path $winsDirPath
+Transfer-File -Src "C:\etc\windows-exporter\windows-exporter.exe" $winsPath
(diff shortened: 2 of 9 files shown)


packages/rancher-windows-exporter/charts/scripts/run.ps1
```diff
@@ -36,9 +36,43 @@ function Transfer-File
...
+}
+
+# format "UDP:4789 TCP:8080"
+$winsExposes = $('TCP:{0}' -f $listenPort)
+
+# format "--a=b --c=d"
+$winsArgs = $('--collectors.enabled={0} --telemetry.addr=:{1} --telemetry.max-requests={2} --telemetry.path=/metrics' -f $enabledCollectors, $listenPort, $maxRequests)
+
+
+wins.exe cli prc run --path $winsPath --exposes $winsExposes --args "$winsArgs"
```
aiyengar2, line 78: Copied from https://github.com/rancher/windows_exporter-package/blob/307e59926e9b0b6b43767360f62df53538786e0d/package/windows/entry.ps1#L5-L28.

I just needed to change the winsPath https://github.com/rancher/windows_exporter-package/blob/307e59926e9b0b6b43767360f62df53538786e0d/package/windows/entry.ps1#L2

nickgerace (approved):

luthermonson (dismissed):

packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
```diff
@@ -294,8 +294,8 @@
 +    ## Every Windows host must have a wins version of 0.1.0+ to use this chart (default as of Rancher 2.5.8).
 +    ## To upgrade wins versions on Windows hosts, see https://github.com/rancher/wins/tree/master/charts/rancher-wins-upgrader.
 +    ##
-+    # windows:
-+    #   enabled: false
++    windows:
++      enabled: false
```
aiyengar2, line 298: Uncommented this since leaving it commented out caused windowsExporter to be deployed by default.

packages/rancher-windows-exporter/charts/templates/configmap.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.clients }}{{ if .Values.clients.enabled }}
```
aiyengar2, line 1: Gated windows-exporter on `.Values.client.enabled` so disabling it on the rancher-monitoring level would allow you to disable deploying windows specific components

packages/rancher-windows-exporter/charts/templates/service.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if and .Values.serviceMonitor .Values.clients }}{{- if and .Values.serviceMonitor.enabled .Values.clients.enabled }}
```
aiyengar2, line 1: service should be deployed whether or not a serviceMonitor is enabled.

packages/rancher-windows-exporter/charts/scripts/run.ps1
```diff
@@ -36,9 +36,43 @@ function Transfer-File
...
+Transfer-File -Src "C:\etc\windows-exporter\windows-exporter.exe" -Dst "C:\host\etc\windows-exporter\windows-exporter.exe"
 
-# Run wins
-Invoke-Expression -Command c:\entry.ps1
+# Copy binary into prefix path, since wins expects the same path on the host and on the container
+$prefixPath = 'C:\'
+if ($env:CATTLE_PREFIX_PATH) {
+    $prefixPath = $env:CATTLE_PREFIX_PATH
+}
+$winsDirPath = $('{0}etc\windows-exporter' -f $prefixPath)
```
nickgerace, line 48: nit: the path below has a path separator after the variable. Should they both match for UX consistency?

nickgerace (approved):
LGTM barring nit

packages/rancher-windows-exporter/charts/scripts/run.ps1
aiyengar2: No need. The reason why the first doesn’t have a path separator is because the CATTLE_PREFIX_PATH has a \ at the end.

By definition, the winsDirPath does not have a \ at the end so we need to add it to construct winsPath based on it.

aiyengar2: This is also in a powershell script, so there’s no UX involved

luthermonson (approved):

PR #1169, opened 2021-04-29
[DNM 2.6] Add RKE Windows path prefix for logging

nickgerace:
Relevant issue: https://github.com/rancher/rancher/issues/32268
Based off: https://github.com/rancher/wins/pull/35

PR #1121, opened 2021-04-15
[DNM 2.6] Add buffer configuration options for fluentbit to logging

nickgerace:
Relevant issue: https://github.com/rancher/rancher/issues/28401

PR #1112, opened 2021-04-11
[DNM v2.5.9/v2.6] Add additional config options for fluentd

dbason:
This change should meet the requirements for https://github.com/rancher/rancher/issues/31807

The replicas setting may be one worth plumbing through to the UI, but I haven't done this yet.

For tolerations and nodeSelectors, fluentd will use the chart wide settings unless overridden with fluent specific settings.

PR #1182, opened 2021-05-05
Add rancher-logging to 2.6 among multiple changes

nickgerace:
## Description
This PR is split into two commits: a copy commit, similar to PRs in [rancher/system-charts](https://github.com/rancher/system-charts), and a commit with the desired changes.

## Objectives
- Add the `rancher-logging` package from `dev-v2.5-source` with no changes
- Add RKE Windows PathPrefix and standardize Windows paths within the logging chart (continuation of #1169)
- Add buffer configuration for `fluent-bit` for logging (continuation of #1121)
- Add `fluentd` options from frequent requests (originally from @dbason; continuation of #1112)
- Tidy the Logging CRs in the chart for style consistency and general cleanup (it also addresses: https://github.com/rancher/charts/pull/1182#commitcomment-50398194)

## Why are these changes in one PR rather than multiple?
1. All continuations of PRs were already approved and reviewed
1. Given the slated timing for 2.6 release, QA may not be able to test individual RCs for each commit (making the split for separate PRs less effective)
1. All major changes are unrelated to one another, so if regressions occur, they will be isolated

## Relevant issues
- https://github.com/rancher/rancher/issues/28401
- https://github.com/rancher/rancher/issues/31807
- https://github.com/rancher/rancher/issues/32268
- https://github.com/rancher/rancher/issues/32541

### DIFF
--- a/packages/rancher-logging/generated-changes/overlay/templates/loggings/aks/logging.yaml
+++ b/packages/rancher-logging/generated-changes/overlay/templates/loggings/aks/logging.yaml
@@ -1,0 +1,58 @@
+{{- if .Values.additionalLoggingSources.aks.enabled }}
+apiVersion: logging.banzaicloud.io/v1beta1
+kind: Logging
+metadata:
+  name: {{ .Release.Name }}-aks
+  namespace: {{ .Release.Namespace }}
+  labels:
+{{ include "logging-operator.labels" . | indent 4 }}
+spec:
+  controlNamespace: {{ .Release.Namespace }}
+  fluentbit:
+    image:
+      repository: {{ template "system_default_registry" . }}{{ .Values.images.fluentbit.repository }}
+      tag: {{ .Values.images.fluentbit.tag }}
+    inputTail:
+      Tag: "aks"
+      Path: "/var/log/azure/kubelet-status.log"
+    {{- if .Values.global.psp.enabled }}
+    security:
+      podSecurityPolicyCreate: true
+      roleBasedAccessControlCreate: true
+    {{- end }}
+  {{- $total_tolerations := concat (.Values.tolerations) (.Values.fluentbit.tolerations) }}
+  {{- with $total_tolerations }}
+    tolerations: {{- toYaml . | nindent 6 }}
+  {{- end }}
+  {{- with .Values.nodeSelector }}
+    nodeSelector: {{- toYaml . | nindent 6 }}
+  {{- end }}
+  {{- with .Values.fluentbit.resources }}
+    resources: {{- toYaml . | nindent 6 }}
+  {{- end }}
+  fluentd:
+    image:
+      repository: {{ template "system_default_registry" . }}{{ .Values.images.fluentd.repository }}
+      tag: {{ .Values.images.fluentd.tag }}
+    configReloaderImage:
+      repository: {{ template "system_default_registry" . }}{{ .Values.images.config_reloader.repository }}
+      tag: {{ .Values.images.config_reloader.tag }}
+    disablePvc: {{ .Values.disablePvc }}
+    {{- if .Values.global.psp.enabled }}
+    security:
+      podSecurityPolicyCreate: true
+      roleBasedAccessControlCreate: true
+    {{- end }}
+    {{- with .Values.tolerations }}
+    tolerations: {{- toYaml . | nindent 6 }}
+    {{- end }}
+    {{- with .Values.nodeSelector }}
+    nodeSelector: {{- toYaml . | nindent 6 }}
+    {{- end }}
+    {{- with .Values.fluentd.resources }}
+    resources: {{- toYaml . | nindent 6 }}
+    {{- end }}
+    {{- with .Values.fluentd.livenessProbe }}
+    livenessProbe: {{- toYaml . | nindent 6 }}
+    {{- end }}
+{{- end }}
--- a/packages/rancher-logging/generated-changes/overlay/templates/loggings/aks/logging.yaml
+++ b/packages/rancher-logging/generated-changes/overlay/templates/loggings/aks/logging.yaml
@@ -15,21 +15,43 @@ spec:
     inputTail:
       Tag: "aks"
       Path: "/var/log/azure/kubelet-status.log"
+      {{- if .Values.fluentbit.inputTail.Buffer_Chunk_Size }}
+      Buffer_Chunk_Size: {{ .Values.fluentbit.inputTail.Buffer_Chunk_Size }}
+      {{- end }}
+      {{- if .Values.fluentbit.inputTail.Buffer_Max_Size }}
+      Buffer_Max_Size: {{ .Values.fluentbit.inputTail.Buffer_Max_Size }}
+      {{- end }}
+      {{- if .Values.fluentbit.inputTail.Mem_Buf_Limit }}
+      Mem_Buf_Limit: {{ .Values.fluentbit.inputTail.Mem_Buf_Limit }}
+      {{- end }}
+      {{- if .Values.fluentbit.inputTail.Multiline_Flush }}
+      Multiline_Flush: {{ .Values.fluentbit.inputTail.Multiline_Flush }}
+      {{- end }}
+      {{- if .Values.fluentbit.inputTail.Skip_Long_Lines }}
+      Skip_Long_Lines: {{ .Values.fluentbit.inputTail.Skip_Long_Lines }}
+      {{- end }}
+    {{- if not .Values.disablePvc }}
+    {{- with .Values.fluentbit.bufferStorage }}
+    bufferStorage: {{- toYaml . | nindent 6 }}
+    {{- end }}
+    {{- with .Values.fluentbit.bufferStorageVolume }}
+    bufferStorageVolume: {{- toYaml . | nindent 6 }}
+    {{- end }}
+    {{- end }}
     {{- if .Values.global.psp.enabled }}
     security:
       podSecurityPolicyCreate: true
       roleBasedAccessControlCreate: true
     {{- end }}
-  {{- $total_tolerations := concat (.Values.tolerations) (.Values.fluentbit.tolerations) }}
-  {{- with $total_tolerations }}
+    {{- with (concat (.Values.tolerations) (.Values.fluentbit.tolerations)) }}
     tolerations: {{- toYaml . | nindent 6 }}
-  {{- end }}
-  {{- with .Values.nodeSelector }}
+    {{- end }}
+    {{- with .Values.nodeSelector }}
     nodeSelector: {{- toYaml . | nindent 6 }}
-  {{- end }}
-  {{- with .Values.fluentbit.resources }}
+    {{- end }}
+    {{- with .Values.fluentbit.resources }}
     resources: {{- toYaml . | nindent 6 }}
-  {{- end }}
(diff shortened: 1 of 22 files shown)


packages/rancher-logging/generated-changes/overlay/templates/loggings/root/logging.yaml
```diff
@@ -49,6 +64,34 @@ spec:
     image:
       repository: {{ template "system_default_registry" . }}{{ .Values.images.fluentbit.repository }}
       tag: {{ .Values.images.fluentbit.tag }}
+    {{- if or .Values.fluentbit.inputTail.Buffer_Chunk_Size .Values.fluentbit.inputTail.Buffer_Max_Size .Values.fluentbit.inputTail.Mem_Buf_Limit .Values.fluentbit.inputTail.Multiline_Flush .Values.fluentbit.inputTail.Skip_Long_Lines }}
```
nickgerace, line 67: Relevant comment: https://github.com/rancher/charts/pull/1121#discussion_r614376734

nickgerace:
Assigned all reviewers from the previous PRs.

paynejacob (dismissed):

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke/daemonset.yaml
```diff
@@ -0,0 +1,124 @@
...
+      serviceAccountName: "{{ .Release.Name }}-rke-aggregator"
+      {{- $total_tolerations := concat (.Values.tolerations) (.Values.fluentbit.tolerations) }}
+      {{- with $total_tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.nodeSelector }}
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
```
nickgerace, line 60: Requires cleanup

packages/rancher-logging/generated-changes/overlay/templates/loggings/rke2/daemonset.yaml
```diff
@@ -0,0 +1,110 @@
...
+              name: machine-id
+              readOnly: true
+      {{- with .Values.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.nodeSelector }}
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
```
nickgerace, line 42: Requires cleanup

nickgerace:
This requires one more validating for the PathPrefix due to internal Windows standards changes: moving out of review temporarily.

packages/rancher-logging/generated-changes/patch/values.yaml.patch
```diff
@@ -119,5 +119,6 @@
 +  # logging operator.
 +  psp:
 +    enabled: true
++  rkeWindowsPathPrefix: "c:"
```
aiyengar2, line 131: I think this and all default calls should be `c:\\`, not `c:`.

packages/rancher-logging/generated-changes/patch/templates/_helpers.tpl.patch
```diff
@@ -22,3 +22,11 @@
 +  {{- end }}
 +  {{- end }}
 +{{- end }}
++
++{{- define "windowsPathPrefix" }}
++{{- $pathPrefix := (required "Must provide value for .Values.global.cattle.rkeWindowsPathPrefix" .Values.global.cattle.rkeWindowsPathPrefix) -}}
```
aiyengar2: `.Values.global.cattle.rkeWindowsPathPrefix` can be empty string and should be defaulted. The dashboard UI will specifically override the default for `""` due to the way it is currently defined.

packages/rancher-logging/generated-changes/patch/templates/_helpers.tpl.patch
```diff
@@ -22,3 +22,11 @@
...
++{{- if (contains "//" $pathPrefix) -}}
++{{- fail ".Values.global.cattle.rkeWindowsPathPrefix must not contain forward-slashes" -}}
```
aiyengar2: We should handle forward-slashes, we just expect the default to be backslashes

packages/rancher-logging/generated-changes/patch/templates/_helpers.tpl.patch
```diff
@@ -22,3 +22,11 @@
...
++{{- end -}}
++{{- dir .Values.global.cattle.rkeWindowsPathPrefix }}
```
aiyengar2: Not sure why we are using [`dir`](https://helm.sh/docs/chart_template_guide/function_list/#dir) since it would return one level up from what the pathPrefix defined is (e.g. `\opt\rke` would return `\opt`)

packages/rancher-logging/generated-changes/overlay/templates/loggings/aks/logging.yaml
```diff
@@ -15,6 +15,31 @@ spec:
     inputTail:
```
aiyengar2, line 16: If we are copying and pasting the exact same values, I think these should either belong in a helper function or we should define a `_logging.tpl` that we call for each of these logging.yaml files.

e.g. see the [`_pod.tpl`](https://github.com/grafana/helm-charts/blob/main/charts/grafana/templates/_pod.tpl) pattern in Grafana

packages/rancher-logging/generated-changes/overlay/templates/loggings/k3s/logging-k3s-openrc.yaml
```diff
@@ -1,4 +1,5 @@
 {{- if and .Values.additionalLoggingSources.k3s.enabled (eq .Values.additionalLoggingSources.k3s.container_engine "openrc")}}
+{{- $total_fluentbit_tolerations := concat (.Values.tolerations) (.Values.fluentbit.tolerations) }}
```
aiyengar2, line 2: Put this in a helper function rather than calling it at the top of the file each time

aiyengar2 (changes requested):
@nickgerace can you squash everything except the base copy PR into 1 commit? It was hard to tell what had or hadn't been changed in your revisions.

Also, if possible I think the logging CRs are repetitive enough to justify an `_logging.tpl` template instead of maintaining the same code and making changes multiple times.

packages/rancher-logging/generated-changes/patch/templates/_helpers.tpl.patch
nickgerace: This is an old commit. I'll squash.

packages/rancher-logging/generated-changes/patch/templates/_helpers.tpl.patch
nickgerace: This is an old commit. I'll squash.

packages/rancher-logging/generated-changes/overlay/templates/loggings/aks/logging.yaml
nickgerace: Filed as https://github.com/rancher/tasks/issues/191. Planned for this PR.

packages/rancher-logging/generated-changes/overlay/templates/loggings/k3s/logging-k3s-openrc.yaml
nickgerace: Fixed by avoiding the use of the variable.

nickgerace:
@aiyengar2 - Decided to move https://github.com/rancher/tasks/issues/191 to its own PR since the diff for this PR is already large. I'd like to confirm with QA if they would like to test the templating system as well, and I think leaving both apart will be best for minimizing a bottleneck later. Thus, this PR is ready to be re-reviewed.

packages/rancher-logging/generated-changes/patch/templates/_helpers.tpl.patch
```diff
@@ -14,11 +14,20 @@
...
-+  {{- end }}
-+  {{- end }}
-+  {{- end }}
++{{- end }}
++{{- end }}
++{{- end }}
++{{- end }}
++
++{{- define "windowsPathPrefix" }}
++{{- $temp := (default "c:\\" .Values.global.cattle.rkeWindowsPathPrefix | replace "\\" "/" | replace "//" "/") }}
```
aiyengar2, line 27: Can we rename this to `$prefixPath`?

aiyengar2 (changes requested):
Some minor changes

packages/rancher-logging/generated-changes/patch/templates/_helpers.tpl.patch
nickgerace: Not sure if I like that idea actually since the RKE Linux prefix path can be different.

packages/rancher-logging/generated-changes/overlay/templates/loggings/gke/logging.yaml
```diff
@@ -15,38 +15,69 @@ spec:
...
+      {{- end }}
+      {{- if .Values.fluentbit.inputTail.Mem_Buf_Limit }}
+      Mem_Buf_Limit: {{ .Values.fluentbit.inputTail.Mem_Buf_Limit }}
+      {{- end }}
+      {{- if .Values.fluentbit.inputTail.Multiline_Flush }}
+      Multiline_Flush: {{ .Values.fluentbit.inputTail.Multiline_Flush }}
+      {{- end }}
+      {{- if .Values.fluentbit.inputTail.Skip_Long_Lines }}
+      Skip_Long_Lines: {{ .Values.fluentbit.inputTail.Skip_Long_Lines }}
+      {{- end }}
```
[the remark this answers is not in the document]
nickgerace: We only want to individually expose settings. This is purposeful actually. However, we may want to discuss in another PR addressing rancher/tasks#191.

packages/rancher-logging/generated-changes/overlay/templates/loggings/aks/logging.yaml
```diff
@@ -15,38 +15,69 @@ spec:
...
       roleBasedAccessControlCreate: true
     {{- end }}
-    {{- with .Values.tolerations }}
+    {{- with (default .Values.tolerations .Values.fluentd.tolerations) }}
     tolerations: {{- toYaml . | nindent 6 }}
     {{- end }}
-    {{- with .Values.nodeSelector }}
+    {{- with (default .Values.nodeSelector .Values.fluentd.nodeSelector) }}
     nodeSelector: {{- toYaml . | nindent 6 }}
     {{- end }}
```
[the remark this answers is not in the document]
nickgerace: This is intentional. Fluentd falls back to default, whereas fluentbit uses concat.

aiyengar2 (approved):
LGTM

packages/rancher-logging/generated-changes/overlay/templates/loggings/gke/logging.yaml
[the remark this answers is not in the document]
nickgerace: Linked your comment @aiyengar2: https://github.com/rancher/tasks/issues/191#issuecomment-839972171

PR #1186, opened 2021-05-05
Add nginx-ingress pushprox and fix pushprox namespaceOverride

aiyengar2:
Related Issue: https://github.com/rancher/rancher/issues/32563


packages/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/exporters/ingress-nginx/service.yaml
```diff
@@ -1,24 +0,0 @@
-{{- if .Values.ingressNginx.enabled }}
```
aiyengar2, line 1: The Service + ServiceMonitor is created by PushProx and is no longer necessary

packages/rancher-monitoring/generated-changes/patch/Chart.yaml.patch
```diff
@@ -48,7 +48,7 @@
  kubeversion: '>=1.16.0-0'
  dependencies:
  - name: grafana
-@@ -185,7 +195,7 @@
```
aiyengar2, line 51: nit change due to adding an extra dependency on the pushprox chart

packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
```diff
@@ -233,37 +233,20 @@
 +##
 +ingressNginx:
 +  enabled: false
-+
```
aiyengar2, line 236: ingressNginx has never been exposed in a released chart (it is new to 2.5.8) so it's fine to completely change its YAML spec. There will be a dashboard change to update the necessary overrides.

packages/rancher-pushprox/charts/templates/_helpers.tpl
```diff
@@ -49,7 +49,7 @@ provider: kubernetes
 {{- if .Values.clients.proxyUrl -}}
 {{ printf "%s" .Values.clients.proxyUrl }}
 {{- else -}}
-{{ printf "http://%s.%s.svc:%d" (include "pushProxy.proxy.name" .) .Release.Namespace (int .Values.proxy.port) }}
+{{ printf "http://%s.%s.svc:%d" (include "pushProxy.proxy.name" .) (include "pushprox.namespace" .) (int .Values.proxy.port) }}
```
aiyengar2, line 52: These were all misses in PushProx since we had never tested a case where a pushprox was deployed in a non-default namespace until now.

brendarearden (approved):
lgtm

PR #1189, opened 2021-05-10
[Live] Migrate to v0.2.0 version of the charts scripts

aiyengar2:
Related Issue: https://github.com/rancher/rancher/issues/32199


scripts/version
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
 set -e
 
-CHARTS_BUILD_SCRIPT_VERSION=v0.0.4
+CHARTS_BUILD_SCRIPTS_REPO=https://github.com/aiyengar2/charts-build-scripts.git
+CHARTS_BUILD_SCRIPT_VERSION=v0.2.0
```
aiyengar2, line 5: The only change that will be left from this commit is this one line

aiyengar2:
Will need to revert the temp commit in favor of the one-line change once the charts-build-script changes are merged in.

cbron (approved):
Seems fine since it doesn't affect the assets. We should do release-v2.6 first though.

README.md
```diff
@@ -19,33 +19,62 @@ charts/
...
+STAGING_BRANCH=dev-v2.x
+LIVE_BRANCH=release-v2.x
+FORKED_BRANCH=release-v2.x.y
+git fetch upstream
+git checkout upstream/${LIVE_BRANCH} -b ${FORKED_BRANCH}
+git branch -u origin/${FORKED_BRANCH}
+git checkout upstream/${STAGING_BRANCH} -- charts assets index.yaml
+git add charts assets index.yaml
+git commit -m "Releasing chart"
+git push --set-upstream origin ${FORKED_BRANCH}
```
StrongMonkey, line 38: Nit: I would not put origin here as it points out to the rancher/charts repo. I would expect release caption to push to their fork first then create pull requests. I don't think we want forked branch to be pushed accidently on rancher/charts.

aiyengar2: @StrongMonkey above it says `# Assuming that your upstream remote (e.g. https://github.com/rancher/charts.git) is named upstream`, so I think the expectation is that origin is their fork.

We can modify the documentation for the new description, but I think that needs to be done as a rancher/charts-build-scripts change to the template that we can then pull in.

StrongMonkey: ok that's fine.

StrongMonkey: Just make sure we can call it out in confluence. Mostly likely people have origin pointing upstream git repo.

StrongMonkey (approved):

PR #1200, opened 2021-05-13
Add NoExecute toleration to rke2Etcd and remove key

aiyengar2:
No other PushProx exporter keys the tolerations, so rke2Etcd's default is being modified for consistency.

Workaround for users who already have Monitoring V2 deployed would be to just override the value in the values.yaml.

If a user needs more fine-tuned tolerations, they can always choose to override that as well.

Related Issue: https://github.com/rancher/rancher/issues/32245


packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
```diff
@@ -225,8 +225,9 @@
 +    nodeSelector:
 +      node-role.kubernetes.io/etcd: "true"
 +    tolerations:
++      - effect: "NoExecute"
++        operator: "Exists"
 +      - effect: "NoSchedule"
-+        key: node-role.kubernetes.io/master
 +        operator: "Equal"
```
dramich, line 231: Should this get updated to `exists`?

aiyengar2: yes. Updating

dramich (dismissed):

aiyengar2 (commented):

aiyengar2:
Rebased to dev-v2.5

dramich (approved):

brendarearden (approved):

PR #1203, opened 2021-05-14
Mimic Helm CRD install process by introducing wait for established

aiyengar2:
As part of the Helm CRD installation process shown [here](https://github.com/helm/helm/blob/43853ea772e779b617b011a8d4e69205636fa4f9/pkg/action/install.go#L160-L162), charts that both attempt to install CRDs and then install resources (e.g. `len(totalItems) > 0`) invalidate the cache and attempt to get the CRDs again to `Give time for the CRD to be recognized.` before continuing the Helm installation.

Because Monitoring V2 currently does not observe this invalidation and wait, flakey race conditions documented in issues like https://github.com/rancher/rancher/issues/32025 and https://github.com/rancher/rancher/issues/29171 can be caused. This commit introduces that invalidation process.


aiyengar2 (commented):

aiyengar2:
Rebased to dev-v2.5

charts/rancher-monitoring/rancher-monitoring-crd/14.5.101+up14.5.0/templates/jobs.yaml
```diff
@@ -45,7 +45,18 @@ spec:
...
+            echo "Waiting for CRDs to be recognized before finishing installation...";
+            
+            {{- range $path, $_ :=  (.Files.Glob  "crd-manifest/**.yaml") }}
+            {{- $apiGroup := get (get ($.Files.Get $path | fromYaml) "spec") "group" }}
+            rm -rf $HOME/.kube/cache/discovery/*/{{ $apiGroup }}
+            {{- end }}
+
+            {{- range $path, $_ :=  (.Files.Glob  "crd-manifest/**.yaml") }}
+            {{- $crd := get (get ($.Files.Get $path | fromYaml) "metadata") "name" }}
+            kubectl get crd {{ $crd }}
```
dramich, line 58: This call can/will fail, if we are attempting to duplicate what helm does it should be an error check and retry loop until the CRDs are available through the API 
https://github.com/helm/helm/blob/43853ea772e779b617b011a8d4e69205636fa4f9/pkg/kube/wait.go#L52:18

aiyengar2:
Changed PR to dev-v2.6

aiyengar2:
Probably resolves https://github.com/rancher/rancher/issues/32827 as well.

aiyengar2:
PR is ready to be re-reviewed. 

I modified this PR according to @dramich's comment in https://github.com/rancher/charts/pull/1203#discussion_r638106337 and updated the initContainer jobs to also check whether the CRDs are established before proceeding.

One note: while Helm upstream checks whether either [`NamesAccepted=False || Established=True`](https://github.com/helm/helm/blob/43853ea772e779b617b011a8d4e69205636fa4f9/pkg/kube/wait.go#L307), my changes to the CRD chart strictly check if the CRDs are established before continuing since we cannot proceed with the installation if there is a naming conflict.

dramich (commented):
@aiyengar2 Can we use a helper method here? This looks like a similar pattern in many places

aiyengar2:
> @aiyengar2 Can we use a helper method here? This looks like a similar pattern in many places

@dramich addressed in latest commit. Good suggestion!

kinarashah (approved):
PR already reviewed by Dan. Went through the new helper changes for crd.established and looks good!

jiaqiluo (approved):
LGTM

PR #1211, opened 2021-05-18
Add sriov chart

manuelbuil:
Signed-off-by: Manuel Buil <mbuil@suse.com>


packages/rancher-sriov/charts/templates/_helpers.tpl
```diff
@@ -0,0 +1,7 @@
+{{- define "system_default_registry" -}}
+{{- if .Values.global.systemDefaultRegistry -}}
```
jiaqiluo, line 2:
```suggestion
{{- if .Values.global.cattle.systemDefaultRegistry -}}
```

packages/rancher-sriov/charts/templates/view-role.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: ClusterRole
+metadata:
+  name: sriov-view
+rules:
+- apiGroups:
+  - sriovnetwork.openshift.io
+  resources:
+  - '*'
+  verbs:
+  - ["get", "watch", "list"]
```
jiaqiluo, line 11:
```suggestion
  verbs:
  - "get"
  - "watch"
  - "list"

```

jiaqiluo (changes requested):

packages/rancher-sriov/charts/Chart.yaml
```diff
@@ -0,0 +1,21 @@
...
+keywords:
+  - sr-iov
+  - Networking
+sources:
+  - https://github.com/rancher/charts
+maintainers:
+  - name: Rancher Labs
+    email: charts@rancher.com
+annotations:
+  catalog.cattle.io/certified: rancher
```
aiyengar2, line 16: Just to confirm, does this need a custom UI? e.g. `catalog.cattle.io/ui-component`

packages/rancher-sriov/charts/Chart.yaml
```diff
@@ -0,0 +1,21 @@
...
+  catalog.cattle.io/experimental: true
```
aiyengar2, line 17: Does this need `catalog.cattle.io/display-name`?

packages/rancher-sriov/charts/app-README.md
```diff
@@ -0,0 +1,13 @@
+This chart is based on the upstream [k8snetworkplumbingwg/sriov-network-operator](https://github.com/k8snetworkplumbingwg/sriov-network-operator) project. The chart deploys the SR-IOV Operator and its CRDs. It allows users to deploy SR-IOV CNI, which can be used, together with Multus, to provide high performing extra network interfaces to pods.
```
aiyengar2, line 1: Needs title? Multus should also have a link IMO.

```suggestion
# Rancher SR-IOV Network Operator

This chart is based on the upstream [k8snetworkplumbingwg/sriov-network-operator](https://github.com/k8snetworkplumbingwg/sriov-network-operator) project. The chart deploys the SR-IOV Operator and its CRDs. It allows users to deploy SR-IOV CNI, which can be used with Multus to provide high performing extra network interfaces to pods.
```

packages/rancher-sriov/charts/app-README.md
```diff
@@ -0,0 +1,13 @@
...
+
+
+The chart installs the following components:
+
+    SR-IOV Operator - The operator helps provision and configure SR-IOV CNI plugin and Device plugin. It requires Multus to operate correctly.
+    SR-IOV Network Config Daemon - Deployed by the operator, this daemon is responsible for discovering the SR-IOV NICs on each node
+
+Note that SR-IOV requires NICs that support SR-IOV and the activation of specific configuration options in the operating system. Nodes that fultill these requirements should be labeled with:
```
aiyengar2, line 9: nit:
```suggestion
Note that SR-IOV requires NICs that support SR-IOV and the activation of specific configuration options in the operating system. Nodes that fulfill these requirements should be labeled with:
```

packages/rancher-sriov/charts/crds/sriov-network-operator-sriovibnetworks_crd.yaml
```diff
@@ -0,0 +1,74 @@
+apiVersion: apiextensions.k8s.io/v1
```
aiyengar2, line 1: Is it possible to add a `# Source: ` to the upstream?

packages/rancher-sriov/package.yaml
```diff
@@ -0,0 +1,9 @@
+url: local
+packageVersion: 01
+releaseCandidateVersion: 00
```
aiyengar2, line 3: No longer necessary
```suggestion
```

packages/rancher-sriov/package.yaml
```diff
@@ -0,0 +1,9 @@
+url: local
```
aiyengar2, line 1: Why are we doing this as a local chart? This chart exists in an upstream https://github.com/k8snetworkplumbingwg/sriov-network-operator/tree/9b59e4ed4785bddeaea67837bd8506163a37a713/deployment/sriov-network-operator.

IMO, we should provide the Github Repository based URL for this chart and maintain just the necessary patches on top of it, unless there is a strong reason why we should not do so.

aiyengar2 (changes requested):
Did not do a deep review, would like to get clarification on why we aren't declaring this as a Git-based forked chart first. https://github.com/k8snetworkplumbingwg/sriov-network-operator/tree/9b59e4ed4785bddeaea67837bd8506163a37a713/deployment/sriov-network-operator

This would be highly preferable as we only need to maintain a patch on upstream.

packages/rancher-sriov/charts/templates/_helpers.tpl
manuelbuil: good catch!

packages/rancher-sriov/charts/templates/view-role.yaml
manuelbuil: done

packages/rancher-sriov/charts/Chart.yaml
manuelbuil: I don't think so

packages/rancher-sriov/charts/Chart.yaml
manuelbuil: According to confluence `  catalog.cattle.io/display-name: "Feature Name" # Only for Charts with custom UI`, as sriov does not have a custom UI, I guess we don't need that

packages/rancher-sriov/package.yaml
manuelbuil: That's a great point. I think the person that started this work missed that there was already a helm chart upstream (possibly the work started before the upstream guys released it). I have just modified my commit and now it is based on the upstream helm chart (following the confluence docs). I tested it and works but please review ;)

packages/rancher-sriov/generated-changes/patch/templates/clusterrolebinding.yaml.patch
```diff
@@ -0,0 +1,18 @@
+--- charts-original/templates/clusterrolebinding.yaml
++++ charts/templates/clusterrolebinding.yaml
```
jiaqiluo, line 2: Is the patch name for the `charts/templates/clusterrolebinding.yaml` file?  
It more looks like for the `charts/templates/clusterrole.yaml` file according to the contents. 
Can you double check with it?

manuelbuil: Good catch :). It's a bug upstream. I noticed it while creating this patch and fixed it upstream ==> https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/138, but it is not yet in the release we are using

jiaqiluo: oh, interesting bug 😄 

We need to fix in our patch file, otherwise the `make prepare` will fail 
```
➜  charts git:(sriov-chart2) ✗ PACKAGE=rancher-sriov make prepare
./bin/charts-build-scripts prepare
INFO[0000] Pulling k8snetworkplumbingwg/sriov-network-operator@v1.0.0[path=deployment/sriov-network-operator] from upstream into charts 
INFO[0020] Loading dependencies for chart               
INFO[0020] Applying changes from generated-changes      
INFO[0020] Applying: generated-changes/patch/Chart.yaml.patch 
INFO[0020] Applying: generated-changes/patch/templates/NOTES.txt.patch 
INFO[0020] Applying: generated-changes/patch/templates/_helpers.tpl.patch 
INFO[0020] Applying: generated-changes/patch/templates/clusterrolebinding.yaml.patch 
ERRO[0020] 
patching file templates/clusterrolebinding.yaml
Hunk #1 FAILED at 15.
1 out of 1 hunk FAILED -- saving rejects to file templates/clusterrolebinding.yaml.rej 
FATA[0020] Encountered error while preparing main chart: Encountered error while trying to apply changes to charts: Unable to generate patch with error: exit status 1 
make: *** [prepare] Error 1
```

packages/rancher-sriov/generated-changes/exclude/crds/k8s.cni.cncf.io_networkattachmentdefinitions_crd.yaml
```diff
@@ -0,0 +1,57 @@
+# Copyright 2020 NVIDIA
```
jiaqiluo, line 1: I see `network-attachment-definition` is listed in the clusterrole `sriov-network-operator`. 
It is ok to include it in the clusterrole even the CRD does not exist. I just need to confirm that we do want to exclude this CRD.

packages/rancher-sriov/generated-changes/exclude/crds/k8s.cni.cncf.io_networkattachmentdefinitions_crd.yaml
```diff
@@ -0,0 +1,57 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: network-attachment-definitions.k8s.cni.cncf.io
```
jiaqiluo, line 17: I see `network-attachment-definition` is listed in the clusterrole `sriov-network-operator`. 
It is ok to include it in the clusterrole even the CRD does not exist. I just need to confirm that we do want to exclude this CRD.

jiaqiluo:
@manuelbuil Can you rebuild the charts by using the `charts-build-scripts` version v0.2.0? 

You can update it by running `make pull-scripts` locally.
```
> make pull-scripts
./scripts/pull-scripts
Pulling in charts-build-scripts version https://github.com/rancher/charts-build-scripts.git@v0.2.0
charts-build-scripts version v0.2.0 (cae97e4)
```

The rebuild the charts by running `PACKAGE=rancher-sriov make charts`.
The chart version in the final chart will be `0.1.1+up0.1.0`.

packages/rancher-sriov/generated-changes/patch/values.yaml.patch
```diff
@@ -0,0 +1,52 @@
...
++    key: cattle.io/os
++    operator: Equal
++    value: linux
+   nodeSelector:
+-    node-role.kubernetes.io/master: ""
++    node-role.kubernetes.io/master: "true"
+   nameOverride: ""
+   fullnameOverride: ""
+-  resourcePrefix: "openshift.io"
++  resourcePrefix: "rancher.io"
```
aiyengar2, line 22: What are the implications of changing the resourcePrefix? Can we safely do this without running into any other issues (e.g. CRDs still have openshift.io on them right?)

aiyengar2 (changes requested):

charts/rancher-sriov/sriov/1.0.1/Chart.yaml
```diff
@@ -0,0 +1,21 @@
...
+keywords:
+- sr-iov
+- Networking
+maintainers:
+- email: charts@rancher.com
+  name: Rancher Labs
+name: sriov
+sources:
+- https://github.com/rancher/charts
+version: 1.0.1
```
cbron, line 21: Needs icon in here pointing at https://charts.rancher.io/assets/logos/.....

packages/rancher-sriov/generated-changes/exclude/crds/k8s.cni.cncf.io_networkattachmentdefinitions_crd.yaml
manuelbuil: SR-IOV requires Multus. Multus allows to set up multiple interfaces in a pod. Multus becomes the main CNI plugin and can propagate calls to different "real" CNI plugins that are registered. The way to register those CNI plugins is through the crd: network-attachment-definitions.k8s.cni.cncf.io. However, that CRD is part of Multus design and thus it should be installed by Multus, not by SR-IOV. In fact, we added a few lines to detect that and if the crd is not installed, then we ask the user to install Multus first: https://github.com/rancher/charts/pull/1211/files#diff-c0260e86c59f0bb92c1629b68aad1d2f97629b9d4d1978783d810ce3470acb02R1. If we install that CRD via SR-IOV, then SR-IOV will start but we will not be able to properly allow pods to use it because we will be missing Multus binary and config which is installed by Multus helm chart. BTW, this is documented

packages/rancher-sriov/generated-changes/overlay/templates/admin-role.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: sriov-admin
+rules:
+- apiGroups:
+  - sriovnetwork.openshift.io
```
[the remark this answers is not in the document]
manuelbuil: As explained above, I'd not do it because it is not part of SR-IOV but of Multus

packages/rancher-sriov/generated-changes/overlay/templates/admin-role.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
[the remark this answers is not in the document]
manuelbuil: Yes, I can do that

packages/rancher-sriov/generated-changes/patch/templates/operator.yaml.patch
```diff
@@ -0,0 +1,70 @@
...
++  namespace: {{ .Release.Namespace }}
+   labels:
+   {{- include "sriov-network-operator.labels" . | nindent 4 }}
+ spec:
+@@ -25,7 +40,7 @@
+       serviceAccountName: {{ include "sriov-network-operator.fullname" . }}
+       containers:
+         - name: {{ include "sriov-network-operator.fullname" . }}
+-          image: {{ .Values.images.operator }}
++          image: {{ include "system_default_registry" . }}{{ .Values.images.operator }}
```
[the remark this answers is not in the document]
manuelbuil: It's ugly but it's in the value already:
```
images:
  operator: rancher/hardened-sriov-network-operator:v1.0.0-build20210429
 ```

File: packages/rancher-sriov/generated-changes/patch/values.yaml.patch (hunk 13 above)
Comment (manuelbuil) [reply packages/rancher-sriov/generated-changes/patch/values.yaml.patch:22]: Yes, this is safe. CRDs still have the openshift.io suffix. The sriov resources are named $resourcePrefix/$nic_model, e.g. rancher.io/intel

File: index.yaml [hunk 18]
```diff
@@ -3282,6 +3282,48 @@ entries:
...
+    - Networking
+    maintainers:
+    - email: charts@rancher.com
+      name: Rancher Labs
+    name: sriov
+    sources:
+    - https://github.com/rancher/charts
+    urls:
+    - assets/rancher-sriov/sriov-1.0.1.tgz
+    version: 1.0.1
```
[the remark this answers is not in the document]
manuelbuil: I recreated index. yaml but it seems there are two packages, right?
```
    urls:    - assets/rancher-sriov/sriov-1.0.1.tgz    version: 1.0.1
```
and
```
    urls:    - assets/rancher-sriov/sriov-0.1.1+up0.1.0.tgz    version: 0.1.1+up0.1.0
```
should I remove the one without `up`?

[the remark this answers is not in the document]
jiaqiluo: Yes, you need to remove the one without `up` in the following places:
- the index.yaml file
- the `assets/rancher-riov` directory

jiaqiluo (changes requested):
overall LGTM, we just need to do the following changes:
- remove assets/rancher-sriov/sriov-1.0.1.tgz
- remove assets/rancher-sriov/sriov-crd-1.0.1.tgz
- update the `index.yaml`

packages/rancher-sriov/generated-changes/overlay/templates/admin-role.yaml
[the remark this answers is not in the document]
aiyengar2: So going back to the explanation in https://github.com/rancher/charts/pull/1211#discussion_r639711405, even if we are not installing that CRD as part of this chart, should a user who has sriov-admin permissions be able to also register and unregister CNI plugins or do we expect that to be a permission that Multus will need to provide?

packages/rancher-sriov/generated-changes/patch/templates/operator.yaml.patch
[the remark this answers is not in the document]
aiyengar2: I think we should patch this chart to have the values.yaml be more consistent with other charts, w.r.t. the image tag, such as:

```
images:
  operator:
    image: rancher/hardened-sriov-network-operator
    tag: v1.0.0-build20210429
    os: linux # optional, but if we eventually support windows this will be necessary for airgap too
```

We can then patch all references here accordingly, e.g. 
```suggestion
+          image: {{ include "system_default_registry" . }}{{ .Values.images.operator.image }}{{ .Values.images.operator.tag }}
```

packages/rancher-sriov/generated-changes/overlay/templates/admin-role.yaml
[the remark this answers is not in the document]
aiyengar2: Yeah I agree with putting it in a clusterrole.yaml, but if would be nice if we could put this under `templates/rancher-sriov` similar to what we do with Monitoring here https://github.com/rancher/charts/tree/dev-v2.5/charts/rancher-monitoring/rancher-monitoring/14.5.101%2Bup14.5.0/templates/rancher-monitoring.

It's way nicer to maintain an overlay file that a big patch 😅 , but clusterrole.yaml makes more sense to contain clusterroles that role.yaml.

packages/rancher-sriov/generated-changes/overlay/templates/secrets.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if not .Values.cert_manager -}}
+{{- if .Values.operator.enableAdmissionController }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: operator-webhook-service
+  namespace: {{ .Release.Namespace }}
+data: {{ include "sriov_operator_cert" . | indent 2 }}
```
aiyengar2, line 8: need the newline right?
```suggestion
data: {{ include "sriov_operator_cert" . | nindent 2 }}
```

packages/rancher-sriov/generated-changes/overlay/templates/secrets.yaml
```diff
@@ -0,0 +1,20 @@
...
+{{- end }}
+---
+{{- if .Values.operator.enableAdmissionController }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: network-resources-injector-secret
+  namespace: {{ .Release.Namespace }}
+data: {{ include "sriov_resource_injector_cert" . | indent 2 }}
```
aiyengar2, line 17:
```suggestion
data: {{ include "sriov_resource_injector_cert" . | nindent 2 }}
```

packages/rancher-sriov/generated-changes/overlay/templates/admin-role.yaml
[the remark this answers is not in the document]
aiyengar2: Oh I see you already added it. Can you drop these overlay files then?

aiyengar2 (changes requested):

manuelbuil:
> overall LGTM, we just need to do the following changes:
> 
>     * remove assets/rancher-sriov/sriov-1.0.1.tgz
> 
>     * remove assets/rancher-sriov/sriov-crd-1.0.1.tgz
> 
>     * update the `index.yaml`

Done! Thanks a lot for your patience and comprehensive reviews

charts/rancher-sriov/sriov/0.1.1+up0.1.0/.helmignore
```diff
@@ -0,0 +1,23 @@
+# Patterns to ignore when building packages.
```
aiyengar2, line 1: Assets need to be regenerated. `0.1.1` is incorrect, right?

packages/rancher-sriov/generated-changes/overlay/app-README.md
```diff
@@ -0,0 +1,14 @@
+# Rancher SR-IOV Network Operator
+
+This chart is based on the upstream [k8snetworkplumbingwg/sriov-network-operator](https://github.com/k8snetworkplumbingwg/sriov-network-operator) project. The chart deploys the SR-IOV Operator and its CRDs. It allows users to deploy SR-IOV CNI, which can be used, together with [Multus](https://github.com/k8snetworkplumbingwg/multus-cni), to provide high performing extra network interfaces to pods.
+
+
+The chart installs the following components:
+
+ - SR-IOV Operator - The operator helps provision and configure SR-IOV CNI plugin and Device plugin. It requires Multus to operate correctly.
+ - SR-IOV Network Config Daemon - Deployed by the operator, this daemon is responsible for discovering the SR-IOV NICs on each node
```
aiyengar2, line 9: nit: even if you don't take these changes, make sure you either add punctuation to the second bullet point or remove it from the first.

```suggestion
 - SR-IOV Operator - An operator that helps provision and configure CNI or Device plugins (requires Multus)
 - SR-IOV Network Config Daemon - A Daemon deployed by the Operator that discovers SR-IOV NICs on each node
```

packages/rancher-sriov/generated-changes/overlay/app-README.md
```diff
@@ -0,0 +1,14 @@
+# Rancher SR-IOV Network Operator
+
+This chart is based on the upstream [k8snetworkplumbingwg/sriov-network-operator](https://github.com/k8snetworkplumbingwg/sriov-network-operator) project. The chart deploys the SR-IOV Operator and its CRDs. It allows users to deploy SR-IOV CNI, which can be used, together with [Multus](https://github.com/k8snetworkplumbingwg/multus-cni), to provide high performing extra network interfaces to pods.
```
aiyengar2, line 3:
```suggestion
This chart is based on the upstream [k8snetworkplumbingwg/sriov-network-operator](https://github.com/k8snetworkplumbingwg/sriov-network-operator) project. The chart deploys the SR-IOV Operator and its CRDs, which is designed to help the user to provision and configure SR-IOV CNI and Device plugins in a cluster that uses [Multus CNI](https://github.com/k8snetworkplumbingwg/multus-cni) to provide high performing extra network interfaces to pods.
```

packages/rancher-sriov/generated-changes/overlay/templates/certmanagercerts.yaml
```diff
@@ -0,0 +1,43 @@
+{{- if .Values.operator.enableAdmissionController -}}
+{{- if .Values.cert_manager -}}
+{{- if not (.Capabilities.APIVersions.Has "cert-manager.io/v1") -}}
```
aiyengar2, line 3: Are these intentionally separated or could we use an `and` here to omit having three `end` blocks?

packages/rancher-sriov/generated-changes/overlay/templates/certmanagercerts.yaml
```diff
@@ -0,0 +1,43 @@
...
+{{- required "cert-manager is required but not found" "" -}}
+{{- end -}}
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: sriov-network-operator-selfsigned-issuer
```
aiyengar2, line 9: Should the naming scheme for these be based on `{{ include "sriov-network-operator.fullname" . }}`?

packages/rancher-sriov/generated-changes/overlay/templates/secrets.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if not .Values.cert_manager -}}
+{{- if .Values.operator.enableAdmissionController }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: operator-webhook-service
```
aiyengar2, line 6: Should the naming scheme for these be based on `{{ include "sriov-network-operator.fullname" . }}`?

packages/rancher-sriov/generated-changes/patch/templates/clusterrole.yaml.patch
```diff
@@ -0,0 +1,61 @@
...
+     resources: ["mutatingwebhookconfigurations", "validatingwebhookconfigurations"]
+@@ -52,3 +52,42 @@
+   - apiGroups: [""]
+     resources: ["pods/eviction"]
+     verbs: ["create"]
++---
++apiVersion: rbac.authorization.k8s.io/v1
++kind: ClusterRole
++metadata:
++  name: sriov-admin
```
aiyengar2, line 27: Do we expect these clusterroles to automatically aggregate into the default k8s admin, edit, and view roles? If so, we need annotations here, e.g.

https://github.com/rancher/charts/blob/7836aff0c1b2846c3fbcf8ed2adc6963d21c20b1/charts/rancher-monitoring/rancher-monitoring/14.5.101%2Bup14.5.0/templates/rancher-monitoring/clusterrole.yaml#L7-L9

aiyengar2 (changes requested):
Still need some more changes but the finish line is in sight! 😄  

Also, IMO it might make sense to temporarily split your changes into two commits (one for `packages` and the other for the output of `make charts`) and squash once we have finished reviews. That way you can just redo the last commit after we have reviewed the `packages` changes and squash it into the first.

charts/rancher-sriov/sriov/0.1.1+up0.1.0/.helmignore
manuelbuil: The upstream version is 0.1.0 --> https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/master/deployment/sriov-network-operator/Chart.yaml#L3. Not sure how is the last 1 filled, any idea?

packages/rancher-sriov/generated-changes/overlay/app-README.md
manuelbuil: I partially applied your suggestion. I think we should have `provision and configure SR-IOV CNI plugin and SR-IOV Device plugin`. If we remove SR-IOV, the user could think that any CNI or any Device Plugin can be deployed.

What do you mean by adding punctuation to the second bullet?

packages/rancher-sriov/generated-changes/overlay/templates/certmanagercerts.yaml
manuelbuil: good idea. I can remove one end combining the first two

packages/rancher-sriov/generated-changes/overlay/templates/certmanagercerts.yaml
manuelbuil: I did not want to add another delta with regards to upstream. I'd suggest we wait until upstream implements this

packages/rancher-sriov/generated-changes/overlay/templates/secrets.yaml
manuelbuil: I did not want to add another delta with regards to upstream. I'd suggest we wait until upstream implements this

packages/rancher-sriov/generated-changes/patch/templates/clusterrole.yaml.patch
manuelbuil: Not sure about this. I added this roles because it was required based on https://confluence.suse.com/pages/viewpage.action?spaceKey=EN&title=Feature+Charts+Helm+Requirements, i.e. we don't need them to run this chart correctly. Do you normally recommend to add that to charts?

packages/rancher-sriov/templates/crd-template/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v2
+version: 1.0.0
```
aiyengar2, line 2: Version should match v0.1.0 or I think the `catalog.cattle.io/auto-install: sriov-crd=match` annotation will break

https://github.com/k8snetworkplumbingwg/sriov-network-operator/blob/07398ed56de01f6e7d2a869743c6d5ce8298ca5b/deployment/sriov-network-operator/Chart.yaml#L3

charts/rancher-sriov/sriov/0.1.1+up0.1.0/.helmignore
aiyengar2: Gotcha. So the fix is to the CRD chart version, not to this chart. Will leave a comment for that in my next review, but resolving this.

packages/rancher-sriov/generated-changes/overlay/app-README.md
aiyengar2: You resolved the issue. I mean that the first had a `.` at the end and the second did not 😄

packages/rancher-sriov/generated-changes/patch/templates/clusterrole.yaml.patch
aiyengar2: So the reason why these annotations are necessary is because they [aggregate these ClusterRoles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles) to the [default k8s roles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles), which are bound to corresponding Rancher role ([Project Owner -> admin](https://github.com/rancher/rancher/blob/9c9491b2879de82454ab53d083492e530e519693/pkg/data/management/role_data.go#L229), [Project Member -> edit](https://github.com/rancher/rancher/blob/9c9491b2879de82454ab53d083492e530e519693/pkg/data/management/role_data.go#L265), [Read-only -> view](https://github.com/rancher/rancher/blob/9c9491b2879de82454ab53d083492e530e519693/pkg/data/management/role_data.go#L297)).

This is a choice made for every chart, so I think it depends on your RBAC expectations for this chart. Do you expect the person who owns or is part of the project that contains `cattle-sriov-system` to have these permissions by default, or should they be manually bound as and when necessary?

cc: @cbron if you have any thoughts.

aiyengar2 (changes requested):
Need to regenerate assets once you get the CRD chart versioning change in.

packages/rancher-sriov/generated-changes/overlay/templates/certmanagercerts.yaml
manuelbuil: You are right, this is currently us supporting it. There are a couple of PRs upstream adding the certmanager stuff and they are not yet merged. Until then, we must support it ourselves. I have just looked at why it does the `{{ include "sriov-network-operator.fullname" . }}` instead of adding the string directly. I see that it is just in case there is a fullnameOverride in the values. We could add this here too but I don't see a big benefit of doing it

packages/rancher-sriov/generated-changes/overlay/templates/secrets.yaml
manuelbuil: Same as above. If you believe this is important, I could spend time adding new functions in _helpers but I don't see the benefit

packages/rancher-sriov/generated-changes/patch/templates/clusterrole.yaml.patch
manuelbuil: Thanks for the explanation. I can see users having only "sriov-view" permissions, because this chart modifies the network configuration of the host and that can create connectivity issues across nodes. Therefore, only the "admin" of the cluster should be able to have full control of this. I guess these should be made available by default. Let me add those

packages/rancher-sriov/generated-changes/patch/templates/clusterrole.yaml.patch
```diff
@@ -0,0 +1,54 @@
...
++  resources:
++  - '*'
++  verbs:
++  - '*'
++---
++apiVersion: rbac.authorization.k8s.io/v1
++kind: ClusterRole
++metadata:
++  name: sriov-edit
++  {{- if .Values.global.rbac.userRoles.aggregateToDefaultRoles }}
```
aiyengar2, line 27: Does a default need to be patched in the values.yaml? e.g.

```yaml
global:
  rbac:
    userRoles:
      aggregateToDefaultRoles: true|false
```

If a `helm template` does not cause a failure and the default for this is expected to be false, then we can continue to omit this. But I'm not sure what would happen since `global.rbac.userRoles` is undefined right now.

packages/rancher-sriov/generated-changes/overlay/templates/certmanagercerts.yaml
aiyengar2: If we are going to deprecate our certmanager stuff in favor of an eventual upstream change, I think it's fine to leave it as-is. Resolving this comment.

index.yaml
```diff
@@ -3386,6 +3386,67 @@ entries:
...
+      catalog.cattle.io/release-name: sriov-crd
+    apiVersion: v2
+    created: "2021-05-27T21:06:06.912831309+02:00"
+    description: Installs the CRDs for rke2-sriov.
+    digest: ef9e1ad98796590a04447a094a399e41cba27bc48bf5172d58c1f95a675e5a03
+    name: sriov-crd
+    type: application
+    urls:
+    - assets/rancher-sriov/sriov-crd-1.0.1+up0.1.0.tgz
+    version: 1.0.1+up0.1.0
```
aiyengar2, line 3434: should not exist?

aiyengar2:
@manuelbuil Thank you for being patient with all the requested changes 😄 

I'm **almost** ready to approve this PR. Before I do, can you:
- [ ] Rebase to the latest upstream (we just fixed CI workflows for charts)
- [ ] Remove the `index.yaml`, `charts/`, and `asset/` changes entirely from the commit and re-run `make charts` to avoid issues like https://github.com/rancher/charts/pull/1211#discussion_r643386272.
- [ ] Verify `helm template assets/rancher-sriov/sriov-crd-0.1.1+up0.1.0.tgz` is successful and prints out a k8s manifest containing just your CRDs
- [ ] Verify `helm template assets/rancher-sriov/sriov-0.1.1+up0.1.0.tgz` is successful and prints out a k8s manifest corresponding to the expected resources (to ensure something like https://github.com/rancher/charts/pull/1211#discussion_r643384352 does not cause an issue)
- [ ] (Optional, but recommended) Verify that a helm install of the chart onto a Rancher downstream cluster with Multus is successful. You can do this by navigating to the Apps & Marketplace UI of the downstream cluster, adding a Charts Repository on Rancher with url `https://github.com/manuelbuil/charts` and branch `sriov-chart2`, and trying to install the chart.

Let's get this merged! 🎉

packages/rancher-sriov/generated-changes/patch/templates/clusterrole.yaml.patch
manuelbuil: Yes, we need this

manuelbuil:
> @manuelbuil Thank you for being patient with all the requested changes smile
> 
> I'm **almost** ready to approve this PR. Before I do, can you:
> 
>     * [x]  Rebase to the latest upstream (we just fixed CI workflows for charts)
> 
>     * [x]  Remove the `index.yaml`, `charts/`, and `asset/` changes entirely from the commit and re-run `make charts` to avoid issues like [#1211 (comment)](https://github.com/rancher/charts/pull/1211#discussion_r643386272).
> 
>     * [x]  Verify `helm template assets/rancher-sriov/sriov-crd-0.1.1+up0.1.0.tgz` is successful and prints out a k8s manifest containing just your CRDs
> 
>     * [x]  Verify `helm template assets/rancher-sriov/sriov-0.1.1+up0.1.0.tgz` is successful and prints out a k8s manifest corresponding to the expected resources (to ensure something like [#1211 (comment)](https://github.com/rancher/charts/pull/1211#discussion_r643384352) does not cause an issue)
> 
>     * [ ]  (Optional, but recommended) Verify that a helm install of the chart onto a Rancher downstream cluster with Multus is successful. You can do this by navigating to the Apps & Marketplace UI of the downstream cluster, adding a Charts Repository on Rancher with url `https://github.com/manuelbuil/charts` and branch `sriov-chart2`, and trying to install the chart.
> 
> 
> Let's get this merged! tada

Thank you for you patience!

I had an old deployment of rancherd v2.5.8. I added `https://github.com/manuelbuil/charts` with branch `sriov-chart2` and I see sriov-crd and sriov. When I try to install any of them I get:

```
NotFound (404)

catalogtemplateversions.management.cattle.io "myrepo-sriov-crd-0.1.1+up0.1.0" not found
```
and
```
NotFound (404)

catalogtemplateversions.management.cattle.io "myrepo-sriov-0.1.1+up0.1.0" not found
```
Does this point to an error in my PR? Or is it because using rancherd v2.5? BTW, is it normal to get both helm charts (crd and non-crd)?

packages/rancher-sriov/generated-changes/patch/templates/clusterrole.yaml.patch
aiyengar2: Sounds good.

charts/rancher-sriov/sriov/0.1.1+up0.1.0/values.yaml
```diff
@@ -0,0 +1,54 @@
+operator:
+  tolerations:
+  - effect: NoSchedule
+    key: node-role.kubernetes.io/control-plane
```
jiaqiluo, line 4:
```suggestion
    key: node-role.kubernetes.io/controlplane
```

charts/rancher-sriov/sriov/0.1.1+up0.1.0/values.yaml
```diff
@@ -0,0 +1,54 @@
...
+    operator: Exists
+  - effect: NoSchedule
+    key: cattle.io/os
+    operator: Equal
+    value: linux
+  nodeSelector:
+    node-role.kubernetes.io/control-plane: "true"
```
jiaqiluo, line 11:
```suggestion
    node-role.kubernetes.io/controlplane: "true"
```

jiaqiluo (changes requested):

manuelbuil:
We tried to deploy with @jiaqiluo using rancher 2.5 and realized that taint and label changes between rke1 and rke2 from `node-role.kubernetes.io/controlplane` to `node-role.kubernetes.io/control-plane`. We were wondering if there is a way to only show the app when the cluster is rke2 and how we could test it with rancher 2.6 and rke2. rancher 2.5 only deploys rke1

aiyengar2:
> I had an old deployment of rancherd v2.5.8. I added `https://github.com/manuelbuil/charts` with branch `sriov-chart2` and I see sriov-crd and sriov. When I try to install any of them I get:
> 
> ```
> NotFound (404)
> 
> catalogtemplateversions.management.cattle.io "myrepo-sriov-crd-0.1.1+up0.1.0" not found
> ```
> 
> and
> 
> ```
> NotFound (404)
> 
> catalogtemplateversions.management.cattle.io "myrepo-sriov-0.1.1+up0.1.0" not found
> ```

Where are you seeing this issue on an install? I was able to trigger an install on your chart successfully and get the CRDs installed on a downstream cluster in Rancher 2.5.8, although it failed shy of installing the full chart since I didn't deploy it on a cluster with Multus installed. Does the failure you encountered happen after this step?

Screen Shot 2021-06-02 at 12 12 29 PM

> Does this point to an error in my PR? Or is it because using rancherd v2.5? BTW, is it normal to get both helm charts (crd and non-crd)?

My assumption is **no**, your PR looks fine. Maybe it could be because you are using rancherd v2.5, but I'm not sure. It's possible that you needed to manually refresh the catalog before proceeding with the install for some reason.

It is normal to get both Helm charts since our UI automatically installs the CRD chart before the main chart.

aiyengar2:
> We tried to deploy with @jiaqiluo using rancher 2.5 and realized that taint and label changes between rke1 and rke2 from `node-role.kubernetes.io/controlplane` to `node-role.kubernetes.io/control-plane`. We were wondering if there is a way to only show the app when the cluster is rke2 and how we could test it with rancher 2.6 and rke2. rancher 2.5 only deploys rke1

To my knowledge, there is no way to specify that a chart is only shown to specific cluster types. However, we could introduce a custom UI that overrides the above values based on the cluster type; this is something we do for Monitoring today.

packages/rancher-sriov/generated-changes/overlay/app-README.md
```diff
@@ -0,0 +1,14 @@
+# Rancher SR-IOV Network Operator
+
+This chart is based on the upstream [k8snetworkplumbingwg/sriov-network-operator](https://github.com/k8snetworkplumbingwg/sriov-network-operator) project. The chart deploys the SR-IOV Operator and its CRDs, which is designed to help the user provision and configure the SR-IOV CNI in a cluster that uses [Multus CNI](https://github.com/k8snetworkplumbingwg/multus-cni), to provide high performing extra network interfaces to pods.
```
aiyengar2, line 3: Based on offline conversation, we need to make sure this file calls out that:
1) This chart is expected to be deployed on an RKE2 cluster. It has not been tested or designed for any other cluster type at the moment, although it can be deployed in other clusters with some modifications. (Addresses https://github.com/rancher/charts/pull/1211#issuecomment-853288767)
2) A user who is deploying this chart is expected to be fairly knowledgable about Multus, SR-IOV, and the use case behind this chart as it is only meant for **advanced use cases** where multiple CNIs and high performing network interfaces on pods are required. Users who do not need these features are not advised to install this chart.

cc: @cjellick @cbron

aiyengar2 (changes requested):
Need to make the changes Jack suggested for nodeSelectors, specifically for RKE 2, and re-check this list: https://github.com/rancher/charts/pull/1211#issuecomment-852359393.

Also, we need to make an app-README.md change due to offline conversation.

packages/rancher-sriov/generated-changes/overlay/app-README.md
jiaqiluo: In the case of the sriov chart, although we can fix the pod tolerations to make it installed successfully in a rke1 cluster,  I think we will keep it as it is today so the failure can also bring the user’s attention and realize the cluster is not an RKE2 cluster.

manuelbuil:
> > I had an old deployment of rancherd v2.5.8. I added `https://github.com/manuelbuil/charts` with branch `sriov-chart2` and I see sriov-crd and sriov. When I try to install any of them I get:
> > ```
> > NotFound (404)
> > 
> > catalogtemplateversions.management.cattle.io "myrepo-sriov-crd-0.1.1+up0.1.0" not found
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and
> > ```
> > NotFound (404)
> > 
> > catalogtemplateversions.management.cattle.io "myrepo-sriov-0.1.1+up0.1.0" not found
> > ```
> 
> Where are you seeing this issue on an install? I was able to trigger an install on your chart successfully and get the CRDs installed on a downstream cluster in Rancher 2.5.8, although it failed shy of installing the full chart since I didn't deploy it on a cluster with Multus installed. Does the failure you encountered happen after this step?
> 
> Screen Shot 2021-06-02 at 12 12 29 PM
> 
> > Does this point to an error in my PR? Or is it because using rancherd v2.5? BTW, is it normal to get both helm charts (crd and non-crd)?
> 
> My assumption is **no**, your PR looks fine. Maybe it could be because you are using rancherd v2.5, but I'm not sure. It's possible that you needed to manually refresh the catalog before proceeding with the install for some reason.
> 
> It is normal to get both Helm charts since our UI automatically installs the CRD chart before the main chart.

I was using rancherd v2.5, that was the issue. Once I moved to the newer verison, it worked :)

manuelbuil:
>     * sriov-chart2



> @manuelbuil Thank you for being patient with all the requested changes smile
> 
> I'm **almost** ready to approve this PR. Before I do, can you:
> 
>     * [x]  Rebase to the latest upstream (we just fixed CI workflows for charts)
> 
>     * [x]  Remove the `index.yaml`, `charts/`, and `asset/` changes entirely from the commit and re-run `make charts` to avoid issues like [#1211 (comment)](https://github.com/rancher/charts/pull/1211#discussion_r643386272).
> 
>     * [x]  Verify `helm template assets/rancher-sriov/sriov-crd-0.1.1+up0.1.0.tgz` is successful and prints out a k8s manifest containing just your CRDs
> 
>     * [x]  Verify `helm template assets/rancher-sriov/sriov-0.1.1+up0.1.0.tgz` is successful and prints out a k8s manifest corresponding to the expected resources (to ensure something like [#1211 (comment)](https://github.com/rancher/charts/pull/1211#discussion_r643384352) does not cause an issue)
> 
>     * [ ]  (Optional, but recommended) Verify that a helm install of the chart onto a Rancher downstream cluster with Multus is successful. You can do this by navigating to the Apps & Marketplace UI of the downstream cluster, adding a Charts Repository on Rancher with url `https://github.com/manuelbuil/charts` and branch `sriov-chart2`, and trying to install the chart.
> 
> 
> Let's get this merged! tada
image

packages/rancher-sriov/package.yaml
```diff
@@ -0,0 +1,10 @@
+url: https://github.com/k8snetworkplumbingwg/sriov-network-operator.git
+subdirectory: deployment/sriov-network-operator
+commit: c5c943cc214f4113b6f6be8d7c48ac837aa1669c
+packageVersion: 01
```
jiaqiluo, line 4: nit. 

```suggestion
packageVersion: 1
```

jiaqiluo (dismissed):
LGTM

packages/rancher-sriov/generated-changes/overlay/app-README.md
```diff
@@ -0,0 +1,13 @@
+# Rancher SR-IOV Network Operator
+
+This chart is based on the upstream [k8snetworkplumbingwg/sriov-network-operator](https://github.com/k8snetworkplumbingwg/sriov-network-operator) project. The chart deploys the SR-IOV Operator and its CRDs, which is designed to help the user provision and configure the SR-IOV CNI in a cluster that uses [Multus CNI](https://github.com/k8snetworkplumbingwg/multus-cni), to provide high performing extra network interfaces to pods. This chart is expected to be deployed on an RRKE2 cluster and only meant for advanced use cases where multiple CNI plugins and high performing network interfaces on pods are required. Users who do not need these features are not advised to install this chart.
```
aiyengar2, line 3: nit:
```suggestion
This chart is based on the upstream [k8snetworkplumbingwg/sriov-network-operator](https://github.com/k8snetworkplumbingwg/sriov-network-operator) project. The chart deploys the SR-IOV Operator and its CRDs, which are designed to help the user provision and configure the SR-IOV CNI in a cluster that uses [Multus CNI](https://github.com/k8snetworkplumbingwg/multus-cni), to provide high performing extra network interfaces to pods. This chart is expected to be deployed on an RKE2 cluster and only meant for advanced use cases where multiple CNI plugins and high performing network interfaces on pods are required. Users who do not need these features are not advised to install this chart.
```

packages/rancher-sriov/generated-changes/overlay/app-README.md
```diff
@@ -0,0 +1,13 @@
...
+
+The chart installs the following components:
+
+ - SR-IOV Operator - An operator that helps provision and configure the SR-IOV CNI plugin and SR-IOV Device plugin (requires Multus)
```
aiyengar2, line 7: nit: `requires Multus` is also mentioned above. Do we need this?
```suggestion
 - SR-IOV Operator - An operator that helps provision and configure the SR-IOV CNI plugin and SR-IOV Device plugin
```

packages/rancher-sriov/generated-changes/patch/templates/role.yaml.patch
```diff
@@ -0,0 +1,7 @@
+--- charts-original/templates/role.yaml
++++ charts/templates/role.yaml
+@@ -109,3 +109,4 @@
+       - configmaps
+     verbs:
+       - get
++
```
aiyengar2, line 7: nit: do we need this file if it is just adding a newline?

aiyengar2 (approved):
LGTM. I have a couple of nitpicks below that should be addressed before merging (specifically to the app-README.md), but I'm good to sign off on this. Can approve again if you make the changes.

Thanks again for making all the changes! Let's get this merged 🎉

aiyengar2:
Also, @manuelbuil, unless I overlooked the relevant PSPs, we are assuming that this initial version of the chart will **not** be deployable on a hardened cluster that pass a CIS 1.16 scan right? 

I don't think we should introduce any more fixes into this PR, but we may need additional changes at some point to support that use case. If a ticket does not currently exist, we should open one up.

manuelbuil:
> Also, @manuelbuil, unless I overlooked the relevant PSPs, we are assuming that this initial version of the chart will **not** be deployable on a hardened cluster that pass a CIS 1.16 scan right?
> 
> I don't think we should introduce any more fixes into this PR, but we may need additional changes at some point to support that use case. If a ticket does not currently exist, we should open one up.

Yes, you are right. I haven't take into account anything related to CIS. When you say ticket, do you mean an issue in this repo?

packages/rancher-sriov/generated-changes/overlay/app-README.md
manuelbuil: Thanks

jiaqiluo (approved):

aiyengar2 (approved):

cbron (commented):
Few minor things, basically gtg besides permission aggregation situation

charts/rancher-sriov/sriov/0.1.1+up0.1.0/Chart.yaml
```diff
@@ -0,0 +1,25 @@
...
+  catalog.cattle.io/experimental: "true"
+  catalog.cattle.io/namespace: cattle-sriov-system
+  catalog.cattle.io/os: linux
+  catalog.cattle.io/release-name: sriov
+apiVersion: v2
+appVersion: 1.0.0
+description: SR-IOV network operator configures and manages SR-IOV networks in the
+  kubernetes cluster
+home: https://github.com/k8snetworkplumbingwg/sriov-network-operator
+icon: https://charts.rancher.io/assets/logos/sr-iov.svg
```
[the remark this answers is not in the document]
manuelbuil: You mean in `remotes/origin/release-v2.6`? What should I put there, the result of doing `PACKAGE=rancher-sriov make charts`, i.e. what gets created in charts/ and assets/?

charts/rancher-sriov/sriov/0.1.1+up0.1.0/values.yaml
```diff
@@ -0,0 +1,57 @@
...
+    tag: v1.0.0-build20210310
+  sriovDevicePlugin: 
+    image: rancher/hardened-sriov-network-device-plugin
+    tag: v3.3.1-build20210310
+  resourcesInjector: 
+    image: rancher/hardened-sriov-network-resources-injector
+    tag: v1-build20210310
+  webhook:
+    image: rancher/hardened-sriov-network-webhook
+    tag: v1.0.0-build20210429
```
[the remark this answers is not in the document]
manuelbuil: Yes, they are all hardened. It's basically taking the upstream code but building them using our golang build image that uses goboring

manuelbuil:
@cbron The configuring of some of the sriov crds ends up configuring hardware resources from the nodes. That is pretty intrusive, so I'd suggest only the cluster admin to be capable of doing that. In that case, should I remove the:
```
  {{- if .Values.global.rbac.userRoles.aggregateToDefaultRoles }}
    rbac.authorization.k8s.io/aggregate-to-admin: "true"
  {{- end }}
```
from sriov-adming and sriov-edit?

charts/rancher-sriov/sriov/0.1.1+up0.1.0/Chart.yaml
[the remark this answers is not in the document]
cbron: Oh sorry I just meant the logo svg file, so it doesn't 404, ex: https://charts.rancher.io/assets/logos/logging.svg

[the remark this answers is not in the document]
manuelbuil: I cherry-picked it into release-v2.6 branch: https://github.com/rancher/charts/pull/1269

manuelbuil:
> @cbron The configuring of some of the sriov crds ends up configuring hardware resources from the nodes. That is pretty intrusive, so I'd suggest only the cluster admin to be capable of doing that. In that case, should I remove the:
> 
> ```
>   {{- if .Values.global.rbac.userRoles.aggregateToDefaultRoles }}
>     rbac.authorization.k8s.io/aggregate-to-admin: "true"
>   {{- end }}
> ```
> 
> from sriov-adming and sriov-edit?

Clarified with Caleb. Admin, edit and view will have the same permissions because we want only the cluster admin to set the config of sr-iov

cbron (approved):

PR #1220, opened 2021-05-20
[dev-v2.6] Fix vSphere CPI RBAC conflicts

PennyScissors:
Issue: https://github.com/rancher/rancher/issues/32604

Please check the issue above, [this](https://github.com/rancher/rancher/issues/32604#issuecomment-841470680) comment and [this](https://github.com/rancher/rancher/issues/32604#issuecomment-854171968) other one before reviewing


packages/rancher-vsphere-cpi/charts/templates/vsphere-cpi-rbac.yaml
```diff
@@ -1,14 +1,15 @@
+{{- if .Values.cloudControllerManager.rbac.enabled -}}
 # Source: https://github.com/kubernetes/cloud-provider-vsphere/blob/master/releases/v1.19/vsphere-cloud-controller-manager.yaml
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: cloud-controller-manager
+  name: vsphere-cloud-controller-manager
```
cbron, line 6: These all get updated on upgrade right, curious if any of these are immutable.

packages/rancher-vsphere-cpi/package.yaml
```diff
@@ -1,2 +1,2 @@
 url: local
-packageVersion: 0
+packageVersion: 1
```
cbron, line 2: Fairly sure this needs a rebase. If this is a local chart shouldn't it not use a packageVersion anymore @aiyengar2 / @jiaqiluo ? I know the "new" changes haven't gotten in yet, but for a local chart you would just bump the chart.yaml. 
@PennyScissors I forget how local this chart is, doesn't it exist upstream ? Are we in line with their versions right now ?

jiaqiluo: @aiyengar2 Please correct me if this is not correct:
For the local chart, we should change the value of `version` in the chart.yaml file directly instead of the packageVersion. 
But it will not error out if we use the `packageVersion` as this PR does.

aiyengar2: That's correct @jiaqiluo.

@cbron iirc, vsphere is a local chart that technically has an upstream, similar to Istio, so maybe bumping the packageVersion is the right move here? If there is no upstream, we should just bump the patch version if possible since we want to ideally deprecate the packageVersion anyways.

packages/rancher-vsphere-cpi/charts/templates/vsphere-cpi-rbac.yaml
aiyengar2: Yeah this can be updated. Helm 3's logic is
- Run `helm template`
- For each resource in `helm template`, check if it exists. If not, create. If so, patch.
- If a resource that is currently annotated with this Helm release exists and the `keep` annotation is not specified and it was not updated or created in the prior step (e.g. does not exist in the template), delete it.

This logic is encoded in [the kube package Client.Update](https://github.com/helm/helm/blob/3f283c7c35fffda567894b9bf6cede5e19d43ffa/pkg/kube/client.go#L192-L198), which is [used by the Configuration object](https://github.com/helm/helm/blob/a499b4b179307c267bdf3ec49b880e3dbd2a5591/pkg/action/action.go#L366) and [is called on an Upgrade](https://github.com/helm/helm/blob/a499b4b179307c267bdf3ec49b880e3dbd2a5591/pkg/action/upgrade.go#L317-L321).

So in this case, this is identical to deleting the ServiceAccount in favor of a new one.

jiaqiluo (approved):
LGTM. 
Do not forget to squash commits before merging the RP.

packages/rancher-vsphere-cpi/charts/questions.yaml
```diff
@@ -1,4 +1,15 @@
 questions:
+  - variable: clusterType
+    label: Cluster Type
+    description: Select the type of the cluster where vSphere CPI will be deployed
+    type: enum
+    default: "rke"
+    options:
+      - "rke"
+      - "rke2"
+      - "other"
```
aiyengar2, line 10: On the UI, does this render as `rke` or `RKE`? Is there a way to change the display name of these options (e.g. should `rke2` be `RKE Government`)?

packages/rancher-vsphere-cpi/charts/templates/vsphere-cpi-ds.yaml
```diff
@@ -21,13 +21,17 @@ spec:
...
       {{- with .Values.cloudControllerManager.nodeSelector }}
         {{- toYaml . | nindent 8 }}
       {{- end }}
-      {{- else }}
+      {{- else if eq .Values.clusterType "rke" }}
+      nodeSelector:
         node-role.kubernetes.io/controlplane: "true"
+      {{- else if eq .Values.clusterType "rke2" }}
+      nodeSelector:
+        node-role.kubernetes.io/control-plane: "true"
```
aiyengar2, line 34: This looks good to me, but I wonder if it would be better to specify this via affinityRules rather than nodeSelectors. 

affinityRules would allow us to select a node if one or the other is true, which means we don't have to look for the clusterType at all. (see https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity).

aiyengar2 (commented):
Need to resolve the conversation in https://github.com/rancher/charts/pull/1220/commits/449fe16252b2bd90d72dff8b0709275ee34794dc#r644383268 and I have a question w.r.t. whether we can drop the cluster type stuff entirely if we just switch to affinityRules instead.

packages/rancher-vsphere-cpi/charts/templates/vsphere-cpi-ds.yaml
cbron: Thats interesting question. Can you or @PennyScissors ask RKE2 team about that. Brad may know good reason.

cbron (approved):
Needs two things addressed still, the rke2 UI situation vs node-affinity, and then also will need to be updated version based on new versioning rules but that can happen down the line.

vincent99:
@PennyScissors This is because the values.yaml already has [a value](https://github.com/rancher/charts/pull/1220/files#diff-470487fb7c93fd6c0b1860b1f520906dba328908454af5aa3ed63f2b1cb462e6R21) (of empty string) for those fields; The default from questions are applied after reading in all the values and is used only if there is [no value set](https://github.com/rancher/dashboard/blob/master/components/Questions/Question.js#L50).

packages/rancher-vsphere-cpi/package.yaml
cbron: Is the upstream versioned or just a bunch of manifests @PennyScissors ? If versioned then we should use 100.0.0. If not we can do whatever we want with it, if 2.5 was on 1.x.x then 2.6 would be on 2.x.x.

PennyScissors:
> Needs two things addressed still, the rke2 UI situation vs node-affinity, and then also will need to be updated version based on new versioning rules but that can happen down the line.

Commented out the `clusterType` field based on this [comment](https://github.com/rancher/dashboard/issues/3188#issuecomment-866588466) from Vincent in the UI issue I created for this issue. As for the versioning, it looks like it was updated so I rebased to pull those in.

packages/rancher-vsphere-cpi/charts/templates/vsphere-cpi-ds.yaml
PennyScissors: I agree this approach would be better, but I think this is something we should look into later on after 2.6 to avoid more changes than necessary.

packages/rancher-vsphere-cpi/charts/questions.yaml
jiaqiluo: I am not sure if it renders as rke or RKE.  Leave this question to others.
Yes, rke2 should be `RKE Government`

packages/rancher-vsphere-cpi/charts/values.yaml
```diff
@@ -14,6 +14,11 @@ cloudControllerManager:
     repository: rancher/mirrored-cloud-provider-vsphere-cpi-release-manager
     tag: v1.2.1
   nodeSelector: {}
+  rbac:
+    enabled: true
+
+# Uncomment and set to "rke" or "rke2" to apply the node selector label appropriate for the cluster
+# clusterType: ""
```
jiaqiluo, line 21: Should this line be un-commented? 

My concern is that whether the value will be set if the field does not exist in the value.yaml (since the field is commented out).

jiaqiluo (changes requested):

packages/rancher-vsphere-cpi/charts/values.yaml
PennyScissors: It was, but I needed to comment it due to [this](https://github.com/rancher/dashboard/issues/3188#issuecomment-866588466).

charts/rancher-vsphere-cpi/rancher-vsphere-cpi/100.0.0/values.yaml
```diff
@@ -14,6 +14,11 @@ cloudControllerManager:
     repository: rancher/mirrored-cloud-provider-vsphere-cpi-release-manager
     tag: v1.2.1
   nodeSelector: {}
+  rbac:
+    enabled: true
```
jiaqiluo, line 18: What if the user sets it to be false? 
Will the app fail to install because of the missing service account?

jiaqiluo (changes requested):

charts/rancher-vsphere-cpi/rancher-vsphere-cpi/100.0.0/values.yaml
PennyScissors: No, it doesn't fail installation. This is just to provide the option to disable the chart's RBAC to avoid incompatibilities like those stated in the issue for this PR (conflicts with rke2's RBAC), but then the expectation is the user will set it up themselves. This is not in the chart's questions so it's something they would need to set up manually.

charts/rancher-vsphere-cpi/rancher-vsphere-cpi/100.0.0/templates/vsphere-cpi-rbac.yaml
```diff
@@ -105,23 +106,24 @@ roleRef:
...
-  name: cloud-controller-manager
+  name: vsphere-cloud-controller-manager
 subjects:
 - kind: ServiceAccount
-  name: cloud-controller-manager
+  name: vsphere-cloud-controller-manager
   namespace: {{ .Release.Namespace }}
 - kind: User
-  name: cloud-controller-manager
+  name: vsphere-cloud-controller-manager
```
jiaqiluo, line 128: Where is the `User` CR from? I do not find its definition in the chart.

jiaqiluo (changes requested):

packages/rancher-vsphere-cpi/charts/questions.yaml
PennyScissors: It renders as "rke" and "rke2". Changing that would be a UI task.

PennyScissors:
> Need to resolve the conversation in [449fe16#r644383268](https://github.com/rancher/charts/commit/449fe16252b2bd90d72dff8b0709275ee34794dc#r644383268) and I have a question w.r.t. whether we can drop the cluster type stuff entirely if we just switch to affinityRules instead.

@aiyengar2 I'm not familiar with affinityRules, so I'm not sure how much effort the change you are suggesting will take. Since this is a small fix, I would like to keep it simple and look into this after 2.6, but if it's an easy change then I can include it in this PR.

packages/rancher-vsphere-cpi/package.yaml
PennyScissors: Yes, upstream is just a bunch of manifests.

charts/rancher-vsphere-cpi/rancher-vsphere-cpi/100.0.0/templates/vsphere-cpi-rbac.yaml
```diff
@@ -105,23 +106,24 @@ roleRef:
 subjects:
 - apiGroup: ""
   kind: ServiceAccount
-  name: cloud-controller-manager
+  name: vsphere-cloud-controller-manager
   namespace: {{ .Release.Namespace }}
 - apiGroup: ""
   kind: User
-  name: cloud-controller-manager
+  name: vsphere-cloud-controller-manager
```
PennyScissors, line 113: > Where is the `User` CR from? I do not find its definition in the chart.
@jiaqiluo  These are all changes from upstream that are already present in the chart and I only changed the name to avoid conflicts with rke2. However, is this here what you are referring to?

jiaqiluo: I just found that they are defined in another chart: `rancher-vsphere-csi` 😆

charts/rancher-vsphere-cpi/rancher-vsphere-cpi/100.0.0/templates/vsphere-cpi-rbac.yaml
jiaqiluo: they are defined in another chart: `rancher-vsphere-csi`

charts/rancher-vsphere-cpi/rancher-vsphere-cpi/100.0.0/templates/vsphere-cpi-rbac.yaml
PennyScissors: Ah, that's what you meant. Yeah, cpi and csi go together 😄

jiaqiluo (dismissed):
LGTM

charts/rancher-vsphere-cpi/rancher-vsphere-cpi/100.0.0/templates/vsphere-cpi-rbac.yaml
PennyScissors: resolved [here](https://github.com/rancher/charts/pull/1220#discussion_r661031973)

charts/rancher-vsphere-cpi/rancher-vsphere-cpi/100.0.0/templates/vsphere-cpi-rbac.yaml
```diff
@@ -1,14 +1,15 @@
+{{- if .Values.cloudControllerManager.rbac.enabled -}}
 # Source: https://github.com/kubernetes/cloud-provider-vsphere/blob/master/releases/v1.19/vsphere-cloud-controller-manager.yaml
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: cloud-controller-manager
+  name: vsphere-cloud-controller-manager
```
aiyengar2, line 6: Using `vsphere-` seems fine, but my preference would be to using `{{ .Chart.Name }}` so that we don't have conflicts with any other vsphere resources (e.g. will this potentially conflict with `rancher-vsphere-csi` resources?)

What are your thoughts?

aiyengar2 (approved):
LGTM, but have one last confirmation to make regarding the naming scheme chosen.

charts/rancher-vsphere-cpi/rancher-vsphere-cpi/100.0.0/templates/vsphere-cpi-rbac.yaml
PennyScissors: I don't think it will conflict because they are supposed to be installed together and we probably would've ran into problems by now with the more generic name. I tried `{{ .Chart.Name }}` and `{{ .Release.Name }}` and they end up with super long names so I figured adding vsphere was enough, but I'll go ahead and update it to `{{ .Chart.Name }}` to be safe.

PennyScissors:
@aiyengar2 @jiaqiluo I uncommented `clusterType` and set it to "rke" as the default because even though the chart deploys successfully when it's not set, I noticed `helm template` fails because it can't do the comparison against a value that doesn't exist at that time. I also replaced all the `vsphere-` for `{{ .Chart.Name }}` to address Arvind's comment.

packages/rancher-vsphere-cpi/charts/templates/vsphere-cpi-ds.yaml
```diff
@@ -1,11 +1,11 @@
-# Source: https://github.com/kubernetes/cloud-provider-vsphere/blob/master/releases/v1.19/vsphere-cloud-controller-manager.yaml
+# Source: https://github.com/kubernetes/cloud-provider-vsphere/blob/master/releases/v1.19/{{ .Chart.Name }}-cloud-controller-manager.yaml
```
jiaqiluo, line 1: this is unnecessary 😆 
```suggestion
# Source: https://github.com/kubernetes/cloud-provider-vsphere/blob/master/releases/v1.19/vsphere-cloud-controller-manager.yaml
```

packages/rancher-vsphere-cpi/charts/templates/vsphere-cpi-rbac.yaml
```diff
@@ -1,14 +1,15 @@
-# Source: https://github.com/kubernetes/cloud-provider-vsphere/blob/master/releases/v1.19/vsphere-cloud-controller-manager.yaml
+{{- if .Values.cloudControllerManager.rbac.enabled -}}
+# Source: https://github.com/kubernetes/cloud-provider-vsphere/blob/master/releases/v1.19/{{ .Chart.Name }}-cloud-controller-manager.yaml
```
jiaqiluo, line 2:
```suggestion
# Source: https://github.com/kubernetes/cloud-provider-vsphere/blob/master/releases/v1.19/vsphere-cloud-controller-manager.yaml
```

jiaqiluo (changes requested):

packages/rancher-vsphere-cpi/charts/templates/vsphere-cpi-ds.yaml
PennyScissors: Good catch, did a find and replace and missed that.

jiaqiluo (approved):
LGTM

packages/rancher-vsphere-cpi/charts/values.yaml
```diff
@@ -14,6 +14,11 @@ cloudControllerManager:
...
     tag: v1.2.1
   nodeSelector: {}
+  rbac:
+    enabled: true
+
+# Set to "rke" or "rke2" to apply the node selector label appropriate for the cluster
+clusterType: "rke"
 
 global:
   cattle:
```
cbron, line 24: These changes look fine if its a new chart version. But that begs the question why we are bumping this to 100.0.0 if its a local chart without an upstream version ? Why not move to 2.0.0.
cc @aiyengar2 and @jiaqiluo

PennyScissors: CSI doesn't have an upstream, both CSI and CPI are just a bunch of manifests upstream. As for the version, I thought we did 100.0.0 to be consistent with the others, but either way works so I'll leave that to you guys to decide.

aiyengar2: @PennyScissors is right that the initial purpose was consistency. We moved all charts to 100.0.0 unless specifically directed otherwise since diverting from 100.0.0 is an exception, not the norm, and keeping the chart major version tied to the Rancher release makes it easy for us to call out if a given chart is the right one to be installed in the given version of Rancher.

I think it's fine to move CPI / CSI back to 2.0.0 though if that is preferred for local charts.

PennyScissors: Currently, we have CPI 1.0.0 and CSI 2.1.0 in the release-v2.6 branch. So if we decide to move them back, I think it would need to be CPI 2.0.0 and CSI 3.0.0. I like 100.0.0 because is not only consistent with other charts, but also consistent with each other since they need to be installed together.

packages/rancher-vsphere-cpi/charts/questions.yaml
cbron: For future reference, we should use name of RKE2, not RKE government.

cbron (approved):
changes are fine, we need to address versioning separately

PR #1221, opened 2021-05-21
Add kubelet and nodeExporter pushprox

### LINKED ISSUE #1051: [rancher-monitoring] Add pushprox for kubelet and node-exporter

dbason:
Fixes https://github.com/rancher/rancher/issues/32841

Adds a pushprox for the kubelet and for node-exporter which runs on the host network.

Kubelet ServiceMonitor is still enabled by default, will disable it using values in the dashboard after this is merged. The node-exporter ServiceMonitor is disabled, and the node-exporter pushprox is enabled by default.

This supercedes #1055


packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
```diff
@@ -230,6 +243,35 @@
 +      - effect: "NoSchedule"
 +        operator: "Exists"
 +
++## hardened components PushProx Monitoring
```
aiyengar2, line 246: Can we change this to `## Additional PushProx Monitoring`?

packages/rancher-monitoring/generated-changes/overlay/Chart.yaml.orig
```diff
@@ -0,0 +1,101 @@
+annotations:
```
aiyengar2, line 1: Chart.yaml.orig seems incorrect. This looks like a patch issue.

packages/rancher-monitoring/generated-changes/patch/Chart.yaml.patch
```diff
@@ -16,7 +16,16 @@
  apiVersion: v2
  appVersion: 0.46.0
  dependencies:
-@@ -59,19 +68,17 @@
+@@ -14,7 +23,7 @@
+ - condition: hardenedKubelet.enabled
+   name: hardenedKubelet
+   repository: file://./charts/hardenedKubelet
+-- condition: hardenedNodeExporter.enabled
++- condition: nodeExporter.enabled
```
aiyengar2, line 24: I don't think we should change this condition from `hardenedNodeExporter.enabled` to `nodeExporter.enabled`

packages/rancher-monitoring/generated-changes/patch/templates/exporters/node-exporter/servicemonitor.yaml.patch
```diff
@@ -0,0 +1,8 @@
+--- charts-original/templates/exporters/node-exporter/servicemonitor.yaml
++++ charts/templates/exporters/node-exporter/servicemonitor.yaml
+@@ -1,4 +1,4 @@
+-{{- if .Values.nodeExporter.enabled }}
++{{- if (and .Values.nodeExporter.enabled .Values.nodeExporter.serviceMonitor.enabled) }}
```
aiyengar2, line 5: This should be disabled if PushProx is enabled since PushProx already comes in with a ServiceMonitor

aiyengar2 (changes requested):

packages/rancher-monitoring/generated-changes/patch/Chart.yaml.patch
dbason: Sure I can put it back.  The default value for hardenedNodeExporter is true; my thinking was to always enable the node exporter pushprox when the node exporter is enabled.  I can change that logic if you would prefer to have them separate.

dbason:
@aiyengar2 resolved the minor stuff.  When I was looking through the values.yaml I also found a couple of duplicate entries which I have cleaned up.

Just waiting on confirmation for the nodeExporter settinngs.

packages/rancher-monitoring/generated-changes/patch/Chart.yaml.patch
aiyengar2: Let's keep them separate since the node-exporter flag is reserved for someone who wants to stick with the old approach (no additional pushproxes). This will be an opt-in flag to support deploying node-exporter in hardened clusters where even the nodeExporter port cannot be scraped.

dbason:
@aiyengar2 I think I've covered everything off

- hardenedKubelet, needed to make some RBAC additions for this to work as the kubelet port is authenticated.
- k3sServer: the 10249 port is exposed on agents as well, but unfortunately it doesn't include the cadvisor and probes endpoints that we need for kubelet monitoring, so I have switched to 10250.  Because the kubelet monitoring includes metric path labels this caused inflated numbers for some of the controlplane dashboard counts, so I've adjusted those with a helper.

packages/rancher-monitoring/generated-changes/patch/charts/hardenedKubelet/templates/pushprox-servicemonitor.yaml.patch
```diff
@@ -0,0 +1,54 @@
...
++    honorLabels: true
++{{- if .Values.metricRelabelings }}
++    metricRelabelings:
++{{ tpl (toYaml .Values.metricRelabelings | indent 4) . }}
++{{- end }}
++{{- if .Values.relabelings }}
++    relabelings:
++{{ toYaml .Values.relabelings | indent 4 }}
++{{- end }}
++{{- end }}
```
aiyengar2, line 51: Instead of patching pushprox specifically for hardenedKubelet, could we just update the rancher-pushprox chart to have `.Values.serviceMonitor.metricsConfig` and insert `{{ tpl (toYaml .Values.serviceMonitor.additionalEndpoints) | indent 2 }}` at the bottom of:

https://github.com/rancher/charts/blob/7836aff0c1b2846c3fbcf8ed2adc6963d21c20b1/charts/rancher-pushprox/rancher-pushprox/0.1.300/templates/pushprox-servicemonitor.yaml#L10-L15

We can then add this configuration to the values.yaml. This should come after `{{ tpl (toYaml .Values.serviceMonitor.additionalMetricsEndpointConfig) | indent 4 }}` from above.

packages/rancher-monitoring/generated-changes/patch/charts/hardenedKubelet/templates/pushprox-servicemonitor.yaml.patch
```diff
@@ -0,0 +1,54 @@
...
+     {{- end }}
++    honorLabels: true
++{{- if .Values.metricRelabelings }}
++    metricRelabelings:
++{{ tpl (toYaml .Values.metricRelabelings | indent 4) . }}
++{{- end }}
++{{- if .Values.relabelings }}
++    relabelings:
++{{ toYaml .Values.relabelings | indent 4 }}
++{{- end }}
```
aiyengar2, line 15: Can these values be added under `.Values.serviceMonitor.metricsEndpoint.*` instead of at the root level?

Also, let's add this to the original pushprox chart instead of patching it.

packages/rancher-monitoring/generated-changes/patch/charts/k3sServer/templates/pushprox-clients-rbac.yaml.patch
```diff
@@ -0,0 +1,16 @@
...
+   resourceNames:
+   - {{ template "pushProxy.client.name" . }}
+ {{- if and .Values.clients.https.enabled .Values.clients.https.useServiceAccountCredentials }}
+-- nonResourceURLs: ["/metrics"]
++- nonResourceURLs:
++  - "/metrics"
++  - "/metrics/cadvisor"
++  verbs: ["get"]
++- apiGroups: [""]
++  resources: ["nodes/metrics"]
```
aiyengar2, line 13: same comments as above; let's avoid patching this subchart if it could be a commonly used feature anyways.

packages/rancher-monitoring/generated-changes/patch/charts/k3sServer/templates/pushprox-servicemonitor.yaml.patch
```diff
@@ -0,0 +1,54 @@
...
+     {{- end }}
++    honorLabels: true
++{{- if .Values.metricRelabelings }}
++    metricRelabelings:
++{{ tpl (toYaml .Values.metricRelabelings | indent 4) . }}
++{{- end }}
++{{- if .Values.relabelings }}
++    relabelings:
++{{ toYaml .Values.relabelings | indent 4 }}
++{{- end }}
```
aiyengar2, line 15: same comments as above; let's avoid patching this subchart if it could be a commonly used feature anyways.

packages/rancher-monitoring/generated-changes/patch/charts/k3sServer/templates/pushprox-servicemonitor.yaml.patch
```diff
@@ -0,0 +1,54 @@
...
++{{- end }}
```
aiyengar2, line 51: same comments as above; let's avoid patching this subchart if it could be a commonly used feature anyways.

packages/rancher-monitoring/generated-changes/patch/templates/_helpers.tpl.patch
```diff
@@ -73,6 +91,20 @@
...
++kubelet
++{{- end -}}
++{{- end }}
++
++{{- define "exporter.controlplane.upLabels" -}}
++{{- if .Values.k3sServer.enabled -}}
++, metrics_path=\"/metrics\"
++{{- end -}}
++{{- end }}
++
```
aiyengar2, line 107: I see why you added this, but I think this should not be a helper function. We should patch the expression directly for this, since it only happens a couple of times.

packages/rancher-monitoring/generated-changes/patch/templates/exporters/node-exporter/servicemonitor.yaml.patch
```diff
@@ -0,0 +1,8 @@
+--- charts-original/templates/exporters/node-exporter/servicemonitor.yaml
++++ charts/templates/exporters/node-exporter/servicemonitor.yaml
+@@ -1,4 +1,4 @@
+-{{- if .Values.nodeExporter.enabled }}
++{{- if (and .Values.nodeExporter.enabled (not .Values.hardenedNodeExporter.enabled)) }}
```
aiyengar2, line 5: Instead of this, let's fail if the user has both enabled.
```
{{- if .Values.nodeExporter.enabled }} 
{{- if .Values.hardenedNodeExporter.enabled }}
{{- required "Cannot set both .Values.nodeExporter.enabled and .Values.hardenedNodeExporter.enabled to true" "" }}
{{- end }}
```

packages/rancher-monitoring/generated-changes/patch/templates/grafana/dashboards-1.14/kubelet.yaml.patch
```diff
@@ -9,3 +15,275 @@
    name: {{ printf "%s-%s" (include "kube-prometheus-stack.fullname" $) "kubelet" | trunc 63 | trimSuffix "-" }}
    annotations:
  {{ toYaml .Values.grafana.sidecar.dashboards.annotations | indent 4 }}
+@@ -107,7 +108,7 @@
+                         "tableColumn": "",
+                         "targets": [
+                             {
+-                                "expr": "sum(up{cluster=\"$cluster\", job=\"kubelet\", metrics_path=\"/metrics\"})",
++                                "expr": "sum(up{cluster=\"$cluster\", job=\"{{ include "exporter.kubelet.jobName" . }}\", metrics_path=\"/metrics\"})",
```
aiyengar2, line 23: metrics_path should come from the helper function throughout this file.

packages/rancher-monitoring/generated-changes/patch/templates/grafana/dashboards-1.14/controller-manager.yaml.patch
```diff
@@ -20,7 +20,7 @@
                          "targets": [
                              {
 -                                "expr": "sum(up{job=\"kube-controller-manager\"})",
-+                                "expr": "sum(up{job=\"{{ include "exporter.kubeControllerManager.jobName" . }}\"})",
++                                "expr": "sum(up{job=\"{{ include "exporter.kubeControllerManager.jobName" . }}\"{{ include "exporter.controlplane.upLabels" . }}})",
```
aiyengar2, line 23: directly patch this expression with an if/else if `k3sServer.enabled`.

packages/rancher-monitoring/generated-changes/patch/templates/grafana/dashboards-1.14/proxy.yaml.patch
```diff
@@ -20,7 +20,7 @@
                          "targets": [
                              {
 -                                "expr": "sum(up{job=\"kube-proxy\"})",
-+                                "expr": "sum(up{job=\"{{ include "exporter.kubeProxy.jobName" . }}\"})",
++                                "expr": "sum(up{job=\"{{ include "exporter.kubeProxy.jobName" . }}\"{{ include "exporter.controlplane.upLabels" . }}})",
```
aiyengar2, line 23: directly patch this with an if/else based on `k3sServer.enabled`.

aiyengar2 (changes requested):
Thanks for the changes! 

Overall looks good, but on a high level we should remove the `up` helper and move these changes directly into rancher-pushprox, since adding more functionality to customize the servicemonitor that is deployed would be universally useful, instead of maintaining a patch on the subchart.

packages/rancher-monitoring/generated-changes/patch/templates/grafana/dashboards-1.14/kubelet.yaml.patch
dbason: Metrics path never changes here; this is the upstream value.  The only thing we're changing is the job name

packages/rancher-monitoring/generated-changes/patch/Chart.yaml.patch
```diff
@@ -16,7 +16,16 @@
  apiVersion: v2
  appVersion: 0.46.0
  dependencies:
-@@ -59,19 +68,17 @@
+@@ -38,7 +47,7 @@
+ - condition: prometheus-adapter.enabled
+   name: prometheus-adapter
+   repository: file://./charts/prometheus-adapter
+-- condition: nodeExporter.enabled
++- condition: nodeExporter.enabled,hardenedNodeExporter.enabled
```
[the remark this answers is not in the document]
dbason: Turns out it doesn't quite work like that:
```
Condition - The condition field holds one or more YAML paths (delimited by commas). If this path exists in the top parent's values and resolves to a boolean value, the chart will be enabled or disabled based on that boolean value. Only the first valid path found in the list is evaluated and if no paths exist then the condition has no effect.
```
If the first condition exists, and is falsely then the dependency won't be installed.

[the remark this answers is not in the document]
aiyengar2: @dbason thanks for confirming this. I wasn't aware that the first valid path would be taken, even if it was false.

Instead of using tags, would it work if we switch the order of conditions to `hardenedNodeExporter.enabled,nodeExporter.enabled` and removed `hardenedNodeExporter.enabled` from the values.yaml?

packages/rancher-monitoring/generated-changes/patch/templates/grafana/dashboards-1.14/kubelet.yaml.patch
aiyengar2: Ah sorry. I meant to delete this comment before submitting the review.

packages/rancher-pushprox/charts/templates/pushprox-servicemonitor.yaml
```diff
@@ -13,6 +13,37 @@ spec:
     params:
       _scheme: [https]
     {{- end }}
+    {{- if .Values.endpoint.honorLabels }}
+    honorLabels: true
+    {{- end }}
+    {{- if .Values.endpoint.metricRelabelings }}
+    metricRelabelings:
+{{ tpl (toYaml .Values.endpoint.metricRelabelings | indent 4) . }}
```
aiyengar2, line 21:
```yaml
metricRelabelings: {{ tpl (toYaml .Values.endpoint.metricRelabelings) . | nindent 4 }}
```

packages/rancher-pushprox/charts/templates/pushprox-servicemonitor.yaml
```diff
@@ -13,6 +13,37 @@ spec:
...
+    {{- end }}
+    {{- if .Values.endpoint.relabelings }}
+    relabelings:
+{{ toYaml .Values.endpoint.relabelings | indent 4 }}
```
aiyengar2, line 25: use nindent

packages/rancher-pushprox/charts/templates/pushprox-servicemonitor.yaml
```diff
@@ -13,6 +13,37 @@ spec:
...
+    {{- end }}
+  {{- range .Values.additionalEndpoints }}
+  - port: metrics
```
aiyengar2, line 28: port should also be taken in from additionalEndpoints

packages/rancher-pushprox/charts/templates/pushprox-servicemonitor.yaml
```diff
@@ -13,6 +13,37 @@ spec:
...
+    params:
+      _scheme: [https]
+    {{- end }}
+    path: {{ .path }}
+    {{- if .honorLabels }}
+    honorLabels: true
+    {{- end }}
+    {{- if .metricRelabelings }}
+    metricRelabelings:
+{{ tpl (toYaml .metricRelabelings | indent 4) . }}
```
aiyengar2, line 40: use nindent

packages/rancher-pushprox/charts/templates/pushprox-servicemonitor.yaml
```diff
@@ -13,6 +13,37 @@ spec:
...
+    {{- end }}
+    {{- if .relabelings }}
+    relabelings:
+{{ toYaml .relabelings | indent 4 }}
```
aiyengar2, line 44: use nindent

packages/rancher-pushprox/charts/templates/pushprox-clients-rbac.yaml
```diff
@@ -11,7 +11,12 @@ rules:
   resourceNames:
   - {{ template "pushProxy.client.name" . }}
 {{- if and .Values.clients.https.enabled .Values.clients.https.useServiceAccountCredentials }}
-- nonResourceURLs: ["/metrics"]
+- nonResourceURLs:
+  - "/metrics"
+  - "/metrics/cadvisor"
```
aiyengar2, line 16: `/metrics/cadvisor` should not be hard-coded; instead, it should come from the `path` provided under all `additionalEndpoints`.

packages/rancher-pushprox/charts/templates/pushprox-clients-rbac.yaml
```diff
@@ -11,7 +11,12 @@ rules:
...
+  verbs: ["get"]
+- apiGroups: [""]
+  resources: ["nodes/metrics"]
```
aiyengar2, line 19: Ideally, this should be supplied by the user and not hard-coded; not all ServiceMonitors need this permission.

packages/rancher-monitoring/generated-changes/patch/Chart.yaml.patch
```diff
@@ -16,7 +16,26 @@
...
+ - condition: k3sServer.enabled
+   name: k3sServer
+   repository: file://./charts/k3sServer
+@@ -41,6 +52,9 @@
+ - condition: nodeExporter.enabled
+   name: prometheus-node-exporter
+   repository: file://./charts/prometheus-node-exporter
++  tags:
++  - nodeExporter
++  - hardenedNodeExporter
```
aiyengar2, line 34: I'd like to avoid using tags since this would be a breaking change (for someone who chose to disable nodeExporter before). See https://github.com/rancher/charts/pull/1221#discussion_r643402266.

packages/rancher-monitoring/generated-changes/patch/templates/grafana/dashboards-1.14/controller-manager.yaml.patch
```diff
@@ -20,7 +20,7 @@
                          "targets": [
                              {
 -                                "expr": "sum(up{job=\"kube-controller-manager\"})",
-+                                "expr": "sum(up{job=\"{{ include "exporter.kubeControllerManager.jobName" . }}\"})",
++                                "expr": "sum(up{job=\"{{ include "exporter.kubeControllerManager.jobName" . }}\"{{- if .Values.k3sServer.enabled  }}, metrics_path=\"/metrics\"{{- end }}})",
```
aiyengar2, line 23: do you need the preceding `-` for the `if` and `end`?

aiyengar2 (changes requested):

packages/rancher-monitoring/generated-changes/patch/Chart.yaml.patch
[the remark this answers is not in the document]
dbason: That's what I was thinking of doing initially, but then hardenedNodeExporter would always be installed unless you explicitly set it to false:
> Only the first valid path found in the list is evaluated and if no paths exist then the condition has no effect.

aiyengar2:
@dbason I forked your changes and made a couple of modifications based on the above comments I left behind.

Opened a new PR in https://github.com/rancher/charts/pull/1245; could you give it a review?

aiyengar2:
Closing in favor of https://github.com/rancher/charts/pull/1245

PR #1223, opened 2021-05-21
Add support for the kube audit log to v2 logging

dbason:
This adds support for https://github.com/rancher/rancher/issues/31437 into the chart.

It generates a Logging object to tail the kube-audit log.  It provides the option to define a loggingRef to allow the audit logs to be isolated from other Flows/Outputs,

Includes a helper to calculate the label selector for control plane nodes, which can be overridden in values.

Currently in draft pending some more testing


charts/rancher-logging/rancher-logging/3.9.401+up3.9.4/templates/loggings/kube-audit/logging.yaml
```diff
@@ -0,0 +1,100 @@
...
+      podSecurityPolicyCreate: true
+      roleBasedAccessControlCreate: true
+    {{- end }}
+    {{- $total_tolerations := concat (.Values.tolerations) (.Values.additionalLoggingSources.kubeAudit.tolerations) }}
+    {{- with $total_tolerations }}
+    tolerations: 
+      {{- toYaml . | nindent 6 }}
+    {{- end }}
+    nodeSelector:
+      {{- include "controlplaneSelector" . | nindent 6 }}
```
nickgerace, line 62: nit: We've since removed `$total_tolerations` in other Logging CRs and refactored the style for these two fields

nickgerace: You can probably just use `concat` wrapped in parens after the `with` keyword

packages/rancher-logging/generated-changes/overlay/templates/loggings/kube-audit/logging.yaml
```diff
@@ -0,0 +1,100 @@
...
+  {{- end }}
+  disableKubernetesFilter: true
+  fluentbit:
+    image:
+      repository: {{ template "system_default_registry" . }}{{ .Values.images.fluentbit.repository }}
+      tag: {{ .Values.images.fluentbit.tag }}
+    inputTail:
+      Tag: "kube-audit"
+      Path: {{ $auditLogDir }}/audit-log.json
+      Parser: json
```
nickgerace, line 23: This should probably be adjustable in `values.yaml`

packages/rancher-logging/generated-changes/overlay/templates/loggings/kube-audit/logging.yaml
```diff
@@ -0,0 +1,100 @@
...
+      podSecurityPolicyCreate: true
+      roleBasedAccessControlCreate: true
+    {{- end }}
+    {{- $total_tolerations := concat (.Values.tolerations) (.Values.additionalLoggingSources.kubeAudit.tolerations) }}
+    {{- with $total_tolerations }}
+    tolerations: 
+      {{- toYaml . | nindent 6 }}
+    {{- end }}
+    nodeSelector:
+      {{- include "controlplaneSelector" . | nindent 6 }}
```
nickgerace, line 62: Same as above

packages/rancher-logging/generated-changes/overlay/templates/loggings/kube-audit/logging.yaml
```diff
@@ -0,0 +1,100 @@
+{{- if .Values.additionalLoggingSources.kubeAudit.enabled }}
+{{- $auditLogDir := default "/var/log/kube-audit" .Values.additionalLoggingSources.kubeAudit.logDirectory }}
```
nickgerace, line 2: Will this change based on provider? Will EKS change dirs compared to RKE2, for example?

charts/rancher-logging/rancher-logging/3.9.401+up3.9.4/values.yaml
```diff
@@ -115,6 +115,16 @@ additionalLoggingSources:
...
     enabled: false
+  kubeAudit:
+    enabled: false
+    tolerations:
+    - key: node-role.kubernetes.io/controlplane
+      value: "true"
+      effect: NoSchedule
+    - key: node-role.kubernetes.io/etcd
+      value: "true"
+      effect: NoExecute
```
nickgerace, line 126: Conflicted on whether or not `logDirectory` should be here as a blank value. We may want to change its name to be something with "PathPrefix" to match our style too.

CC: @paynejacob for thoughts too

nickgerace (changes requested):

packages/rancher-logging/generated-changes/overlay/templates/loggings/kube-audit/logging.yaml
dbason: My thinking was to focus on rke/rke2/k3s initially.  I'm not sure whether we even have access to controlplane stuff for hosted providers, I know with AKS we definitely don't.

nickgerace: We do not. Regardless, I think the `values.yaml` field will suffice.

dbason:
@nickgerace based on your comments I've tidied up the config.  I removed the variable assignment at the top of the Logging, as we're not actually doing any modification to the pathPrefix from values, so I figured it would be cleaner just to pass it straight through.

I've also tested and fixed up the new helper function

packages/rancher-logging/generated-changes/overlay/templates/loggings/kube-audit/logging.yaml
paynejacob: Could we default this based on provider and allow for user override?

dbason: Sure, k3s has no default audit logging turned on at all, you have to manually pass the kube-api flags through, so I don't think a default makes sense for that.  I will check what the rke2 behaviour is. Rke is set in the cluster.yaml; is there an accepted way to pull values from there?

dbason: I've refactored the helper I added and added some more to support this.  Also to clean things up I've just hardcoded the destination for the extra mount to `/kube-audit-logs` and then we can pick the file up from there.

For rke2 I've added the default location for a CIS hardened cluster - audit logging is automatically configured as part of the hardened config.

For k3s I've made the values required as there's no way to set a default for that.

nickgerace: Discussed with @dbason offline, but we can assume that we will always know k3s/rke/rke2 with `additionalLoggingSources`. We always know with aks/gke/eks as well, but those are _optional_ via the dashboard checkbox. Thus, it can be a bit confusing (@paynejacob: maybe we should split cloud into its own section in the future?).

We should just use `additionalLoggingSources.$RANCHER_DISTRO.enabled` where possible.

packages/rancher-logging/generated-changes/patch/values.yaml.patch
```diff
@@ -62,6 +62,20 @@
...
++    fluentbit:
++      logTag: kube-audit
++      tolerations:
++      - key: node-role.kubernetes.io/controlplane
++        value: "true"
++        effect: NoSchedule
++      - key: node-role.kubernetes.io/etcd
++        value: "true"
++        effect: NoExecute
++
```
nickgerace, line 78: nit: extra whitespace

packages/rancher-logging/generated-changes/patch/values.yaml.patch
```diff
@@ -62,6 +62,20 @@
 +    enabled: false
 +  gke:
 +    enabled: false
++  kubeAudit:
++    enabled: false
++    pathPrefix: ""
++    auditFilename: ""
++    fluentbit:
```
nickgerace, line 69: nit: alphabetical ordering would be great

packages/rancher-logging/generated-changes/patch/templates/_helpers.tpl.patch
```diff
@@ -34,3 +34,58 @@
 +{{- define "windowsInputTailMount" -}}
 +{{- (include "windowsPathPrefix" .) | replace "C:" "" -}}
 +{{- end -}}
++
```
nickgerace, line 37: This file should reflect this discussion: https://github.com/rancher/charts/pull/1223/commits/dd1b289ef2618c1c158274e9f675673c402294cb#r638422713

paynejacob (approved):

nickgerace (approved):

dbason:
@paynejacob @nickgerace  Just a reminder I don't have merge permissions ;)

paynejacob:
@dbason can you backport this to 2.5 also?

PR #1227, opened 2021-05-24
Add push-prox sub-chart for ingress-nginx in rke and rke2 clusters

jiaqiluo:
Issue: https://github.com/rancher/rancher/issues/32563

This PR contains the following 2 major changes:
- Update rancher-pushprox to support overriding namespace, switching the workload type of the client between Deployment and DaemonSet by adding replicas and affinity in pod.spec
- Update rancher-monitoring to add push-prox sub-charts for ingress-nginx in rke and rke2 clusters


charts/rancher-monitoring/rancher-monitoring/14.5.101+up14.5.0/templates/rancher-monitoring/exporters/ingress-nginx/service.yaml
```diff
@@ -1,24 +0,0 @@
-{{- if .Values.ingressNginx.enabled }}
```
jiaqiluo, line 1: This file is no longer needed. The push-prox chart includes it.

charts/rancher-monitoring/rancher-monitoring/14.5.101+up14.5.0/templates/rancher-monitoring/exporters/ingress-nginx/servicemonitor.yaml
```diff
@@ -1,33 +0,0 @@
-{{- if .Values.ingressNginx.enabled }}
```
jiaqiluo, line 1: same here
This file is no longer needed. The push-prox chart includes it.

packages/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/exporters/ingress-nginx/service.yaml
```diff
@@ -1,24 +0,0 @@
-{{- if .Values.ingressNginx.enabled }}
```
jiaqiluo, line 1: This file is no longer needed. The push-prox chart includes it.

packages/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/exporters/ingress-nginx/servicemonitor.yaml
```diff
@@ -1,33 +0,0 @@
-{{- if .Values.ingressNginx.enabled }}
```
jiaqiluo, line 1: This file is no longer needed. The push-prox chart includes it.

packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
```diff
@@ -230,41 +245,32 @@
...
++          requiredDuringSchedulingIgnoredDuringExecution:
++            - labelSelector:
++                matchExpressions:
++                  - key: "app.kubernetes.io/component"
++                    operator: "In"
++                    values:
++                      - "controller"
++              topologyKey: "kubernetes.io/hostname"
++              namespaces:
++                - "kube-system"
```
jiaqiluo, line 268: The nginx-controller is 
- a Deployment with 1 pod in RKE2 cluster which version <= 1.20
- a Daemonset in RKE2 cluster which version since 1.21

The podAffinity makes the pod to be assigned to the node where the nginx-controller's pod runs.

packages/rancher-pushprox/charts/values.yaml
```diff
@@ -16,6 +16,8 @@ global:
   cattle:
     systemDefaultRegistry: ""
 
+namespaceOverride: ""
```
jiaqiluo, line 19: This new variable is added but not used in the monitoring chart in this PR.

nickgerace:
> This PR contains the following 2 major changes:
>
> - Update rancher-pushprox to support overriding namespace, switching the workload type of the client between Deployment and DaemonSet by adding replicas and affinity in pod.spec
> - Update rancher-monitoring to add push-prox sub-charts for ingress-nginx in rke and rke2 clusters

@jiaqiluo is it possible to split this into two or more commits? Want to make sure the diff is not too large for first round of reviews.

jiaqiluo:
@nickgerace I split the changes into multiple commits and it should be more clear to review. 
The reason why the `make charts` commit changes too many files is that we changed the rancher-pushprox chart, so it updates every dependent charts that are based on the rancher-pushprox chart.

charts/rancher-pushprox/rancher-pushprox/0.1.300/templates/pushprox-clients.yaml
```diff
@@ -1,18 +1,33 @@
+{{- if .Values.clients.deployment.replicas }}
+  {{- if empty .Values.clients.deployment.affinity }}
```
nickgerace, line 2: nit: We can make combine this statements

packages/rancher-pushprox/charts/templates/pushprox-clients.yaml
```diff
@@ -1,18 +1,33 @@
+{{- if .Values.clients.deployment.replicas }}
+  {{- if empty .Values.clients.deployment.affinity }}
```
nickgerace, line 2: nit: we can combine these

nickgerace (dismissed):
LGTM barring nits

charts/rancher-pushprox/rancher-pushprox/0.1.300/templates/pushprox-clients.yaml
jiaqiluo: The condition will be super complex if we want to combine those two lines because 2 of 4 cases are what we want.

<table>
<tr>
	<td> replicas
	<td> affinity
	<td> action
<tr>
	<td> is 0
	<td> empty
	<td> pass
<tr>
	<td> is 0
	<td> not empty
	<td> error out
<tr>
	<td> not 0
	<td> empty
	<td> error out
<tr>
	<td> not 0
	<td> not empty
	<td> pass
</table>

For example, a way to express it is 

{{- if not or (and .Values.clients.deployment.replicas .Values.clients.deployment.affinity) (and  (not .Values.clients.deployment.replicas) (not .Values.clients.deployment.affinity) ) }}
{{- fail ".Values.clients.deployment.affinity can not be empty. Please set either nodeAffinity, inter-pod affinity or anti-affinity." }}
{{- end }}

@nickgerace  And idea how we can simplify the above expression?

packages/rancher-pushprox/charts/values.yaml
```diff
@@ -52,6 +54,15 @@ clients:
...
     caCertFile: ""
 
+  # By default, the client is a DaemonSet, but if the component's pods run on some, not all, nodes,
+  # the client should be a Deployment which pods run on the nodes where the component's run.
+  deployment:
+    # The number of pods the client Deployment should have. 0 means the client is a DaemonSet.
+    replicas: 0
+    # Options to select nodes based on nodeAffinity, inter-pod affinity and anti-affinity.
+    # This can not be empty when replicas is not 0.
+    affinity: {}
```
aiyengar2, line 64: Can a DaemonSet specify `affinity` too? Ideally this should be located in the section where nodeSelectors and tolerations are described.

packages/rancher-pushprox/charts/values.yaml
```diff
@@ -52,6 +54,15 @@ clients:
     keyFile: ""
     caCertFile: ""
 
+  # By default, the client is a DaemonSet, but if the component's pods run on some, not all, nodes,
+  # the client should be a Deployment which pods run on the nodes where the component's run.
+  deployment:
+    # The number of pods the client Deployment should have. 0 means the client is a DaemonSet.
+    replicas: 0
```
aiyengar2, line 61: I don't think we should infer this value based on the default `deployment.replicas` set.

Let's add a `deployment.enabled` flag so that the user specifically needs to set `clients.deployment.enabled=true` to get this working.

Also, for the comment, I'd like to slightly re-word it to make it more clear why a user would ever set this value:
```yaml
# The default intention of rancher-pushprox clients is to scrape hostNetwork metrics across all nodes.
# This can be used to scrape internal Kubernetes components or DaemonSets of hostNetwork Pods in 
# situations where a cloud provider firewall prevents Pod-To-Host communication but not Pod-To-Pod.
#
# However, if the underlying hostNetwork Pod that is being scraped is managed by a Deployment instead,
# this advanced option enables users to deploy the client as a Deployment instead of a DaemonSet.
#
# If a user deploys this feature and the underlying Deployment's number of replicas changes, the user will
# be responsible for upgrading this chart accordingly to the right number of replicas.
```

packages/rancher-pushprox/charts/values.yaml
aiyengar2: Need to add documentation to the README.md for this variable and the others.

aiyengar2 (changes requested):

packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
```diff
@@ -230,41 +245,32 @@
...
++rke2IngressNginx:
 +  enabled: false
-+
-+  ## The namespace to search for your nginx-ingress-controller
-+  ##
-+  namespace: ingress-nginx
-+  
-+  service:
-+    port: 9913
-+    targetPort: 10254
```
aiyengar2, line 244: We cannot make a breaking change to an already released chart, so `ingressNginx` should still exist / be respected.

I think we should follow the pattern of https://github.com/rancher/charts/pull/1245, where:
1. We cannot set `rkeIngressNginx.enabled=true` unless `ingressNginx.enabled=true`. Fail on create if so.
2. If `rkeIngressNginx.enabled=true`, we don't deploy the ServiceMonitor or Service for `ingressNginx`
3. No need to change the ConfigMap since `rkeIngressNginx.enabled` implies `ingressNginx.enabled`.

packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
aiyengar2: Can you add this as a comment to the values.yaml itself?

Also, do we need to specify this in the UI ticket to infer what the value of `rke2IngressNginx.clients.deployment.[replicas|enabled]?` should be based on Cluster Type + k8s version? I'm not sure we support that today. 

What happens if a user has RKE 1.20 cluster and upgrades to a RKE 1.21 cluster? Do we need to add documentation that Monitoring needs to be updated in these situations? 

It slightly worries me since transitioning from a Deployment -> DaemonSet in rke2 k8s 1.21 wouldn't actually show any targets down on the UI on an upgrade, even though you are completely omitting metrics in the latter case.

Maybe the right option here would be to set `rke2IngressNginx.clients.deployment.[replicas|enabled]=[0|false]` by default, which would still work since the affinityRules should work on a DaemonSet (but will leave pods pending), and specify in the docs that an RKE2 cluster with a version <=1.20 can choose to override the deployment flag to prevent Pods from being stuck on Pending (with a giant warning sign that says that they must upgrade monitoring on a k8s upgrade).

aiyengar2 (changes requested):

charts/rancher-pushprox/rancher-pushprox/0.1.300/templates/pushprox-clients.yaml
aiyengar2: Do we need to validate that the user provides affinity rules correctly for Deployments? I think it should be fine to say "use at your own risk, here be dragons" if we call it out as an advanced feature (see comment below): https://github.com/rancher/charts/pull/1227#discussion_r645805780

packages/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/exporters/ingress-nginx/service.yaml
jiaqiluo: this file is brought back because we do not want to introduce breaking changes.

packages/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/exporters/ingress-nginx/servicemonitor.yaml
jiaqiluo: this file is brought back because we do not want to introduce breaking changes.

packages/rancher-pushprox/charts/values.yaml
jiaqiluo: changes are made based on your comments.

packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
jiaqiluo: changes are made based on your comments.

packages/rancher-pushprox/charts/values.yaml
jiaqiluo: changes are made based on your comments.

charts/rancher-pushprox/rancher-pushprox/0.1.300/templates/pushprox-clients.yaml
jiaqiluo: @aiyengar2  You are right. 
I will leave it to the user to make sure the affinity rules is not empty. Also, there are default values in the value.yaml .

packages/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/dashboards/addons/ingress-nginx-dashboard.yaml
```diff
@@ -1,12 +1,12 @@
-# Source: 
-{{- if and .Values.grafana.enabled .Values.grafana.defaultDashboardsEnabled .Values.ingressNginx.enabled }}
+{{- if and .Values.grafana.enabled .Values.grafana.defaultDashboardsEnabled .Values.ingressNginx.enabled (or .Values.rkeIngressNginx.enabled .Values.rke2IngressNginx.enabled) }}
```
aiyengar2, line 1: We don't need this right? `.Values.rkeIngressNginx.enabled` can only be set to true if `.Values.ingress.enabled=true`, so we can omit this line's change.

packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
```diff
@@ -84,6 +84,21 @@
...
++  clients:
++    port: 10015
++    useLocalhost: true
++    tolerations:
++      - effect: "NoExecute"
++        operator: "Exists"
++      - effect: "NoSchedule"
++        operator: "Exists"
++    nodeSelector:
++      node-role.kubernetes.io/worker: "true"
```
aiyengar2, line 100: Just to confirm; this means that the ingress-controller cannot exist on non-worker nodes? Where does this requirement come from?

packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
```diff
@@ -260,6 +275,38 @@
...
++    tolerations:
++      - effect: "NoExecute"
++        operator: "Exists"
++      - effect: "NoSchedule"
++        operator: "Exists"
++    # in the RKE2 cluster, the ingress-nginx-controller is deployed as
++    # a Deployment with 1 pod when RKE2 version is <= 1.20,
++    # a DaemonSet when RKE2 version is >= 1.21
++    deployment:
++      enabled: false
```
aiyengar2, line 294: For future reference / other reviewers: https://github.com/rancher/dashboard/issues/3189

@jiaqiluo we should also figure out if we should release note this requirement or add it somewhere to our docs.

packages/rancher-monitoring/package.yaml
```diff
@@ -1,7 +1,7 @@
 url: https://github.com/prometheus-community/helm-charts.git
 subdirectory: charts/kube-prometheus-stack
 commit: 3ca6ba66032a1efce0500f9ad6f83351ad0604b8
-packageVersion: 1
+packageVersion: 2
```
aiyengar2, line 4: No need to bump the packageVersion; we only bump once per release and 14.5.101 has not been release yet

packages/rancher-pushprox/charts/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-version: 0.1.4
+version: 0.1.5
```
aiyengar2, line 2: Same comment as above. v0.1.4 has not been released yet.

packages/rancher-pushprox/charts/README.md
```diff
@@ -42,6 +43,9 @@ The following tables list the configurable parameters of the rancher-pushprox ch
 | `clients.https.keyFile` | The path to the TLS key file located within `clients.https.certDir`. Required and only used if `clients.https.enabled` is set | `""` |
 | `clients.https.caCertFile` | The path to the TLS cacert file located within `clients.https.certDir`. Required and only used if `clients.https.enabled` is set | `""` |
 | `clients.rbac.additionalRules` | Additional permissions to provide to the ServiceAccount bound to the client. This can be used to provide additional permissions for the client to scrape metrics from the k8s API. Only enabled if clients.https.enabled and clients.https.useServiceAccountCredentials are true | `[]` |
+| `clients.deployment.enable` | Deploy the client as a Deployment if the underlying hostNetwork Pod that is being scraped is managed by a Deployment | `false` |
```
aiyengar2, line 46:
```suggestion
| `clients.deployment.enabled` | Deploys the client as a Deployment (generally used if the underlying hostNetwork Pod that is being scraped is managed by a Deployment) | `false` |
```

packages/rancher-pushprox/charts/templates/pushprox-clients.yaml
```diff
@@ -1,18 +1,28 @@
...
+  {{- end }}
   selector:
     matchLabels: {{ include "pushProxy.client.labels" . | nindent 6 }}
   template:
     metadata:
       labels: {{ include "pushProxy.client.labels" . | nindent 8 }}
     spec:
+      {{- if .Values.clients.deployment.enabled }}
+      affinity: {{ toYaml .Values.clients.deployment.affinity | nindent 8 }}
+      {{- end }}
```
aiyengar2, line 25: affinity should be under `.Values.clients.affinity`, re: https://github.com/rancher/charts/pull/1227#discussion_r645795496

aiyengar2 (changes requested):

packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
jiaqiluo: I created an RKE2 cluster with 1.20 and found it on the ingress-nginx-controller Deployment

packages/rancher-pushprox/charts/README.md
jiaqiluo: fixed.

packages/rancher-pushprox/charts/templates/pushprox-clients.yaml
jiaqiluo: fixed.

packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
jiaqiluo: we can add it to both the release note and the documentation.  I will follow up on it.

packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
aiyengar2: Do you mean RKE1?

aiyengar2 (approved):
LGTM

packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
jiaqiluo: yes, RKE1 cluster.

jiaqiluo: the ingress-nginx Daemonset on RKE1 has the following nodeAffinity

```
    spec:
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: beta.kubernetes.io/os
                operator: NotIn
                values:
                - windows
              - key: node-role.kubernetes.io/worker
                operator: Exists
```

nickgerace (approved):

packages/rancher-monitoring/generated-changes/patch/values.yaml.patch
jiaqiluo: https://github.com/rancher/docs/issues/3326
https://github.com/rancher/dashboard/issues/3201

PR #1228, opened 2021-05-24
[dev-v2.6] update the rancher-k8s-components panel

jiaqiluo:
Issue: https://github.com/rancher/rancher/issues/32559

Changes:
- add the unit to the label when it is applicable 
- show the legends by default

Result:
image


cbron (commented):
lgtm but I have no experience in grafana, so I'll leave it to others to approve. Neat that we only change the one tarball and a couple files.

cbron:
@bashofmann you might be interested in reviewing this

aiyengar2 (changes requested):
@jiaqiluo I don't think this resolves the linked issue. While the `"show": true` fixes LGTM, the titles aren't what need to be changed; the units on the graphs themselves are what are missing.

e.g. the fix requested looks a little bit more like https://github.com/rancher/charts/pull/1154, where we are expecting the API Server Request Rate to have something like [`requests/sec`](https://github.com/grafana/grafana/blob/20d356947c2bfaf461594bca451a43fe1b676a34/packages/grafana-data/src/valueFormats/categories.ts#L380). These will show up on the legend you `show` by default now too.

You may also need to check out something like custom units (https://grafana.com/docs/grafana/latest/panels/field-options/standard-field-options/) for `connections/sec` for Ingress Controller Connections.

But I think we should omit the title changes in this PR.

jiaqiluo:
Close this PR since it is not the right way to fix the issue.

PR #1233, opened 2021-05-26
Kube audit logging v2.5

dbason:
Backport of audit logging to v2.5 as per https://github.com/rancher/charts/pull/1223#issuecomment-848367650


dbason:
Not backporting feature to v2.5

PR #1238, opened 2021-05-29
Fix 2.6 Github Workflow

aiyengar2:


cbron (approved):
Do we need to validate anything in past ? Could something have slipped by that is now in code, or would every new scan check that.

aiyengar2:
> Do we need to validate anything in past ? Could something have slipped by that is now in code, or would every new scan check that.

Every new scan would check that.

PR #1247, opened 2021-06-04
Fix/issue 32301

paynejacob:
https://github.com/rancher/rancher/issues/32301

# Problem
Rancher CIS Benchmark uses an api version that is deprecated in Kubernetes 1.22

# Solution
Updated non compliant resources in compliance with https://kubernetes.io/docs/reference/using-api/deprecation-guide/


superseb (dismissed):

packages/rancher-cis-benchmark/charts/crds/clusterscanreport.yaml
```diff
@@ -1,40 +1,39 @@
...
+                nullable: true
+                type: string
+              lastRunTimestamp:
+                nullable: true
+                type: string
+              reportJSON:
+                nullable: true
+                type: string
+            type: object
+        type: object
```
prachidamle, line 39: newline removed?

prachidamle:
@paynejacob have we tested the cis scan feature with updated CRD chart?

prachidamle (approved):

PR #1259, opened 2021-06-09
Kiali 1.35, Istio 1.10

brendarearden:
**Issue(s)**
https://github.com/rancher/rancher/issues/32330
https://github.com/rancher/rancher/issues/32529
https://github.com/rancher/rancher/issues/32795

**Problem || Solution** 
Adding a recent release Kiali v1.35.0
Adding most recent Istio v1.10.1 release


packages/rancher-istio-1.9/package.yaml
```diff
@@ -1,2 +1,2 @@
 url: local
-packageVersion: 1
```
nickgerace, line 2: I'm a bit confused, do we need to bump Kiali within Istio, which is what caused the package bump?

brendarearden: Yes, with every new kiali version we have to add it to our istio charts which results in an istio chart bump

nickgerace (dismissed):
Reviewed the non-copy and non-`make charts` commits.

packages/rancher-istio-1.9/charts/values.yaml
```diff
@@ -80,6 +78,3 @@ kiali:
 tracing:
   enabled: false
   contextPath: "/jaeger"
-  jaeger:
-    repository: rancher/mirrored-jaegertracing-all-in-one
-    tag: 1.20.0
```
aiyengar2, line 85: does this also exist in the jaeger chart?

packages/rancher-istio-1.9/charts/values.yaml
```diff
@@ -65,8 +65,6 @@ kiali:
     strategy: anonymous
   deployment:
     ingress_enabled: false
-    repository: rancher/mirrored-kiali-kiali
-    tag: v1.32.0
```
aiyengar2, line 69: fyi @nickgerace, brenda and I discussed this offline and we are removing the tag here since it is already tracked in the subchart's values.yaml.

packages/rancher-istio-1.10/charts/Chart.yaml
```diff
@@ -13,7 +13,7 @@ annotations:
   catalog.cattle.io/release-name: rancher-istio
   catalog.cattle.io/ui-component: istio
   catalog.cattle.io/provides-gvr: networking.istio.io.virtualservice/v1beta1
-  catalog.cattle.io/auto-install: rancher-kiali-server-crd=1.32.100-rc01
+  catalog.cattle.io/auto-install: rancher-kiali-server-crd=1.35.1
```
aiyengar2, line 16: @brendarearden can you double check that, when adding your fork as a Chart Repository in Rancher's UI, the Istio chart triggers the CRD install as expected?

It should be the case that this works, but I'm wondering whether our auto-install logic handles the fact that it should ignore the build metadata (e.g. +up) when trying to find a chart with this version string.

aiyengar2 (approved):
LGTM, but have some clarification questions.

packages/rancher-istio-1.9/charts/values.yaml
nickgerace: Makes sense to me! That's what I had thought, so thank you both for checking up on that.

packages/rancher-istio-1.10/charts/Chart.yaml
brendarearden: `This chart requires rancher-kiali-server-crd=1.35.1 but no matching chart was found` When I attempted to install it in the UI

packages/rancher-istio-1.10/charts/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: 1.9.5
+appVersion: 1.10.0
 description: A basic Istio setup that installs with the istioctl. Refer to https://istio.io/latest/ for details.
 name: rancher-istio
-version: 1.9.5
+version: 1.10.0
```
cbron, line 5: Did upgrade notes show anything new and important or breaking ?

packages/rancher-istio-1.9/charts/values.yaml
cbron: Was this not there for a reason before, not as a default or anything, it just got overridden immediately ?

packages/rancher-istio-1.9/charts/values.yaml
brendarearden: https://github.com/rancher/charts/blob/1c33d37907da925fd25760fb6b36952a19d8899a/packages/rancher-tracing/charts/values.yaml#L24 << yep, exact same situation as kiali.

packages/rancher-istio-1.9/charts/values.yaml
brendarearden: From what I remember, we had these there to ensure it was picked up for airgap image list, but we are handling that differently now. These are the same values that are being set in the subchart so there is no reason to override it here.

packages/rancher-istio-1.10/charts/Chart.yaml
brendarearden: @aiyengar2 I went ahead and updated the auto-install to include the metadata at this time, https://github.com/rancher/dashboard/issues/3206 << issue opened here to correct it in the ui so that future charts do not need to include the metadata

packages/rancher-istio-1.10/charts/Chart.yaml
brendarearden: Things that were deprecated: 

- values.global.jwtPolicy=first-party-jwt << we do not use this in our base config file so we did not need to remove it.
- values.global.arch << we do not use this in our base config file so we did not need to remove it. 
- Deprecated the `remote` installation profile << we do not point to this profile, ours is the `default` profile

From upgrade notes:
- PILOT_ENABLE_INBOUND_PASSTHROUGH=false  << this was added and is enabled by default. We could add it to our base config, but its something, that if needed could be set via the overlay file. This is something that should be release noted because if you have pods that bind to `localhost` they will no longer be exposed to external pods. You can run `istioctl experimental precheck` to see whats going to be impacted - this something we could consider adding temporarily to our istio-installer image but I am not keen on that idea
- They also changed how the sidecar injector selector works. You can set `values.sidecarInjectorWebhook.useLegacySelectors=true` if you want to continue with the legacy support. Essentially if you can now have a namspace without an istio-injection label, but set the pod annotation to `sidecar.istio.io/inject=true`. So people shouldn't see changes really, but its possible that they took the label off a namespace but left it on a pod, and now that pod would get picked up in a 1.10 world. 

These were the only super noteworthy things I saw when looking through the 1.10 upgrade notes and release annoucements. None of them really required changes to our existing base-config file.

brendarearden:
I am kinda of confused on the versioning - does the metadata not actually get calculated into the version equation? it seems to me like you could have a 100.0.0+up1.32.0 and then have a 100.0.0+up1.35.0 and they would not conflict in terms of version. I went ahead and bumped it to 100.1.0+up1.35.0 when I bumped the kiali version, but I just want to clarify that is the correct way to version bump in the new system

brendarearden:
https://github.com/rancher/rancher/issues/32831 << to fix this issue, we have to have a new version of istio 1.9.5 to correct the auto-install issue that happened as part of the new build process. This chart already exists in production. This was fixed when Jack bumped all the versions to 100.0.0. The versions I have in this pr are a net new istio 1.10.1 and kiali 1.35.0, both now with a rancher charts version of 100.1.0.

packages/rancher-kiali-server/package.yaml
```diff
@@ -1,5 +1,5 @@
-url: https://kiali.org/helm-charts/kiali-server-1.32.0.tgz
-version: 100.0.0
+url: https://kiali.org/helm-charts/kiali-server-1.35.0.tgz
+version: 100.1.0
```
cbron, line 2: > This was fixed when Jack bumped all the versions to 100.0.0. The versions I have in this pr are a net new istio 1.10.1 and kiali 1.35.0, both now with a rancher charts version of 100.1.0.

So to clarify, a 100.0.0 already exists for 1.32.0 ? Wouldn't you need to split up the kiali package into two folders the same way we do istio ? Its possible/likely that we will wipe out whats in dev and then re-generate current dev + release-v2.6. If that happens then 100.0.0 would go away. Can we just do 1.35.0 (100.0.0+up1.35.0) for both istio 1.9 and 1.10, does it work for both ?

packages/rancher-istio-1.9/package.yaml
```diff
@@ -1,2 +0,0 @@
-url: local
-version: 100.0.0
```
cbron, line 2: Same question here. Thought we were publishing 1.9 this release. If so we need to keep both. We can't just use the generated asset.

cbron (changes requested):
If I understand correctly the versioning will be: 
100.0.0=istio 1.9 & kiali 1.32 (maybe 1.35)
100.1.0=istio 1.10 & kiali 1.35
That makes sense, just need to actually have both here.

brendarearden:
@cbron the new release process is a copy of the assets & charts dir over from the dev branch to the release branch. This means that since the new `100.0.0=istio 1.9 & kiali 1.32` is already in the assets dir it will be released, even though there is not a package for it - we are not rebuilding anything during the release process. I will not be releasing anymore 1.9 versions in 2.6 so its okay to remove the package.

brendarearden:
Update:

- We are not releasing new versions of istio 1.8, so I am removing that package
- We are not releasing new versions of istio 1.9, so I am removing that package
- I am releasing Istio 1.10.1 so it will start at version 100.0.0
- I am releasing Kiali 1.35.0 so it will start at version 100.0.0

cbron (approved):
lgtm, chance we need to add 1.9 back in future but for now we'll just add those to 2.5

aiyengar2 (approved):
LGTM

jiaqiluo (approved):

jiaqiluo:
LGTM

PR #1261, opened 2021-06-09
Force convert non-string buffer values for Logging 3.10.0

nickgerace:
Relevant issue: https://github.com/rancher/rancher/issues/28401


paynejacob:
@nickgerace   Have you tested this with a quoted string in values.yaml?

packages/rancher-logging/package.yaml
```diff
@@ -1,5 +1,5 @@
 url: https://kubernetes-charts.banzaicloud.com/charts/logging-operator-3.10.0.tgz
-packageVersion: 0
+packageVersion: 1
```
aiyengar2, line 2: I agree with bumping the packageVersion to 1 here (since a packageVersion of 0 doesn't sound right), but we shouldn't bump the packageVersion > once per release.

aiyengar2 (commented):
LGTM, but add a commit to clean up https://github.com/rancher/charts/tree/dev-v2.6/charts/rancher-logging/rancher-logging/3.10.0%2Bup3.10.0.

nickgerace:
@paynejacob: Yes. Here is an example with the double quote used directly in `values.yaml`.
```
  spec:
    controlNamespace: cattle-logging-system
    fluentbit:
      image:
        repository: rancher/mirrored-fluent-fluent-bit
        tag: 1.7.4
      inputTail:
        Multiline_Flush: "8"
```

paynejacob (approved):

packages/rancher-logging/package.yaml
nickgerace: Done

aiyengar2 (approved):
LGTM

aiyengar2 (approved):
LGTM

PR #1256, opened 2021-06-09
[2.6 Staging] Migrate to v0.2.1 of the charts scripts

jiaqiluo:
issue https://github.com/rancher/charts/issues/1264
As the commit messages indicate, this PR includes the following changes:
- update the `chart-build-scripts` to v0.2.1
- update templates 
- update all influenced charts to follow new pattern and requirements 
- pass `./scripts/package-ci` and `make validate`

To Reviewers:
The changes are split into smaller commits to make it easier to review.

PR #1265, opened 2021-06-10
Change autio-install version, delete existing assets

brendarearden:
**Issue**
https://github.com/rancher/rancher/issues/32831

**Problem**
When installing 1.9.501 and 1.8.601, error `This chart requires rancher-kiali-server-crd=1.32.100-rc01 but no matching chart was found` occurs

**Solution**
Correct the chart.yaml for both charts to point to the correct rancher-kiali-server-crd and rebuilt the assets


nickgerace:
Does this need new `make charts` generation?

brendarearden:
@nickgerace it has it, I just squashed it already on accident, but left it since I was only changing those 1 lines. If you go into the commit, and then click the 3 dots, you can see the longer commit message includes the makes. I can separate it back out if I need to.

nickgerace (approved):

brendarearden:
https://github.com/rancher/charts/pull/1256 fixed as part of this PR

PR #1277, opened 2021-06-16
fix the permission issue for the job using the image rancher/shell

jiaqiluo:
https://github.com/rancher/rancher/issues/32358

According to the offline discussion, we will use the `rancher/kubectl` image to avoid the requirements for root user permission.

Also,rancher-alerting-drivers 100.0.0 is a local chart and version is not changed, so there is not update in the `index.yaml` file.


nickgerace (commented):
Does this fail SELinux? Or do we care about SELinux?

index.yaml
```diff
@@ -642,7 +642,7 @@ entries:
       repository: file://./charts/sachet
     description: The manager for third-party webhook receivers used in Prometheus
       Alertmanager
-    digest: 7d164438ac543b36e1b2f7aed50c7b45551464a357e60ac42a79160f934c3558
+    digest: ad5d4a31db37f433b3381981b8ff6e26d421f5b1920101c81e452e4e2760fc5a
```
jiaqiluo, line 645: This is expected since we updated the content of the chart, so a new digest value is generated.

packages/rancher-alerting-drivers/charts/templates/hardened.yaml
```diff
@@ -83,7 +82,7 @@ spec:
   hostIPC: false
   hostPID: false
   runAsUser:
-    rule: 'MustRunAsNonRoot'
+    rule: 'RunAsAny'
   seLinux:
     rule: 'RunAsAny'
```
jiaqiluo, line 87: we need RunAsAny because the pod needs to run as root to be able to use inClusterConfig

nickgerace (dismissed):

packages/rancher-alerting-drivers/charts/values.yaml
```diff
@@ -7,8 +7,8 @@ global:
     # the registry where all images will be pulled from
     systemDefaultRegistry: ""
   kubectl:
-    repository: rancher/shell
-    tag: v0.1.6
+    repository: rancher/kubectl
+    tag: v1.20.2
```
jiaqiluo, line 11: According to the offline discussion, we will use the `rancher/kubectl` image to avoid the requirements for root user permission.

thedadams (approved):

cbron (approved):
At first thought the tar got deleted but just a small change.

PR #1279, opened 2021-06-18
Add fleet v0.3.5 chart back

StrongMonkey:


jiaqiluo (changes requested):
I see fleet 0.3.500 in the release-v2.5 branch https://github.com/rancher/charts/tree/release-v2.5/assets/fleet

should we add 0.3.500 instead of 0.3.500+up0.3.5?

StrongMonkey:
@jiaqiluo Not sure if it is possible given the current chart build script version. But.. I can manually pick changes from previous version, not sure if this matters.

jiaqiluo:
@StrongMonkey  we can get the release-v2.6 synced with the release-v2.5 first, then you just need to run `make validate` to get the 0.3.500 in your branch.

aiyengar2 (changes requested):
@strongmonkey, as discussed offline the fix to add back in this chart will be done in https://github.com/rancher/charts/pull/1281.

Can you close out this PR in favor of that (unless there are any other concerns?)

PR #1283, opened 2021-06-18
Revert istio version change

brendarearden:
**Issue**
https://github.com/rancher/rancher/issues/32831

TLDR;
This reverts the version changes for istio 1.9.5, we do not need to release a new 1.9.5 version in 2.6

**Problem**
Since we are introducing istio 1.10 in the 2.6 release, we do not want to bump the version for istio 1.9.5. The auto-install version needs updated because the rc versions are no longer getting dropped. 

**Solution**
Do not bump the istio 1.9.5 version, but we build the existing version with the correct auto-install option. The dev-v2.5 branch version works, so I don't think there shouldn't be a need to bump this version to fix the auto-install version

**What is currently released in v2.5**
https://github.com/rancher/charts/blob/release-v2.5/charts/rancher-istio-1.9/rancher-istio/1.9.500/Chart.yaml
https://github.com/rancher/charts/tree/release-v2.5/charts/rancher-kiali-server/rancher-kiali-server
**What is currently in release v2.6**
https://github.com/rancher/charts/blob/release-v2.6/charts/rancher-istio-1.9/rancher-istio/1.9.500/Chart.yaml
https://github.com/rancher/charts/tree/release-v2.6/charts/rancher-kiali-server/rancher-kiali-server

**The remaining chart changes for istio 1.9.5 happened due to the chart build scripts changing and do not need to be released because these chart changes will carry over into the new 1.10.x chart:**
https://github.com/rancher/charts/commit/8845d9ae9f554a26559e63b8da6bbcb88e6f73d5#diff-8c8f823e4ea6b99f20a772bee097b7fcc47e8893b26b51f7f1a74fb70a54aa94
https://github.com/rancher/charts/commit/09f8f4b5bfb750ecb7468f76827ef6baf2f8b0b9#diff-8c8f823e4ea6b99f20a772bee097b7fcc47e8893b26b51f7f1a74fb70a54aa94


packages/rancher-istio-1.9/package.yaml
```diff
@@ -1,2 +1,2 @@
 url: local
-version: 100.0.0
```
jiaqiluo, line 2: If I understand it correctly, we want to keep the rancher-istio-1.9 chart identical to the dev-v2.5 branch, right? 
If so, could we make this file identical to its counterpart in [dev-v2.5](https://github.com/rancher/charts/blob/dev-v2.5/packages/rancher-istio-1.9/package.yaml)

packages/rancher-kiali-server/package.yaml
```diff
@@ -1,5 +1,5 @@
 url: https://kiali.org/helm-charts/kiali-server-1.32.0.tgz
-version: 100.0.0
```
jiaqiluo, line 2: Same here, can we make this file identical to https://github.com/rancher/charts/blob/dev-v2.5/packages/rancher-kiali-server/package.yaml ?

jiaqiluo (changes requested):

packages/rancher-istio-1.9/package.yaml
brendarearden: I just updated the PR to include why I believe we do not need to do that:

Essentially the version that is in production right now is istio 1.9.500 and kiali 1.32.100. Since then there have been two PRs that made changes to the istio chart but both are due to chart build script changes and not actual istio or kiali upstream changes (8845d9a#diff-8c8f823e4ea6b99f20a772bee097b7fcc47e8893b26b51f7f1a74fb70a54aa94
09f8f4b#diff-8c8f823e4ea6b99f20a772bee097b7fcc47e8893b26b51f7f1a74fb70a54aa94). With this in mind, I think this is just part of the divergence that we will start seeing from the two branches and we should only keep them the same up to what has been released in production.

jiaqiluo: I agree with you. 
And here is my 2 cents:
- we are not changing or editing the released chart in this PR
- either your current changes and my "keeping identical to the dev-2.5" approach will generate the same charts with the same versions
- if we have no plan to release new 1.9.x charts from this branch (dev-v2.6), then I thought it might be easier to just leave the files identical to the dev-v2.5 branch. 

But your changes and the reasons make sense to me as well.

brendarearden: That makes sense - I can upgrade it to match the dev-v2.5 branch exactly.

jiaqiluo:
@brendarearden  We also need to clean up the unused `100.0.0+up` charts for Istio and Kiali

based on your PR, can you also:
- delete `rancher-kiali-server-100.0.0+up1.32.0`, `rancher-kiali-server-crd-100.0.0+up1.32.0`, and `rancher-istio-100.0.0` from the dir charts/ and assests/
- delete `index.yaml`
- run `make charts`

packages/rancher-istio-1.9/package.yaml
brendarearden: @jiaqiluo I just realized - dev 2.5 has packageVersion and 2.6 is using version - wont that cause a problem if I try to make it exactly like 2.5?

jiaqiluo: @brendarearden It should continue working. packageVersion is stilled supported, we just avoid it in new charts.

index.yaml
```diff
@@ -12,13 +12,13 @@ entries:
...
     appVersion: 0.3.6-rc2
-    created: "2021-06-18T15:16:27.617227-07:00"
+    created: "2021-06-21T15:54:59.221518-07:00"
     dependencies:
     - condition: gitops.enabled
       name: gitjob
       repository: file://./charts/gitjob
     description: Fleet Manager - GitOps at Scale
-    digest: e0851254ef96f419c2d2bd0976ffe1651e85b57a3a3c895ae83c7b3df972e09f
+    digest: 5254dee03378d776bccdf95cc307801b98154bcbc14931b08a431fd347f0007b
```
jiaqiluo, line 21: This is a known issue that @arvind and I are aware of and need to investigate. It does not affect the PR.

jiaqiluo (approved):
LGTM

jiaqiluo (approved):

brendarearden:
We are no longer going to revert these version changes.

PR #1286, opened 2021-06-21
Update rancher-webhook to 0.1.1-rc1

ryansann:
Issue:
- https://github.com/rancher/rancher/issues/33286


aiyengar2 (changes requested):
needs to clean up old resources under `0.1.0`

index.yaml
```diff
@@ -12,9 +12,9 @@ entries:
       catalog.cattle.io/release-name: fleet
     apiVersion: v2
     appVersion: 0.3.5
-    created: "2021-06-11T16:24:22.11119-07:00"
+    created: "2021-06-21T15:12:36.065586-07:00"
     description: Fleet Manager - GitOps at Scale
-    digest: 2c1aadc3bf10e1d5d732ebc6a8a52730f33c9d159362561df9ac2f5d5b0a493c
+    digest: 6d0dbc8d82b1a485136dbb3588bf58bc88b610ecc7901557c6a65d70f12ec87b
```
aiyengar2, line 17: @jiaqiluo and I have seen these digests changing on regenerating the index.yaml for certain charts but have not been able to isolate why this is the case. This is a known issue but should not impact this PR.

index.yaml
```diff
@@ -837,10 +837,10 @@ entries:
       catalog.cattle.io/ui-component: rancher-cis-benchmark
     apiVersion: v1
     appVersion: v1.0.4
-    created: "2021-06-11T16:24:24.721891-07:00"
+    created: "2021-06-21T15:12:36.089308-07:00"
     description: The cis-operator enables running CIS benchmark security scans on
       a kubernetes cluster
-    digest: 72d5e03b9b75c36507ba56bbe762baddca7348167bf2b811a57b21810f2be064
+    digest: ef7c8837f09bcf669bcb7aef5aec2dca84a27e2c93da5b44194fb0113a1c527d
```
aiyengar2, line 843: same comment as above; known issue, not a problem afaik

index.yaml
```diff
@@ -942,9 +942,9 @@ entries:
       catalog.cattle.io/namespace: cis-operator-system
       catalog.cattle.io/release-name: rancher-cis-benchmark-crd
     apiVersion: v1
-    created: "2021-06-11T16:24:24.722162-07:00"
+    created: "2021-06-21T15:12:36.090248-07:00"
     description: Installs the CRDs for rancher-cis-benchmark.
-    digest: 29cdcf281d00393368c81e8b90154b2d2971ee91806b31ec3842522827d4a72a
+    digest: df1bf2270629356a5ad545053da7f18b1782a5441ed98c66d6030c41a3d421d0
```
aiyengar2, line 947: same comment

index.yaml
```diff
@@ -1103,10 +1103,10 @@ entries:
       catalog.cattle.io/ui-component: gatekeeper
     apiVersion: v1
     appVersion: v3.3.0
-    created: "2021-06-11T16:24:25.450931-07:00"
+    created: "2021-06-21T15:12:36.096886-07:00"
     description: Modifies Open Policy Agent's upstream gatekeeper chart that provides
       policy-based control for cloud native environments
-    digest: 62ece8260448bec8536b8c096c5a0ec0fff8dc9c417234c7d29f4c4ff9b6ff73
+    digest: 5f4530d14b1ede9d9b0c82d3c2dcda473ee45eb87cc5e5140e5a07c381814670
```
aiyengar2, line 1109: same comment

index.yaml
```diff
@@ -1116,8 +1116,8 @@ entries:
     sources:
     - https://github.com/open-policy-agent/gatekeeper.git
     urls:
-    - assets/rancher-gatekeeper/rancher-gatekeeper-3.3.1+up3.3.0.tgz
-    version: 3.3.1+up3.3.0
+    - assets/rancher-gatekeeper/rancher-gatekeeper-3.3.001.tgz
+    version: 3.3.001
```
aiyengar2, line 1120: Nothing actually changed with gatekeeper; the ordering was just flipped when regenerating the index.yaml. See below for the exact change applied to the original location of `3.3.1+up3.3.0`.

index.yaml
```diff
@@ -3176,10 +3176,10 @@ entries:
       catalog.rancher.io/release-name: rancher-pushprox
     apiVersion: v1
     appVersion: 0.1.0
-    created: "2021-06-11T16:26:32.981122-07:00"
+    created: "2021-06-21T15:12:36.323586-07:00"
     description: Sets up a deployment of the PushProx proxy and a DaemonSet of PushProx
       clients.
-    digest: 6621b424fd9b1a94ff271c0bd0e06f42a85a02d6b1162c2e2005ac2491d06513
+    digest: f1f1e0d4d24a20b4b995a8ecc360ff0db310d4af92ec7fcdf87e4b9d5f977dd4
```
aiyengar2, line 3182: same comment as above; unsure why the digest changed

index.yaml
```diff
@@ -3259,9 +3259,9 @@ entries:
       catalog.cattle.io/release-name: rancher-sachet
     apiVersion: v2
     appVersion: 0.2.3
-    created: "2021-06-11T16:26:33.081767-07:00"
+    created: "2021-06-21T15:12:36.324442-07:00"
     description: A Helm chart for Sachet based on the upstream https://github.com/messagebird/sachet
-    digest: 63522af80700a095f1a18de7045e1410ffebc9111ca94e50e6a793fe0adbecf6
+    digest: 0c438e1f63546d35f8e48fbafdc5ef2bbe0fa8d59383f7a36b4f601c99677029
```
aiyengar2, line 3264: same comment as above

index.yaml
```diff
@@ -3376,13 +3391,13 @@ entries:
...
+    created: "2021-06-21T15:12:36.328532-07:00"
     description: ValidatingAdmissionWebhook for Rancher types
-    digest: de5fa7d48cd66c158f30dd02a3ef0c3b1ed6e431e31bf5b9929141a6a0af0e45
+    digest: 3c20cc0a6b0fdc9672d7fa84be46f74b208a00664397d1abcfd8acd9ae10ff0e
     name: rancher-webhook
     urls:
-    - assets/rancher-webhook/rancher-webhook-0.1.0.tgz
-    version: 0.1.0
+    - assets/rancher-webhook/rancher-webhook-0.1.000.tgz
+    version: 0.1.000
```
aiyengar2, line 3400: It seems like dev-v2.5 had an incorrect index.yaml for quite a bit of time...

https://github.com/rancher/charts/blob/dev-v2.5/index.yaml#L3371-L3385 shows that dev-v2.5 is missing 0.1.000, which exists in release-v2.5 https://github.com/rancher/charts/blob/037989f43e8090a362f86e3ad6a21c981337a0f8/index.yaml#L3129-L3143. It replaces 0.1.000 with 0.1.0, which was incorrect.

The current PR rewrites 0.1.0 => 0.1.000 and then adds the new version, so the changes in the index.yaml also look correct here to rectify a prior mistake.

cc: @jiaqiluo not sure if there was a missing step here during a migration.

index.yaml
```diff
@@ -3494,15 +3509,15 @@ entries:
...
-    digest: d12341801bbe55648737273240485a98ac1da7d21a67d3a8ade45a82aa648528
+    digest: 8baa5c330cc152b3d3d87f918ed3ff96b927efad412f4b97bd7db90445e28602
     home: https://rio.io
     icon: https://charts.rancher.io/assets/logos/rio.svg
     name: rio
     urls:
-    - assets/rio/rio-0.8.1.tgz
-    version: 0.8.1
+    - assets/rio/rio-0.8.001.tgz
+    version: 0.8.001
```
aiyengar2, line 3520: Another artifacts of my above comment, possibly a migration miss that was resolved by this PR.

aiyengar2 (approved):
LGTM. The index.yaml has some weird changes, but those rectified mistakes that were possibly made during the migration. The current contents of the index.yaml look like the right contents though, so this PR should be good to merge and resolve other issues too.

PR #1300, opened 2021-06-24
Remove istio 1.9.501 as it will not be released

brendarearden:
Since we are releasing 1.9.600 next, we will not be releasing 1.9.501


jiaqiluo (approved):
LGTM

aiyengar2 (approved):

PR #1304, opened 2021-06-25
Istio 1.9.6

brendarearden:
Fixing the auto-install version to point to the version of kiali that is in production.


cbron (dismissed):

cbron (commented):
Istio part here is right, we only want one version of Istio going in to release branch and no versions of Kiali. But something is wrong with index.yaml.

brendarearden:
index.yaml was not updating as expected, so I deleted it and ran `helm repo index assets/ --url assets/ `. This resulted in a index.yaml with additional entries that were previously missing.

index.yaml
```diff
@@ -12,7 +12,26 @@ entries:
       catalog.cattle.io/release-name: fleet
     apiVersion: v2
     appVersion: 0.3.5
-    created: "2021-06-24T10:32:13.535986-07:00"
+    created: "2021-06-24T21:41:35.636627-07:00"
```
aiyengar2, line 15: @jiaqiluo @brendarearden and I had a long discussion about the changes in the index.yaml and this looks like the right change to make. Thank you @brendarearden for catching this!

### TLDR 

This was a **bug in the migration process to v0.2.0 of the charts-build-scripts** in handling semver build metadata versions within the index.yaml. 

If you take a look at the index.yaml from our initial PR, Fleet 0.3.500 was in release-v2.5 but never existed in dev-v2.5 as it was replaced by 0.3.500+up0.3.5.

https://github.com/rancher/charts/blob/f3efb3d739ffe493e9d2cea2c18bc9b73b866823/index.yaml#L4-L22

### Is this a bug with upstream Helm?

To be clear, this is not a Helm issue; this is an issue with the difference between Rancher's expectations and Helm's expectations, in that Rancher treats charts with different build metadata versions as **different charts** v.s. Helm treats them **as the same chart**. 

(side note: this is also related to a previous issue @brendarearden has flagged in rancher/dashboard in that we do not use semver contraints for the `catalog.cattle.io/auto-install` annotation)

### Why does this happen?

If you take a look at [the way Helm does index.yaml merges](https://github.com/helm/helm/blob/a499b4b179307c267bdf3ec49b880e3dbd2a5591/pkg/repo/index.go#L238-L248), it only modifies the index.yaml for a chart if `!i.Has(cv.Name, cv.Version)`, which in turn runs `i.Get(name, version)`.

`i.Get(name, version)` will return the version [based on whether the version matches a semver Constraint](https://github.com/helm/helm/blob/a499b4b179307c267bdf3ec49b880e3dbd2a5591/pkg/repo/index.go#L212-L214), **not** based on a string match. This was different than our expectation since this means that if Helm sees a chart like `0.3.500+ANYTHING` and is trying to add `0.3.500` to the index, it will **keep** the `0.3.500+ANYTHING` in the index.yaml on a `helm repo index --merge`. This behavior is **not** observed on a `helm repo index` since that generates the index.yaml from scratch, in which case this semver-based constraint check is not performed.

### How can we resolve this in the future?

We need to make broader changes to the way charts-build-scripts manages in the index.yaml file to make it easier for devs to use (e.g. automatically dropping untracked charts). This should be prioritized as soon as possible since it would also reduce the amount of time we're spending on PRs. cc: @cbron 

### So... is this PR good to go?

Yes. The Helm logic would only prevent a 0.3.500 from being added into the chart if 0.3.500+up0.3.5 already exists; if both are in the index.yaml it does not merge it.

However, if someone does need to bump the index.yaml in the future by entirely recreating it, they should use `helm repo index` like brenda did, not `helm repo index --merge` or relying on the `make charts` command as previously recommended.

aiyengar2 (approved):
LGTM. See notes for why the index.yaml changes happened.

jiaqiluo (approved):
LGTM

PR #1290, opened 2021-06-22
Remove old Monitoring assets

aiyengar2:
Forgot to delete old assets in https://github.com/rancher/charts/pull/1274

PR #1274, opened 2021-06-15
Rebase Monitoring to 16.x.x

aiyengar2:
Related Issue: https://github.com/rancher/rancher/issues/32519

This PR absorbs the changes from https://github.com/rancher/charts/pull/1250, https://github.com/rancher/charts/pull/1251, https://github.com/rancher/charts/pull/1252, and https://github.com/rancher/charts/pull/1258

---

- [x] Image Mirror PR: https://github.com/rancher/image-mirror/pull/125

PR #1307, opened 2021-06-26
Update Grafana to latest that uses Apache license

aiyengar2:
This PR contains changes to shift Grafana's base chart to 6.11.0 of the upstream Helm chart.

It also makes additional changes to bump images beyond what is tracked in upstream, namely:
- grafana itself (moved to 7.5.8, latest version of 7.5.x line)
- curl
- grafana-image-renderer
- k8s-sidecar

Related Issue: https://github.com/rancher/rancher/issues/33210

---

- [x] https://github.com/rancher/image-mirror/pull/133


nickgerace (approved):

packages/rancher-grafana/package.yaml
```diff
@@ -1,4 +1,4 @@
 url: https://github.com/grafana/helm-charts.git
 subdirectory: charts/grafana
-commit: 837b4ace32278eee14310f1d654d651d0898bb19
+commit: bf7e1110a5ee9258190d0377fea319bb8e764e62
```
jiaqiluo, line 3: this commit is used in Grafana helm chart starting at 6.12.0 https://github.com/grafana/helm-charts/commit/bf7e1110a5ee9258190d0377fea319bb8e764e62

and Grafana chart 6.12.0 uses Grafana image tag 8.0.0

But the original issue says we want to use chart 6.11.0 https://github.com/rancher/rancher/issues/33210

@aiyengar2 Can you double-check the versions?

jiaqiluo (changes requested):

packages/rancher-grafana/package.yaml
aiyengar2: @jiaqiluo Grafana is 6.11.0 in this commit:
https://github.com/grafana/helm-charts/blob/bf7e1110a5ee9258190d0377fea319bb8e764e62/charts/grafana/Chart.yaml#L3-L4.

We don't want to upgrade to 8.x.x since >7.x.x is APGLv3 instead of Apache licensed.

jiaqiluo (changes requested):
@aiyengar2  Why do we need to delete the charts for rancher-kube-state-metrics, rancher-node-exporter, and rancher-prometheus-adapter?

aiyengar2:
> @aiyengar2 Why do we need to delete the charts for rancher-kube-state-metrics, rancher-node-exporter, and rancher-prometheus-adapter?

TLDR is:
1) Neither https://github.com/rancher/charts/pull/1274 or the followup in https://github.com/rancher/charts/pull/1290 removed the old assets from the released dependencies
2) The index.yaml was not updated to point at the right chart with the +up annotation as a result of https://github.com/rancher/charts/pull/1304#discussion_r658930831

jiaqiluo (approved):
LGTM

PR #1314, opened 2021-06-29
Change CRD install image to rancher/shell and bump nginx

aiyengar2:
Switches from using rancher/rancher-agent, which is versioned based on Rancher releases, to rancher/shell.

Also bumps the nginx image to the latest

rancher/shell does require us to run as root user (as opposed to non-root), so the PSP and securityContext has been changed accordingly. 

@jiaqiluo and I talked about this offline and decided this doesn't have high security risk since: 
1) we only run this job / have this Pod available for a short period of time during an upgrade
2) Other than being able to run as the root user, you still don't have the ability to do anything on the host since this is not a privileged pod or one with access to the hostNetwork.

---

- [x] https://github.com/rancher/image-mirror/pull/134


brendarearden (approved):
lgtm

jiaqiluo (approved):
LGTM

PR #1317, opened 2021-06-30
Remove rancher-operator

thedadams:
rancher-operator is no longer needed in Rancher 2.6+, so it should be
removed here.


jiaqiluo:
@thedadams 
I think we need to keep those assets and charts of the released `rancher-operator` to make `dev-v2.6` a superset of `release-v2.6`. 

Although rancher-operator is deprecated in rancher v2.5, we still need to keep them in `release-v2.6` for users who upgraded from v2.5 to v2.6

thedadams:
@jiaqiluo I am going to remove from release-v2.6 as well. Does the fact that rancher-operator is being uninstalled on upgrade change your opinion on this?

jiaqiluo:
@thedadams  I worry if there will be a race condition.
we are doing two things during or right after upgrading rancher:
1. helm uninstall rancher-operator
2. change the branch of `rancher-charts` from `release-v2.5` to `release-v2.6` 

Helm needs to look up the chart prior to uninstalling it. If no.2 happens before no.1, Helm will fail to uninstall rancher-operator because it cannot find the chart's info. 

Do you know the order of those 2 operations? I am ok with your change if we ensure that no.1 always happens beofer no.2

thedadams:
Closing this because of the chart deprecation issue.

PR #1318, opened 2021-06-30
Remove rancher-operator

thedadams:
rancher-operator is no longer used in Rancher v2.6+ and is uninstalled on
upgrade. Therefore, it is no longer needed and should be removed.


thedadams:
Closing this because of the chart deprecation issue.

PR #1085, opened 2021-04-06
Rancher-alerting-drivers and sub-charts

jiaqiluo:
This PR contains the following 3 charts:
- rancher-alerting-drivers
- rancher-sachet
- rancher-prom-to-teams

Issue: https://github.com/rancher/rancher/issues/29951

PR #1319, opened 2021-06-30
Remove alertmanager pre-install job in favor of Helm annotations

aiyengar2:
Thanks @jiaqiluo for coming up with the idea for this fix in https://github.com/rancher/charts/pull/1085#discussion_r612018643!

p.s. this has the added benefit of speeding up the monitoring install process since we no longer have to wait on a pre-install job 😄 

Related Issue: https://github.com/rancher/rancher/issues/32113


packages/rancher-monitoring/generated-changes/patch/templates/alertmanager/secret.yaml.patch
```diff
@@ -1,164 +1,28 @@
...
-+  name: alertmanager-{{ template "kube-prometheus-stack.fullname" . }}-pre-install
++  name: {{ $secretName }}
    namespace: {{ template "kube-prometheus-stack.namespace" . }}
 -{{- if .Values.alertmanager.secret.annotations }}
    annotations:
-+    "helm.sh/hook": pre-install
-+    "helm.sh/hook-delete-policy": hook-succeeded, hook-failed
++    "helm.sh/hook": pre-install, pre-upgrade
 +    "helm.sh/hook-weight": "3"
++    "helm.sh/resource-policy": keep
```
jiaqiluo, line 18: Why do we want to keep the secret now?
If it is ok to keep it, why we deleted it previously?

nickgerace (approved):

packages/rancher-monitoring/generated-changes/patch/templates/alertmanager/secret.yaml.patch
aiyengar2: We didn't delete it previously. Before, we kept it by default but provided a way to run a job on uninstall to delete it.

However, I removed that cleanup job because 1) we never exposed it on the UI anyways and I don't think it was specifically requested and 2) it required adding a dependency on another image, which doesn't seem to be necessary when all you would need to do is manually clean up the secret with a single kubectl command.

jiaqiluo (approved):
LGTM

PR #1324, opened 2021-07-01
Add rewrite rule to Prometheus and Grafana for /k8s/clusters

aiyengar2:
The root cause of the issue could have been a result of https://github.com/prometheus/prometheus/pull/8142.

In `main` today, the pathPrefix is literally embedded into the template that is rendered, as seen in https://github.com/prometheus/prometheus/blob/90976e7505c225a44d4f0d4b6978f2b42fe0dcb8/console_libraries/prom.lib#L4-L13.

Previously, our approach was to modify the Javascript env var `PATH_PREFIX` [here](https://github.com/prometheus/prometheus/blob/90976e7505c225a44d4f0d4b6978f2b42fe0dcb8/console_libraries/prom.lib#L16), as seen in our nginx config file:

https://github.com/rancher/charts/blob/release-v2.6/charts/rancher-monitoring/rancher-monitoring/14.5.100/templates/prometheus/nginx-config.yaml#L59

However, with the new UI, it seems like this is no longer the option.

Therefore, the solution in this PR is to add a rewrite rule to nginx where, if it receives a path that matches `^/k8s/clusters/.*/proxy(.*)`, we strip it from the path.

Since we know that Prometheus and Grafana's APIs don't support something as specific as a `/k8s/clusters` path, we shouldn't have any conflicts with them.

Related Issue: https://github.com/rancher/rancher/issues/32884

----

NOTE: As a future-proofing measure and to keep the nginx configs in sync, I added the same rewrite rule to Grafana, even though there is no current issue with this today.


nickgerace (dismissed):
Good that it's moved to the chart itself, LGTM

jiaqiluo (approved):
LGTM

nickgerace (approved):

PR #1325, opened 2021-07-01
Update vSphere CSI to v2.2.1

PennyScissors:
Update vSphere CSI chart to upstream's latest v2.2.1 (This is just image updates addressing CVEs, see references to issues below)

image-mirror PR: https://github.com/rancher/image-mirror/pull/137


jiaqiluo (approved):
LGTM

PennyScissors:
@cbron Do I need to backport this update to 2.5?

aiyengar2 (approved):
LGTM

charts/rancher-vsphere-csi/rancher-vsphere-csi/100.0.0/values.yaml
```diff
@@ -25,7 +25,7 @@ csiController:
     enabled: false
```
cbron, line 25: Old asset should get deleted and index needs updating.

aiyengar2: @cbron the version isn't changing so the index.yaml doesn't need to be modified. The asset also got modified, which means that it did change

cbron (approved):
lgtm, but versioning may change based on discussion.

PR #1335, opened 2021-07-07
update fleet to v0.3.6-rc4

kinarashah:
https://github.com/rancher/fleet/issues/424


nickgerace (changes requested):
It appears that there's an index.yaml conflict

nickgerace (changes requested):
I think we need to run make charts again even with the gitjob dep change

kinarashah:
@nickgerace I did run `make charts` but got no changes. Comparing with https://github.com/rancher/charts/pull/1337/commits/049891f22b6d6dc5f9334b53c127cc7cfba32ad8, it's only fleet chart updates which were already included in the previous commit, so this looks expected?

nickgerace:
I think we are good then!

nickgerace (approved):

PR #1347, opened 2021-07-12
update system-upgrade-controller to v0.7.1

kinarashah:


superseb (approved):
I assume we are using the same chart version because its unreleased?

aiyengar2 (approved):
LGTM

kinarashah:
@superseb thanks for bringing this up, I wasn't sure either but @aiyengar2 confirmed that for charts like system-upgrade-controller that are directly checked into the repo - we can just update `/packages/${PACKAGE}/charts/values.yaml` and update index with `make charts`.

kinarashah:
https://github.com/rancher/rancher/issues/32904

PR #1357, opened 2021-07-13
Add external-ip-webhook package for 1.0.0

cmurphy:


jiaqiluo (approved):
LGTM

packages/rancher-external-ip-webhook/package.yaml
```diff
@@ -0,0 +1,2 @@
+url: https://github.com/rancher/externalip-webhook/releases/download/v1.0.0/rancher-externalip-webhook-1.0.0.tgz
+version: 1.0.0
```
cmurphy, line 2: Added new package.yaml. Did not include the Chart.yaml/values.yaml patches that were in 2.5.

index.yaml
```diff
@@ -1356,6 +1356,35 @@ entries:
...
+    - security
+    maintainers:
+    - email: raul@rancher.com
+      name: rawmind0
+    name: rancher-externalip-webhook
+    sources:
+    - https://github.com/rancher/externalip-webhook
+    urls:
+    - assets/rancher-external-ip-webhook/rancher-externalip-webhook-1.0.0.tgz
+    version: 1.0.0
```
cmurphy, line 1387: This is the version the scripts generate, it didn't add +up1.0.0

jiaqiluo (approved):

index.yaml
jiaqiluo: right, this is expected because the value of `version` is the same as the upstream version.

packages/rancher-external-ip-webhook/package.yaml
cbron: Why not ? We still need to switch to mirror versions and do name chaning: https://github.com/rancher/charts/blob/dev-v2.5/packages/rancher-external-ip-webhook/generated-changes/patch/Chart.yaml.patch
The reason for name changing is because this chart also exists in another spot for 2.4, and we have to support both chart names, so we named them differently.

cbron (commented):
Added @rawmind0 as reviewer as he made first one

packages/rancher-external-ip-webhook/package.yaml
cmurphy: Fixed

charts/rancher-external-ip-webhook/rancher-external-ip-webhook/1.0.0/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+  Deploy the external-ip-webhook to mitigate k8s CVE-2020-8554
+home: https://github.com/rancher/externalip-webhook
+keywords:
+- cve
+- externalip
+- webhook
+- security
+maintainers:
+- email: raul@rancher.com
+  name: rawmind0
```
rawmind0, line 20: (optional) May be we should update the maintainers??

index.yaml
```diff
@@ -1272,6 +1272,34 @@ entries:
...
+    digest: b8e46e3b23bbbcf9cb832f5908627f4a534bf5455f1c38a850c2e0d59b4eb17d
+    home: https://github.com/rancher/externalip-webhook
+    keywords:
+    - cve
+    - externalip
+    - webhook
+    - security
+    maintainers:
+    - email: raul@rancher.com
+      name: rawmind0
```
rawmind0, line 1296: (optional) May be we should update the maintainers??

rawmind0 (approved):
Beside the optional commented action, LGTM

cbron (approved):

PR #1361, opened 2021-07-15
Update aks-operator to rc16

cmurphy:
https://github.com/rancher/rancher/issues/33536
https://github.com/rancher/rancher/issues/33603


thedadams (changes requested):
The rc15 assets and charts should be removed.

thedadams (approved):

cbron (approved):

PR #1374, opened 2021-07-27
Istio 1.10 fixes

brendarearden:
**Issues**
https://github.com/rancher/rancher/issues/33742
https://github.com/rancher/rancher/issues/33737

**Problem**
https://github.com/kiali/helm-charts/pull/75 << in this upstream PR, while they were refactoring fullNameOverride to deployment.instance_name, the `app.kubernetes.io/instance` label was changed to `"kiali-server.fullname"`. While this change make sense from a `deployment.instance_name` should set the `instance` label, the release.name value was and can be VERY different from the fullNameOverride.

new permissions are required for installing jaeger

**solution**
Change the  `app.kubernetes.io/instance` back to `release.name` so that upgrade does not fail with immutable issues
Add telemetry.istio.io to the clusterrole apis with `*` permissions


packages/rancher-kiali-server/generated-changes/patch/templates/_helpers.tpl.patch
```diff
@@ -1,5 +1,14 @@
 --- charts-original/templates/_helpers.tpl
 +++ charts/templates/_helpers.tpl
+@@ -51,7 +51,7 @@
+ */}}
+ {{- define "kiali-server.selectorLabels" -}}
+ app.kubernetes.io/name: kiali
+-app.kubernetes.io/instance: {{ include "kiali-server.fullname" . }}
++app.kubernetes.io/instance: {{ .Release.Name }}
```
aiyengar2, line 8: Not sure if I agree with this change. We should be using the newer label on fresh installs; we should only preserve the older label if we can detect that the deployment already exists without the right label.

This can be done by checking if the [`.Release.IsUpgrade`](https://helm.sh/docs/chart_template_guide/builtin_objects/) and doing a [Helm `lookup`](https://helm.sh/docs/chart_template_guide/function_list/#lookup).

The result of the lookup should give you access to the fields of the k8s resource so you can pull the label that should be placed here from there. Alternatively, maybe if you just omit the label in this case it will preserve the existing one? Not sure though.

aiyengar2 (changes requested):

packages/rancher-kiali-server/generated-changes/patch/templates/_helpers.tpl.patch
brendarearden: Side note: https://helm.sh/docs/chart_best_practices/labels/ it does look like the best practice for app.kubernetes.io/instance label is to use the {{ .Release.Name}}

brendarearden: @aiyengar2 I updated this per your recommendation and I agree with your reasonings for why I shouldn't just revert back to using {{ .Release.Name}}. My new code now checks if its .Release.IsUpgrade and then I grab the labels from the existing deployment and get the value for the `/instance` label that has been impacted. If the value matches the release name, I use that, if it doesn't then I default to the new label value (which is kiali normally). and then on fresh install its using the new label value.

charts/rancher-istio/rancher-istio/100.0.0/charts/kiali/templates/_helpers.tpl
```diff
@@ -50,8 +50,20 @@ app.kubernetes.io/part-of: "kiali"
 Selector labels
 */}}
 {{- define "kiali-server.selectorLabels" -}}
+{{- $releaseName := .Release.Name -}}
+{{- $fullName := include "kiali-server.fullname" . -}}
 app.kubernetes.io/name: kiali
-app.kubernetes.io/instance: {{ include "kiali-server.fullname" . }}
+{{- if .Release.IsUpgrade }}
+    {{- $labels := (lookup "apps/v1" "Deployment" .Release.Namespace $fullName).metadata.labels }}
```
aiyengar2, line 57: Since these labels are contained in a helper function, can we assume that the deployment is the only resource we need to check to get a preserved `$instanceValue`? Or do we need to do multiple lookups?

aiyengar2, line 57: What happens if the lookup fails? 

e.g. the user deletes the Deployment after deploying and triggers an upgrade.

I believe the current code would cause an error on trying to render the template on an upgrade right?

I think we should modify this logic slightly to say something like:

```
$label := $fullName
lookup ...
if isUpgrade and resource exists:
  get label value
  if label value is not nil
    set $label to that value
  end
end
app.kubernetes.io/instance: {{ $label }}
```

aiyengar2: If Helm supports preseving omitted labels automatically, it would be nice if we can drop the stuff in the top-level if above entirely.

jiaqiluo (approved):
LGTM

aiyengar2 (changes requested):

charts/rancher-istio/rancher-istio/100.0.0/charts/kiali/templates/_helpers.tpl
aiyengar2: possibly should look at the service instead, based on offline discussion

aiyengar2 (approved):
LGTM

charts/rancher-kiali-server/rancher-kiali-server/100.0.0+up1.35.0/templates/_helpers.tpl
```diff
@@ -50,8 +50,15 @@ app.kubernetes.io/part-of: "kiali"
 Selector labels
 */}}
 {{- define "kiali-server.selectorLabels" -}}
+{{- $releaseName := .Release.Name -}}
+{{- $fullName := include "kiali-server.fullname" . -}}
+{{- $deployment := (lookup "apps/v1" "Deployment" .Release.Namespace $fullName) -}}
 app.kubernetes.io/name: kiali
-app.kubernetes.io/instance: {{ include "kiali-server.fullname" . }}
+{{- if (and .Release.IsUpgrade $deployment)}}
```
cbron, line 57: So if its an upgrade and the deployment with our fullName exists then just use the existing instance annotation ?

cbron (approved):
lgtm if other comments are addressed

charts/rancher-kiali-server/rancher-kiali-server/100.0.0+up1.35.0/templates/_helpers.tpl
brendarearden: Correct - we just carry that value over on upgrade to avoid conflicts. there is an issue getting the instance annotation value, we default to the $fullName since that is the new template value from upstream. 

Scenarios tested: 

1. **upgrade from istio 1.9.6**
    - deployed istio 1.9.6 - instance label value is rancher-istio
    - upgraded to istio 100.0.0 - instance label was carried over and is still rancher-istio
2. **upgrade from istio 1.9.6 with deleted deployment**
    - deployed istio 1.9.6 - instance label value is rancher-istio
    - deleted the kiali deployment
    - upgraded to istio 100.0.0 - instance label is now kiali
3. **upgrade from istio 100.0.0**
    - deployed istio 100.0.0 - instance label value is kiali
    - upgraded istio 100.0.0 - instance label was carried over and is still kiali
4. **upgrade from istio 100.0.0 with deleted deployment**
    - deployed istio 100.0.0 - instance label value is kiali
    - deleted the kiali deployment
    - upgraded istio 100.0.0 - instance label was carried over and is still kiali

PR #1376, opened 2021-07-28
Fixed WordPress Username label typo

simonflood:
Noticed during recent Rancher Rodeo - WordPress Username label had a typo (Usernname). Corrected for all versions.


recena:
Closing since **wordpress** is not part of `rancher/charts` anymore.

PR #1381, opened 2021-07-29
Backport istio 1.10.2 and Kiali 1.35.0

brendarearden:
**Issue**
https://github.com/rancher/rancher/issues/33824

**dev-v2.6 PR** 
https://github.com/rancher/charts/pull/1374

**Steps for Backport** 
1. Check new branch from  dev-v2.5 
2. Check out the kiali and istio package directors from dev-v2.6
3. Change the versioning to match the dev-v2.5 branch version schema
4. Make charts


charts/rancher-istio/rancher-istio/1.10.200/Chart.yaml
```diff
@@ -0,0 +1,21 @@
+annotations:
```
cbron, line 1: This needs updated: https://github.com/rancher/rancher/issues/33528
It looks like 1.8.6 is in that list already but not 1.9.x ? 

Failing CI too.

brendarearden: I will get the CI updated and update the 2.5.10

We included 1.9.x in 2.5.9 charts because it was OOB between 2.5.8 and 2.5.9 https://github.com/rancher/rancher/issues/33557

cbron: I'm guessing that included the "version scheme change" but 1.8 didn't, so now there is a new 1.8 but not 1.9. Only 1.10 now would be airgapped.

index.yaml
```diff
@@ -440,8 +440,8 @@ entries:
     - https://github.com/longhorn/longhorn-tests
     - https://github.com/longhorn/backing-image-manager
     urls:
-    - assets/longhorn/longhorn-1.1.100.tgz
-    version: 1.1.100
+    - assets/longhorn/longhorn-1.1.100+up1.1.1.tgz
+    version: 1.1.100+up1.1.1
```
jiaqiluo, line 444: Interesting, it flips the order of 2 longhron charts.

index.yaml
```diff
@@ -665,35 +665,35 @@ entries:
...
     description: Installs the CRDs for longhorn.
-    digest: ee37cf4c632dd56bc409f104c49916f61e9350652ac997138af729d9e7560a38
+    digest: eec1b37ef0f12930cac4b8dc812e0c08af32ec5af8afb37114f3932e333bb5b6
     name: longhorn-crd
     type: application
     urls:
-    - assets/longhorn/longhorn-crd-1.1.100+up1.1.1.tgz
-    version: 1.1.100+up1.1.1
+    - assets/longhorn/longhorn-crd-1.1.100.tgz
+    version: 1.1.100
```
jiaqiluo, line 689: Interesting, it flips the order of 2 longhorn charts.

index.yaml
```diff
@@ -751,16 +751,18 @@ entries:
...
+      name: prom2teams
       repository: file://./charts/prom2teams
     - condition: sachet.enabled
       name: sachet
       repository: file://./charts/sachet
     description: The manager for third-party webhook receivers used in Prometheus
       Alertmanager
-    digest: 5a382220965c57d384b9a66cf09606ec76219242d5b48562c270e5c9f9a5e327
+    digest: f76948784dce142ca88002a9a6a3af1d22c5f91079d1330cb14ebb87c2cbea93
+    icon: https://charts.rancher.io/assets/logos/alerting-drivers.svg
```
jiaqiluo, line 765: Your changes should be nothing about the prom2teams chart. 
I guess maybe your index.yaml is from a wrong branch?

jiaqiluo (changes requested):

index.yaml
brendarearden: https://github.com/rancher/charts/blob/941bef117c51d9d838678a3f97f9411428134d0e/charts/rancher-alerting-drivers/rancher-alerting-drivers/1.0.200/Chart.yaml#L16 so it looks like the icon is infact in the chart, so the index.yaml must've had old data, so when I re-did it, it corrected

charts/rancher-istio/rancher-istio/1.10.200/Chart.yaml
brendarearden: Yes, the OOB release for 1.9 did include the version changes, except for we did not bump the chart annotation for auto install. To make sure we are all up to date and all the new version schemes are on the same page, I did go ahead and bump the chart annotation in 1.9.6 from `catalog.cattle.io/auto-install: rancher-kiali-server-crd=1.32.100` to `catalog.cattle.io/auto-install: rancher-kiali-server-crd=1.32.100+up1.32.0`

packages/rancher-kiali-server-1.35/package.yaml
```diff
@@ -0,0 +1,8 @@
+url: https://kiali.org/helm-charts/kiali-server-1.35.0.tgz
```
cbron, line 1: Why'd we make this package (the folder) version specific instead of reusing the standard one ? I'm assuming its related to my previous question, you have to introduce two kiali versions. But even if you do, is 1 kiali version not compatible across 1.9 and 1.10 istio ?

brendarearden: We are not making chart changes to 1.8. or 1.9 in this release, except for moving them to the new chart scripts versioning schema. Also, with the refresh issues in kiali 1.35, it would be preferable to leave istio 1.8 and 1.9 on the current kiali version so that we aren't introducing that issue to more istio versions than we have to.

charts/rancher-istio-1.9/rancher-istio/1.9.601/Chart.yaml
```diff
@@ -0,0 +1,21 @@
+annotations:
```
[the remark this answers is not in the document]
brendarearden: The commits have been squashed

packages/rancher-istio-1.9/charts/Chart.yaml
```diff
@@ -13,7 +13,7 @@ annotations:
   catalog.cattle.io/release-name: rancher-istio
   catalog.cattle.io/ui-component: istio
   catalog.cattle.io/provides-gvr: networking.istio.io.virtualservice/v1beta1
-  catalog.cattle.io/auto-install: rancher-kiali-server-crd=1.32.100
+  catalog.cattle.io/auto-install: rancher-kiali-server-crd=1.32.100+up1.32.0
```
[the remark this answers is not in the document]
brendarearden:
```
So why are we adjusting them to the new schema versioning and bumping the package, when they are still using the same version ?
```
This question could be asked for any of the charts that we are releasing with just the version schema change. This gives us a fresh starting point where all the charts in the branch are updated to the same new schema. 

When I did the OOB release for 1.9.6, I rolled back Jack's schema version change for the kiali-server-crd version from `1.32.100+up1.32.0` to `1.32.100` so it makes sense to add that change back in. https://github.com/rancher/charts/commit/223629aeefb6982af83dc3566c355cb1b349001f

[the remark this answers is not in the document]
cbron: Bringing this up in separate issue. We can leave as-is here.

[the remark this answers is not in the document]
cbron: Can we just not bump the package version and not introduce the version with this new format ? Then that would avoid the new kiali chart version below.

brendarearden:
Updates: 
I added DoNotRelease to 1.8 and 1.9 Istio so that we are not releasing any new charts for istio due to schema version changes only. We only have one kiali package again. 

There is still something weird going on with the index.yaml generation when running make charts. After I deleted the assets that we no longer need, one of them remained in the index.yaml even though it wasn’t in the assets dir anymore so I fully regenerated the index.yaml again just to make sure. I confirmed that I didn't have the code open in another code editor that could be caching the data. 🤷  somethings definitely going on.

packages/rancher-istio-1.9/package.yaml
```diff
@@ -1,2 +1,3 @@
 url: local
-packageVersion: 0
+packageVersion: 1
```
cbron, line 2: This should still be 0

packages/rancher-istio-1.9/charts/Chart.yaml
cbron, line 16: This should be undone

packages/rancher-kiali-server/templates/crd-template/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v2
-version: 1.32.1
+version: 1.35.1
```
cbron, line 2: Is this 1.35.1 instead of 1.35.0 because of the whole 1==100 situation ? Meaning if upstream is 0 we use 0-99.

cbron (commented):
Asset changes look right to me now, just a few things.

packages/rancher-kiali-server/templates/crd-template/Chart.yaml
brendarearden: Correct, for the leading zero problem. We did this for the previous version of Kiali as well.

packages/rancher-istio-1.8/package.yaml
```diff
@@ -1,2 +1,3 @@
 url: local
 packageVersion: 1
+doNotRelease: true
```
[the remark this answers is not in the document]
brendarearden: Correct - the changes for 1.8.601 and 1.32.100+up1.32.0 were both slated for 1.5.10 and are both version schema changes only.

packages/rancher-istio-1.9/package.yaml
brendarearden: reverted back to 0

packages/rancher-istio-1.9/charts/Chart.yaml
brendarearden: removed +up

index.yaml
```diff
@@ -412,9 +412,9 @@ entries:
       catalog.cattle.io/ui-component: longhorn
     apiVersion: v1
     appVersion: v1.1.1
-    created: "2021-06-24T21:41:35.652793-07:00"
+    created: "2021-08-04T15:23:53.922038-07:00"
     description: Longhorn is a distributed block storage system for Kubernetes.
-    digest: bf329bb6f2429e47489851d7355e474c72b590a0123b90fb3ea6eaed5d6d449b
+    digest: e256933b6743544cf832f496c44d391043fe37e6461ac1693cf3c999e0e73378
```
cbron, line 417: Some weird digest changes in there that are unrelated to PR

index.yaml
```diff
@@ -2600,7 +2602,7 @@ entries:
     description: Collects several related Helm charts, Grafana dashboards, and Prometheus
       rules combined with documentation and scripts to provide easy to operate end-to-end
       Kubernetes cluster monitoring with Prometheus using the Prometheus Operator.
-    digest: 41b2bff949e5b318196f9ca028b1cf97e8739d535dc0758ba2be6da89960d947
+    digest: 88c5fbe910af3fe218bbc14e92b357fe24918ba60ab6b77759a6dcc8d3303551
```
cbron, line 2605: same here, not sure whats going on, might try rebasing and re-generating

cbron (commented):
lgtm besides other comments and odd index.yaml issues

index.yaml
cbron: This may be related to comments below, something odd going on with index generation

index.yaml
cbron: Ohh, it just flipped them, didn't change them.

index.yaml
brendarearden: Steps followed: 
1. `git fetch upstream`
2. `git pull upstream dev-v2.5 --rebase`
3. noted that `Current branch backport-istio is up to date.`
4. `make charts`
5. `git status`
6. noted that `nothing to commit, working tree clean` 
7. `helm repo index --url assets assets`
8. `mv assets/index.yaml .`
9. `chmod +x index.yaml`
10. `git diff` and noted that the only changes that occured to the index.yaml were the `created` timestamps 

Generally I don't run `make charts` without setting the `export PACKAGE=<whatever chart package I am working on>` unless there was an issue with the checks passing for the PR, which did happen with this PR a few times. When I ran `make charts` it is possible that the digest changed on some of the charts, even if the content didn't change.

jiaqiluo (dismissed):
LGTM

index.yaml
aiyengar2: https://github.com/rancher/charts-build-scripts/issues/54 should address this since we should be producing a sorted index.yaml so that it remains consistent on each make charts call

index.yaml
aiyengar2: Ensuring that the digest did not change is one of the reasons why I still think we should prioritize https://github.com/rancher/charts-build-scripts/issues/56 to ensure that the inner contents of TGZ files have not been modified. 

Right now, we make an assumption that, the assets/ directory did not change [if nothing changed for that chart in charts/](https://github.com/rancher/charts-build-scripts/blob/deb66c8dd574ddd1f3b58271fec97b7b141588d0/main.go#L193).

cc: @cbron

aiyengar2 (commented):
Couple of questions:
1. Why are we deleting rancher-istio-1.8.601 from dev-v2.5?
  - Are we no longer releasing that patch to 1.8.x in Rancher 2.5?
  - If so, I think we should either a) call that out in the PR title or b) move that into another PR. IMO, the latter option makes more sense since backporting charts from dev-v2.6 and removing charts from dev-v2.5 are independent actions
2. Why does the index.html have a mode change? I think that needs to be removed from this PR.
3. If we are removing rancher-istio-1.8 assets, should we also remove rancher-istio-1.8's packages/ entry?

index.yaml
cbron: Confirmed that in dev-v2.5 the icon is not on 1.0.100 or 1.0.200. It only got added to 1.0.200 here but thats likely because thats where package currently is.

index.yaml
cbron: The short-term fix here to avoid changes in future would be to delete the charts that need to change from the index.yaml instead of regenerating entirely.

cbron:
1. On 1.8.601, that was never released. Afaik it was only generated to handle the new charts build scripts. We don't want to release any more 1.8 versions if we can avoid it: https://github.com/rancher/charts/commits/dev-v2.6/assets/rancher-istio-1.8/rancher-istio-1.8.601.tgz
We could pull that into separate PR, but it wasn't introduced in any issue besides the scripts changes, so removing it isn't affecting anything functionally. @brendarearden can correct me if we had added it at some point. 

3. That would be the alternative to the "DoNotRelease" option. Fine with me.

brendarearden:
@aiyengar2 

2) good call out - removed the mode change. 
3) I left the package in because I was not sure if we were going to add the schema change in across all charts again. We wont be release any new changes to the chart chart of istio-1.8 because its past support date. The reason the 1.8 chart changes and 1.9 chart changes were included in this PR is due to the fact that we don't want to have two packages for kiali to manage so we either need to just remove 1.8 all together (include assets,charts,package) as part of this PR OR only add the DoNotRelease tag and leave the package behind but clean up the assets/charts resources.

aiyengar2 (dismissed):
LGTM

index.yaml
```diff
@@ -1750,15 +1750,15 @@ entries:
...
-    digest: ef85e326a6dcc4b1d7f867f8e9cf43fe3ec82c9b08ca9458004fc60d06e37e3e
+    digest: 4774d1f117a1ce47230e2004dc23609f54ed49c6b45100b2438bf920822f7ca1
     icon: https://charts.rancher.io/assets/logos/istio.svg
     keywords:
     - networking
     - infrastructure
     name: rancher-istio
     urls:
-    - assets/rancher-istio-1.8/rancher-istio-1.8.601.tgz
-    version: 1.8.601
```
cbron, line 1761: This is another very frustrating diff to review but afai can tell this is the only thing actually getting removed. All the 1.9.x's stay as is and we are adding 1.10.2. Then for kiali we are swapping 1.32 for 1.35

cbron (approved):
lgtm

aiyengar2 (approved):
LGTM

jiaqiluo (approved):
LGTM

PR #1388, opened 2021-07-30
Cluster tools annotation

brendarearden:
**Issue**
https://github.com/rancher/rancher/issues/33841

**Problem**
new Cluster Tools page needs a way to display specific charts. 

**Solution**
Add a new chart annotation that can be used to determine what charts should be displayed: `catalog.cattle.io/type=cluster-tool`

**Note**
I will be doing Rancher Backups in its own PR since it has an upstream PR thats its dependent on


aiyengar2 (approved):
LGTM

jiaqiluo (approved):
LGTM

cbron (approved):
lgtm, though its odd monitoring actually lost bytes when this is only adding things

brendarearden:
@aiyengar2 caleb has a good point - does that normally happen when building the monitoring asset?

aiyengar2:
@cbron @brendarearden I can’t remember if I’ve seen the asset reduce in size like this, but given the fact that it is a compressed asset, I don’t think adding or removing bytes means anything. 

It’s possible adding this line allowed the overall file to be compressed in a slightly more efficient way, but I’m not sure. Either way, if CI is passing I think we should be good to go.

PR #1390, opened 2021-08-02
Add rancher-version to rancher-operator

cmurphy:
rancher-operator was merged into Rancher[1] and so the image and chart
won't be used for Rancher 2.6 going forward. This change adds a patch to
pin the rancher version to <2.6 so it won't be used in new installs and
won't be picked up by image-scanning. 

Partial forward port of https://github.com/rancher/charts/pull/1391


cmurphy:
Not sure if I should add back the package.yaml for this chart?

jiaqiluo:
@cmurphy The charts in the dev-v2.6 branch should be identical to and forward-ported from the ones in the dev-v2.5 branch. So could you do it this way:
- update the chart in the dev-v2.5 branch ==> this will be another PR to dev-v2.5
- forward-port the new chart to dev-2.6

cmurphy:
Thanks @jiaqiluo, I made a PR to dev-v2.5: https://github.com/rancher/charts/pull/1391

jiaqiluo (approved):
LGTM

cbron:
If we are forwardporting, why do we need this PR ?

cmurphy:
@cbron this PR is now the forwardport itself

cbron:
@jiaqiluo / @aiyengar2 can we do forwardports from dev to dev ? I thought they'd have to go into release-v2.5 and then across to release-v2.6 and finally down to dev-v2.6.

jiaqiluo:
@cbron Technically the porting can happen between any two branches. 
In this PR we forward port the chart from `dev-v2.5` to `dev-v2.6` because we want to validate that the chart will not show in rancher v2.6 setup before the chart is released in rancher v2.5 
cc @aiyengar2

cbron:
Won't work as image script runs today

PR #1391, opened 2021-08-02
Add rancher-version to rancher-operator

cmurphy:
rancher-operator was merged into Rancher[1] and so the image and chart
won't be used for Rancher 2.6 going forward. This change adds a patch to
pin the rancher version to <2.6 so it won't be used in new installs and
won't be picked up by image-scanning. The change is made directly to the
chart because the package was removed in 143c0fec.


jiaqiluo (approved):
LGTM.

@cmurphy  Can you update the version for rancher-operator in https://github.com/rancher/rancher/issues/33528

@cbron Should we remove the version `0.1.400+up0.1.4` from the branch? Its content is identical to the released version `0.1.400` except the version as it was generated when we introduced the new version schema to the dev-v2.5 branch.

cbron (approved):
lgtm, not 100% sure this will actually fix the problem.

cbron:
Oh and yes lets do what @jiaqiluo said, delete the unreleased 0.1.400 version from dev-v2.5. We should only ever release one version.

cbron (changes requested):
Oh real quick lets do what @jiaqiluo said, delete the unreleased 0.1.400 version from dev-v2.5. We should only ever release one version.

cmurphy:
Removed the 0.1.400+up0.1.4 charts and assets

index.yaml
```diff
@@ -3413,6 +3414,24 @@ entries:
...
+      catalog.cattle.io/release-name: rancher-operator
+    apiVersion: v2
+    appVersion: 0.1.4
+    created: "2021-06-24T21:41:35.866839-07:00"
+    description: Control Rancher using GitOps
+    digest: 881ecadac8830da4ed862934a98ae047c957e4ccfb57ba0189a8dd759e81db68
+    name: rancher-operator
+    urls:
+    - assets/rancher-operator/rancher-operator-0.1.400+up0.1.4.tgz
+    version: 0.1.400+up0.1.4
```
cbron, line 3434: The two `0.1.400+up0.1.4`'s are still in the index

cmurphy: This was what was generated by `make charts`. Manually removed it.

cbron:
Won't work as image script runs today

PR #1392, opened 2021-08-03
Bump aks-operator to rc18

cmurphy:
https://github.com/rancher/rancher/issues/33097


index.yaml
```diff
@@ -1773,38 +1773,6 @@ entries:
...
-      for details.
-    digest: daab9ccb018dc9f0b9f13153307dc7d3c8dc961e44e8b85ca443fb17287a7343
-    icon: https://charts.rancher.io/assets/logos/istio.svg
-    keywords:
-    - networking
-    - infrastructure
-    name: rancher-istio
-    urls:
-    - assets/rancher-istio-1.9/rancher-istio-1.9.601.tgz
-    version: 1.9.601
```
cbron, line 1807: Uhh whats going on here, cc @brendarearden

cmurphy: Not related to aks, splitting it into its own commit

brendarearden: Not entirely sure what happened here. Chart version 1.9.601 is a chart in the 2.5 branch so it makes sense that this is being removed as part of this PR and I approve of this change becuase that asset does not exist in 2.6. 

I was working on changes to the 2.5 branch for a backport, so its possible that the chart asset was cached in one code editor, while I was working on the cluster-tool changes in the 2.6 branch in a different code editor which showed up in the generated index.yaml. I know we have had some other weirdness with the index.yaml so maybe that explains those situations as well..

cbron: 1.9.601 doesn't exist in any of the other branches (dev-v2.5, release-v2.5/6) so removing seems right.

cbron (approved):
lgtm, we'll address istio situation in 2.5 backport PR, the CRD swap to v1 I believe we already talked about and confirmed is ok.

index.yaml
brendarearden: https://github.com/rancher/charts/pull/1381/files << it exists in my backport to 2.5 which is why I believe it was a caching issue. But I agree - the index.yaml for this aks operator bump looks accurate and good to merge!

thedadams (approved):

PR #1393, opened 2021-08-03
Delete rancher-operator

cmurphy:
rancher-operator was merged into rancher[1] and won't be installed
separately going forward on 2.6. This deletes the charts so they won't
be erroneously caught by image-scanning.


cbron (dismissed):
Lgtm, but we need to do in release-v2.6 also

cmurphy:
rebased

index.yaml
```diff
@@ -3520,171 +3520,6 @@ entries:
     urls:
     - assets/rancher-node-exporter/rancher-node-exporter-1.16.201.tgz
     version: 1.16.201
-  rancher-operator:
```
jiaqiluo, line 3523: As we know we are deleting this whole section, can you just delete it manually instead of running `make charts` to avoid all those timestamp changes?

jiaqiluo (changes requested):

index.yaml
cmurphy: good idea, done

jiaqiluo (approved):
LGTM

cbron (approved):

PR #1394, opened 2021-08-04
Delete rancher-operator

cmurphy:
rancher-operator was merged into rancher[1] and won't be installed
separately going forward on 2.6. This deletes the charts so they won't
be erroneously caught by image-scanning.


cbron (approved):

cbron:
Will merge once Arvind approves

aiyengar2 (approved):
LGTM. We should also clean these up in dev-v2.6 @cmurphy

PR #1400, opened 2021-08-06
update rancher/shell to v0.1.9 and add comments

jiaqiluo:
issue: https://github.com/rancher/rancher/issues/33995
 
**Cause**:
The image `rancher/shell:v0.1.8` is missing because it is only used by the `monitoring-crd` chart, and we have [a logic in the package script to skip the crd charts](https://github.com/rancher/rancher/blob/8df979bdd5e0d4be5fb41b288fcedfc5e376696b/scripts/package#L56-L67) so it will not be picked.
We did not run into this bug before simply because the image&tag is used by other charts so it is included in the rancher-images.txt

**Solution**:
Update the tag to v0.1.9 which is used by Rancher charts which is already included in the rancher-images.txt
also, add comment as a reminder for keeping this tag inline with rancher chart


kinarashah (approved):

charts/rancher-monitoring/rancher-monitoring-crd/100.0.0+up16.6.0/values.yaml
```diff
@@ -7,5 +7,6 @@ global:
     systemDefaultRegistry: ""
 
 image:
+  # keep this inline with whatever Rancher's chart uses to avoid missing tag in rancher-images.txt
```
cbron, line 10:
```suggestion
  # keep this inline with both the primary monitoring chart and the rancher shell setting to avoid missing tag in rancher-images.txt
```

packages/rancher-monitoring/templates/crd-template/values.yaml
```diff
@@ -7,5 +7,6 @@ global:
     systemDefaultRegistry: ""
```
cbron, line 7: I don't think this crd-template/values.yaml would get picked up by the images.txt script. I meant to add it here, this will get picked up: https://github.com/rancher/charts/blob/dev-v2.6/charts/rancher-monitoring/rancher-monitoring/100.0.0%2Bup16.6.0/values.yaml

Add comment something like:
> This is used in the CRD chart, keep this inline with that, and the rancher shell setting to avoid missing tag in rancher-images.txt

cbron (changes requested):

aiyengar2 (changes requested):
Why are we making the change in rancher/charts instead of making the change in rancher/rancher to pull images from CRD charts too, if they have a values.yaml?

If we proceed with this change, we will also need to do the same for Gatekeeper: https://github.com/rancher/charts/blob/941bef117c51d9d838678a3f97f9411428134d0e/charts/rancher-gatekeeper/rancher-gatekeeper-crd/3.3.1%2Bup3.3.0/values.yaml#L10-L11

jiaqiluo:
Close this PR as we will fix the bug in rancher/rancher.

PR #1410, opened 2021-08-11
The defaults are too low

Tejeev:
We are still seeing high severity tickets with Monitoring crashing due to WAL compaction exceeding the default limits in place.  I noticed the limits in Monitoring by default are lower than any of the recommended.

I've updated the default to reflect that suggested in the table here: https://rancher.com/docs/rancher/v2.5/en/monitoring-alerting/#setting-resource-limits-and-requests
I've also created a PR to add best practice of tuning defaults to the docs here: https://github.com/rancher/docs/pull/3436

It's odd to me that, in the docs, there is a "known issue" to increase the Prometheus memory limits to those noted in the table above...

### DIFF
--- a/charts/rancher-monitoring/values.yaml
+++ b/charts/rancher-monitoring/values.yaml
@@ -2280,10 +2280,10 @@ prometheus:
     ##
     resources:
       limits:
-        memory: 1500Mi
+        memory: 2500Mi
         cpu: 1000m
       requests:
-        memory: 750Mi
+        memory: 1750Mi
         cpu: 750m
 
     ## Prometheus StorageSpec for persistent data


aiyengar2:
@Tejeev I believe this has already been bumped in the latest released monitoring https://github.com/rancher/charts/blob/8503a72b749021e6994fb1cce2542407face5168/charts/rancher-monitoring/rancher-monitoring/100.1.0%2Bup19.0.3/values.yaml#L2943-L2949

The requests have remained the same though.

If a fix is no longer necessary, can you close out this PR? If not, the correct branch to make this fix in should be `dev-v2.6`.

nicholasSUSE:
Hello @Tejeev 

Can I close this old PR, is this still going to be merged?

Tejeev:
@nicholasSUSE Sorry I'd not seen Arvind's reply.  Yes I think we can close this as defaults were bumped up

PR #1417, opened 2021-08-17
Fix/issue 33328

paynejacob:
https://github.com/rancher/rancher/issues/33046
https://github.com/rancher/rancher/issues/33328

The default journald path in ubuntu is `/var/log/journal`.  With 2.6 we are going to support provisioning RKE2 and K3s on ubuntu.  This change will make logging work out of the box for these cluster types.


packages/rancher-logging/generated-changes/patch/values.yaml.patch
```diff
@@ -43,7 +43,7 @@
 +
 +disablePvc: true
 +
-+systemdLogPath: "/run/log/journal"
++systemdLogPath: "/var/log/journal"
```
snasovich, line 46: Is it OK to just change this outright?
https://sematext.com/blog/journald-logging-tutorial/#toc-journald-storage-4

paynejacob: we are not configuring systemd we are reading from it.  We have a fluent-bit pod that can read journald logs.  This pod needs to know where journald is writing the logs to be able to mount it and read it.

paynejacob: You can see our usage here https://github.com/rancher/charts/blob/dev-v2.6/charts/rancher-logging/rancher-logging/100.0.0%2Bup3.12.0/templates/loggings/k3s/configmap.yaml#L21

snasovich: Right, so previously we used `/run/log/journal` and now we changing it in all cases, not just on ubuntu to `/var/log/journal` - so I'm just confirming it's OK.
Also realized I should have better commented on the .patch file, but we can continue here.

nickgerace (approved):
This looks good to me. While this may breaks some existing users, journal detection would be unreliable because the location can change, it is distro dependent, and depends on distro settings. I believe a release note with documentation will suffice.

snasovich:
> This looks good to me. While this may breaks some existing users, journal detection would be unreliable because the location can change, it is distro dependent, and depends on distro settings. I believe a release note with documentation will suffice.

@nickgerace , "breaks some existing users" doesn't give me warm & fuzzy feeling. Are you sure we're OK?

packages/rancher-logging/generated-changes/patch/values.yaml.patch
paynejacob: Yes this will change the default in all cases. If approved I will add a release note about this.   `/run/log/journal` is the default on selinux installations.  Previously this value was only used for rke2 which typically was installed on selinux distros.  Now that k3s uses it and we are supporting provisioning with rke2 and k3s my thinking is the sensible default is no longer sensible.

paynejacob: This is also going out with a minor version bump so this is the time to break it if we are going to.  As it is right now if you provision a k3s cluster in 2.6 on ubuntu it will not work out of the box.

snasovich: Understood. Added `release-note` label to the issue to track.
@cbron , please review the convo above.

nickgerace:
@snasovich There is a bit of nuance here. k3s has switched how it logs, and we have switched our default to mimic that. The location can be different as a result.

One might think that we could just key off of the version, but the issue is that the location can be changed. This is distro dependent and settings dependent. As a result, changing the default location for provisioning as well as new RKE2 and k3s clusters appears to be the path of least resistance.

Moreover, the previous defaults were only shipped in a patch release or two.

In an ideal world, I would love for us to be able to dynamically tell the journaling location, however I think this might be best left to a docs change with instructions for users.

Our defaults should likely match provisioning defaults as a result.

packages/rancher-logging/generated-changes/patch/values.yaml.patch
deniseschannon: Is this a value of the chart that could be changed if needed depending on your cluster and where it's located? Therefore, yes, we could change the default, BUT it would be configurable for the cases where the default is different? 

I agree with @snasovich that changing it outright doesn't sound ideal esp for users who upgrade charts and their logging was working but now it's not and they have no idea why.

paynejacob: Ubuntu uses `/var/log/journal` while RHEL uses `/run/log/journal`  This is a challenging value to default because if you create the `/var/log/journal` directory on a system that is configured to use `/run/log/journal` journald will start to use the `/var/log/journal` on the next boot.  My thinking with changing the default value here is more uses use ubuntu so it would work out of the box for more folks.

paynejacob: You can always override the default by settings it in values.  It should be note the behavior above is the default behavior.  it is possible to configure any of these os to use the other format.

deniseschannon: Is this option available in the UI?

paynejacob: no

paynejacob:
After discussing with @deniseschannon we are going to leave the default as is and give the users some guidance in the ui and docs.  I will close this pr and open another one with documentation instead.

PR #1418, opened 2021-08-18
bump rancher-webhook to 0.2.0-beta05

jiaqiluo:
issues
 https://github.com/rancher/rancher/issues/33988
https://github.com/rancher/rancher/issues/34188


StrongMonkey (dismissed):

aiyengar2 (approved):
LGTM

charts/rancher-webhook/rancher-webhook/1.0.0+up0.2.0-beta04/values.yaml
```diff
@@ -12,4 +12,10 @@ capi:
...
 mcm:
-  enabled: true
\ No newline at end of file
+  enabled: true
+
+preDelete:
+  enabled: true
+  image:
+    repository: rancher/kubectl
+    tag: v0.2.0-beta04
```
cbron, line 21: This isn't correct, looks like it grabbed the webhook tag, which is odd since it looks correct in the chart: https://github.com/rancher/webhook/blob/master/charts/rancher-webhook/values.yaml#L21

https://hub.docker.com/r/rancher/kubectl/tags?page=1&ordering=last_updated

cbron (changes requested):
looks right except one thing

charts/rancher-webhook/rancher-webhook/1.0.0+up0.2.0-beta04/values.yaml
aiyengar2: @cbron if you download the tgz file and look at the file contents, it does show `v0.2.0-beta04`

Screen Shot 2021-08-18 at 10 51 54 AM

aiyengar2: This is probably because we replace all tags with the HELM_TAG here

https://github.com/rancher/webhook/blob/efe9b4b5743347554cc6c332d1b615de4c321b66/scripts/package-helm#L20-L22

Where the HELM_TAG is set to have the RC based on:

https://github.com/rancher/webhook/blob/245c6bc9362b0204c3bcad7c923ade36087ea055/scripts/version#L22

jiaqiluo: Thank you for the investigation. I will fix it in rancher/webhook

StrongMonkey (approved):

cbron (approved):

PR #1422, opened 2021-08-19
Move suc from local to source from system-charts/k3s-upgrader

kinarashah:


kinarashah:
https://github.com/rancher/rancher/issues/34071

packages/system-upgrade-controller/generated-changes/patch/templates/configmap.yaml.patch
```diff
@@ -0,0 +1,17 @@
+--- charts-original/templates/configmap.yaml
++++ charts/templates/configmap.yaml
+@@ -2,13 +2,12 @@
+ kind: ConfigMap
+ metadata:
+   name: system-upgrade-controller-config
+-  namespace: cattle-system
```
aiyengar2, line 7: If we drop the namespace, won't this be created in the default namespace? If you want this in the namespace specified for the helm chart, this should be `{{ .Release.Namespace }}`

packages/system-upgrade-controller/generated-changes/patch/templates/configmap.yaml.patch
```diff
@@ -0,0 +1,17 @@
...
+ data:
+   SYSTEM_UPGRADE_CONTROLLER_DEBUG: {{ .Values.systemUpgradeControllerDebug | default "false" | quote }}
+   SYSTEM_UPGRADE_CONTROLLER_THREADS: {{ .Values.systemUpgradeControllerThreads | default "2" | quote }}
+   SYSTEM_UPGRADE_JOB_ACTIVE_DEADLINE_SECONDS: {{ .Values.systemUpgradeJobActiveDeadlineSeconds | default "900" | quote }}
+   SYSTEM_UPGRADE_JOB_BACKOFF_LIMIT: {{ .Values.systemUpgradeJobBackoffLimit | default "99" | quote }}
+-  SYSTEM_UPGRADE_JOB_IMAGE_PULL_POLICY: {{ .Values.systemUpgradeJobImagePullPolicy | default "IfNotPresent" | quote }}
++  SYSTEM_UPGRADE_JOB_IMAGE_PULL_POLICY: {{ .Values.systemUpgradeJobImagePullPolicy | default "Always" | quote }}
```
aiyengar2, line 14: Should not be patched right?

packages/system-upgrade-controller/generated-changes/patch/templates/deployment.yaml.patch
```diff
@@ -0,0 +1,40 @@
+--- charts-original/templates/deployment.yaml
++++ charts/templates/deployment.yaml
+@@ -2,7 +2,6 @@
+ kind: Deployment
+ metadata:
+   name: system-upgrade-controller
+-  namespace: cattle-system
```
aiyengar2, line 7: Same comment as above. Maybe we need `{{ .Release.Namespace }}` here too

packages/system-upgrade-controller/generated-changes/patch/templates/deployment.yaml.patch
```diff
@@ -0,0 +1,40 @@
...
+-                    - "true"
+-            weight: 100
+-      tolerations:
+-        operator: "Exists"
+-      serviceAccountName: system-upgrade
++          requiredDuringSchedulingIgnoredDuringExecution:
++            nodeSelectorTerms:
++              - matchExpressions:
++                  - {key: "node-role.kubernetes.io/master", operator: In, values: ["true"]}
++      serviceAccountName: system-upgrade-controller
```
aiyengar2, line 37: should be dropped right?

packages/system-upgrade-controller/generated-changes/patch/templates/serviceaccount.yaml.patch
```diff
@@ -0,0 +1,7 @@
+--- charts-original/templates/serviceaccount.yaml
++++ charts/templates/serviceaccount.yaml
+@@ -2,4 +2,3 @@
+ kind: ServiceAccount
+ metadata:
+   name: system-upgrade-controller
+-  namespace: cattle-system
```
aiyengar2, line 7: Same comment about release namespace as above

aiyengar2 (changes requested):

charts/system-upgrade-controller/system-upgrade-controller/100.0.0+up0.3.0/Chart.yaml
```diff
@@ -1,14 +1,14 @@
-apiVersion: v1
-name: system-upgrade-controller
-description: General purpose controller to make system level updates to nodes
-home: https://github.com/rancher/system-charts/charts/system-upgrade-controller
-sources:
-  - "https://github.com/rancher/system-charts/charts/system-upgrade-controller"
-version: 0.7.3
-appVersion: v0.7.3
 annotations:
   catalog.cattle.io/certified: rancher
```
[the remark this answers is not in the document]
kinarashah: removed

kinarashah:
@aiyengar2 Thanks, all the patches were unnecessary, removed them. The only one we want is `packages/system-upgrade-controller/generated-changes/patch/Chart.yaml.patch`

packages/system-upgrade-controller/generated-changes/patch/Chart.yaml.patch
```diff
@@ -0,0 +1,23 @@
...
+-  Users do not need to manually upgrade this app. It will be automatically upgraded to the latest version when upgrading a cluster.
+-home: https://github.com/rancher/system-charts/blob/dev-v2.6/charts/rancher-k3s-upgrader
++name: system-upgrade-controller
++description: General purpose controller to make system level updates to nodes
++home: https://github.com/rancher/system-charts/charts/system-upgrade-controller
+ sources:
+-  - "https://github.com/rancher/system-charts/blob/dev-v2.6/charts/rancher-k3s-upgrader"
+-version: 0.3.0
++  - "https://github.com/rancher/system-charts/charts/system-upgrade-controller"
++version: 0.7.4
```
aiyengar2, line 16: This patch will not make a difference since it is overridden by the `version` in the package.yaml (see produced Chart.yaml under `charts/` directory above). We might as well remove it.

aiyengar2 (changes requested):
last nit, otherwise LGTM

index.yaml
```diff
@@ -4187,15 +4187,15 @@ entries:
...
     name: system-upgrade-controller
     sources:
     - https://github.com/rancher/system-charts/charts/system-upgrade-controller
     urls:
-    - assets/system-upgrade-controller/system-upgrade-controller-100.0.0.tgz
-    version: 100.0.0
-generated: "2021-08-06T16:57:33.967930468-07:00"
+    - assets/system-upgrade-controller/system-upgrade-controller-100.0.0+up0.3.0.tgz
+    version: 100.0.0+up0.3.0
+generated: "0001-01-01T00:00:00Z"
```
aiyengar2, line 4201: nit: this seems really weird. Why did this timestamp get generated?

I don't think it makes a difference though w.r.t. charts, so I will not block this PR on it.

aiyengar2 (dismissed):
LGTM

index.yaml
kinarashah: I deleted the full blob of system-upgrade-controller in index.yaml and ran `make charts` again. Should I keep it as is?

aiyengar2: yeah keep it as-is

aiyengar2 (approved):

Issue #1352, opened 2021-07-13
Please add new Helm chart for Longhorn 1.1.2


rgutierrez2004:
Longhorn 1.1.2 was released 6 days ago, please add new Helm chart for Longhorn 1.1., so it can be installed from Rancher UI

dhawton:
PR already submitted https://github.com/rancher/charts/pull/1336

beathorst:
I'm confuced. Which is the correct Branch for the Chart Repo in Rancher Apps & Marketplace?

https://git.rancher.io/charts
- master
- dev-v2.5
- release-v2.5

I cannot find the longhorn version 1.1.2

konstantin-921:
I use the default Cluster Explorer repositories. In all my clusters, the repository https://git.rancher.io/charts is configured by default to use the release-v2.5 branch.

At the moment I would like to update the Longhorn version. I see that version 1.1.2 is available in the master branch but not in the release-v2.5 branch. I am waiting for release 1.1.2 to be added to the release-v2.5 branch because I cannot stop using the release-v2.5 branch and go to the master (I will then be asked to reinstall 
Longhorn instead of updating which will most likely cause an error).

Can we expect to add Longhorn 1.1.2 to the release-v2.5 branch?

beathorst:
> I use the default Cluster Explorer => Apps & Marketplace repositories. In all my clusters, the repository https://git.rancher.io/charts is configured by default to use the release-v2.5 branch.

Thank you

innobead:
> I use the default Cluster Explorer => Apps & Marketplace repositories. In all my clusters, the repository https://git.rancher.io/charts is configured by default to use the release-v2.5 branch.
> 
> At the moment I would like to update the Longhorn version. I see that version 1.1.2 is available in the master branch but not in the release-v2.5 branch. I am waiting for version 1.1.2 to be added to the release-v2.5 branch because I cannot stop using the release-v2.5 branch and go to the master (I checked this out - I was asked to reinstall Longhorn instead of updating, which will most likely cause an error).
> 
> Can we expect to add Longhorn 1.1.2 to the release-v2.5 branch?

cc @cbron

cbron:
[...]

PR #1425, opened 2021-08-23
Add longhorn 1.2.0 chart for cluster manager

innobead:

### DIFF
--- a/charts/longhorn/latest/Chart.yaml
+++ b/charts/longhorn/latest/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v1
 name: longhorn
-version: 1.1.2
-appVersion: v1.1.2
-kubeVersion: ">=v1.16.0-r0"
+version: 1.2.0
+appVersion: v1.2.0
+kubeVersion: ">=1.18.0-0"
 description: Longhorn is a distributed block storage system for Kubernetes.
 keywords:
 - longhorn
@@ -18,7 +18,6 @@ sources:
 - https://github.com/longhorn/longhorn-engine
 - https://github.com/longhorn/longhorn-instance-manager
 - https://github.com/longhorn/longhorn-share-manager
-- https://github.com/longhorn/backing-image-manager
 - https://github.com/longhorn/longhorn-manager
 - https://github.com/longhorn/longhorn-ui
 - https://github.com/longhorn/longhorn-tests
--- a/charts/longhorn/latest/README.md
+++ b/charts/longhorn/latest/README.md
@@ -18,15 +18,45 @@ Longhorn is 100% open source software. Project source code is spread across a nu
 ## Prerequisites
 
 1. A container runtime compatible with Kubernetes (Docker v1.13+, containerd v1.3.7+, etc.)
-2. Kubernetes v1.16+
+2. Kubernetes v1.18+
 3. Make sure `bash`, `curl`, `findmnt`, `grep`, `awk` and `blkid` has been installed in all nodes of the Kubernetes cluster.
 4. Make sure `open-iscsi` has been installed, and the `iscsid` daemon is running on all nodes of the Kubernetes cluster. For GKE, recommended Ubuntu as guest OS image since it contains `open-iscsi` already.
 
+## Installation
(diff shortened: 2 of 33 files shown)


khushboo-rancher:
Validated with the branch `add-longhorn-1.2.0-chart-master`

Validation - **Pass**

aiyengar2:
@innobead we should close out this PR since it's to the wrong branch

innobead:
> @innobead we should close out this PR since it's to the wrong branch

Which branch is the correct one?

charts/longhorn/latest/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 name: longhorn
-version: 1.1.2
-appVersion: v1.1.2
-kubeVersion: ">=v1.16.0-r0"
+version: 1.2.0
+appVersion: v1.2.0
+kubeVersion: ">=v1.18.0"
```
shuo-wu, line 5: May need to change this to `"kubeVersion: >=1.18.0-r0"`

See https://github.com/longhorn/longhorn/issues/2963

aiyengar2:
> @innobead we should close out this PR since it's to the wrong branch

Sorry for the confusion; just checked in with @cbron about this and this is the right branch.

innobead:

aiyengar2 (commented):
One clarification: why does longhorn place the latest chart under `latest/` instead of under `1.2.0`? Seems like longhorn is the only chart that does this.

Otherwise, these changes LGTM

cbron:
@innobead are we going to wait to merge until its equivalent in cluster-explorer goes out with rancher release ? To avoid situation we hit last time about users asking when its going live.

jiaqiluo (approved):

shuo-wu (approved):

innobead:
> @innobead are we going to wait to merge until its equivalent in cluster-explorer goes out with rancher release ? To avoid situation we hit last time about users asking when its going live.

We don't need to wait to merge, because we don't know how long it will take and at least let users using cluster manager able to install/upgrade longhorn earlier. 

As you mentioned, I believe people will ask a similar question why 1.2.0 is still unavailable in cluster explorer, but you already have a good explanation for 1.1.2 about the difference between cluster manager and cluster explore in https://github.com/rancher/charts/issues/1352#issuecomment-887999551, so we can probably share this with users if they have the question.

innobead:
@khushboo-rancher we did a few updates for this PR, so please help verify again to ensure the PR working as expected. Thank you!

meldafrawi:
I've validated the chart again, **LGTM**.

jiaqiluo (approved):
LGTM

khushboo-rancher:
https://github.com/longhorn/longhorn/issues/2964 is fixed with this PR

aiyengar2 (approved):
Already gave an LGTM but re-approving since it got dismissed.

@innobead @shuo-wu @cbron I think we still need to clarify the following:

> One clarification: why does longhorn place the latest chart under `latest/` instead of under `1.2.0/`? Seems like longhorn is the only chart that does this.

yasker:
@aiyengar2 It was required to do so when we started publishing the chart a few years back. I guess somewhere along the way Rancher has changed the chart layout? I found at least it was still like this in late 2019. https://github.com/rancher/charts/tree/325754b5123b6e3358504c72a453dad2aea39d82 

It's much better if we can skip the `latest` folder.

aiyengar2:
Thanks @yasker! Yes I don't think it's a blocker. 

w.r.t. the fix, I'm also worried that a fix will introduce an error into Rancher due to some code dependency that may have necessitated it in the first place (which is why I asked about the context). But since this branch is practically deprecated in Rancher 2.6.0 (since Cluster Manager is no longer supported), I think it's fine to leave it as-is.

yasker:
Thanks @aiyengar2 .

I noticed the CI failed. Is it a blocker?

yasker:
@cbron Can you take a final look? Thanks!

cbron (approved):
Branches lgtm, not reviewing content since its already approved by longhorn team.

PR #1426, opened 2021-08-24
Add harvester-cloud-provider and harvester csi

yaocw2020:
- Add harvester-cloud-provider 0.1.1
- Add harvester-csi-driver 0.1.0


gitlawr (approved):
LGTM

packages/harvester-cloud-provider/package.yaml
```diff
@@ -0,0 +1,3 @@
+url: https://github.com/harvester/charts.git
+subdirectory: charts/harvester-cloud-provider
+commit: c7746e5bd688d41ca7b984e545166f887e9eefc2
```
aiyengar2, line 3: Need to add version:
https://github.com/rancher/charts/blob/33db4185684ece7b2ca1575cdc66af4841967764/packages/longhorn/package.yaml#L4

packages/harvester-csi-driver/package.yaml
```diff
@@ -0,0 +1,3 @@
+url: https://github.com/harvester/charts.git
+subdirectory: charts/harvester-csi-driver
+commit: 40656f4e5024041a668b1c847ccb69250c3fd017
```
aiyengar2, line 3: Need to add version: https://github.com/rancher/charts/blob/33db4185684ece7b2ca1575cdc66af4841967764/packages/longhorn/package.yaml#L4

aiyengar2 (changes requested):
Need to add version. Should be 100.0.0 for both of these since this is the first release in 2.6.x cc: @cbron 

Also, @yaocw2020 can you split this into commits in a slightly different way?

Commit 1: Add the package.yaml for harvester-cloud-provider only
Commit 2: Add the package.yaml for harvester-csi-driver only
Commit 3: Run make charts, producing the charts, assets, and index.yaml changes for **both** charts in one commit

packages/harvester-cloud-provider/package.yaml
aiyengar2: No need to change the Chart.yaml; scripts will take care of that automatically. Just need to add this flag to package.yaml

packages/harvester-csi-driver/package.yaml
aiyengar2: No need to change the Chart.yaml; scripts will take care of that automatically. Just need to add this flag to package.yaml

yaocw2020: I find that the generated chart version is 100.0.0+up0.1.1, rather than 0.1.1. If I want to add it in the rancher kontainer-driver-metadata, which version is appropriate, 100.0.0+up0.1.1 or 0.1.1?

guangbochen:
as discussed with Craig in the https://github.com/rancher/rke2/pull/1679#discussion_r699811231, we will move those charts to the rke2-charts repo, thanks.

PR #1343, opened 2021-07-09
Longhorn: Update dev-v2.6 for v1.1.2 release

shuo-wu:
longhorn/longhorn#2724

PR #1428, opened 2021-08-24
Add longhorn 1.2.0 chart dev 2.6

innobead:

cc @khushboo-rancher @meldafrawi @shuo-wu


shuo-wu (changes requested):
1. `package/longhorn/templates/crd-template/Chart.yaml` is not updated.
2. Missing update for `index.yaml`. But I am not sure if we need to remove 100.0.0+up1.1.2 first. I will leave Rancher team to answer it. https://github.com/rancher/charts/pull/1343#issuecomment-878635783

jiaqiluo:
@innobead Do you want longhorn 1.2.0-rc2 to be available in Rancher for testing? Otherwise, the PR can be updated when the longhorn 1.2.0 is release.

It looks like longhorn-100.0.0+up1.1.2 is not in the release/v2.6 branch, so we need to delete it from this branch. 

What you need to do is:
- remove the longhorn-100.0.0+up1.1.2 from assets/ , charts/, and its entry in the index.yaml
- update the packages/longhorn to bring in longhorn 1.2.0
- run `PACKAGE=longhorn make charts` to generate the new charts

Unless we decided to keep the chart version the same as the upstream, it should be `100.0.0+up1.2.0`

innobead:
> @innobead Do you want longhorn 1.2.0-rc2 to be available in Rancher for testing? Otherwise, the PR can be updated when the longhorn 1.2.0 is release.

This PR will be updated when 1.2.0 released.
> 
> ~It looks like longhorn-100.0.0+up1.1.2 is not in the release/v2.6 branch, so we need to delete it from this branch.~ It looks like longhorn-100.0.0+up1.1.2 will be released with rancher 2.6.0

Yes, we need 1.1.2 and 1.2.0 all in 2.6
> 
> What you need to do is:
> 
> * ~remove the longhorn-100.0.0+up1.1.2 from assets/ , charts/, and its entry in the index.yaml~
> * update the packages/longhorn to bring in longhorn 1.2.0
> * run `PACKAGE=longhorn make charts` to generate the new charts
> 
> Unless we decided to keep the chart version the same as the upstream, it should be `100.0.0+up1.2.0`

If we don’t update the version (current: 100.0.0) in longhorn/package.yaml, as I mentioned, the index.yaml will not be updated to include 1.2.0. (it will be ignored to merge the index file because 100.0.0 already exists then keep 100.0.0+up1.2.0 in the index only)

Any suggestion?

jiaqiluo:
how about this 
longhorn  1.1.2  ==> 100.0.0+up1.1.2
longhorn  1.2.0  ==> 101.0.0+up1.2.0

innobead:
> how about this
> longhorn 1.1.2 ==> 100.0.0+up1.1.2
> longhorn 1.2.0 ==> 101.0.0+up1.2.0

How to decide to increase major/minor/patch? If we have 2.0.0, what version it should look? And the version pattern also ensure the upgrade path work like 1.1.* able to upgrade to 1.2.*.

jiaqiluo:
Oh, I did not realize we have longhorn 2.0.0, it can start at a larger major number like 120.0.0

innobead:
> Oh, I did not realize we have longhorn 2.0.0, it can start at a larger major number like 120.0.0

What about 1.1.3? 101.0.0+up1.1.3? If yes, does this version pattern allows below upgrade? (Or exact question is does the version impacting the upgrade path?)

1.1.2 able to upgrade to 1.1.3
1.1.2/1.1.3 able to upgrade to 1.2.0

jiaqiluo:
upgrading should not be a problem as long as we keep the 100 versions align to the upstream,
for example,
longhorn 1.1.2 ==> 100.0.0+up1.1.2
longhorn 1.1.3 ==> 100.3.0+up1.1.3
longhorn 1.2.0 ==> 101.0.0+up1.2.0
longhorn 2.0.0 ==> 120.0.0+up2.0.0

helm checks the major.minor.patch version to determine the order of version. 
the +up-version part is metadata and is not considered when comparing versions

innobead:
> upgrading should not be a problem as long as we keep the 100 versions align to the upstream,
> for example,
> longhorn 1.1.2 ==> 100.0.0+up1.1.2
> longhorn 1.1.3 ==> 100.3.0+up1.1.3
> longhorn 1.2.0 ==> 101.0.0+up1.2.0
> longhorn 2.0.0 ==> 120.0.0+up2.0.0
> 
> helm checks the major.minor.patch version to determine the order of version.
> the +up-version part is metadata and is not considered when comparing versions

Perfect then all good.

innobead:
ping @jiaqiluo review again based on your suggestion.

shuo-wu (commented):
In general LGTM. But there are lots of differences caused by the meaningless indenting. Should we remove those indent changes?

innobead:
> In general LGTM. But there are lots of differences caused by the meaningless indenting. Should we remove those indent changes?

Which part do you mean? All generated and copied instead of manual operation, so probably we don't need to tackle anything here.

shuo-wu:
> > In general LGTM. But there are lots of differences caused by the meaningless indenting. Should we remove those indent changes?
> 
> Which part do you mean? All generated and copied instead of manual operation, so probably we don't need to tackle anything here.

The modification in `packages/longhorn/generated-changes/overlay/crds/crds.yaml`

innobead:
> > > In general LGTM. But there are lots of differences caused by the meaningless indenting. Should we remove those indent changes?
> > 
> > 
> > Which part do you mean? All generated and copied instead of manual operation, so probably we don't need to tackle anything here.
> 
> The modification in `packages/longhorn/generated-changes/overlay/crds/crds.yaml`

Fixed :+1:

khushboo-rancher:
Validated with this branch

Validation - **Pass**

shuo-wu (approved):

innobead:

packages/longhorn/package.yaml
```diff
@@ -1,7 +1,7 @@
 url: https://github.com/longhorn/charts.git
 subdirectory: charts/longhorn
-commit: 37f3263c5f7d3361550521c9e36968b5f3d9b7b4
-version: 100.0.0
+commit: f629c0352211bb04eb0e7a97472c779078e971c3
+version: 101.0.0
```
aiyengar2, line 4:
```suggestion
version: 100.0.1
```

We only bump major / minor version during Rancher minor releases

index.yaml
```diff
@@ -609,6 +651,20 @@ entries:
...
+      catalog.cattle.io/release-name: longhorn-crd
+    apiVersion: v1
+    created: "2021-09-02T23:33:08.754460728+08:00"
+    description: Installs the CRDs for longhorn.
+    digest: 0da94608691f075178a186a75b2b12b7fe4208ea32fc44dcce7c101fc1fd01d8
+    name: longhorn-crd
+    type: application
+    urls:
+    - assets/longhorn/longhorn-crd-101.0.0+up1.2.0.tgz
+    version: 101.0.0+up1.2.0
```
aiyengar2, line 667: needs to be regenerated after version is changed

index.yaml
```diff
@@ -363,6 +363,48 @@ entries:
...
+    - https://github.com/longhorn/longhorn-engine
+    - https://github.com/longhorn/longhorn-instance-manager
+    - https://github.com/longhorn/longhorn-share-manager
+    - https://github.com/longhorn/longhorn-manager
+    - https://github.com/longhorn/longhorn-ui
+    - https://github.com/longhorn/longhorn-tests
+    - https://github.com/longhorn/backing-image-manager
+    urls:
+    - assets/longhorn/longhorn-101.0.0+up1.2.0.tgz
+    version: 101.0.0+up1.2.0
```
aiyengar2, line 407: needs to be regenerated after version is changed

aiyengar2 (changes requested):
version is incorrect

aiyengar2:
W.r.t. the questions on versioning scheme above, the rules for versioning outlined by Denise have been detailed in https://confluence.suse.com/display/EN/Feature+Charts+Versioning+and+Branching (see `Versioning Rules`).

We only bump the major version of a chart during a Rancher release.

We only bump the minor version of a chart for two reasons:
1. when a new minor release is cut (e.g. when 2.7 is released, the first chart backported from 2.7 to 2.6 will have its minor version bumped from 100.0.x -> 100.1.0)
2. when there is a large chart change to the upstream chart, e.g. if we may need to have multiple release lines
  - For example, let's say longhorn releases 2.0.0 and we still need to maintain the ability to patch and release 1.x.x since there is some functionality loss on upgrade that some customers may want. Then you can bump the minor version
  - For example, let's say longhorn releases 2.0.0 and it deprecates support for k8s 1.18 and below clusters while Rancher still supports that. Then you can bump the minor version

Otherwise, always bump the patch version only.

aiyengar2:
FYI, here is a known and accepted issue with this versioning scheme for Rancher:

A user who has 100.1.x+up2.x.x installed can upgrade Rancher from 2.6 -> 2.7 and then upgrade the Helm chart from the newer chart back to an older chart (since 100.1.x+up2.x.x < 101.0.0+up1.1.x, even though 2.x.x > 1.1.x). 

However, it is the expectation that the user would not perform this upgrade in favor of one that takes them to the appropriate 2.x.x version in 2.7, since there will never be a case when 2.x.x is available in Rancher 2.6 but not in 2.7.

innobead:
> W.r.t. the questions on versioning scheme above, the rules for versioning outlined by Denise have been detailed in https://confluence.suse.com/display/EN/Feature+Charts+Versioning+and+Branching (see `Versioning Rules`).
> 
> We only bump the major version of a chart during a Rancher release.
> 
> We only bump the minor version of a chart for two reasons:
> 
> 1. when a new minor release is cut (e.g. when 2.7 is released, the first chart backported from 2.7 to 2.6 will have its minor version bumped from 100.0.x -> 100.1.0)
> 2. when there is a large chart change to the upstream chart, e.g. if we may need to have multiple release lines
> 
> * For example, let's say longhorn releases 2.0.0 and we still need to maintain the ability to patch and release 1.x.x since there is some functionality loss on upgrade that some customers may want. Then you can bump the minor version
> * For example, let's say longhorn releases 2.0.0 and it deprecates support for k8s 1.18 and below clusters while Rancher still supports that. Then you can bump the minor version
> 
> Otherwise, always bump the patch version only.
> 
> In this case, I don't think we need to support an older release line, so a patch version bump seems fine. @innobead correct me if I'm wrong, in which case we may need the version to be 100.1.0.

For longhorn, we need patch releases for each minor release, so for example, we should keep supporting 1.1.* and start supporting 1.2.* after this PR merged. As per your suggestion, we should use `100.1.0` instead, right?

index.yaml
innobead: As I mentioned in https://github.com/rancher/charts/pull/1428#issuecomment-912224971, we should use `100.1.0` instead.

shuo-wu (approved):

shuo-wu:
> W.r.t. the questions on versioning scheme above, the rules for versioning outlined by Denise have been detailed in https://confluence.suse.com/display/EN/Feature+Charts+Versioning+and+Branching (see `Versioning Rules`).
> 
> We only bump the major version of a chart during a Rancher release.
> 
> We only bump the minor version of a chart for two reasons:
> 
> 1. when a new minor release is cut (e.g. when 2.7 is released, the first chart backported from 2.7 to 2.6 will have its minor version bumped from 100.0.x -> 100.1.0)
> 2. when there is a large chart change to the upstream chart, e.g. if we may need to have multiple release lines
> 
> * For example, let's say longhorn releases 2.0.0 and we still need to maintain the ability to patch and release 1.x.x since there is some functionality loss on upgrade that some customers may want. Then you can bump the minor version
> * For example, let's say longhorn releases 2.0.0 and it deprecates support for k8s 1.18 and below clusters while Rancher still supports that. Then you can bump the minor version
> 
> Otherwise, always bump the patch version only.
> 
> In this case, I don't think we need to support an older release line, so a patch version bump seems fine. @innobead correct me if I'm wrong, in which case we may need the version to be 100.1.0.

Then when can we bump up the patch version `100.0.x`? It seems that `100.0.x` controls if the Chart is compatible with the current Rancher version whereas the application version determines `+up1.x.x` only.

meldafrawi:
Validated on Rancher v2.6: **LGTM**

jiaqiluo:
Based on the conversation, I think those are the versions we will use:
```
longhorn 1.1.2 ==> 100.0.0+up1.1.2
longhorn 1.1.3 ==> 100.0.1+up1.1.3
longhorn 1.2.0 ==> 100.1.0+up1.2.0 
longhorn 1.2.1 ==> 100.1.1+up1.2.1 
longhorn 2.0.0 ==> 101.0.0+up2.0.0 
```

jiaqiluo (changes requested):

khushboo-rancher:
With this PR https://github.com/longhorn/longhorn/issues/2964 is fixed

packages/longhorn/templates/crd-template/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-version: 1.1.2
+version: 1.2.0
```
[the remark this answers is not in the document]
innobead: The main chart version will be overridden to 1.1.2 if we don't have this change, so we still need to update the longhorn-crd chart version to 1.2.0.

I would suggest keeping this change. WDYT? @jiaqiluo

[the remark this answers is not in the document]
innobead: You are right :+1:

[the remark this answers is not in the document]
shuo-wu: Then do we still need to set this field? Retaining this field to `1.1.2` for v1.2.0 Chart looks weird and may lead to confusion.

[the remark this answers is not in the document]
innobead: True, but it seems something needed to fix in https://github.com/rancher/charts-build-scripts because currently, it will check the version field by using helm lib during the `make prepare`.

This should be possible to enhance in the future. Leave to Rancher team.

aiyengar2:
> Based on the conversation, I think those are the versions we will use:
> 
> ```
> longhorn 1.1.2 ==> 100.0.0+up1.1.2
> longhorn 1.1.3 ==> 100.0.1+up1.1.3
> longhorn 1.2.0 ==> 100.1.0+up1.2.0 
> longhorn 1.2.1 ==> 100.1.1+up1.2.1 
> longhorn 2.0.0 ==> 101.0.0+up2.0.0 
> ```

I agree with all of this except the last line.

```
longhorn 2.0.0 ==> 100.2.0+up2.0.0 
```

If we anticipate that we will release and maintain a `1.3.x`, `1.4.x`... `1.9.x` **within Rancher 2.6**, then we should do:

```
longhorn 2.0.0 ==> 100.10.0+up2.0.0 
```

But the major version bump should be reserved for Rancher minor version only.

cc: @cbron

aiyengar2:
> Then when can we bump up the patch version 100.0.x? It seems that 100.0.x controls if the Chart is compatible with the current Rancher version whereas the application version determines +up1.x.x only.

As @jiaqiluo mentioned, the patch version will get bumped when you make a patch release. 

So 1.1.2 -> 1.1.3 must require a dev to bump the Rancher patch version.

aiyengar2 (changes requested):
Since we are planning to actively maintain 2 release lines for longhorn (1.1.x and 1.2.x) within Rancher 2.6, it might make more sense to duplicate the `packages/longhorn` to track them as separate packages.

This way we don't need to flip flop between versions on trying to patch the 1.1.x line when 1.2.x is checked in.

More specifically, I would recommend the following:
- Add an initial commit that copies `packages/longhorn -> packages/longhorn-1.2`. Title this commit "Base copy longhorn 1.2"
- Add a second commit that moves `packages/longhorn -> packages/longhorn-1.1`. Title this commit "Rename longhorn to longhorn-1.1"
- Refactor your current `packages/longhorn` changes to be on `packages/longhorn-1.2`. Title this commit "Update longhorn 1.2.0 from 1.1.2"
- Delete any `charts/`, `assets`, or `index.yaml` entries for longhorn charts that don't exist on `release-v2.5`. Title this commit "Remove old entries"
- Run `make charts` and title the commit "make charts"

The end result of this is that the contents of `charts/longhorn` should be identical to the contents of https://github.com/rancher/charts/tree/release-v2.5/charts/longhorn (no new charts introduced) and instead a `charts/longhorn-1.1` and `charts/longhorn-1.2` directory should be created with the two charts you want to release in this release.

innobead:
> > Based on the conversation, I think those are the versions we will use:
> > ```
> > longhorn 1.1.2 ==> 100.0.0+up1.1.2
> > longhorn 1.1.3 ==> 100.0.1+up1.1.3
> > longhorn 1.2.0 ==> 100.1.0+up1.2.0 
> > longhorn 1.2.1 ==> 100.1.1+up1.2.1 
> > longhorn 2.0.0 ==> 101.0.0+up2.0.0 
> > ```
> 
> I agree with all of this except the last line.
> 
> ```
> longhorn 2.0.0 ==> 100.2.0+up2.0.0 
> ```
> 
> If we anticipate that we will release and maintain a `1.3.x`, `1.4.x`... `1.9.x` **within Rancher 2.6**, then we should do:
> 
> ```
> longhorn 2.0.0 ==> 100.10.0+up2.0.0 
> ```
> 
> But the major version bump should be reserved for Rancher minor version only.
> 
> cc: @cbron

This is kind of like a general question and guidance people will ask. I would suggest having some guidance/rules to describe how to name versions between upstream and downstream chart in the future, maybe README is a good place.

packages/longhorn/generated-changes/patch/values.yaml.patch
```diff
@@ -1,52 +0,0 @@
---- charts-original/values.yaml
-+++ charts/values.yaml
-@@ -8,38 +8,38 @@
- image:
-   longhorn:
-     engine:
--      repository: longhornio/longhorn-engine
-+      repository: rancher/mirrored-longhornio-longhorn-engine
```
aiyengar2, line 8: Is this change in the wrong commit?

packages/longhorn/generated-changes/patch/questions.yml.patch
```diff
@@ -1,101 +0,0 @@
---- charts-original/questions.yml
-+++ charts/questions.yml
```
aiyengar2: Is this change in the wrong commit?

aiyengar2 (changes requested):
nit: first commit seems to contain changes that do not belong there.

Otherwise, LGTM

aiyengar2 (approved):
LGTM

#### Note

The index.yaml will still point longhorn `100.0.0+up1.1.2` at `assets/longhorn/longhorn-100.0.0+up1.1.2.tgz` instead of `assets/longhorn-1.1/longhorn-100.0.0+up1.1.2.tgz`.

https://github.com/rancher/charts/blob/9f7e7f8da5ecd9ff60a6b90e5a2a518cb00345d7/index.yaml#L449-L450

This is expected since that version is already released; however, the first time that longhorn pushes a change into the 1.1.x line, we should delete that orphaned chart.

For now, we need to leave it in since CI will fail without it.

khushboo-rancher:
Revalidated the chart with this branch - **LGTM**

Screen Shot 2021-09-09 at 4 14 23 PM

index.yaml
```diff
@@ -363,6 +363,48 @@ entries:
     - assets/fleet-crd/fleet-crd-0.3.000.tgz
```
cbron, line 363: > The index.yaml will still point longhorn 100.0.0+up1.1.2 at assets/longhorn/longhorn-100.0.0+up1.1.2.tgz instead of assets/longhorn-1.1/longhorn-100.0.0+up1.1.2.tgz.
> This is expected since that version is already released; however, the first time that longhorn pushes a change into the 1.1.x line, we should delete that orphaned chart.

We need to add a rancher-max of `2.6.0` to 100.0.0+up1.1.2, @aiyengar2 does that require a re-release of the chart or are we just going to modify it on the release branch ?

packages/longhorn-1.2/templates/crd-template/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+version: 1.2.0
```
cbron, line 2: Just an fyi that only this version will be airgapped, not 1.1.x.

jiaqiluo (dismissed):

index.yaml
aiyengar2: We have to modify it in the release branch; PR workflows won't pass in dev-v2.6 if we're trying to modify an already released chart.

cbron: Ok, that can be handled as part of this: https://github.com/rancher/rancher/issues/34479#issuecomment-918619403

@innobead here we just need to add `rancher-version` annotation on both of these tars being introduced:
* 1.1.2 should be min of 2.6.0 and max of 2.6.1 (the max because of the upgrade-to-older-version problem)
* 1.2.0 should be min of 2.6.1 and no max

See linked issue above for more info.

innobead: @cbron could u help explain what the upgrade-to-older-version problem here? Because it looks u wants to have max version for 1.1.2 as 2.6.1, does that means users can’t install 1.1.2 after 2.6.1? If that’s the case, then why?

packages/longhorn-1.2/templates/crd-template/Chart.yaml
innobead: I thought all feature charts in cluster explorer (2.6 or 2.5) all air-gapped. 

Could u explain more? What case will be air-gapped and what case will not?

cbron: Only the latest version of any chart is air-gapped. There is an issue to add support where it goes off of the rancher-version instead, so you could have multiple if they both specify this upcoming version.

index.yaml
cbron: Because of our 100 versioning scheme. If we introduce 1.2.0 in 2.5 then a user could "upgrade" from 1.2.0 to 100.0.0+up1.1.2 and actually perform a downgrade from 1.2.0->1.1.2. If we limit what is available with min/max and the users go to the latest rancher 2.6.1, then the "older" version won't be available and they would stay on 1.2.0. We also try to help people install the latest version and not older ones.

innobead: Got your point for 100 versioning.

But as I mentioned above, right now Longhorn supports multiple releases for different minor releases. (Currently it’s 1.1.x and 1.2.x)

So we should allow users staying in 1.1.x and being able to choose to upgrade to the next patch version (1.1.2) instead of only upgrading to the latest minor release (1.2.0).

Also in the future, we will have 1.2.1 and 1.1.3 in following releases. 

So the question is whether users are able to install or upgrade to 1.1.2 in 2.6.1 if we set max versions for 1.1.2 as u suggested?

packages/longhorn-1.2/templates/crd-template/Chart.yaml
innobead: Got it. But how do we handle multiple releases case not just one latest version? Like I mentioned above if we have ongoing releases for more than one release line like 1.1.x and 1.2.x?

cbron: We can't, we only airgap one line for now. So you can release both of them in 2.6.1, but only 1.2 will get air-gapped.

index.yaml
cbron: They would not be able to upgrade to 1.1.2 if we set the max at less than the rancher version. For other charts thats how we handle it, we just move to the latest. If you want to keep both lines open to install/upgrade, then the tradeoff is that they can accidentally downgrade. I suppose longhorn can be an exception here too, perhaps you should warn about it in longhorn docs or release notes.

innobead: Thanks, so what we really need to add based on our discussion is:

1.1.2/1.2.0 should be min of 2.6.0 and no max, correct?

Also, for longhorn releases starting from 2.6.0, I think we only need to add min of 2.6.0, because the upgrade issue will not happen for charts in 2.6, because the upgrade will rely on the versions we discussed here (https://github.com/rancher/charts/pull/1428#issuecomment-914651849) w/ corresponding minor/patch updates of downstream charts for the upstream charts.

packages/longhorn-1.2/templates/crd-template/Chart.yaml
innobead: Thanks for the explanation. One more question, do we have any documentation about this? I believe some people will have similar questions in the future.

cbron: Not that I'm aware of. There is an issue to allow multiple per release so it will change soon anyway.

index.yaml
cbron: > 1.1.2/1.2.0 should be min of 2.6.0 and no max, correct?

If you are ok with the potential downfall of the false-upgrade here then yes. But we shouldn't do it long term. 

>  for longhorn releases starting from 2.6.0, I think we only need to add min of 2.6.0 w/o max, because the upgrade issue will not happen for charts in 2.6 and the chart upgrade will rely on the versions we discussed

Doing this permanently is going to bring up problems, besides the false-upgrade situation. If there are rancher requirements in place that longhorn needs a specific version of rancher to be compatible with, for example which versions of kubernetes rancher does and does not support. Also once we do have the fix for airgap in place we'll have to be more specific about what we allow in airgap, so the size of the total-airgap images doesn't get too large.

innobead: Agree, no perfect solution if we want support multiple maintenance line releases and also keep releasing in 2.5 and 2.6.

But the proposal is:
1. Keep the latest releases of each release line
2. Disable previous releases of each release line (rerelease or update the released one directly?)

This can avoid false upgrade for each line (ex: 1.1.x at rancher 2.5 to 1.1.<latest> at rancher 2.6 or 1.2.x at rancher 2.5 to 1.2.<latest> at rancher 2.6) but cross line false upgrade can't (ex: 1.2.x at 2.5 to 1.1.x at 2.6). This part needs to tackle by documentation).

WDYT? Any better ideas are welcome.

packages/longhorn-1.2/templates/crd-template/Chart.yaml
innobead: Ok this part is clear now, resolved

index.yaml
cbron: That sounds like a good compromise for now. We should note that in here: https://github.com/rancher/rancher/issues/34589

2.6.1: 
1.1.2-min of 2.6.0
1.2.0-min of 2.6.0 

2.6.2: 
1.1.2-min of 2.6.0, max of 2.6.1
1.2.0-min of 2.6.0, max of 2.6.1
1.1.3-min of 2.6.2
1.2.1-min of 2.6.2

innobead: The example is what I thought now but if we don’t have any release in 2.6.2 but 2.6.3 instead, the example will be like this: (however the same concept)

2.6.3:
1.1.2-min of 2.6.0, max of 2.6.2
1.2.0-min of 2.6.0, max of 2.6.2
1.1.3-min of 2.6.3
1.2.1-min of 2.6.3

If no other concerns, lets do this way first but also need to share this info with PM to let them know this change introduced by 2.6 introduction, different version scheming between 2.5 n 2.6, multiple releases support and continuous releases to 2.5 n 2.6.

I will update the PR. Thanks @cbron

cbron: Yes that is correct, we won't add the previous version's max until new versions are going out. We don't have great mechanism to handle that (updating the older versions) right now but should soon.

innobead:
@cbron updated, please review. But I didn't update 1.1.2 because it was released (ref: https://github.com/rancher/charts/pull/1428#pullrequestreview-750768023). 

Any suggestion or should it be tackled by https://github.com/rancher/charts/pull/1468?

cbron (dismissed):

cbron:
Yes the already released one will be handled separately by us. I'll let your team merge and handle QA as necessary.

khushboo-rancher:
On Rancher `v2.6.1-rc1`, the chart version v1.2.0 is not appearing
Screen Shot 2021-09-16 at 5 21 24 PM

innobead:
> Yes the already released one will be handled separately by us. I'll let your team merge and handle QA as necessary.

Sounds good :+1:

innobead:
> On Rancher `v2.6.1-rc1`, the chart version v1.2.0 is not appearing
> Screen Shot 2021-09-16 at 5 21 24 PM

It should be related to `-rc1` meta info. @cbron ?

cbron:
Unlikely, if it was 2.6.0-rc1 that would make sense, but any 2.6.1-rc/alpha is greater than 2.6.0. 
Is that pulling in this branch specifically ? If that doesn't work you could update it to `rancher-version: '>= 2.6.0-alpha1'`
Example tests that run showing versions: https://github.com/rancher/rancher/blob/cb7de4e6c3d7e783828dc662b1142c1f9ae5edbe/pkg/catalogv2/content/content_test.go#L63-L68

innobead:
> Unlikely, if it was 2.6.0-rc1 that would make sense, but any 2.6.1-rc/alpha is greater than 2.6.0.
> Is that pulling in this branch specifically ? If that doesn't work you could update it to `rancher-version: '>= 2.6.0-alpha1'`
> Example tests that run showing versions: https://github.com/rancher/rancher/blob/cb7de4e6c3d7e783828dc662b1142c1f9ae5edbe/pkg/catalogv2/content/content_test.go#L63-L68

Thanks for the example. Just took a look, as I said, it was caused by `rc1` prerelease meta. I updated the annotation again to `catalog.cattle.io/rancher-version: '>= 2.6.0-0'`, it should work now for `v2.6.0-rc1`. 

I didn't use `2.6.0-alpha1` because it is too specific for a milestone.

ref: https://github.com/Masterminds/semver
```
SemVer comparisons using constraints without a prerelease comparator will skip prerelease versions. For example, >=1.2.3 will skip prereleases when looking at a list of releases while >=1.2.3-0 will evaluate and find prereleases.
```

charts/longhorn-1.1/longhorn/100.0.0+up1.1.2/Chart.yaml
```diff
@@ -0,0 +1,40 @@
+annotations:
+  catalog.cattle.io/auto-install: longhorn-crd=match
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/display-name: Longhorn
+  catalog.cattle.io/namespace: longhorn-system
+  catalog.cattle.io/os: linux
+  catalog.cattle.io/provides-gvr: longhorn.io/v1beta1
+  catalog.cattle.io/rancher-version: '>= 2.6.0'
```
cbron, line 8: This one is missing the `-0`

cbron (commented):
Ahh, forgot about that part. Looks like we are missing it in one place, but otherwise lgtm. We can merge whenever longhorn team wants.

charts/longhorn-1.1/longhorn/100.0.0+up1.1.2/Chart.yaml
innobead: fixed. thanks!

innobead:
> Ahh, forgot about that part. Looks like we are missing it in one place, but otherwise lgtm. We can merge whenever longhorn team wants.

Sounds good. @khushboo-rancher will check again and sign off first before merging.

cbron (approved):

khushboo-rancher:
Re-validated the Longhorn with this branch - LGTM

innobead:
> Re-validated Longhorn with this branch - LGTM
> 
> Validated Installation, upgrade (v1.1.2 to v1.2.0), uninstallation and P0 cases.

Thank you @khushboo-rancher 

@cbron please help merge. Thanks.

PR #1340, opened 2021-07-08
Longhorn:  Update dev-v2.5 for v1.1.2 release

shuo-wu:
https://github.com/longhorn/longhorn/issues/2724

PR #1429, opened 2021-08-24
Add longhorn 1.2.0 chart dev 2.5

innobead:

cc @khushboo-rancher @meldafrawi @shuo-wu


khushboo-rancher:
Validated with this branch.

Validation - **Pass**

index.yaml
```diff
@@ -440,8 +482,8 @@ entries:
     - https://github.com/longhorn/longhorn-tests
     - https://github.com/longhorn/backing-image-manager
     urls:
-    - assets/longhorn/longhorn-1.1.100.tgz
-    version: 1.1.100
```
shuo-wu, line 444: Can we modify the version for existing entries? Similar below
Need to wait for Rancher team's review.

index.yaml
```diff
@@ -651,28 +693,28 @@ entries:
...
     description: Installs the CRDs for longhorn.
-    digest: a0bcc369ba8daa495df155bb459d826f711af13a4d82974f195bd0e16508a385
+    digest: 16484e129b9c24fd677dcece4b1ce82a9222e9d665be1076ddbf811a3499d837
     name: longhorn-crd
     type: application
     urls:
-    - assets/longhorn/longhorn-crd-1.1.200+up1.1.2.tgz
-    version: 1.1.200+up1.1.2
+    - assets/longhorn/longhorn-crd-1.2.0.tgz
+    version: 1.2.0
```
shuo-wu, line 703: I think we need to add a new entry rather than modifying the existing one.

index.yaml
innobead: This is automatically generated. 

@jiaqiluo any idea for this or what else we need to change in package.yaml?

innobead: cc @aiyengar2

innobead:

index.yaml
jiaqiluo: see my previous comment.

index.yaml
jiaqiluo: The items are ordered from the largest version to the smallest, so we see those changes here.
the real changes are the additions of longhorn 1.2.0 and longhorn-crd 1.2.0, which is expected.

jiaqiluo (approved):
LGTM.

khushboo-rancher:
Validated the chart - LGTM

innobead:
> Validated the chart - LGTM
> 
> The issue [longhorn/longhorn#2964](https://github.com/longhorn/longhorn/issues/2964) is solved with this PR
> 
> @innobead are the below versions as expected?
> Screen Shot 2021-09-07 at 3 11 37 PM

I think yes, but need @jiaqiluo to confirm again.

jiaqiluo:
I am sure that `1.1.100+up1.1.1` should be removed as it is a duplication of `1.1.100` (we plan to trim all those unneeded charts before releasing 2.5.10)

But I am not sure about `1.1.200+up1.1.2` @innobead  Do we have a plan to release it for 2.5.10? 

If not, could you remove both of them(assets, packages, and entries in the index.yaml)?

innobead:
> I am sure that `1.1.100+up1.1.1` should be removed as it is a duplication of `1.1.100` (we plan to trim all those unneeded charts before releasing 2.5.10)
> 
> But I am not sure about `1.1.200+up1.1.2` @innobead Do we have a plan to release it for 2.5.10?
> 
> If not, could you remove both of them(assets, packages, and entries in the index.yaml)?

We should have 1.1.2 because it's the patch release released.

jiaqiluo:
> > I am sure that `1.1.100+up1.1.1` should be removed as it is a duplication of `1.1.100` (we plan to trim all those unneeded charts before releasing 2.5.10)
> > But I am not sure about `1.1.200+up1.1.2` @innobead Do we have a plan to release it for 2.5.10?
> > If not, could you remove both of them(assets, packages, and entries in the index.yaml)?
> 
> We should have 1.1.2 because it's the patch release already released.

Oh, ok. I just found that it was released as 100.0.0+up1.1.2 in the [release-v2.6 branch ](https://github.com/rancher/charts/tree/release-v2.6/charts/longhorn/longhorn). Let's keep `1.1.200+up1.1.2` in this branch.

aiyengar2 (dismissed):
LGTM

aiyengar2:
FYI I have the same comment on this as I did on https://github.com/rancher/charts/pull/1428#pullrequestreview-749556036. If you plan to maintain two release lines for longhorn in 2.5, it would be better to split the `packages/` entries into two, as described in the comment.

packages/longhorn/generated-changes/patch/questions.yml.patch
```diff
@@ -1,101 +0,0 @@
---- charts-original/questions.yml
```
aiyengar2: Is this in the wrong commit?

packages/longhorn/generated-changes/patch/values.yaml.patch
```diff
@@ -1,52 +0,0 @@
---- charts-original/values.yaml
```
aiyengar2: Is this in the wrong commit?

aiyengar2 (changes requested):
Need to delete `charts/`, `assets/`, and `index.yaml` entries for the following orphaned charts, since they have been relocated to their own packages:

- [ ] longhorn-crd 1.1.100+up1.1.1
- [ ] longhorn-crd 1.1.200+up1.1.2
- [ ] longhorn 1.1.100+up1.1.1
- [ ] longhorn 1.1.200+up1.1.2

Can you add this to a separate commit post make charts?

innobead:
> Need to delete `charts/`, `assets/`, and `index.yaml` entries for the following orphaned charts, since they have been relocated to their own packages:
> 
> * [ ]  longhorn-crd 1.1.100+up1.1.1
> * [ ]  longhorn-crd 1.1.200+up1.1.2
> * [ ]  longhorn 1.1.100+up1.1.1
> * [ ]  longhorn 1.1.200+up1.1.2
> 
> Can you add this to a separate commit post make charts?

Please note that after removing` longhorn 1.1.200+up1.1.2` from the old longhorn package, the `index.yaml` required to update to make 1.1.2 release point to the new location `assets/longhorn-1.1/longhorn-crd-1.1.200+up1.1.2.tgz`.

Because 1.1.2 chart was introduced before (https://github.com/rancher/charts/pull/1340) even it's not released yet (release-2.5), we need to revalidate 1.1.2 as per this change.

aiyengar2:
> > Need to delete `charts/`, `assets/`, and `index.yaml` entries for the following orphaned charts, since they have been relocated to their own packages:
> > 
> > * [ ]   longhorn-crd 1.1.100+up1.1.1
> > * [ ]   longhorn-crd 1.1.200+up1.1.2
> > * [ ]   longhorn 1.1.100+up1.1.1
> > * [ ]   longhorn 1.1.200+up1.1.2
> > 
> > Can you add this to a separate commit post make charts?
> 
> Please note that after removing` longhorn 1.1.200+up1.1.2` from the old longhorn package, the `index.yaml` required to update to make 1.1.2 release point to the new location `assets/longhorn-1.1/longhorn-crd-1.1.200+up1.1.2.tgz`.
> 
> Because 1.1.2 chart was introduced before (#1340) even it's not released yet (release-2.5), we need to revalidate 1.1.2 as per this change. cc @khushboo-rancher

yes, we need to ensure the index.yaml at the end contains the new chart in the new asset location. 

One way to do this would be to add the above changes to the commit **before** make charts instead of after; then, when you run make charts, it will automatically detect that there is no chart entry for the asset and add it to the index.yaml

aiyengar2 (dismissed):
LGTM

khushboo-rancher:
Revalidated the chart with v1.1.2 and v1.2.0 with this branch - **LGTM**

Screen Shot 2021-09-09 at 4 42 33 PM

cbron (dismissed):
lgtm, the versions look weird but thats ok, it makes sense why it ended up that way.

index.yaml
```diff
@@ -393,14 +393,13 @@ entries:
...
-    - https://github.com/longhorn/backing-image-manager
     - https://github.com/longhorn/longhorn-manager
     - https://github.com/longhorn/longhorn-ui
     - https://github.com/longhorn/longhorn-tests
     - https://github.com/longhorn/backing-image-manager
     urls:
-    - assets/longhorn/longhorn-1.1.200+up1.1.2.tgz
-    version: 1.1.200+up1.1.2
+    - assets/longhorn-1.2/longhorn-1.2.0.tgz
+    version: 1.2.0
```
jiaqiluo, line 402: The changes in this file are actually:
- add the 1.2.0 
- remove the 1.1.100+up1.1.1

jiaqiluo (approved):
LGTM

innobead:
Resolved the index.yaml conflict. 

@cbron Please help review again. Thanks.

jiaqiluo (approved):

jiaqiluo (approved):

khushboo-rancher:
Revalidated the chart with this branch - **LGTM**

Verified installation, uninstallation, upgrade (v1.1.2- v1.2.0) and attaching/detaching, mounting of rwo and rwx volume.

cc: @innobead

innobead:
@cbron @jiaqiluo QA already verified it, so please feel free to merge. Thanks.

PR #1453, opened 2021-08-30
[2.6.2] bump rancher/mirrored-messagebird-sachet tag to 0.2.6; bump rancher-sachet chart version

jiaqiluo:
Issue https://github.com/rancher/rancher/issues/34558

Bump both `rancher-sachet` and `rancher-alerting-drives` chart to `100.0.1` since the only change is a bump of the image tag of `rancher/mirrored-messagebird-sachet`

--------
Note: this PR will be combined with https://github.com/rancher/charts/pull/1474


packages/rancher-sachet/package.yaml
```diff
@@ -1,2 +1 @@
 url: local
-version: 100.0.0
```
jiaqiluo, line 2: Since it is a local chart, so we can change its version in the `Chart.yaml` directly. Otherwise, the chart version will be wrong inside the `rancher-alerting-drivers/charts/sachet` sub-chart

brendarearden: https://github.com/rancher/charts/blob/33db4185684ece7b2ca1575cdc66af4841967764/packages/rancher-istio/package.yaml#L2 so this is how we handled istio since it is a local chart - is there a consistent way to handle this that we should document for local charts?

aiyengar2: @jiaqiluo I don't think it makes a difference if the version is incorrect in the subchart since no user will ever see it and we refer to sachet as a local chart. Nonetheless, I think we should just keep the version in the package.yaml since that's what we shipped in 2.6.0

aiyengar2: @brendarearden for most local charts, +up is not required; Istio is an edge case in that you have an upstream set of manifests, so you are manually patching the version.

I don't think there is any general, consistent guidance we would need but what would you propose?

packages/rancher-sachet/charts/Chart.yaml
```diff
@@ -7,5 +7,5 @@ apiVersion: v2
 name: rancher-sachet
 description: A Helm chart for Sachet based on the upstream https://github.com/messagebird/sachet
 type: application
-version: 1.0.1
```
aiyengar2, line 10: no need to change this version since it's already overridden by the package.yaml version

aiyengar2 (commented):
We should move the rancher-sachet version change back from Chart.yaml to package.yaml. Otherwise, LGTM since it's a super small bump + make charts.

@jiaqiluo when you redo the PR, could you change it to the following commits:
1. Commit where you just bump the package.yaml versions of **both** charts
2. Commit where you bump the tag of the sachet image
3. make charts

packages/rancher-sachet/package.yaml
brendarearden: @aiyengar2 my example was just meant to show that the version was still in package.yaml for istio (didn't mean for the +up to throw the example off) and not removed completely, so it sounds like that is consistently what we want to do . 

Mainly I was just concerned that if we were going to have charts that didn't have versions in the package.yaml then we would want to document in the readme why that is. Sounds like we will not be doing that so I think the current documentation is pretty good as is.

packages/rancher-sachet/charts/Chart.yaml
jiaqiluo: fixed.

packages/rancher-sachet/package.yaml
jiaqiluo: Thank you @brendarearden and @aiyengar2 for bringing this question up! 
I agree with you two that we should keep the version in the `package.yaml` as the source of truth for the chart version.

aiyengar2 (approved):

brendarearden (approved):

PR #1455, opened 2021-08-31
Istio 1.9.8

brendarearden:
This PR adds support for 1.9.8 as part of an OOB release

#### Add Istio 1.9 back into branch
Because `packages/rancher-istio-1.9` was removed in a previous commit (see historical context), the `packages/rancher-istio-1.9` needed to be added back into the project. This was done by checking out the directory into the current branch from the following commit: 13c579933e27806768ac704ff8e530edc57a80c4 

#### Add changes for istio 1.9.8
- Make changes in `packages/rancher-istio-1.9` to support 1.9.8
- Change the dependency path for kiali in `packages/rancher-istio-1.9` to point to the released version of https://github.com/rancher/charts/blob/release-v2.5/assets/rancher-kiali-server/rancher-kiali-server-crd-1.32.100.tgz. This was a great option suggeted by @aiyengar2 since we did not need to make kiali changes and it keeps us from temporarily having to track two kiali server versions. 

#### Make charts
- exported package for `rancher-istio-1.9`, ran `make charts` and made a commit

### Historical Context

The most recent PR for istio in the dev-v2.5 branch was #1381 
#### What did PR #1381 Do? 
- It removed `packages/rancher-istio-1.8`
- It removed `packages/rancher-istio-1.9`
- It updated `ackages/rancher-istio` to support Istio 1.10.2
- It updated `packages/rancher-kiali-server` to support kaili 1.35.0

Because the 1.9 package was removed, any OOB release for 1.9.x would require the package to be added back in if there was an OOB release

### Post OOB Release Cleanup
- [ ] Delete `packages/rancher-istio-1.9`


packages/rancher-kiali-server/package.yaml
```diff
@@ -1,4 +1,4 @@
-url: https://kiali.org/helm-charts/kiali-server-1.35.0.tgz
```
jiaqiluo, line 1: > I had to temporarily track two rancher-kiali-server versions, but this will be reverted when I complete the 1.10.4 backport

why do you need two rancher-kiali-server versions at the same time? why not update the existing one to 1.35.1?

jiaqiluo (changes requested):

packages/rancher-kiali-server/package.yaml
brendarearden: Normally for a new version I would update it if possible, but since we are doing an OOB release I want to introduce a new istio version and nothing else.

brendarearden: Also its only temporary until the version is moved to the release branch. I will be cleaning up the repo afterwards to only have 1 kiali version.

jiaqiluo: ok, it makes sense to me to have a temporary Kiali chart. 
My next question is why do you need to change the existing Kiali package?

brendarearden: Mainly it was my attempt to "not make any changes to the existing kaili in production" - so I renamed the existing package to rancher-kiali-server-1.35 in the dev branch so that I could check out the 1.32 version from the commit history at the 1.9.6 deployment commit. That way I knew that the rancher-kiali-server package I was using for istio 1.9.8 is the same one that was used for the istio currently in production. 

The rancher-kiali-server package I renamed has not been released yet for 2.5.x release so it made sense that that would be the one to temporarily change.  

If it would make it easier to reivew, I could revert this PR https://github.com/rancher/charts/pull/1381 and then redo this one after the revert gets merged.

jiaqiluo: Thanks for the explanation, it is clear to me now. 
As a note to other reviewers:
- the `packages/rancher-kiali-server-1.35` is the original `packages/rancher-kiali-server`
- the `packages/rancher-kiali-server` is modified to be Istio 1.32 and used for Istio 1.9.8
- the `packages/rancher-kiali-server-1.35` will be moved back to `packages/rancher-kiali-server`

charts/rancher-istio-1.9/rancher-istio/1.9.800/Chart.yaml
```diff
@@ -0,0 +1,21 @@
+annotations:
+  catalog.cattle.io/auto-install: rancher-kiali-server-crd=1.32.100
```
jiaqiluo, line 2: Here we still use the rancher-kiali-server-crd chart 1.32.100 which already existing in the repo. 

The new charts `rancher-kiali-server-1.32.100+up1.32.0` and `rancher-kiali-server-1.35.100+up1.35.0` will be removed in future.

jiaqiluo:
@brendarearden Why there is only the new entry for Istio in the index.yaml file? Did you delete the ones for `rancher-kiali-server-1.32.100+up1.32.0` and `rancher-kiali-server-1.35.100+up1.35.0` and etc. ?

brendarearden:
@jiaqiluo so we don't want to delete `rancher-kiali-server-1.35.100+up1.35.0` from the index.yaml because it is still in use by https://github.com/rancher/charts/blob/1fa344921a1d97f1c131e7897401eb964c2a7c20/packages/rancher-istio/charts/Chart.yaml#L16 -

and while istio 1.9.8 is not is not pointing to auto-install the 1.32.000+up version of kiali, to build the Istio chart, the package has be be available in package/rancher-kiali-server which means rancher-kiali-server has to be built and will end up in the index.yaml as the +up version so once we merge this in, I would do another PR to clean up the index.yaml.

brendarearden:
@jiaqiluo I updated the PR description with some additional context in a more structured format so that might help with context.

packages/rancher-istio/generated-changes/dependencies/kiali/dependency.yaml
```diff
@@ -1,2 +1,2 @@
 workingDir: ""
-url: packages/rancher-kiali-server
+url: packages/rancher-kiali-server-1.35
```
aiyengar2, line 2: just change this URL to point to the released assert. dependency.yaml supports everything package.yaml supports, so you can specify it as a release tgz or a git repository (pointing to release-v2.5, with subdirectory and commit).

This way we can avoid having an rancher-kiali-server release or a rancher-kiali-server-crd release. Also, no need to change annotations.

packages/rancher-kiali-server-1.35/generated-changes/overlay/templates/psp.yaml
```diff
@@ -0,0 +1,67 @@
+{{- if .Values.global.rbac.pspEnabled }}
```
aiyengar2, line 1: unnecessary change; see comment on dependency.yaml

packages/rancher-istio-1.9/charts/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
-appVersion: 1.9.6
+appVersion: 1.9.8
```
aiyengar2, line 2: with the dependency.yaml change below, we shouldn't need to modify the annotation for CRD install. However, if we do not go that route, you will need to bump the Chart.yaml here

aiyengar2 (changes requested):
Omit recreating the kiali-server chart since it's not necessary if no changes are required to the kiali 1.32.x line. See comments

packages/rancher-kiali-server/package.yaml
brendarearden: per arvind's suggestion, we are no longer tracking two kiali versions.

packages/rancher-istio/generated-changes/dependencies/kiali/dependency.yaml
brendarearden: awesome! I have redone the PR to use this suggestion!

packages/rancher-kiali-server-1.35/generated-changes/overlay/templates/psp.yaml
brendarearden: resolved in most recent PR commits.

packages/rancher-istio-1.9/generated-changes/dependencies/kiali/dependency.yaml
```diff
@@ -1,2 +1,2 @@
 workingDir: ""
-url: packages/rancher-kiali-server
+url: https://github.com/rancher/charts/raw/release-v2.5/assets/rancher-kiali-server/rancher-kiali-server-crd-1.32.100.tgz
```
aiyengar2:
```yaml
url: https://github.com/rancher/charts.git
subdirectory: charts/rancher-kiali-server/rancher-kiali-server/1.32.100
commit: 0ae672a2bb1d2679145b2c2297ec6830a5b4365e
```

aiyengar2 (changes requested):
pointed to wrong kiali-server subchart

packages/rancher-istio-1.9/generated-changes/dependencies/kiali/dependency.yaml
aiyengar2: commit is 0ae672a2bb1d2679145b2c2297ec6830a5b4365e

brendarearden: Yep that was my bad - fixed in the most recent commits!

jiaqiluo (approved):
LGTM

aiyengar2 (approved):
LGTM

cbron (changes requested):

charts/rancher-istio-1.9/rancher-istio/1.9.800/Chart.yaml
```diff
@@ -0,0 +1,21 @@
+annotations:
```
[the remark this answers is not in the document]
brendarearden: Yes, I am planning on addressing that as part of https://github.com/rancher/rancher/issues/33824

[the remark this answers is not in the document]
brendarearden: https://github.com/rancher/charts/pull/1462

cbron (approved):

PR #1456, opened 2021-09-01
Revert "Backport istio 1.10.2 and Kiali 1.35.0"

brendarearden:
Reverts rancher/charts#1381

Reasons why reverting this is the best option: 

We are doing an OOB release for 1.9.8 and this PR removed the 1.9 packages. 
We are not releasing 1.10.2 and be doing another backport for istio 1.10.4 instead.


brendarearden:
we do not need this revert

PR #1462, opened 2021-09-02
Remove istio 1.10.2 and add donNotRelease to package

brendarearden:
We are not going to be releasing a new 1.10.x for istio during the 2.5.10 release so this needs to be removed.


cbron (approved):
lgtm, but wait for others

jiaqiluo (approved):

PR #1473, opened 2021-09-15
Change vsphere-csi tolerations to match upstream and allow override from user

aiyengar2:
Related Issue: https://github.com/rancher/rancher/issues/34751


pennyscissors (approved):

packages/rancher-vsphere-csi/charts/templates/vsphere-csi-controller-deployment.yaml
```diff
@@ -39,25 +39,17 @@ spec:
...
-        # node becomes not-ready or unreachable. Default is 300 seconds if not specified.
-        #- key: node.kubernetes.io/not-ready
-        #  operator: Exists
-        #  effect: NoExecute
-        #  tolerationSeconds: 30
-        #- key: node.kubernetes.io/unreachable
-        #  operator: Exists
-        #  effect: NoExecute
-        #  tolerationSeconds: 30
+      {{- end }}
```
cbron, line 52: So to be clear, there isn't any functional changes but we put it in values.yaml to be easier to use ? And also `key: node-role.kubernetes.io/master` is always there ?

aiyengar2: There is a functional change. 

We are changing the default annotation from the one set here:

https://github.com/rancher/charts/blob/4067090e7f9bc351897f8c46a9ae8ddf92f3cabe/charts/rancher-vsphere-csi/rancher-vsphere-csi/100.0.0/templates/vsphere-csi-controller-deployment.yaml#L47-L50

To the one commented out here:

https://github.com/rancher/charts/blob/4067090e7f9bc351897f8c46a9ae8ddf92f3cabe/charts/rancher-vsphere-csi/rancher-vsphere-csi/100.0.0/templates/vsphere-csi-controller-deployment.yaml#L44-L46

aiyengar2: The reason why this change is necessary is since currently the CSI node will not get rescheduled if the underlying node dies, since we're allowing it to stay on the node even if it is marked NoSchedule or NoExecute. Instead, we should be forcibly moving the node to another node if NoExecute is specified, although it's fine to schedule on a NoSchedule master node.

The reason for the values.yaml change is also to unblock users from just overriding the tolerations field in the future.

cbron: Ok. But there are two tolerations nodes in the yaml now, on lines 43 and 48, those would technically overwrite each other.

cbron (approved):

PR #1476, opened 2021-09-15
bump rancher-webhook to 0.2.1-rc1

jiaqiluo:
issue: https://github.com/rancher/rancher/issues/34691


index.yaml
```diff
@@ -3990,6 +3990,25 @@ entries:
...
+    dependencies:
+    - condition: capi.enabled
+      name: capi
+      repository: ""
+    description: ValidatingAdmissionWebhook for Rancher types
+    digest: 2144f7ad506cf197d8013bdbf824a881d03b8fc13b50669bcf91389fec0053c0
+    name: rancher-webhook
+    urls:
+    - assets/rancher-webhook/rancher-webhook-1.0.1+up0.2.1-rc1.tgz
+    version: 1.0.1+up0.2.1-rc1
```
jiaqiluo, line 4011: the version will become `1.0.1+up0.2.1` when the upstream rancher-webhook 0.2.1 is cut.

packages/rancher-webhook/package.yaml
```diff
@@ -1,2 +1,2 @@
-url: https://github.com/rancher/webhook/releases/download/v0.2.0/rancher-webhook-0.2.0.tgz
-version: 1.0.0
+url: https://github.com/rancher/webhook/releases/download/v0.2.1-rc1/rancher-webhook-0.2.1-rc1.tgz
+version: 1.0.1
```
jiaqiluo, line 2: rancher-webhook `1.0.0+up0.2.0` is released https://github.com/rancher/charts/blob/release-v2.6/charts/rancher-webhook/rancher-webhook/1.0.0%2Bup0.2.0/Chart.yaml, 
so bump this version to `1.0.1`

cbron (approved):
lgtm. Current live version is 1.0.0+up0.2.0. Don't need to add min because webhook version is pinned in rancher. Already added to new-charts issue.

nickgerace (commented):
We do not have to remove an older chart right? This is the first RC, index.yaml looks good, and 0.2.0 already shipped.

jiaqiluo:
> We do not have to remove an older chart right? This is the first RC, index.yaml looks good, and 0.2.0 already shipped.

Correct, it is the 1st RC

PR #1480, opened 2021-09-17
Add max rancher version for 2.5.x charts

jakefhyde:
All 2.5.x charts have been updated to add the `catalog.cattle.io/rancher-version: < 2.6.1-0` annotation. Charts have been regenerated from updated assets. ~~The indentation level on yaml sequences is from `yq`, not sure how important it is but I can update if need be; I have generation scripts for this so it shouldn't be too much of a hassle~~. Files can be editied using a previous version of `yq`, I had success with 2.4.1. Confirmed that 2.5.x charts cannot be installed on 2.6.1-rc1.

Related Issue: https://github.com/rancher/rancher/issues/34479


brendarearden:
Screen Shot 2021-09-17 at 8 56 20 AM

I pulled this branch into my rancher ui and I am still seeing the older versions. Does this have to do with the legacy flag?

jakefhyde:
@brendarearden I noticed the same thing for a bit, but actually trying to install the charts on a lower version fails. I checked again just now and this is what it shows: 
istio

jakefhyde:
@brendarearden  This issue seems to be related to downstream clusters. I've added my charts repo on the local cluster, and no 2.5.x images show up. When adding to a downstream, it doesn't appear the versions are being checked.

rmweir (changes requested):

rmweir (changes requested):
Last feedback was not addressed. The commit `Add max rancher version for 2.5.x charts` should only contain the added rancher version constraint.

rmweir (approved):
LGTM

charts/longhorn/longhorn/1.0.200/Chart.yaml
```diff
@@ -6,30 +6,31 @@ annotations:
   catalog.cattle.io/release-name: longhorn
   catalog.cattle.io/ui-component: longhorn
   catalog.cattle.io/os: linux
+  catalog.cattle.io/rancher-version: "< 2.5.99-0"
```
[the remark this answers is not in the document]
rmweir: Complicated because file in question contains other changes.

cbron (commented):
This looks ok to me: 
* Internal charts like fleet/aks-operator/webhook don't need this
* CRD charts don't need this

The formatting is unfortunate but its in both the asset and the charts yaml, so I suppose thats ok. @aiyengar2 you agree ? 

One open question I have is, will doing this break anyone currently running a chart in 2.6 ? For example if you run istio 1.7.100 and this goes to release branch, then you try to edit chart what happens ?

jiaqiluo (approved):
LGTM

snasovich:
@jakefhyde , commit https://github.com/rancher/charts/pull/1480/commits/42349a772c76a4a243d7d44b1b9e2cabce1af883 (Add max version 2.6.1-0) has only *.tgz changes. I don't believe this is expected, so need to sort out commits here to have actual changes and generated ones separately in appropriately named commits.

jiaqiluo (approved):

jiaqiluo (approved):
LGTM. 
Note that the annotation is added to some of the old backup-crd charts, which is unnecessary but also harmless.

brendarearden (approved):
LGTM!

aiyengar2 (changes requested):
index.yaml needs to be regenerated so that the new annotations make it in the index. Currently, chart annotations on the index.yaml are out of date.

```bash
rm index.yaml
helm repo index --merge ./index.yaml --url assets assets;
mv assets/index.yaml index.yaml
```

aiyengar2 (approved):
LGTM

index.yaml
```diff
@@ -3835,9 +3879,9 @@ entries:
       catalog.cattle.io/release-name: vsphere-csi
     apiVersion: v1
     appVersion: 2.2.0
-    created: "2021-08-25T15:50:58.902011206-07:00"
+    created: "2021-09-29T14:12:43.383483308-04:00"
     description: vSphere Cloud Storage Interface (CSI)
-    digest: 8db3385ae16017f9f354a4e746ec44519eeadec8a0452b8a4833b224a7aac404
+    digest: a20bc110bca55c0059658daa7e5d5981ce2fcbbdcd2c957f0941170fa43c0afa
```
cbron, line 3884: Why'd the current charts digest change if no changes were made ?  This is for vsphere-csi 100.0.0, which is new for 2.6.0.

jiaqiluo: This is a known issue that the digest value changed each time we run the `make charts` even this chart is not touched.

cbron (approved):

PR #1481, opened 2021-09-17
Drop the rancher-version annotation from the rancher-webhook chart

jiaqiluo:
Drop the rancher-version annotation from the rancher-webhook chart since it's an internal chart and not user-facing.


packages/rancher-webhook/generated-changes/patch/Chart.yaml.patch
```diff
@@ -4,7 +4,7 @@
    catalog.cattle.io/namespace: cattle-system
    catalog.cattle.io/os: linux
    catalog.cattle.io/release-name: rancher-webhook
-+  catalog.cattle.io/rancher-version: ">= 2.5.10 <2.5.99"
++  catalog.cattle.io/rancher-version: ">= 2.5.10-0 <2.5.99-0"
```
cbron, line 7: This should happen in the upstream chart

jiaqiluo: right, it should be in the upstream chart, and this was a quick fix last time we did last time. 

I will close this PR and move the fix to the upstream chart.

cbron: This is 2.5.10, so we have until next week

charts/rancher-webhook/rancher-webhook/0.1.200-rc1+up0.1.2-rc1/Chart.yaml
```diff
@@ -3,7 +3,6 @@ annotations:
   catalog.cattle.io/hidden: "true"
   catalog.cattle.io/namespace: cattle-system
   catalog.cattle.io/os: linux
-  catalog.cattle.io/rancher-version: '>= 2.5.10 <2.5.99'
```
cbron, line 6: We may need this in the future for airgap purposes, but we can remove until then.

cbron (approved):

PR #1485, opened 2021-09-20
Update istio to work with new nginx spoof image

brendarearden:
**Issue**

**Depedent PR**
https://github.com/rancher/istio-installer/pull/32

**Problem**
istioctl upgrade reaches out to external github to download manifests, which breaks in an airgapped environment. The istio-installer image has been updated to intercept that call and download the manifests from the istio-installer image. 

**Solution**
The chart has been upgraded to pass in the proper environment variables for the new image mirror nginx server. It has also been updated to to allow a user to mount a manifest tar to the job pod to be served up by the nginx container to cover instances where an manifest tar wasn't bundled with the istio-installer


packages/rancher-istio/charts/values.yaml
```diff
@@ -1,11 +1,19 @@
 overlayFile: ""
 tag: 1.10.4
-##Setting forceInstall: true will remove the check for istio version < 1.6.x and will not analyze your install cluster prior to install
-forceInstall: false
 
 installer:
-  repository: rancher/istio-installer
-  tag: 1.10.4-rancher1
+  repository: brendarearden/istio-test
```
jiaqiluo, line 5: Should we use a rancher's repo here?

jiaqiluo (changes requested):

packages/rancher-istio/charts/values.yaml
brendarearden: Yes, I will switch this over once the new image is tagged .

charts/rancher-istio/rancher-istio/100.0.1+up1.10.4/templates/istio-uninstall-job.yaml
```diff
@@ -39,7 +42,9 @@ spec:
             name: istio-installer-overlay
         {{ end }}
       serviceAccountName: istio-installer
+      {{- if .Values.global.rbac.pspEnabled }}
       securityContext:
         runAsUser: 101
         runAsGroup: 101
+      {{- end }}
```
jiaqiluo, line 49: 1. We are not checking the value of `.Values.global.rbac.pspEnabled` at this place in the `istio-install-job.yaml` file. Should we add it there? 
2. In both files, the `securityContext` at the container lever will be applied anyway, which makes the check useless. I think we should either remove the `securityContext` at the container level or move the checks there.

jiaqiluo (changes requested):

packages/rancher-istio/charts/templates/istio-install-job.yaml
```diff
@@ -41,6 +55,11 @@ spec:
           configMap:
             name: istio-installer-overlay
         {{- end }}
+        {{- if .Values.installer.releaseMirror.provideAdditionalIstioTag }}
+        - name: istio-additional-tag-volume
+          hostPath:
```
cbron, line 60: hostPath is not a good idea for a few reasons, one is that the tar won't exist on every node and we don't yet have nodeSelectors. Lets drop the volume part for now and release not that this only supports upgrading from versions available from rancher 2.5.0-2.5.9

packages/rancher-istio/charts/templates/istio-uninstall-job.yaml
```diff
@@ -39,7 +42,9 @@ spec:
             name: istio-installer-overlay
         {{ end }}
       serviceAccountName: istio-installer
+      {{- if .Values.global.rbac.pspEnabled }}
       securityContext:
         runAsUser: 101
         runAsGroup: 101
+      {{- end }}
```
cbron, line 49: Why do we wrap this one in conditional but not the above one ?

charts/rancher-istio/rancher-istio/100.0.1+up1.10.4/templates/istio-uninstall-job.yaml
brendarearden: I have moved the security context to the container level for this and am no longer doing the rbac.pspEnabled check as it should always run as user 101

charts/rancher-istio/rancher-istio/100.0.1+up1.10.4/templates/istio-install-job.yaml
```diff
@@ -20,7 +24,9 @@ spec:
           - name: ISTIO_NAMESPACE
             value: {{ template "istio.namespace" . }}
           - name: FORCE_INSTALL
-            value: {{ .Values.forceInstall | default "false" | quote }}
+            value: {{ .Values.installer.forceInstall | default "false" | quote }}
+          - name: RELEASE_MIRROR_ENABLED
+            value: {{ .Values.installer.releaseMirror.enabled | default "true" | quote }}
```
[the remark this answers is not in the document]
brendarearden: Yeah I thought that was weird too - I have removed it and am now just relying on the values.yaml to be set

packages/rancher-istio/charts/values.yaml
```diff
@@ -1,11 +1,19 @@
...
+  tag: spoof-2.0.4
+  ##Setting forceInstall: true will remove the check for istio version < 1.6.x and will not analyze your install cluster prior to install
+  forceInstall: false
+  ##releaseMirror are configurations for istio upgrades. 
+  ##Setting releaseMirror.enabled: true will cause istio to use bundled in images from rancher/istio-installer to perfom an upgrade - this is ideal
+  ##for airgap setups. Setting releaseMirror.enabled: false will allow istio to use its default call to external github to get the required images.
+  releaseMirror:
+    enabled: true
```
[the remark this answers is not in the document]
brendarearden: this is now set to false

packages/rancher-istio/charts/values.yaml
```diff
@@ -1,11 +1,19 @@
...
 installer:
-  repository: rancher/istio-installer
-  tag: 1.10.4-rancher1
+  repository: brendarearden/istio-test
+  tag: spoof-2.0.4
+  ##Setting forceInstall: true will remove the check for istio version < 1.6.x and will not analyze your install cluster prior to install
+  forceInstall: false
+  ##releaseMirror are configurations for istio upgrades. 
+  ##Setting releaseMirror.enabled: true will cause istio to use bundled in images from rancher/istio-installer to perfom an upgrade - this is ideal
+  ##for airgap setups. Setting releaseMirror.enabled: false will allow istio to use its default call to external github to get the required images.
```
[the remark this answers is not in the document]
brendarearden: Updated the wording with your suggestion.

packages/rancher-istio/charts/templates/istio-install-job.yaml
```diff
@@ -11,6 +11,10 @@ spec:
   backoffLimit: 1
   template:
     spec:
+      hostAliases:
+      - ip: "127.0.0.1"
+        hostnames:
+        - "github.com"
```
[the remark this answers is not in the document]
brendarearden: That is correct, this now replaces the need for the /etc/host. I have now wrapped it in an if statement for mirror enabled

packages/rancher-istio/charts/templates/base-config-map.yaml
```diff
@@ -4,4 +4,4 @@ metadata:
   name: istio-installer-base
   namespace: {{ template "istio.namespace" . }}
 data:
-{{ tpl (.Files.Glob "configs/*").AsConfig . | indent 2 }}
+{{ tpl (.Files.Glob "configs/istio-base.yaml").AsConfig . | indent 2 }}
```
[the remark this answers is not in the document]
brendarearden: I have reverted this change. I had originally changed it so that I could store more files in the `configs/`

packages/rancher-istio/charts/templates/istio-install-job.yaml
brendarearden: https://github.com/rancher/rancher/issues/34858 I have opened an issue here to cover this scenario and I am removing the hostPath change from the chart.

packages/rancher-istio/charts/templates/istio-uninstall-job.yaml
brendarearden: Originally the security context was only set when PSP was enabled, but now it needs to be on all the time - I have removed the job level security context and I am only keeping the container level securityContext so this is not longer an issue.

jiaqiluo (dismissed):

packages/rancher-istio/charts/templates/istio-uninstall-job.yaml
```diff
@@ -29,6 +29,9 @@ spec:
             mountPath: /app/overlay-config.yaml
             subPath: overlay-config.yaml
           {{ end }}
+          securityContext:
+            runAsUser: 101
+            runAsGroup: 101
```
aiyengar2, line 34: same comment as above re: choice of user / group

packages/rancher-istio/charts/templates/istio-install-job.yaml
```diff
@@ -32,6 +40,9 @@ spec:
             mountPath: /app/overlay-config.yaml
             subPath: overlay-config.yaml
           {{- end }}
+          securityContext:
+            runAsUser: 101
+            runAsGroup: 101
```
aiyengar2, line 45: Should this be configurable (and also update the PSP)? Why was it set to 101?

aiyengar2: nit: why not just leave this where it originally was under serviceAccountName?

packages/rancher-istio/charts/templates/istio-install-job.yaml
```diff
@@ -20,7 +26,9 @@ spec:
           - name: ISTIO_NAMESPACE
             value: {{ template "istio.namespace" . }}
           - name: FORCE_INSTALL
-            value: {{ .Values.forceInstall | default "false" | quote }}
+            value: {{ .Values.installer.forceInstall | quote }}
```
aiyengar2, line 29: Why did we remove the default?

packages/rancher-istio/charts/templates/istio-install-job.yaml
```diff
@@ -20,7 +26,9 @@ spec:
...
+          - name: RELEASE_MIRROR_ENABLED
+            value: {{ .Values.installer.releaseMirror.enabled | quote }}
```
aiyengar2, line 31: should a default be introduced here too?

packages/rancher-istio/charts/values.yaml
```diff
@@ -56,7 +61,7 @@ global:
   defaultPodDisruptionBudget:
     enabled: true
   rbac:
-    pspEnabled: false
+    pspEnabled: true
```
aiyengar2, line 64: 👍  this changes istio to have the same default as monitoring

aiyengar2 (commented):
needs to change image before approval

packages/rancher-istio/charts/templates/istio-install-job.yaml
brendarearden: default cannot be introduced here because the value is used for an if check else where and I don't want inconsistencies in how that evaluates. Its also set by default in the values.yaml so unless the user removes the value all together then it will be set.

packages/rancher-istio/charts/templates/istio-install-job.yaml
cbron: Because its set in the values.yaml, whats point in keeping it both places ?

brendarearden: this is added back in with the most recent commits - I originally removed it because we are setting this by default in the values.yaml so doesn't make sense to also have a default value - but it also doesn't hurt in this situation.

packages/rancher-istio/charts/templates/istio-install-job.yaml
brendarearden: Moved back to global job setting instead of per contian. This should not be configurable because the istio-installer image does not run as root so I do not want people to attempt to run the container as root. The user in the istio-installer is a 101 user.

packages/rancher-istio/charts/templates/istio-uninstall-job.yaml
brendarearden: see answer to other security context question.

packages/rancher-istio/charts/templates/istio-install-job.yaml
brendarearden: https://github.com/nginxinc/docker-nginx/blob/d496baf859613adfe391ca8e7615cc7ec7966621/mainline/alpine/Dockerfile#L17 additional context of how the user is generated - this is from the alpine image, but its the same across all nginx user setups.

cbron (commented):
lgtm, but wait for @aiyengar2 and need to tag image

brendarearden:
Image has been updated to the most recent istio-installer tag and is ready for final review.

charts/rancher-istio/rancher-istio/100.0.1+up1.10.4/templates/istio-uninstall-job.yaml
brendarearden: Moved back to job context instead of container context.

jiaqiluo (approved):

packages/rancher-istio/charts/templates/istio-install-job.yaml
aiyengar2: ah yes this makes sense. Thanks for the clarification!

packages/rancher-istio/charts/templates/istio-install-job.yaml
aiyengar2: sounds good to me. I don't think dropping it is that much of a concern.

aiyengar2 (approved):
LGTM

PR #1490, opened 2021-09-22
[Backport 2.5] Backport vsphere-csi 2.2.0 chart

aiyengar2:
Related Issue: https://github.com/rancher/rancher/issues/34751


aiyengar2 (commented):
@PennyScissors I did a straight backport of this chart by just copying the charts/ from 2.6 into here. Is there anything that needs to be changed for 2.5 only? Also, are there associated backport issues you can tag on this so that QA can test those use cases?

cbron:
Seems like this is ongoing, removing assignment for now

PR #1493, opened 2021-09-23
fixing dns resolution issues in fluent-bit

paynejacob:
https://github.com/rancher/rancher/issues/34772

There is a bug in older versions of fluent-bit that can cause dns resolution failures, It has to do with 0 value timeouts meaning the issues can be transient.  I updated to the latest fluent-bit as it contains  a fix for these issues.  I tested on RKE v1.21 and did not see any dns issues with the new image.  

https://github.com/fluent/fluent-bit/issues/4050


paynejacob:
https://github.com/rancher/image-mirror/pull/164/files

snasovich (approved):

nickgerace (approved):

PR #1494, opened 2021-09-24
Fix for OPA Gatekeeper 3.6.0 charts.

rayandas:
This is a fix for https://github.com/rancher/rancher/issues/34897


PennyScissors (approved):
LGTM, but let's wait until @aiyengar2 approves as well before merging this.

aiyengar2 (approved):
LGTM

aiyengar2:
FYI this change is necessary since our image build scripts currently have a dependency on `values.yaml` using the labels `repository` and `tag` for all images, as shown here: https://github.com/rancher/rancher/blob/cb7de4e6c3d7e783828dc662b1142c1f9ae5edbe/pkg/image/resolve.go#L119-L120.

Therefore, a field like `crdRepository` needs to be refactored to `crdImage.repository` and `crdImage.tag` to specifically call out which image needs to be pulled in.

cbron (approved):

PR #1497, opened 2021-09-24
Add longhorn 1.2.1 chart dev 2.5

innobead:


meldafrawi:
Validated installation/upgrade on rancher v2.5.9, LGTM
Ref: longhorn/longhorn#3090

PR #1499, opened 2021-09-24
Release Feature Charts for v2.6.1

nickgerace:
https://github.com/rancher/rancher/issues/34589


index.yaml
```diff
@@ -375,7 +473,7 @@ entries:
       catalog.cattle.io/ui-component: longhorn
     apiVersion: v1
     appVersion: v1.1.2
-    created: "2021-09-29T14:12:43.175318108-04:00"
+    created: "2021-08-25T15:50:58.486087084-07:00"
```
aiyengar2, line 476: Note: Longhorn 1.1 version is correct, it is not actually being released. We just moved longhorn to its own directory but made no other changes. Therefore, a new asset under `assets/longhorn-1.1` for `100.0.0+up1.1.2` is expected (which matches this: https://github.com/rancher/charts/tree/release-v2.6/charts/longhorn/longhorn/100.0.0%2Bup1.1.2), not `100.0.1+up1.1.2`. 

The index.yaml should also not be modified for this additional chart.

aiyengar2 (changes requested):
Why did `assets/fleet-agent/fleet-agent-0.3.000.tgz ` and `charts/fleet-agent/fleet-agent/0.3.000/Chart.yaml` have permission changes? Can we remove those?

Otherwise, LGTM

index.yaml
aiyengar2: @cbron should we consider putting `doNotTrack` on the longhorn 1.1 chart in order to avoid releasing a second, identical version of the same chart? It should have no impact since our index.yaml does not reference it, but we could avoid generating the new chart just for good measure.

nickgerace:
> Why did assets/fleet-agent/fleet-agent-0.3.000.tgz and charts/fleet-agent/fleet-agent/0.3.000/Chart.yaml have permission changes? Can we remove those?

@aiyengar2 no idea whatsoever. Could it related to the new change @cbron made? I ran the following command for this PR from scratch:

```sh
STAGING_BRANCH=dev-v2.6
LIVE_BRANCH=release-v2.6
FORKED_BRANCH=release-v2.6.1
git fetch upstream
git checkout upstream/${LIVE_BRANCH} -b ${FORKED_BRANCH}
git branch -u origin/${FORKED_BRANCH}
git checkout upstream/${STAGING_BRANCH} -- charts assets index.yaml
git restore --staged .
git diff -p --diff-filter=ACM \
    | grep -E '^(diff|old mode|new mode)' \
    | sed -e 's/^old/NEW/;s/^new/old/;s/^NEW/new/' \
    | git apply
git checkout assets/README.md charts/README.md
git add index.yaml assets/
git commit -m "add index.yaml and assets"
git add charts
git commit -m "add charts"
git push --set-upstream origin ${FORKED_BRANCH}
```

nickgerace:
Fixing with @aiyengar2 ...

charts/fleet-agent/fleet-agent/100.0.1+up0.3.7/Chart.yaml
```diff
@@ -0,0 +1,12 @@
+annotations:
```
cbron, line 1: One thing we missed was to add the rancher-max to previous charts that are user facing:

* OPA gatekeeper - needs max, last version was 100.0.0+up3.5.1.tgz
* CIS - needs max since underlying image versions are changing
* Backup - needs max since underlying image versions are changing. Technically 2.0.0 is greater than any 1.x.0, but you could still "downgrade" any backports. This one and CIS are actually worse because its not obvious to user that its a downgrade.
* Istio - doesn't actually matter since its same underlying version and its just a bug fix. We'll have to max both in future, but now it doesn't matter. 

So 3 charts from 2.6.0 need max of < 2.6.1-0.

index.yaml
nickgerace: Done

charts/fleet-agent/fleet-agent/100.0.1+up0.3.7/Chart.yaml
nickgerace: This is being addressed in a follow-up PR by @jakefhyde

aiyengar2 (approved):
LGTM, discussed https://github.com/rancher/charts/pull/1499#discussion_r719612012 offline and it seems like we're doing that in an independent PR.

nickgerace:
Finished self-review after comments that've been addressed. LGTM

index.yaml
```diff
@@ -3005,12 +3268,11 @@ entries:
       catalog.cattle.io/namespace: cattle-monitoring-system
       catalog.cattle.io/os: linux
       catalog.cattle.io/provides-gvr: monitoring.coreos.com.prometheus/v1
-      catalog.cattle.io/rancher-version: < 2.6.1-0
```
jiaqiluo, line 3008: I think we should not drop the annotaoin in this PR, right?

jiaqiluo (changes requested):

index.yaml
jiaqiluo: The drop happens in multiple places& charts.

nickgerace: @aiyengar2 is this addressed by https://github.com/rancher/charts/pull/1520?

jiaqiluo: Your PR removes much more than the above PR adds back.

nickgerace: Well, only for that field though

nickgerace: Other than the annotation being dropped here (and wherever it exists), the PR is updating everything in place in `index.yaml` or is adding the new charts/assets. That's the only issue I see now, is that the same for you?

jiaqiluo: that is right. All newly-added charts look right; the only issue is the drop of annotations on exiting charts

nickgerace:
@cbron @aiyengar2 @jiaqiluo this is ready for re-review.

charts/rancher-backup/rancher-backup/2.0.0/Chart.yaml
```diff
@@ -9,7 +9,6 @@ annotations:
   catalog.cattle.io/scope: management
   catalog.cattle.io/type: cluster-tool
   catalog.cattle.io/ui-component: rancher-backup
-  catalog.cattle.io/rancher-version: < 2.6.1-0
```
jiaqiluo, line 12: this should NOT be dropped. 
I doubt your base is not the latest release-v2.6 ?

jiaqiluo (changes requested):

charts/rancher-backup/rancher-backup/2.0.0/Chart.yaml
nickgerace: It is. It's the same issue `max version` again.
https://github.com/rancher/charts/pull/1541#pullrequestreview-775444057

aiyengar2 (approved):
fleet 100.0.1+up0.3.7
- [x] assets added?
- [x] index.yaml entry added?
longhorn 100.0.0+up1.1.2
- [x] assets added?
- [x] index.yaml entry added?
longhorn 100.1.0+up1.2.2
- [x] assets added?
- [x] index.yaml entry added?
aks-operator 100.0.1+up1.0.2
- [x] assets added?
- [x] index.yaml entry added?
rancher-backup 2.0.1
- [x] assets added?
- [x] index.yaml entry added?
cis-benchmark 2.0.1
- [x] assets added?
- [x] index.yaml entry added?
opa-gatekeeper 100.0.1+up3.6.0
- [x] assets added?
- [x] index.yaml entry added?
rancher-istio 100.0.1+up1.10.4
- [x] assets added? 
- [x] index.yaml entry added?
rancher-webhook 1.0.1+up0.2.1
- [x] assets added?
- [x] index.yaml entry added?

`catalog.cattle.io/rancher-version` inconsistencies in `index.yaml` were noticed and discussed offline. Since all the rancher-versions annotations seem to include 2.6.0 and we will anyways revisit those annotations when releasing 2.6.2 (to add the rancher max versions), we are not blocking this release PR on these inconsistencies.

LGTM

cbron (approved):
Charts changing match the "new charts" issue. Index matches too.

PR #1500, opened 2021-09-25
[dev-v2.5] Refactor and update CSI to v2.3.0, and make volume expansion in storage class configurable 

PennyScissors:
Issues:
https://github.com/rancher/rancher/issues/34630
https://github.com/rancher/rancher/issues/34797

1st commit backports the CSI chart from dev-v2.6 since there were a few fixes we made that weren't present in the 2.5 line.

2nd commits refactors the chart's template filenames to shorter names since the `vsphere-csi` was redundant. Also, breaks down resources into their own files to improve organization.

3rd commit is the actual update of the chart to the v2.3.0 upstream

4th commit is to fix issue https://github.com/rancher/rancher/issues/34797


packages/rancher-vsphere-csi/charts/questions.yaml
```diff
@@ -110,6 +110,11 @@ questions:
         default: true
         type: boolean
 
+      - variable: storageClass.allowVolumeExpansion
+        label: Allow Volume Expansion
+        description: Allows resizing the volume by editing the corresponding PVC object
+        type: string
```
jiaqiluo, line 116: Why is the type `string`, not `boolean`?

jiaqiluo (changes requested):
`doNotRelease: true` should be removed from the `package.yaml` file.

PennyScissors:
> `doNotRelease: true` should be removed from the `package.yaml` file.

@jiaqiluo Which commit do you see this added? I don't see this in the package.yaml file.

packages/rancher-vsphere-csi/charts/questions.yaml
PennyScissors: Good catch, fixed

jiaqiluo:
@PennyScissors  Here is the line I mentioned https://github.com/rancher/charts/blob/785da257bf3073474a422dc424c6177a509bbde1/packages/rancher-vsphere-csi/package.yaml#L3

PennyScissors:
Waiting on https://github.com/rancher/charts/pull/1510 to be merged so that I can rebase and amend my backport commit.

aiyengar2:
> Waiting on #1510 to be merged so that I can rebase and amend my backport commit.

it's merged.

packages/rancher-vsphere-csi/charts/questions.yaml
```diff
@@ -5,46 +5,83 @@ questions:
...
         description: Password for vCenter
         type: password
-        group: Configuration
 
   - variable: vCenter.configSecret.name
     label: CSI vSphere Config Secret Name
     description: Name of the Secret that contains a CSI vSphere config and credentials (Will not be visible in the API. More info in the README)
     type: string
-    group: Configuration
+    group: vCenter Configuration
```
aiyengar2, line 35: The overall PR looks good to me, but I'm a little concerned about the fact that we are doing a heavy refactor of the chart within the 2.5 release line of Rancher, which will presumably be forward-ported only in 2.6.2's release later this year.

Specifically, here is the upgrade scenario that concerns me:
- User is currently on Rancher 2.5 using the older version of the vsphere chart
- User upgrades to Rancher 2.5.10
- User upgrades to latest vsphere chart version 2.3.0
- User upgrades Rancher to 2.6.1

In this situation, will the user be able to upgrade from `2.3.0` -> `100.0.0`? I presume the answer here is no, since that will cause them to downgrade vsphere-cpi versions.

In addition, if a user chooses to keep `2.3.0` in Rancher 2.6 since there is no path for upgrade, is there a way to do an in-place Helm upgrade / re-install after migrating? The dev-v2.5 ChartRepository will no longer exist, so I do not think this is an option either.

aiyengar2: From what I can tell, this probably would merit an OOB release for vsphere after 2.5.10 is released into the 2.6 charts release line to ensure that this type of upgrade path is still supported for users to upgrade to 2.6.1.

aiyengar2 (approved):
Changes in this PR look good to me, but I have concerns about the upgrade path to 2.6 (in comment below).

PennyScissors:
Rebased to pull in the tolerations that were reverted, and reorganized the templates a little bit more since the file names were still a bit long.

jiaqiluo (approved):
The PR looks good to me. And we need to address the upgrading issue Arvind mentioned:
- do an OOB release to fork this chart to release-v2.6
- forward port the fixes to dev-v2.6, as a new chart

aiyengar2 (approved):

PR #1502, opened 2021-09-27
Bump suc to 0.8.0-rc.1 and chart to 100.0.1

thedadams:
Related issues:
https://github.com/rancher/rancher/issues/34908
https://github.com/rancher/rancher/issues/32633


packages/system-upgrade-controller/generated-changes/exclude/questions.yml
```diff
@@ -1 +1 @@
-rancher_min_version: 2.4.0-rc1
```
jiaqiluo, line 1: Did you change this file manually? If so can you move it to the 1st commit? 
Also, since this file is excluded from the generated chart, why did you update it?

packages/system-upgrade-controller/generated-changes/patch/Chart.yaml.patch
```diff
@@ -12,8 +12,8 @@
  sources:
 -  - "https://github.com/rancher/system-charts/blob/dev-v2.6/charts/rancher-k3s-upgrader"
 +  - "https://github.com/rancher/system-charts/charts/system-upgrade-controller"
- version: 0.3.0
- appVersion: v0.7.5
+ version: 0.3.1
+ appVersion: v0.8.0-rc.1
```
jiaqiluo, line 16: Did you change this file manually? If so can you move it to the 1st commit?

jiaqiluo (changes requested):

packages/system-upgrade-controller/generated-changes/exclude/questions.yml
thedadams: I didn't change this at all. The file was updated upstream, so the exclude file was here was updated also.

packages/system-upgrade-controller/generated-changes/patch/Chart.yaml.patch
thedadams: I didn't change this file. It was changed by `make patch`

charts/system-upgrade-controller/system-upgrade-controller/100.0.1+up0.3.1/Chart.yaml
```diff
@@ -0,0 +1,14 @@
...
+  catalog.cattle.io/namespace: cattle-system
+  catalog.cattle.io/os: linux
+  catalog.cattle.io/release-name: system-upgrade-controller
+apiVersion: v1
+appVersion: v0.8.0-rc.1
+description: General purpose controller to make system level updates to nodes
+home: https://github.com/rancher/system-charts/charts/system-upgrade-controller
+name: system-upgrade-controller
+sources:
+- https://github.com/rancher/system-charts/charts/system-upgrade-controller
```
jiaqiluo, line 13: The home URL and the sources URL is invalid, 404 error is returned

jiaqiluo (changes requested):

charts/system-upgrade-controller/system-upgrade-controller/100.0.1+up0.3.1/Chart.yaml
thedadams: I didn't change those, they are the same as the previous version. Maybe @kinarashah can comment on that.

kinarashah: @jiaqiluo What are the home and source URLs used for? This is sourced from upstream rancher/system-charts https://github.com/rancher/charts/blob/dev-v2.6/packages/system-upgrade-controller/package.yaml, and the released version 0.3.0 also has the same values https://github.com/rancher/charts/blob/dev-v2.6/charts/system-upgrade-controller/system-upgrade-controller/100.0.0%2Bup0.3.0/Chart.yaml.

kinarashah: Original PR for more context: https://github.com/rancher/charts/pull/1422

jiaqiluo: @kinarashah Those are general information of the chart, which are not related to any functionality of the chart. I think they are just out of date in the upstream chart, and we need to update them sooner or later.

kinarashah: @jiaqiluo I see. Looks like we missed setting those correctly when sourcing from upstream for 0.3.0? 
I'm still confused on what the values should be, upstream chart by itself has the right path https://github.com/rancher/system-charts/blob/dev-v2.6/charts/rancher-k3s-upgrader/0.3.1/Chart.yaml#L5. Should these URLs point to `charts/system-upgrade-controller` or `system-charts/rancher-k3s-upgrader`?

packages/system-upgrade-controller/generated-changes/patch/Chart.yaml.patch
```diff
@@ -12,8 +12,8 @@
  sources:
 -  - "https://github.com/rancher/system-charts/blob/dev-v2.6/charts/rancher-k3s-upgrader"
 +  - "https://github.com/rancher/system-charts/charts/system-upgrade-controller"
```
jiaqiluo, line 14: We should drop those 2 lines to keep the source URL unchanged. 

Also, change the above home URL to `https://github.com/rancher/system-charts/blob/dev-v2.6/charts/rancher-k3s-upgrader`

jiaqiluo (changes requested):

charts/system-upgrade-controller/system-upgrade-controller/100.0.1+up0.3.1/Chart.yaml
jiaqiluo: I think the url should be the `charts/system-upgrade-controller` one. see my comment below.

packages/system-upgrade-controller/generated-changes/patch/Chart.yaml.patch
thedadams: I believe this means that we will leave the source and home URLs unchanged from the "upstream" chart. I will update this.

jiaqiluo (changes requested):
LGTM, but I have one last thing to confirm:
It is the official 0.3.1 release for the upstream chart, but the `appVersion` and the image tag is the weird value `v0.8.0-rc.1`.  @thedadams Do you know why?

thedadams:
@jiaqiluo I'm not sure if an RC version of the chart is necessary in this case because the chart version (v0.3.1) and the image version (v0.8.0-rc.1) are separate. The plan was to replace this chart release with the non-RC image once we validate the related issues.

Do you think we need an RC version of the chart?

jiaqiluo:
> The plan was to replace this chart release with the non-RC image once we validate the related issues.

@thedadams I am totally fine if this is the plan

jiaqiluo (approved):

kinarashah (approved):

PR #1503, opened 2021-09-27
Remove Red Sky Ops chart

jgustie:


charts/redskyops/v0.1.2/questions.yml
```diff
@@ -1,45 +0,0 @@
-labels:
```
PennyScissors, line 1: Add `rancher_max_version: 2.5.10` at the top of this file for both charts.

PennyScissors (changes requested):
Hello @jgustie, we don't allow deleting charts since it could affect users using those charts currently. Instead, please implement the constraint I suggested in the questions.yaml of both charts and it should hide them in future Rancher releases post 2.5.10.

PennyScissors:
This PR should only contain 1 commit (remove all other commits) to add `rancher_max_version: 2.5.10` at the top of [v0.1.1's questions.yaml](https://github.com/rancher/charts/blob/master/charts/redskyops/v0.1.1/questions.yml) and [v0.1.2's questions.yaml](https://github.com/rancher/charts/blob/master/charts/redskyops/v0.1.2/questions.yml) in order to hide them both. Everything else in the charts should remain unmodified.

PR #1505, opened 2021-09-27
[2.6.3] Expose Merge_Log field to disable JSON parsing in logs

a-blender:
Fixes [34720](https://github.com/rancher/rancher/issues/34720) and exposes related default fields. 

#### Root cause

There was no way for the user to configure the Merge_Log field for the fluentd pod configuration, so the user always got their logs JSON parsed. They didn't want that.
 
#### What was fixed, or what changes have occurred

- Add Merge_Log (and other Merge keys) to the fluentd configuration in the rancher-logging chart
- Update rancher-logging version to 3.101+
 
#### Areas or cases that should be tested

- Rancher logging chart can be installed on a local Rancher cluster
- Rancher logging upgrade works
- Check that Merge_Log field is set to “” in fluentd pod configuration (you can check this in the Rancher UI under secrets in the fluentd pod config)


packages/rancher-logging/generated-changes/overlay/templates/loggings/aks/logging.yaml
```diff
@@ -52,19 +52,21 @@ spec:
...
       {{- end }}
-      {- if .Values.fluentbit.filterKubernetes.Merge_Log_Trim }}
+      {{- if .Values.fluentbit.filterKubernetes.Merge_Log_Trim }}
       Merge_Log_Trim: "{{ .Values.fluentbit.filterKubernetes.Merge_Log_Trim }}"
       {{- end }}
-      {- if .Values.fluentbit.filterKubernetes.Merge_Parser }}
+      {{- if .Values.fluentbit.filterKubernetes.Merge_Parser }}
       Merge_Parser: "{{ .Values.fluentbit.filterKubernetes.Merge_Parser }}"
       {{- end }}
+    {{- end}}
```
nickgerace, line 69: These need a space at the end `{{- end }}`. This includes the other files

nickgerace (changes requested):
I think with my last comment and your commits squashed into the following...

1. your changes in one commit
2. `make charts` in another commit

...we should be good!

nickgerace (dismissed):

nickgerace:
Awaiting review from @paynejacob and Red Team.

paynejacob (dismissed):

prachidamle:
@annablender, Can we test upgrade of logging chart on a downstream cluster:
- install 100.0.0 first and then upgrade to new version 100.0.1 and ensure no issues with chart upgrade.

annablender:
Upgrade test 100.0.0 -> 100.0.1 needs to be done

index.yaml
```diff
@@ -3014,21 +2990,7 @@ entries:
...
-    - assets/rancher-logging/rancher-logging-crd-100.0.1+up3.12.0.tgz
-    version: 100.0.1+up3.12.0
-  - annotations:
-      catalog.cattle.io/certified: rancher
-      catalog.cattle.io/hidden: "true"
-      catalog.cattle.io/namespace: cattle-logging-system
-      catalog.cattle.io/release-name: rancher-logging-crd
-    apiVersion: v1
-    created: "2021-08-25T15:50:58.609939054-07:00"
+    created: "2021-09-30T14:07:17.987241483-04:00"
```
nickgerace: I believe are missing the package now with this merge commit.
We should probably perform an interactive rebase (`git rebase`) and then manually fix the confict.
cc @paynejacob

nickgerace (changes requested):

index.yaml
annablender: The change I'm pulling in from dev-v2.6 regenerates the index.yaml. Should I not accept the incoming change? That's what I did using the git UI

paynejacob (dismissed):
once commits are addressed, lgtm

index.yaml
annablender: Ok done

packages/rancher-logging/generated-changes/patch/templates/_helpers.tpl.patch
```diff
@@ -93,4 +93,12 @@
 +    Time_Key          timestamp
 +    Time_Format       %FT%H:%M:%S.%L
 +{{- end -}}
-\ No newline at end of file
++
++{{/*
++Set kubernetes log options if they are configured
++*/}}
++{{- define "requireFilterKubernetes" -}}
++{{- if or .Values.fluentbit.filterKubernetes.Merge_Log .Values.fluentbit.filterKubernetes.Merge_Log_Key .Values.fluentbit.filterKubernetes.Merge_Trim .Values.fluentbit.filterKubernetes.Merge_Parser}}
```
nickgerace, line 101: Looks like we either missed this or it's new, but the end of the line sprig formatter should have a whitespace and a dash to ensure we don't add blank newlines.
```
Merge_Parser -}}
```

nickgerace (changes requested):

paynejacob (dismissed):

paynejacob (approved):

nickgerace (approved):
🚀

PR #1508, opened 2021-09-28
Add instructions to separate multiple images inside values.yaml for dev-v2.6

rayandas:
This PR adds an instruction about how/when to separate images inside `packages/<package-name>/charts/values.yaml` after running `make prepare`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,31 @@ If this `major.minor.patch` (e.g. `0.0.1`) version of the Chart has never been r
 
 If this `major.minor.patch` (e.g. `0.0.1`) version of the Chart has been released before, increment the `packageVersion`.
 
+#### Separating multiple images within `charts/values.yaml`
+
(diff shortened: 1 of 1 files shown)


nickgerace (approved):
Awesome! Thank you

rayandas:
cc: @PennyScissors

aiyengar2 (changes requested):
@rayandas we can't modify this README.md today since it is currently managed by `make template`, as the source of these docs exist in upstream here: https://github.com/rancher/charts-build-scripts/tree/579e62f8a7bc7aebf61905efe7f674e0c6e9a382/templates/template.

However, I am working https://github.com/rancher/charts-build-scripts/issues/60, which will modify these docs anyways, so I can put the fix to the docs in in that PR.

rayandas:
Closing this as it's WIP in https://github.com/rancher/charts-build-scripts/issues/62

PR #1511, opened 2021-09-28
[dev-v2.6] Add max rancher version for 2.5.x charts

jakefhyde:
`dev-v2.6` port of this pr: https://github.com/rancher/charts/pull/1480


jiaqiluo (changes requested):
Can you remove those 2 new entries for longhorn from the index.yaml file? They are generated because the assets were not cleaned up properly in a pervious PR, and we will fix it later in a separate PR.

brendarearden (changes requested):
Looks good except can you revert the chmod changes on index.yaml?

jiaqiluo (dismissed):
LGTM. 
One small nit: there is no need to add the annotation to the backup-crd charts, however adding the annotation does not break anything.

jiaqiluo (approved):
The CI build fails because of the dispatch of the index.yaml file which is caused by the duplicated longhorn 100.0.0+up1.1.2 charts in two different directories.  We will fix it in another PR.

cbron:
And we won't actually merge this one right, its just a test branch/pr ?

cbron:
These charts are in dev now.

PR #1512, opened 2021-09-28
Change crdRepository into crdImage.* fields

aiyengar2:
This change is necessary since our image build scripts currently have a dependency on `values.yaml` using the labels `repository` and `tag` for all images, as shown here: https://github.com/rancher/rancher/blob/cb7de4e6c3d7e783828dc662b1142c1f9ae5edbe/pkg/image/resolve.go#L119-L120.

Therefore, a field like `crdRepository` needs to be refactored to `crdImage.repository` and `crdImage.tag` to specifically call out which image needs to be pulled in.

Related Issue: https://github.com/rancher/rancher/issues/34897


charts/rancher-gatekeeper/rancher-gatekeeper/100.0.1+up3.6.0/values.yaml
```diff
@@ -25,10 +25,13 @@ postInstall:
...
 image:
   repository: rancher/mirrored-openpolicyagent-gatekeeper
-  crdRepository: rancher/mirrored-openpolicyagent-gatekeeper-crds
   tag: v3.6.0
   pullPolicy: IfNotPresent
   pullSecrets: []
+crdImage:
+  repository: rancher/mirrored-openpolicyagent-gatekeeper-crds
+  tag: v3.6.0
+  pullPolicy: IfNotPresent
```
aiyengar2, line 34: Not a breaking change since this field never existed before:

https://github.com/rancher/charts/blob/5411430480180a486d2e067f46d0906b79938626/charts/rancher-gatekeeper/rancher-gatekeeper/100.0.0%2Bup3.5.1/values.yaml#L23-L27

PR #1516, opened 2021-09-29
Fix/rancher range

paynejacob:


jiaqiluo (changes requested):
the entry in the index.yaml still contains the old annotation. Can you delete that entry then make charts again?

jiaqiluo (approved):

nickgerace (approved):

prachidamle (approved):

PR #1519, opened 2021-09-30
[2.6.1 Blocker] Add new istio 1.10.4 image for airgap support

brendarearden:
**Issue**

**Problem**
User permissions do not work universally in all rke and rke2 environments

**Solution**
Update istio-installer image to [new image](https://github.com/rancher/istio-installer/pull/36) that uses sudo to run nginx to avoid port binding issues.


charts/rancher-istio/rancher-istio/100.0.1+up1.10.4/templates/istio-install-job.yaml
```diff
@@ -29,6 +29,10 @@ spec:
             value: {{ .Values.forceInstall | default "false" | quote }}
           - name: RELEASE_MIRROR_ENABLED
             value: {{ .Values.installer.releaseMirror.enabled | quote }}
+          - name: DEBUG
+            value: {{ .Values.installer.debug.enabled | quote }}
+          - name: SECONDS_SLEEP
```
nickgerace, line 34: Should we only provide this value based on if debug is true or false?

cbron (approved):

charts/rancher-istio/rancher-istio/100.0.1+up1.10.4/templates/istio-install-job.yaml
cbron: I’d also prefer debug to be an int that is the seconds sleep. We can address that later.

PR #1521, opened 2021-09-30
fix the index.yaml file

jiaqiluo:
Problem:
When Helm generates the index.yaml, it skips the existing ones(determined by version). As a result, all those existing charts to which we added the annotation are not updated in the index.yaml

Fix:
wipe the index.yaml and rerun the `make charts` command

### DIFF
--- a/index.yaml
+++ b/index.yaml
@@ -12,13 +12,13 @@ entries:
       catalog.cattle.io/release-name: fleet
     apiVersion: v2
     appVersion: 0.3.7
-    created: "2021-09-29T22:20:49.286746-04:00"
+    created: "2021-09-30T12:30:02.133338-07:00"
     dependencies:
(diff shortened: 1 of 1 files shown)


index.yaml
```diff
@@ -552,13 +553,14 @@ entries:
       catalog.cattle.io/namespace: longhorn-system
       catalog.cattle.io/os: linux
       catalog.cattle.io/provides-gvr: longhorn.io/v1beta1
+      catalog.cattle.io/rancher-version: < 2.6.1-0
```
jiaqiluo, line 556: To Reviewers:
The annotation is added to all applied charts. 
The rest of the changes are timestamps and digest value which we can ignore.

cbron (approved):

aiyengar2 (approved):
LGTM

PR #1522, opened 2021-10-01
[dev-v2.5] Update vSphere CPI to v1.21.0 and add support for K8s 1.18 to 1.21

PennyScissors:
Issue: https://github.com/rancher/rancher/issues/34977

1st commit backports CPI from dev-v2.6 to pull in some changes we made to fix conflicts with rke2 resources' names

2nd and 3rd commit refactors and breaks down chart templates into separate files to improve organization (following the same pattern as their upstream chart).

4th commit updates to the v1.21.0 upstream (releases/1.21 directory they added in tag v1.22.0)

6th commit implements a helper function to determine the image repository and tag corresponding to the cluster's kubernetes version line and this way add support for kubernetes 1.18 to 1.21. The pattern I used was needed in order for the rancher images export scripts to work with it since it looks for maps containing `repository` and `tag` and all versions of the image need to be available in airgap in case a user updates k8s and needs to update the chart as well.


packages/rancher-vsphere-cpi/charts/templates/daemonset.yaml
```diff
@@ -44,20 +45,30 @@ spec:
...
+        # - key: node-role.kubernetes.io/master
+        #   effect: NoSchedule
+        #   operator: Exists
+        # - key: node.kubernetes.io/not-ready
+        #   effect: NoSchedule
+        #   operator: Exists
+        - effect: NoExecute
+          operator: Exists
+        - effect: NoSchedule
+          operator: Exists
```
aiyengar2, line 62: This is the same problem we will encounter in https://github.com/rancher/rancher/issues/34751 for the CPI chart, right?

Could we either a) change this as part of this PR or b) create an issue to track the CPI issue?

packages/rancher-vsphere-cpi/charts/templates/_helpers.tpl
```diff
@@ -5,3 +5,14 @@
...
 {{- end -}}
+
+{{- /* Return the image's repository and tag corresponding to the cluster's kubernetes version. */ -}}
+{{- define "get_image" -}}
+{{- $kubeVersion := printf "%s.%s" .Capabilities.KubeVersion.Major .Capabilities.KubeVersion.Minor -}}
+{{- if hasKey .Values.cloudControllerManager.image $kubeVersion -}}
+{{- $imageForKubeVersion := get .Values.cloudControllerManager.image $kubeVersion -}}
+{{- printf "%s:%s" $imageForKubeVersion.repository $imageForKubeVersion.tag -}}
+{{- else -}}
+{{- required (printf "unsupported Kubernetes version: %s" $kubeVersion) "" -}}
```
aiyengar2, line 16: Can we add which versions are currently supported to this error message?

packages/rancher-vsphere-cpi/charts/templates/_helpers.tpl
```diff
@@ -5,3 +5,14 @@
 {{- "" -}}
 {{- end -}}
 {{- end -}}
+
+{{- /* Return the image's repository and tag corresponding to the cluster's kubernetes version. */ -}}
+{{- define "get_image" -}}
+{{- $kubeVersion := printf "%s.%s" .Capabilities.KubeVersion.Major .Capabilities.KubeVersion.Minor -}}
+{{- if hasKey .Values.cloudControllerManager.image $kubeVersion -}}
```
aiyengar2, line 12: Can we add a check to see if `.Values.cloudControllerManager.image.override.repository` or `.Values.cloudControllerManager.image.override.tag` are set here? If either is set, we should respect a user override over trying to calculate the kubeVersion. If both are set, we shouldn't even try doing the kubeVersion check.

We shouldn't expose the override field in the values.yaml (only accept it if a user provides it), but this would allow us to bake-in a solution in case a user needs to override the image + tag for some reason.

aiyengar2 (changes requested):

packages/rancher-vsphere-cpi/charts/templates/_helpers.tpl
PennyScissors: I agree the user should have the option to override it, but due to the way this chart works (image tags based on the k8s version line), I think the user should set both a repository and tag for the override to work. If they override the repository only, we would need to guess the tag based on the cluster's k8s version which I don't think is something we should do.

packages/rancher-vsphere-cpi/charts/templates/_helpers.tpl
PennyScissors: Adding it in next commit

packages/rancher-vsphere-cpi/charts/templates/daemonset.yaml
PennyScissors: Sounds like is the same thing. I'll verify if we want it fixed in 2.5 since the fix for CSI got reverted. If we do, I'll implement it for both CPI and CSI. If we don't, I'll create an issue for CPI to track it separately

packages/rancher-vsphere-cpi/charts/templates/role-binding.yaml
```diff
@@ -1,129 +1,38 @@
...
-  name: {{ .Chart.Name }}-cloud-controller-manager
-subjects:
-- kind: ServiceAccount
-  name: {{ .Chart.Name }}-cloud-controller-manager
-  namespace: {{ .Release.Namespace }}
-- kind: User
-  name: {{ .Chart.Name }}-cloud-controller-manager
+kind: List
+metadata: {}
+items:
```
jiaqiluo, line 6: Instead of putting the resources in a list, can we move the RoleBinding and ClusterRoleBinding resources at the fist-level?

jiaqiluo (changes requested):

packages/rancher-vsphere-cpi/charts/templates/role-binding.yaml
PennyScissors: They are in a list to match the upstream chart [here](https://github.com/kubernetes/cloud-provider-vsphere/blob/master/charts/vsphere-cpi/templates/role-binding.yaml). We haven't upgraded to use the chart yet, but I figured matching this pattern will make it a bit easier to work with and upgrade later. Is there a specific reason you would like this moved?

jiaqiluo: I did not see this pattern much in other charts, but I am fine if that is what that upstream does.

packages/rancher-vsphere-cpi/charts/templates/daemonset.yaml
PennyScissors: I changed the tolerations to match upstream and added the option to override.

jiaqiluo (dismissed):
LGTM

PennyScissors:
Added kubeVersion since 1.22 is not yet supported for CPI or CSI.

jiaqiluo (approved):

aiyengar2 (approved):

PR #1523, opened 2021-10-01
Add newest istio-installer image

brendarearden:
### Issue

### PR Dependency
https://github.com/rancher/istio-installer/pull/37

### Problem
Istio dashboard configmaps are not getting created with all of the necessary data because the istio-installer is not configured with the proper user permissions.

istio-installer has two flags for enabling debug, and should only have one

### Solution
Use newest istio-installer image that has the fix for user permissions

Only pass in secondsSleep to enable debug.


jiaqiluo (dismissed):
LGTM

charts/rancher-istio/rancher-istio/100.0.1+up1.10.4/values.yaml
```diff
@@ -5,16 +5,15 @@ forceInstall: false
...
   repository: rancher/istio-installer
-  tag: 1.10.4-rancher3
+  tag: 1.10.4-rancher4
   ##releaseMirror are configurations for istio upgrades. 
   ##Setting releaseMirror.enabled: true will cause istio to use bundled in images from rancher/istio-installer to perfom an upgrade - this is ideal
   ##for airgap setups. Setting releaseMirror.enabled to false means istio will call externally to github to fetch the required assets.
   releaseMirror:
     enabled: false
-  ##A flag to run a sleep command `sleep <secondsSleep>s` to allow time to exec into istio-installer pod for debugging
+  ##Set the scondsSleep to run a sleep command `sleep <secondsSleep>s` to allow time to exec into istio-installer pod for debugging
```
aiyengar2, line 14: nit: typo

charts/rancher-istio/rancher-istio/100.0.1+up1.10.4/templates/istio-install-job.yaml
```diff
@@ -29,8 +29,6 @@ spec:
             value: {{ .Values.forceInstall | default "false" | quote }}
           - name: RELEASE_MIRROR_ENABLED
             value: {{ .Values.installer.releaseMirror.enabled | quote }}
-          - name: DEBUG
```
aiyengar2, line 32: Does Istio no longer support the debug flag? Or does it assume debug if secondSleep > 0 now?

charts/rancher-istio/rancher-istio/100.0.1+up1.10.4/values.yaml
```diff
@@ -5,16 +5,15 @@ forceInstall: false
...
   debug:
-    enabled: false
-    secondsSleep: 500
+    secondsSleep: 0
```
aiyengar2, line 16: Why did we drop the default from 500 to 0?

aiyengar2 (changes requested):
While I understand that we’re trying to do an image bump for the original issue, I don’t understand the context behind why we are deprecating / removing the debug field and introducing a change to the default value of secondsSleep. Was this a change in the behavior of the istio-installer image itself?

charts/rancher-istio/rancher-istio/100.0.1+up1.10.4/templates/istio-install-job.yaml
brendarearden: It assumes debug if secondsSleep > 0

charts/rancher-istio/rancher-istio/100.0.1+up1.10.4/values.yaml
brendarearden: I had a default value when we were using a separate flag for enabled/disabled debug mode, now that I removed the flag, any amount of seconds will cause the istio-installer job to sleep for x seconds.

charts/rancher-istio/rancher-istio/100.0.1+up1.10.4/values.yaml
brendarearden: fixed typo in recent commit

charts/rancher-istio/rancher-istio/100.0.1+up1.10.4/templates/istio-install-job.yaml
aiyengar2: Gotcha. I see the change in https://github.com/rancher/istio-installer/commit/08a77173f6e13654490de3896425fb2a163ea8ac.

charts/rancher-istio/rancher-istio/100.0.1+up1.10.4/values.yaml
aiyengar2: So I think we should avoid trying to make a breaking change to the Helm chart's behavior / defaults if possible.

Instead of deprecating the `installer.debug` Helm value, could we do something like this?

```yaml
{{- if .Values.installer.debug.enabled }}
- name: SECONDS_SLEEP
   value: {{ .Values.installer.debug.secondsSleep | quote }}
{{- end }}
```
This way, we can also keep the default secondsSleep at 500 but still remove passing in the unnecessary DEBUG env var.

aiyengar2 (changes requested):

charts/rancher-istio/rancher-istio/100.0.1+up1.10.4/values.yaml
cbron: This is all brand new for 2.6.1/2.5.10. Won't be a breaking change. 
We've gone back and forth between two vars or one, me and Nick liked a single var. It doesn't matter so much as long as it works somehow.

cbron (approved):

charts/rancher-istio/rancher-istio/100.0.1+up1.10.4/values.yaml
brendarearden: [100.0.0+up1.10.4](https://github.com/rancher/charts/tree/release-v2.6/charts/rancher-istio/rancher-istio) is the current released version of Istio, so Caleb is right, the debug flag does not exist in production at this time so it is not a breaking change to make adjustments to how it works. My vote is to leave it as a single var.

aiyengar2 (approved):
LGTM

charts/rancher-istio/rancher-istio/100.0.1+up1.10.4/values.yaml
aiyengar2: sounds good to me 👍

PR #1524, opened 2021-10-04
Add longhorn 1.2.1 for cluster manager

innobead:

### DIFF
--- a/charts/longhorn/v1.2.0/.helmignore
+++ b/charts/longhorn/v1.2.0/.helmignore
@@ -1,0 +1,21 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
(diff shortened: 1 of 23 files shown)


khushboo-rancher:
Verified with the branch, there are some missing setting on the app page.

1. Namespace is appearing as longhorn and is editable field whereas it should be longhorn-system

With the Branch
Screen Shot 2021-10-05 at 12 07 10 PM

Released version
Screen Shot 2021-10-05 at 12 10 54 PM

2. The setting Longhorn images, Private registry Settings, Longhorn Default Settings, Longhorn StorageClass Settings, Services And LoadBalancing, and Other Settings are not available.
Screen Shot 2021-10-05 at 12 07 56 PM

meldafrawi:
Validated installation/upgrade on rancher v2.5.9, LGTM
Ref: https://github.com/longhorn/longhorn/issues/3090

PR #1525, opened 2021-10-04
[dev-2.5] Change CSI tolerations to match upstream and allow override

PennyScissors:
Issue: https://github.com/rancher/rancher/issues/34751

This PR is basically the same as [this PR](https://github.com/rancher/charts/pull/1473), but it was reverted in 2.6.1 so we are adding it in 2.5.10 and forward porting it instead. It also adds a kubeVersion to match their compatibility matrix [here](https://vsphere-csi-driver.sigs.k8s.io/compatiblity_matrix.html#compatibility-matrix-for-vsphere-csi-driver).


PennyScissors:
This PR is the same thing as [this](https://github.com/rancher/charts/pull/1473), but it was reverted in 2.6.1 so we are adding it in 2.5.10 and forward porting it instead.

packages/rancher-vsphere-csi/charts/templates/controller/deployment.yaml
```diff
@@ -39,25 +39,17 @@ spec:
...
+      {{- with .Values.csiController.tolerations }}
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- else }}
+      tolerations:
+        - key: node-role.kubernetes.io/master
+          operator: Exists
           effect: NoSchedule
-        - operator: "Exists"
-          effect: NoExecute
```
jiaqiluo, line 50: Why is this one dropped?

jiaqiluo (changes requested):

packages/rancher-vsphere-csi/charts/templates/controller/deployment.yaml
PennyScissors: To match what's [upstream](https://github.com/kubernetes-sigs/vsphere-csi-driver/blob/e878840857838f84e89784823cdbbd5d6b1c853a/manifests/vanilla/vsphere-csi-driver.yaml#L205-L218)

jiaqiluo: It makes sense to me now.

jiaqiluo (approved):

packages/rancher-vsphere-csi/charts/values.yaml
```diff
@@ -70,6 +81,7 @@ csiNode:
     livenessProbe:
       repository: rancher/mirrored-k8scsi-livenessprobe
       tag: v2.2.0
+  tolerations: {}
```
aiyengar2, line 84: why did we introduce tolerations here?

aiyengar2 (changes requested):

packages/rancher-vsphere-csi/charts/values.yaml
PennyScissors: I left that in there by mistake, I'll remove it.

aiyengar2 (approved):

jiaqiluo (approved):

PR #1526, opened 2021-10-05
[2.5] Backport istio spoof

brendarearden:
### Issue
https://github.com/rancher/rancher/issues/35009

### Solution
I did not cherry-pick specific changes, but I compared the 1.9.8 chart with the 1.10.4 chart from dev-v2.6 and moved the changes over from the following PRs

https://github.com/rancher/charts/pull/1523
https://github.com/rancher/charts/pull/1519
https://github.com/rancher/charts/pull/1485


packages/rancher-istio-1.9/package.yaml
```diff
@@ -1,2 +1,2 @@
 url: local
-packageVersion: 0
+packageVersion: 1
```
cbron, line 2: Needs update here: https://github.com/rancher/rancher/issues/33528

cbron (approved):
One thing, otherwise good

packages/rancher-istio-1.9/package.yaml
brendarearden: I have updated the issue - it now includes the OOB release info, and this new release info.

PR #1527, opened 2021-10-05
revert to last known working fluentbit image

paynejacob:
https://github.com/rancher/rancher/issues/34625
https://github.com/rancher/rancher/issues/34772

PROBLEM:
the newer fluentbit image seems to  be having issues resolving dns.  The issue is intermittent and can be difficult to reproduce

SOLUTION:
reverted to last working fluentbit image.  The newer versions contain features we are not using and performance enhancements.

TESTING:
install the chart on an rke cluster running 1.20.11 and observed no error messages.


nickgerace (approved):
_sigh_, LGTM

charts/rancher-logging/rancher-logging/3.14.0/values.yaml
```diff
@@ -147,10 +147,10 @@ images:
     tag: v0.4.0
   fluentbit:
     repository: rancher/mirrored-fluent-fluent-bit
-    tag: 1.8.7
+    tag: 1.6.10
```
cmurphy, line 150: This seems like quite a step backward, did nothing in the 1.8.x series work? What actually is the issue? Is it this https://github.com/fluent/fluent-bit/issues/4050 ?

paynejacob: I tried a few different 1.8.x version and they all have the same issue.  I don't see it as a step backwards, it is more of opting out of an upgrade.  1.6.10 is what we are currently using in rancher logging.

cmurphy: Looks like it might be fixed in the next release https://github.com/fluent/fluent-bit/issues/4050#issuecomment-934803218 if we can't wait for that can we make sure there's a followup issue to upgrade when we can?

cmurphy (approved):

charts/rancher-logging/rancher-logging/3.14.0/values.yaml
paynejacob: https://github.com/rancher/rancher/issues/35058

PR #1529, opened 2021-10-06
[2.5] Fix CPI missing system prefix in role binding

PennyScissors:
Issue: https://github.com/rancher/rancher/issues/34977


aiyengar2 (approved):
LGTM

PR #1531, opened 2021-10-06
Logging to require Rancher 2.5.11+

paynejacob:
There is a bug in the upstream version that we expect to be fixed in time for 2.5.11.  This pr will delay this version of logging to 2.5.11+ to allow time for the fix to be introduced.


charts/rancher-logging/rancher-logging-crd/3.14.0/Chart.yaml
```diff
@@ -2,6 +2,7 @@ annotations:
   catalog.cattle.io/certified: rancher
```
cbron, line 2: @paynejacob why not just delete it and then revert the PR later ? Releasing a chart early is a bit off, there will be confusion on if we can edit it or if we need to make a 3.14.001

paynejacob:
closing in favor of https://github.com/rancher/charts/pull/1539

PR #1532, opened 2021-10-06
2.6.3 - Add volume mount support for additional istio mirror

brendarearden:
### Issue
https://github.com/rancher/rancher/issues/34858

### Problem
When adding support for istio airgapped upgrades, the option for users to provider their own image was not supported. This support is needed in case the istio version a user is on is not bundled in with the [rancher/istio-installer ](https://github.com/rancher/istio-installer/blob/56b5590819a85b0d14b868f916393d37d946dadf/scripts/fetch_istio_releases.sh#L9). The image needs to be available to the installer to support airgapped upgrades. 

### Solution
Add a volume mount to mount the istio tar file in the correct directory for the nginx spoof to serve it for the upgrade to complete successfully. Since there are a wide rang of volume mounts, the volumes are fully available for editing in the values.yaml


jakefhyde (approved):
LGTM

packages/rancher-istio/charts/templates/istio-install-job.yaml
```diff
@@ -51,6 +56,9 @@ spec:
           configMap:
             name: istio-installer-overlay
         {{- end }}
+{{- if and .Values.installer.releaseMirror.additionalIstioImage.enabled .Values.installer.releaseMirror.additionalIstioImage.volumes }}
```
jiaqiluo, line 59: I think we do not need to check if `.Values.installer.releaseMirror.additionalIstioImage.volumes` exists,  it must exist since it is predefined in the value.yaml file.

In line 45, we do not check it.

jiaqiluo (changes requested):

PR #1533, opened 2021-10-07
Add instructions to separate multiple images inside values.yaml for dev-v2.5

rayandas:
This PR adds an instruction about how/when to separate images inside packages/<package-name>/charts/values.yaml after running make prepare.
Similar to https://github.com/rancher/charts/pull/1508


nickgerace (changes requested):
https://github.com/rancher/charts/pull/1508#pullrequestreview-774272747
cc @aiyengar2

PR #1538, opened 2021-10-07
Add longhorn 1.2.2 chart dev 2.5

innobead:

As discussed with @cbron , we would like to replace 1.2.0 with 1.2.2 in the coming 2.5.10.


charts/longhorn-1.2/longhorn-crd/1.2.200+up1.2.2/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/hidden: "true"
+  catalog.cattle.io/namespace: longhorn-system
+  catalog.cattle.io/release-name: longhorn-crd
+apiVersion: v1
+description: Installs the CRDs for longhorn.
+name: longhorn-crd
+type: application
+version: 1.2.200+up1.2.2
```
shuo-wu, line 10: Can anyone explain why the version prefix `1.2.200+up` is generated here but not in v1.2.0?

shuo-wu (commented):
Others LGTM

charts/longhorn-1.2/longhorn-crd/1.2.200+up1.2.2/Chart.yaml
shuo-wu: But I still don't know why there is a (upstream) package version generated when the Longhorn patch version is not 0...

PR #1539, opened 2021-10-07
Logging to require Rancher 2.5.11+

paynejacob:
There is a bug in the upstream version that we expect to be fixed in time for 2.5.11. This pr will delay this version of logging to 2.5.11+ to allow time for the fix to be introduced.


packages/rancher-logging/templates/crd-template/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-version: 3.14.0
```
cbron, line 2: I still see this version in the assets and charts folders. Assuming its still in the index as well.

cbron: It also still in the 'new charts' issue

paynejacob:
I thought about that, I think this is better incase we need to do an emergency release or the 2.5.11 scope changes

PR #1543, opened 2021-10-08
[2.5.10] bump the rancher-webhook chart to 0.1.2

jiaqiluo:
issue https://github.com/rancher/rancher/issues/33528


brendarearden (approved):
This appears to only be removing the RC tag, so it looks good to me!

PR #1549, opened 2021-10-12
removed istio 1.9.801

paynejacob:
revert istio 1.9.801


packages/rancher-istio-1.9/package.yaml
```diff
@@ -1,2 +1,4 @@
 url: local
 packageVersion: 1
+doNotRelease: true
```
thedadams, line 3: Didn't know this existed. Nice.

PR #1560, opened 2021-10-19
Conditional constraint templates

suryatejaboorlu:


aiyengar2 (changes requested):
This PR directly tries to add a chart, but the expectation in this repository is that we update the packages in https://github.com/rancher/charts/tree/dev-v2.6/packages/rancher-gatekeeper and run `make charts` to generate the appropriate chart, asset, and index.yaml changes.

See https://github.com/rancher/charts/tree/dev-v2.6/packages#common-workflow for more information on packages and the expected developer workflow.

PR #1567, opened 2021-10-20
Rebase Monitoring v2 on upstream kube-prometheus-stack chart 19.0.3

jiaqiluo:
issue https://github.com/rancher/rancher/issues/34577

chart name | upstream chart name | upstream chart version | note
-----------|------------------------|---------------------|------
rancher-monitoring | kube-prometheus-stack | 19.0.3|  crd-chart template is also updated 
rancher-grafana | grafana | 6.16.14| grafana image tag is updated to `7.5.11`, not the one in the upsream chart; internal-chart, not release  
rancher-kube-state-metrics | kube-state-metrics | 3.5.2  | internal-chart, not release 
rancher-prometheus-adapter | prometheus-adapter | 2.17.0 | internal-chart, not release 
rancher-node-exporter | prometheus-node-exporter | 2.2.0 | internal-chart, not release 

To reviewers:
1. The commits are made in an organized and clear way. I strongly recommend reviewing commits one by one to ease your reviewing process. 
2. Basic validation has been done on a k8s 1.21 cluster in the latest rancher v2.6-head:
	- the chart: installation and uninstallation succeed 
	- all components are installed and working 
	- Rancher Cluster Explorer UI: Metrics show up
	- Grafana: all dashboards present 
	- Prometheus: all targets are up
3. Installation and upgrading has been tested on RKE, K3s, hardened RKE cluster 

--------
This PR also includes the fixed for the following issues:
- https://github.com/rancher/rancher/issues/35264
- https://github.com/rancher/rancher/issues/35204
- https://github.com/rancher/rancher/issues/35217
- https://github.com/rancher/rancher/issues/35390
- https://github.com/rancher/rancher/issues/35288
- https://github.com/rancher/rancher/issues/34439
- https://github.com/rancher/rancher/issues/35571
- https://github.com/rancher/rancher/issues/35595
- https://github.com/rancher/rancher/issues/35222


jiaqiluo:
The CI fails because the chart-build-script does not contain the latest changes for supporting the changes in the rancher-monitoring-crd chart.
here is the related PR: https://github.com/rancher/charts-build-scripts/pull/64

jiaqiluo:
**kube-prometheus-stack**

> Based on docs on kube-prometheus-stack, it seems like kube-state-metrics deprecated kube-state-metrics.serviceMonitor.namespaceOverride in favor of a subchart-wide kube-state-metrics.namespaceOverride. Can we ensure that we add something to alert the user if kube-state-metrics.serviceMonitor.namespaceOverride is set? e.g. we could either fail install on seeing that that value is set with an error message that says that "kube-state-metrics.serviceMonitor.namespaceOverride has been deprecated, please ensure that you set kube-state-metrics.namespaceOverride instead."...

A check is added as suggested. 

> kube-state-metrics introduced a breaking change in the fact that collectors now takes in a list instead of a map. Can we ensure that we fail if kube-state-metrics.collectors is a map with an appropriate error?

A check is added as suggested. 

> If you set any of the <subchart>.namespaceOverrides and ignoreNamespaceSelectors is set to true on the Prometheus spec, will we fail to scrape resources? If so, we might need to introduce a check that fails if namespaceOverrides are set but ignoreNamespaceSelectors is set to true (unless the serviceMonitors installed by our chart are disabled). What do you think about this?

This is not a problem in the rancher-monitoring chart: at least from what I find, there is no `serviceMonitor` or `podMonitor` in the sub-charts that are selecting a namespace that is different from where the sub-chart is installed. 


> I think we should call out to QA that we need to check if kubelet monitoring works for all supported cluster types that Monitoring V2 is installed on since the selector argument has been modifed in templates/exporters/kubelet/servicemonitor.yaml; what do you think? Would this be necessary?

Action item: 
- [ ] tell QA to check the kubelet metrics on all supported cluster types 

> templates/grafana/dashboards-1.14/alertmanager-overview.yaml: should this only be deployed if Alertmanager is enabled in the cluster and a servicemonitor to collect alertmanager metrics is enabled? Have you also manually checked that this new dashboard works with no additional changes?

A check is added as suggested. Confirmed that the dashbaord works.

> In general, we need to ensure QA does validation of the dashboard contents of Kubernetes Components for RKE2 / k3s clusters. Caught the following errors, but more may be uncaught even post-review since there are a LOT of changes :
	- templates/grafana/dashboards-1.14/cluster-total.yaml: I believe label_values(up{job=\"kubelet\", metrics_path=\"/metrics/cadvisor\"}, cluster) won't work for RKE2 clusters since the job name is different. You will need to use {{ include "exporter.kubelet.jobName" . }}; see the kubelet.yaml dashboard for more examples.
	- templates/grafana/dashboards-1.14/controller-manager.yaml: same as above for label_values(up{job=\"kube-controller-manager\"}, cluster) and label_values(up{cluster=\"$cluster\", job=\"kube-controller-manager\"}, instance)
	- templates/grafana/dashboards-1.14/k8s-resources-cluster.yaml: same as above label_values(up{job=\"kubelet\", metrics_path=\"/metrics/cadvisor\"}, cluster)
	- templates/grafana/dashboards-1.14/namespace-by-pod.yaml: same as above label_values(up{job=\"kubelet\", metrics_path=\"/metrics/cadvisor\"}, cluster)
	- templates/grafana/dashboards-1.14/namespace-by-workload.yaml: same as above label_values(up{job=\"kubelet\", metrics_path=\"/metrics/cadvisor\"}, cluster)
	- templates/grafana/dashboards-1.14/pod-total.yaml : same as above label_values(up{job=\"kubelet\", metrics_path=\"/metrics/cadvisor\"}, cluster)
	- templates/grafana/dashboards-1.14/scheduler.yaml: same as above label_values(up{job=\"kube-scheduler\"}, cluster), need to use {{ include "exporter.kubeScheduler.jobName" . }} for job

Fixed.

> nit: For the changes in templates/prometheus/rules-1.14/alertmanager.rules.yaml, my concern here is that upstream seems to have switched from explaining what the rule does in the message annotation to description annotation; so similarly, on creating a rule in the Rancher UI today, are both message and description treated as special annotations (outside of the extra annotations that can additionally be written in) or do we just currently bubble up message as a suggested field to fill out?
	- same for templates/prometheus/rules-1.14/general.rules.yaml
	- same for templates/prometheus/rules-1.14/node-exporter.yaml

Confirmed that Rancher UI can handled them properly. 

> nit: For the changes in templates/prometheus/rules-1.14/alertmanager.rules.yaml, do the runbook URL links even work? I feel like we should check whether the runbook URL is set and, if it isn't working, ensure Helm does not even render those annotations.
	- same for templates/prometheus/rules-1.14/general.rules.yaml
	- same for templates/prometheus/rules-1.14/kube-state-metrics.yaml
	- same for templates/prometheus/rules-1.14/node-exporter.yaml
	- same for templates/prometheus/rules-1.14/prometheus.yaml

I feel we should keep those links since the upstream is keeping adding entries to the linked websites.

> templates/prometheus-operator/deployment.yaml: needs private registry support for -thanos-default-base-image=, currently forgot to add template in front of image repo and tag.
	- also, quay.io/thanos/thanos needs to be mirrored.
	- We should also create a ticket to validate Thanos support for Monitoring V2, since it seems like upstream has added a lot of support for it now... shouldn't be done as part of this rebase but should be done in upcoming releases ideally.

The image is mirrowed now. 

Action item:
- [ ] QA tests thanos

> Was the change from rancher/mirrored-jettech-kube-webhook-certgen:v1.5.2 to rancher/mirrored-ingress-nginx-kube-webhook-certgen:v1.0 intentional? Just wanted to make sure it wasn't an accidental copy / paste

Yes, upstream changes the repo:  https://github.com/prometheus-community/helm-charts/commit/f15adcdcfe6bee8eea6afab5a96e5fa5b415b2d0

**Grafana:**

> Grafana made a breaking change in grafana/helm-charts@2bda3e9. First, we need to ensure that this breaking change has been backported to the 7.x.x release line since this may be a Grafana 8.x.x only feature.
	- If 7.x.x does support this, can you patch the Helm chart to ensure that we can still pass in a string for all the *.searchNamespaces so that existing customers aren't broken on a helm upgrade? Also, can you file the docs issue or make the docs change to specify that we support passing in multiple namespaces now as a list?
	- If 7.x.x does not support this, can you patch the README.md on the Grafana Helm chart to specifically indicate this? Perhaps we may need to release note / doc this as well, but I'll leave that call up to you :)

All those 3 String-to-List changes(`sidecar.dashboards.searchNamespace`, `sidecar.datasources.searchNamespace`, `sidecar.notifiers.searchNamespace`) are NOT breaking changes. 
In the case of upgrading from an exiting installing to the new version, the String value will be handled properly by the chart. 
 

> nit: let's patch grafana/helm-charts@bf70570/charts/grafana/values.yaml#L192

Fixed. 

> Seems like Grafana added an option to specify whether or not you want to set automountServiceAccountToken in the _pod.tpl: grafana/helm-charts@b6015e2. I'm not sure if supporting either option would require changes in our PSP; can you ensure that you can deploy and pass a CIS scan in a hardened cluster if either option is set to true?

The default value in the chart is `true` which is also the default of the k8s cluster, so it should NOT affect anything. 
Action Item 2: test Grafana in PSP enabled cluster for the new default value `serviceAccount.autoMount: true` 

> For the change for the PSP that adds in the RunAsAny rule instead of MustRunAsNonRoot, I believe we have a Rancher issue for this since it was brought up by someone in FE / Support. Can we attach that issue to this PR to also be tested?

Action Item 3: 
- [ ]  find the GH issue 

>  Can we disable Grafana test resource by default, as discussed offline?

Fixed: set the default to be false.


**prometheus-adapter**

> Seems like they introduced a setting to specify .podSecurityPolicy. Can we ensure the PSP we have supports any arbitrary values provided in that field? Also, the statement in the values.yaml that says "k8s 1.21 needs fsGroup to be set for non root deployments" seems to imply that we need to release note this / explicitly test this out to see what we need to recommend. Perhaps we can hack a fix for this into the Helm chart itself, if a user does not override .podSecurityPolicy manually and we detect the k8s version is 1.21+?

It is backward compatible. 

**prometheus-node-exporter**

> Same comment as above in grafana re: automountServiceAccountToken

not a breaking change but moving the original value to the new variable

> Metrics port name can now be overridden with .Values.service.portName; can we double check that setting an arbitrary string here does not break Monitoring V2?

Confirmed that it is back-compatible 

> .Values.hostPID can now be set. Will this break our existing PSP or is there something we have to add for it?

No, it does not break anything. The existing PSP has `hostPID: true`

aiyengar2 (changes requested):
nit: need to run make charts

Still need to do full review after this

brendarearden (dismissed):
Looks like Arvind's nits/concerns were addressed, all images appear to be using the system_default_registry and are mirrored. Looks like the make charts needs to be redone, otherwise it looks good to me.

jiaqiluo:
> nit: need to run make charts
> 
> Still need to do full review after this

The PR is rebased to match the head of the `dev-v2.6` branch. The build still fails because the chart and asset are made by using the dev version of the chart-build-scripts, so the failure is expected.

aiyengar2:
https://github.com/rancher/rancher/issues/35288

packages/rancher-monitoring/templates/crd-template/templates/jobs.yaml
```diff
@@ -49,8 +49,12 @@ spec:
           - -c
           - >
             echo "Applying CRDs...";
-            kubectl apply -f /etc/config/crd-manifest.yaml;
-            
+            mkdir -p /etc/crd-manifest;
+            base64 -d /etc/config/crd-manifest.tgz > /etc/crd-manifest/crd-manifest.tgz;
+            tar -xzvf /etc/crd-manifest/crd-manifest.tgz -C /etc/crd-manifest;
+            rm /etc/crd-manifest/crd-manifest.tgz;
+            kubectl apply -Rf /etc/crd-manifest;
```
aiyengar2, line 56: can we just replace this with:

```bash
base64 -d /etc/config/crd-manifest.tgz | tar -xvzfO | kubectl apply -
```

packages/rancher-monitoring/templates/crd-template/templates/manifest.yaml
```diff
@@ -4,11 +4,5 @@ metadata:
...
 data:
-  crd-manifest.yaml: |
-    {{- $currentScope := . -}}
-    {{- $crds := (.Files.Glob  "crd-manifest/**.yaml") -}}
-    {{- range $path, $_ :=  $crds -}}
-    {{- with $currentScope -}}
-    {{ .Files.Get $path | nindent 4 }}
-    ---
-    {{- end -}}{{- end -}}
+  crd-manifest.tgz:
```
aiyengar2, line 7: `crd-manifest.tgz` should maybe be renamed `crd-manifest.tgz.b64`

packages/rancher-monitoring/templates/crd-template/templates/jobs.yaml
```diff
@@ -121,7 +125,12 @@ spec:
           - /bin/sh
           - -c
           - >
-            kubectl delete -f /etc/config/crd-manifest.yaml
+            echo "Deleting CRDs...";
+            mkdir -p /etc/crd-manifest;
+            base64 -d /etc/config/crd-manifest.tgz > /etc/crd-manifest/crd-manifest.tgz;
+            tar -xzvf /etc/crd-manifest/crd-manifest.tgz -C /etc/crd-manifest;
+            rm /etc/crd-manifest/crd-manifest.tgz;
+            kubectl delete -Rf /etc/crd-manifest;
```
aiyengar2, line 139: same comment as above

packages/rancher-monitoring/templates/crd-template/templates/jobs.yaml
```diff
@@ -49,8 +49,12 @@ spec:
           - -c
           - >
             echo "Applying CRDs...";
-            kubectl apply -f /etc/config/crd-manifest.yaml;
-            
+            mkdir -p /etc/crd-manifest;
```
aiyengar2, line 52: note: the fact that the initContainers still use the YAML files in `crd-manifest` means that this CRD chart today relies on the fact that the contents of `crd-manifest.tgz` exactly match that which was unfurled in the `crd-manifest/` directory.

Should keep a note of this in the future, I believe this might have been what you were alluding to in https://github.com/rancher/charts-build-scripts/pull/64#discussion_r741631747?

aiyengar2: no resolution needed for this comment, you can just resolve it with no changes.

aiyengar2 (changes requested):
main chart looks good, CRD chart might need some changes. We could either proceed with reviewing the CRD chart within this PR or spin it off into a new PR? either way works.

packages/rancher-monitoring/templates/crd-template/templates/jobs.yaml
aiyengar2: what does `-Rf` do in the kubectl commmand?

packages/rancher-monitoring/templates/crd-template/templates/manifest.yaml
jiaqiluo: fixed.

packages/rancher-monitoring/templates/crd-template/templates/jobs.yaml
jiaqiluo: The section is updated to reduce unnecessary steps, but we can achieve it in one pipeline because there are directory struct information in the output of `tar` command that makes  `kubectl` fail.

aiyengar2 (approved):
LGTM :shipit:

PR #1570, opened 2021-10-20
Backport to dev rancher-istio 100.0.2+up1.10.4

brendarearden:
### Problem
rancher-istio was OOB released and needed to be backported to the dev branch https://github.com/rancher/charts/pull/1563

It would appear that rancher-istio 100.0.2+up1.10.4 was merged back into the dev-v2.6 branch here https://github.com/rancher/charts/commit/0622f0482b04a9ea0a2efe45bbadea7951f845e7. This only included porting back the charts and assets, so the packages/rancher-istio needed updated to match the currently deployed versions. The release branch also has different file modes, so those had to be adjusted as well.


jakefhyde (approved):
LGTM

PR #1571, opened 2021-10-20
Conditional constraint template

suryatejaboorlu:


aiyengar2:
@suryatejaboorlu does this PR need to be recreated with a base pointing to `dev-v2.5`? `dev-v2.5.9` was a temporary branch for the 2.5.9 release that is no longer in use.

recena:
Closing due to the lack of activity.

PR #1587, opened 2021-10-27
Adding fluentd and fluentbit monitoring into helm template

rpenugonda:
can we have something like this so that it has hooks into installing monitoring metrics

### DIFF
--- a/charts/rancher-logging/rancher-logging/3.9.400/templates/loggings/root/logging.yaml
+++ b/charts/rancher-logging/rancher-logging/3.9.400/templates/loggings/root/logging.yaml
@@ -50,6 +50,9 @@ spec:
     metrics:
       serviceMonitor: true
   {{- end }}
+    positiondb:
+      hostPath:
+        path: /var/lib/fluent-bit
     image:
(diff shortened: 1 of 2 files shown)


aiyengar2:
@rpenugonda we don't directly merge fixes to release-v2.6. Please make the PR to dev-v2.6 by modifying the appropriate `Package`; please see https://github.com/rancher/charts/blob/dev-v2.6/packages/README.md#common-workflow for more information on how to contribute!

tjanson:
This still seems very desirable to me -- being able to enable the service monitor for metrics if desired should definitely be an option, right? Would you accept this change if the PR were cleaned up? (@aiyengar2 @Jono-SUSE-Rancher ?)

recena:
Closing since `2.5.x` is not supported.

tjanson:
Just for the record, this same issue still applies to the current latest `dev-v2.9`: https://github.com/rancher/charts/blob/baf95653c8649e7f4ba7e112c5ba32b423fd26c0/charts/rancher-logging/104.0.0%2Bup3.17.10/templates/loggings/root/logging.yaml

PR #1588, opened 2021-10-28
Conditional constraint template

suryatejaboorlu:


packages/rancher-gatekeeper/generated-changes/overlay/templates/allowedrepos.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.installDefaultConstraintTemplates}}
```
aiyengar2, line 1: nit: add space to the end of all of these Helm blocks. e.g.

```suggestion
{{- if .Values.installDefaultConstraintTemplates }}
```

aiyengar2 (changes requested):
nit review comments.

Let's also refactor these commits to the following
1. "Bump rancher-gatekeeper to 100.0.2": add package.yaml changes
2. "Allow users to disable default constraint templates": add generated-changes changes
3. "make charts": add all change to charts/, assets/ and index.yaml

Once we refactor to these commit and make the nit changes to the chart shown below, this PR should be good to go.

Will need a second review from a chart SME before merge, added @cmurphy for that review.

nicholasSUSE:
Closed due to inactivity and lack of labels

PR #1599, opened 2021-11-08
Backport 1.10.4

brendarearden:
### Issue
https://github.com/rancher/rancher/issues/33824
https://github.com/rancher/rancher/issues/35341
https://github.com/rancher/rancher/issues/35009

### Problem
A version of 1.10.x needed to be backported to dev-v2.5. This version also needed to have the fix for k8s server ip issue and include the air-gapped upgrade fixes.

### Solution
Backporting the 1.10.4 version of istio that is in dev-v2.6, but changing the istio installer version to have the fix for k8s server ip issue. So this is not the exact same chart that is from dev.2.6 because of the istio-installer version change and the rancher-version annotation change, so I did do a chart version bump. 

dev-v2.6 version: [100.0.2+up1.10.4](https://github.com/rancher/charts/tree/release-v2.6/charts/rancher-istio/rancher-istio/100.0.2%2Bup1.10.4)

new version for dev-v2.5: [100.0.3+up1.10.4](https://github.com/rancher/charts/pull/1599/commits/554a1eca71ffe4168c29cbd5aadb917656607d63#diff-161ea49cd45f8da1bce053c75d86279f95b016f538f124496e7028282eae7a73R2)


packages/rancher-istio/package.yaml
```diff
@@ -1,3 +1,2 @@
 url: local
-packageVersion: 0
-doNotRelease: true
\ No newline at end of file
+version: 100.0.3+up1.10.4
```
jiaqiluo, line 2: 1. This chart is removed from dev-2.6 in this [PR](https://github.com/rancher/charts/pull/1600#pullrequestreview-800660559), so it is not a backporting anymore, right? 
2. Do we decide to use the `100.x.x` version schema for Istio in the v2.5 line?

jiaqiluo (changes requested):

packages/rancher-istio/package.yaml
brendarearden: 1. You are correct, because the version of 1.10.4 in the dev-v2.6 branch has the kiali server api bug in it, we did not want to back port it as is. We still wanted to have 1.10.4 in the dev-v2.5 branch, but with an istio-installer version that did not have the kiali server api issue.
2. We did not decide - I had failed to change to the version back to the v2.5 line versioning, will update PR and re-request review.

samkulkarni20 (dismissed):
LGTM

doflamingo721 (approved):
Apart from the version schema that @jiaqiluo mentioned, the rest looks good to me.

jiaqiluo (approved):
LGTM

PR #1603, opened 2021-11-09
[release-v2.5] Hide CPI 1.0.000 and CSI 2.1.000

PennyScissors:
Issue: https://github.com/rancher/rancher/issues/35378

Hides the following charts in the 2.5 line:
CPI 1.0.000 -> because we want to encourage users to upgrade to 1.0.100 since 1.0.0 doesn't use the correct image tag for each k8s version line.
CSI 2.1.000 -> because we needed to set a kubeVersion which cannot be done in place so we introduced CSI 2.1.1 (2.1.1 is a copy of 2.1.000 with a kubeVersion meant to replace 2.1.000).


jiaqiluo (approved):
LGTM 
**Caution: This PR goes to the release-v2.5 branch.**
We need to hold it until new charts are ready to be released.

PennyScissors:
Changing to draft to prevent merging unintentionally

PR #1609, opened 2021-11-11
Hide existing non-partner charts in "library" catalog

rayandas:
Fixes: https://github.com/rancher/rancher/issues/35347

### DIFF
--- a/charts/argo-cd/v2.3.5/questions.yaml
+++ b/charts/argo-cd/v2.3.5/questions.yaml
@@ -1,3 +1,4 @@
+rancher_max_version: 2.6.2
 questions: 
 # Openshift
 - variable: openshift.enabled
--- a/charts/chartmuseum/v2.3.1/questions.yml
+++ b/charts/chartmuseum/v2.3.1/questions.yml
@@ -1,3 +1,4 @@
+rancher_max_version: 2.6.2
 labels:
   io.cattle.role: project # options are cluster/project
 questions:
--- a/charts/cost-analyzer/v1.57.2/questions.yaml
+++ b/charts/cost-analyzer/v1.57.2/questions.yaml
@@ -1,3 +1,4 @@
+rancher_max_version: 2.6.2
 labels:
   io.cattle.role: cluster
 questions:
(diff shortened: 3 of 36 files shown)


PennyScissors (changes requested):
This repository only contains Helm 2 charts which aren't compatible in the new UI. For that reason, you don't need to set the rancher-version annotation, instead add the rancher min/max to the questions.yaml. If a chart doesn't have a questions.yaml file (some really old charts may not have one), you need to create a questions.yaml file and add the min/max to it in order to hide it. This PR should only contain modifications to question.y*ml files, and all charts should have either a modified or newly added questions.y*ml.

rayandas:
@PennyScissors Got it. I will update this PR today with the suggested changes.

charts/longhorn/latest/questions.yml
```diff
@@ -1,3 +1,4 @@
+rancher_max_version: 2.6.2
```
deniseschannon, line 1: We should allow Longhorn as this is one of the actively maintained charts by SUSE/Rancher.

deniseschannon (changes requested):

charts/longhorn/latest/questions.yml
rayandas: alright. I'll remove the rancher_max_version from longhorn.

PennyScissors (changes requested):
The following questions files are missing a rancher_max_version.
```
./charts/etcd-operator/v0.9.0/questions.yml
./charts/vault-operator/v0.1.3/questions.yml
./charts/datadog/v1.33.0/questions.yml
```

deniseschannon:
@rayandas When I add your branch into a v2.6.3-rc1, I see lots of non-partner charts still:

docker-registry
etcd-operator
fluentd-aggregator
grafana
harbor-scanner-trivvy
kubeflow
memcached
openfaas
prometheus
redis

rayandas:
@deniseschannon apologies I missed few of them. I just pushed the fix. these charts shouldn't be visible anymore.
/cc: @PennyScissors

deniseschannon (approved):

deniseschannon (approved):

PR #1610, opened 2021-11-11
Hide existing partner charts in "library" catalog

dhruvmewada15:
Linked issues:
rancher/rancher#35352


PennyScissors (changes requested):
This repository only contains Helm 2 charts which aren't compatible in the new UI. For that reason, you don't need to set the rancher-version annotation, instead add the rancher min/max to the questions.yaml. If a chart doesn't have a questions.yaml file (some really old charts may not have one), you need to create a questions.yaml file and add the min/max to it in order to hide it. This PR should only contain modifications to question.y*ml files, and all charts should have either a modified or newly added questions.y*ml.

PennyScissors (approved):

PR #1622, opened 2021-11-18
security-scan version bump to fix rancher/security-scan#62

paynejacob:
adding changes for https://github.com/rancher/security-scan/issues/62


snasovich (approved):
Changes match some previous PRs where versions got bumped, so should be good.

prachidamle (approved):

PR #1625, opened 2021-11-19
Bump KEv2 operators

cmurphy:
https://github.com/rancher/rancher/issues/35485


paynejacob (dismissed):

packages/rancher-aks-operator-crd/package.yaml
```diff
@@ -1,2 +1,2 @@
-url: https://github.com/rancher/aks-operator/releases/download/v1.0.2/rancher-aks-operator-crd-1.0.2.tgz
+url: https://github.com/rancher/aks-operator/releases/download/v1.0.3-rc1/rancher-aks-operator-crd-1.0.3-rc1.tgz
 version: 100.0.1
```
thedadams, line 2: I think this version should be bumped as well to stay inline with the aks-operator chart.

packages/rancher-gke-operator/charts/Chart.yaml
```diff
@@ -0,0 +1,18 @@
+annotations:
```
thedadams, line 1: Do we want a copy of this chart in the `packages/` directory?

thedadams (changes requested):

packages/rancher-gke-operator/charts/Chart.yaml
cmurphy: :raised_eyebrow: not sure how this ended up here

packages/rancher-aks-operator-crd/package.yaml
paynejacob: that version is typically for the changes specific to the charts repo, if we bump upstream it will be added as  tag to the semver.

thedadams (approved):

paynejacob (changes requested):
there is no need to bump the pacakge version if we are just bumping the upstream.  the version in package.yaml is specifically for any patches we make in charts.

cmurphy:
@paynejacob the AKS chart is the first release since 2.6.2 so I think we do need a new package version

paynejacob:
@cmurphy you are changing the upstream version the up version will  already change. `100.0.0+up1.0.0` -> `100.0.0+up1.0.1`.  We only bump the package version if we are making changes without bumping the upstream version.

paynejacob (approved):
discussed offline, since these are pre-releases they are technically the same version.

PR #1638, opened 2021-11-29
Longhorn 1.2.3 & 1.1.3 for v2.6

innobead:

cc @khushboo-rancher @meldafrawi @kaxing


khushboo-rancher:
@chanow816 has validated the chart v1.1.3 on rancher v2.6-head - LGTM

Validation of chart v1.2.3 is pending.

charts/longhorn-1.1/longhorn/100.0.1+up1.1.3-rc2/Chart.yaml
```diff
@@ -0,0 +1,40 @@
+annotations:
+  catalog.cattle.io/auto-install: longhorn-crd=match
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/display-name: Longhorn
+  catalog.cattle.io/namespace: longhorn-system
+  catalog.cattle.io/os: linux
+  catalog.cattle.io/provides-gvr: longhorn.io/v1beta1
+  catalog.cattle.io/rancher-version: '>= 2.6.0-0'
```
jiaqiluo, line 8: can we also add the following 2 annotations:
```
catalog.cattle.io/kube-version:  <the same value as the kubeVersion in chart.yaml> 
catalog.cattle.io/upstream-version: <longhorn's version>
```

jiaqiluo (changes requested):

aiyengar2 (changes requested):
I haven't checked this one in depth yet, but I assume it will be the same concerns as https://github.com/rancher/charts/pull/1639#pullrequestreview-828184086. Can you address those Chart.yaml changes here as well?

I can do a deeper review of this once those changes have been addressed to see if there are any more concerns.

charts/longhorn-1.1/longhorn/100.0.1+up1.1.3-rc2/Chart.yaml
innobead: Thanks for the info about this new change. Added.

innobead:
> I haven't checked this one in depth yet, but I assume it will be the same concerns as [#1639 (review)](https://github.com/rancher/charts/pull/1639#pullrequestreview-828184086). Can you address those Chart.yaml changes here as well?
> 
> I can do a deeper review of this once those changes have been addressed to see if there are any more concerns.

Fixed.

jiaqiluo (approved):
LGTM

packages/longhorn-1.1/templates/crd-template/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-version: 1.1.2
```
aiyengar2, line 2: Same question as https://github.com/rancher/charts/pull/1639#discussion_r768911771. Can merge if this is okay.

aiyengar2 (approved):
Same concern as in https://github.com/rancher/charts/pull/1639#discussion_r768911771, but approving since the PR is good to go as long as that question does not spawn any more changes.

innobead:
We should not merge this until QA verify. We need to update again when GA. cc @cbron

PR #1647, opened 2021-12-09
Add support for kuberVersionOverrides in rancher-pushprox

brendarearden:
### Issue
https://github.com/rancher/rancher/issues/35571
https://github.com/rancher/rancher/issues/35751

### Problem
Starting in kubernetes 1.22, insecure ports are no longer supported for kube-controller-manager & kube-scheduler so secure port support was added to the latest RKE 1.22.x. This is causes issues because rancher-monitoring needs to support several versions of k8s but previous versions of RKE do not support connection on secure ports for these resources. 

### Solution
Add a kubeVersionOverride value that allows certain `.Values` to be overriden depending on the k8s version. The kubeVersionOverride has been set to override with insecure ports for previous versions of rke / rke2 kube-controller-manager & kube-scheduler


packages/rancher-pushprox/charts/values.yaml
```diff
@@ -53,7 +55,7 @@ clients:
     # If set to true, the client will disable SSL security checks
     insecureSkipVerify: false
     # Directory on host where necessary TLS cert and key to scrape metrics can be found
-    certDir: ""
+    certDir: "/usr/.ssh"
```
brendarearden, line 58: this needs to be removed.

brendarearden: This has been remove and it is back to the original default value of ""

cbron:
Moving this to 2.6.4.

aiyengar2:
Closing in favor of https://github.com/rancher/charts/pull/1766

PR #1648, opened 2021-12-09
Update Fleet charts to v0.3.8-rc3

prachidamle:
Update Fleet charts to v0.3.8-rc3

https://github.com/rancher/fleet/issues/428


StrongMonkey:
@prachidamle can you also delete old rc2 chart?

prachidamle:
> @prachidamle can you also delete old rc2 chart?

Why d we manually remove old charts? Is it because make charts is not modifying the index.yaml otherwise?

StrongMonkey:
@prachidamle it is just for house keeping purpose since we don't need old rcs in charts.

StrongMonkey (approved):

PR #1651, opened 2021-12-14
Remove experimental tag for harvester cloud provider and csi driver

yaocw2020:
The Harvester team maintains the charts in the harvester/charts repo. Here is the related modification https://github.com/harvester/charts/pull/63/files  for this pr.


brendarearden (approved):
👍

PR #1653, opened 2021-12-14
Fix Grafana init container issue 

jiaqiluo:
issue: https://github.com/rancher/rancher/issues/35390

Problem:
The Init container named `init-chown-data` in the grafana pod requires permission to run `chown` command, but the permission is dropped in the built-in PSP ([see this PR](https://github.com/grafana/helm-charts/pull/371))

Solution:
Instead of dropping all, we list out all capabilities that need to be dropped. The list is exactly the same one changed by the above PR.


MbolotSuse (dismissed):
No issues with me here - these changes seem to be in line with what was put into place on [the original PR](https://github.com/helm/charts/pull/10519). 

This being said, it might be a good idea to have someone who is not me (i.e. who is more familiar with this repo) look at the repo maintenance changes made to the non-podsecuritypolicy.yaml files.

charts/rancher-monitoring/rancher-monitoring/100.1.0+up19.0.3/charts/grafana/templates/podsecuritypolicy.yaml
```diff
@@ -13,7 +13,18 @@ spec:
   allowPrivilegeEscalation: false
   requiredDropCapabilities:
     # Default set from Docker, with DAC_OVERRIDE and CHOWN
-      - ALL
+    - FOWNER
```
brendarearden, line 16: What is not dropped by this? does that just leave chown?

MbolotSuse: [K8s docs on this](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#capabilities) redirect you to [this man page](https://man7.org/linux/man-pages/man7/capabilities.7.html). There's a substantial deal of other capabilities that will no longer be dropped.

jiaqiluo: Good question, I just copied the original list from the upstream and did not check if the list covers all. 
I think what I can do is to add all except ` CHOWN` to the list to see if it works.

MbolotSuse: This being said, [docker](https://docs.docker.com/engine/security/) does restrict capabilities to a [limited list](https://github.com/moby/moby/blob/master/oci/caps/defaults.go#L6-L19) and this requires that they all be dropped besides CHOWN and DAC_OVERRIDE. But docker is not the only container engine so this might be a valid concern. I'll remove my approval until we resolve this discussion.

jiaqiluo: I updated the PR to include the `DAC_OVERRIDE` to the list of  `requiredDropCapabilities`. As the result, `CHOWN` is the only permission that is allowed by the PSP.
I do not know why `DAC_OVERRIDE` is preserved in the original chart, but I tested the modified chart in a hardened cluster and can confirm Grafana works properly without `DAC_OVERRIDE`. 

By default, when `AllowedCapabilities` is not set/empty, it means no additional capabilities may be added beyond the default set (reference). I confirmed that [k8s' default capability set](https://kubernetes.io/docs/concepts/security/pod-security-standards/#:~:text=Capabilities,must%20be%20disallowed.) is a subset of [Docker's capability default ](https://docs.docker.com/engine/reference/run/#:~:text=can%20be%20dropped.-,Capability%20Key,Capability%20Description,-AUDIT_WRITE). (`NET_RAW` is in docker's not in k8s').  

As to docker is not the only container engine, since Docker is the only officially supported container engine for Rancher, we only need to confirm the chart working properly when Docker is used.

brendarearden (approved):
LGTM

MbolotSuse (commented):
Question on the changes to the zip files for rancher-monitoring-crd (namely - assets/rancher-monitoring/rancher-monitoring-crd-100.1.0+up19.0.3.tgz and charts/rancher-monitoring/rancher-monitoring-crd/100.1.0+up19.0.3/files/crd-manifest.tgz). Are changes to those files expected there? I don't see any changes to the source, so do we expect the zips to change?

jiaqiluo:
@MbolotSuse the changes on the zip file is expected. The zip file is regenerated each time we build the crd chart, so its metadata is chaned, although there is no changes in its content, Git still marks it as midified.

MbolotSuse (approved):
Conversations have been resolved. No changes needed.

PR #1663, opened 2021-12-16
Update Fleet charts to v0.3.8

prachidamle:
updating Fleet to non-rc version v0.3.8


brendarearden (approved):
LGTM

charts/fleet/fleet/100.0.2+up0.3.8/Chart.yaml
```diff
@@ -8,12 +8,12 @@ annotations:
   catalog.cattle.io/provides-gvr: clusters.fleet.cattle.io/v1alpha1
```
aiyengar2, line 8: @brendarearden do we need `kube-version`, `rancher-version`, or `upstream-version`? I'd expect that we do not need these annotations for Fleet (since it's deployed by Rancher, not by users), but just want to confirm.

aiyengar2 (approved):
LGTM

charts/fleet/fleet/100.0.2+up0.3.8/Chart.yaml
brendarearden: All of these have `catalog.cattle.io/hidden: "true"` so you are correct - its not available in the UI for users so the additional filtering would be unnecessary.

brendarearden (approved):

PR #1675, opened 2021-12-30
[2.5.12] Update CSI documentation

doflamingo721:
Updated the CSI documentation with the workaround for upgrading kubernetes version after csi-migration. 
Issue: https://github.com/rancher/rancher/issues/35102


packages/rancher-vsphere-csi/charts/README.md
```diff
@@ -71,3 +71,18 @@ More information on managing Secrets using kubectl [here](https://kubernetes.io/
...
+- To upgrade the Kubernetes version after Migration Follow the [Upgrading Kubernetes version](https://rancher.com/docs/rancher/v2.5/en/cluster-admin/upgrading-kubernetes/) steps.
+
+- While the cluster is updating, if you have a workload with a bound volume attached, you would find 2n pods as before. The first set will be new and active and the second set will be stuck in **Terminating** state.
+
+- For these pods in **Terminating** state, delete that pod using `kubectl delete pod <pod_id> --force=true --timeout=120s --grace-period=0`
+  * **Note** you must wait for the pod(s) to become stuck in this state. If you remove pods before they have a working duplicate, `--grace-period=0` will remove any existing data including the bound volume.
+
+- The upgrade will continue after the pod(s) stuck in terminating stage are removed.
+
+- If there is more than 1 node where workloads that have bound volumes from the migration, this step must be repeated for each node
```
PennyScissors, line 88: Can we change this to a paragraph rather than bullet points since it all basically reads like a paragraph? 

This could also be simplified to something like this:

Follow the steps in the [Upgrading Kubernetes](https://rancher.com/docs/rancher/v2.5/en/cluster-admin/upgrading-kubernetes/) guide to upgrade the Kubernetes version after migration.

**Important:** If you have a workload with a bound volume attached to it during an upgrade, you will see the pods get redeployed and the old pods will get stuck in a **Terminating** state. You can safely delete the old stuck pods by executing the command `kubectl delete pod <pod_id> --force=true --timeout=120s --grace-period=0` to fix this problem, but it's imperative you verify before deleting a pod, that there is a new active pod corresponding to the old pod you want to delete. Deleting an old pod before a new one has been deployed could result in loss of data and bound volumes (Due to `--grace-period=0`).

PennyScissors (commented):
Validation is failing, but that's expected since you are modifying the chart in place. You will need a committer/admin to force merge this for you once it's ready.

packages/rancher-vsphere-csi/package.yaml
```diff
@@ -1,3 +1,2 @@
 url: local
-packageVersion: 0
-doNotRelease: true
\ No newline at end of file
+packageVersion: 0
```
aiyengar2, line 2: we should bump this to 1 so we make a change that isn't modifying an already released chart

aiyengar2 (changes requested):

packages/rancher-vsphere-csi/charts/README.md
doflamingo721: @PennyScissors I made the changes as you mentioned. Could you please review it again?

packages/rancher-vsphere-csi/package.yaml
doflamingo721: @aiyengar2 I bumped the packageVerison as you mentioned. Can you please review it again?

charts/rancher-vsphere-csi/rancher-vsphere-csi/2.3.1/Chart.yaml
```diff
@@ -0,0 +1,21 @@
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/display-name: vSphere CSI
+  catalog.cattle.io/namespace: kube-system
+  catalog.cattle.io/os: linux
+  catalog.cattle.io/rancher-version: '>= 2.5.0-0 <= 2.5.99-0'
+  catalog.cattle.io/release-name: vsphere-csi
+apiVersion: v1
+appVersion: 2.3.0
```
PennyScissors, line 9: This should be 2.3.1 to match `version`

doflamingo721: @PennyScissors @aiyengar2  Should update the value to 2.3.1?

aiyengar2: Let's keep it since we've always had it. Let's make it 2.3.1

PR #1685, opened 2022-01-12
[2.6] Add vSphere CSI 2.4.0 and CPI 1.22.4

doflamingo721:
Linked issues:
https://github.com/rancher/rancher/issues/33848


PennyScissors:
Additionally to updating image tags, to update CSI and CPI you need to bring changes from the upstream manifests into the charts.

To update CSI to 2.4.1 (Looks like a 2.4.1 was just released):
1. Compare these manifests with our chart's resources to figure out if there are any changes from upstream that need to be pulled into our chart: https://github.com/kubernetes-sigs/vsphere-csi-driver/tree/v2.4.1/manifests/vanilla
2. Ignore the  `csi-snapshot-validatingwebhook.yaml`, `deploy-csi-snapshot-components.sh`, and `deploy-vsphere-csi-validation-webhook.sh`. We removed the webhook from the chart because you need to run those scripts to set it up which we don't want to do.

To update CPI to 1.22.4:
1. Same as 1 from CSI, but this is the upstream for CPI: https://github.com/kubernetes/cloud-provider-vsphere/tree/v1.22.4/manifests/controller-manager
2. The CPI upstream has a branch for each k8s version, but our chart is made to handle all of the k8s versions together. You need to verify that the only difference across those branches is still the CPI image tag. If this is the case then all you need to do is adjust the kubeVersion and image tags accordingly. If there are new changes across those branches (other than the tag), the chart will need to be updated to account for those as well based on the k8s version of the cluster the chart is installed.

packages/rancher-vsphere/rancher-vsphere-csi/charts/templates/controller/role.yaml
```diff
@@ -37,3 +37,21 @@ rules:
...
+    verbs: [ "watch", "get", "list" ]
+  - apiGroups: [ "snapshot.storage.k8s.io" ]
+    resources: [ "volumesnapshotcontents" ]
+    verbs: [ "create", "get", "list", "watch", "update", "delete" ]
+  - apiGroups: [ "snapshot.storage.k8s.io" ]
+    resources: [ "volumesnapshotcontents/status" ]
+    verbs: [ "update", "patch" ]
+  - apiGroups: [ "cns.vmware.com" ]
+    resources: [ "csinodetopologies" ]
+    verbs: ["get", "update", "watch", "list"]
```
aiyengar2, line 57: nit: newline

packages/rancher-vsphere/rancher-vsphere-csi/package.yaml
```diff
@@ -1,2 +1,2 @@
 url: local
-version: 100.1.0+up2.3.0
+version: 100.2.0+up2.4.1
```
aiyengar2, line 2: `100.1.1+up2.4.1` should suffice right? Why do we need to bump the minor version?

packages/rancher-vsphere/rancher-vsphere-cpi/package.yaml
```diff
@@ -1,2 +1,2 @@
 url: local
-version: 100.1.0+up1.0.100
+version: 100.2.0+up1.1.100
```
aiyengar2, line 2: 100.1.1+up1.1.100 should suffice right? Why do we need to bump the minor version

release.yaml
```diff
@@ -12,3 +12,7 @@ rancher-webhook:
 - 1.0.3+up0.2.4-rc3
 - 1.0.3+up0.2.4-rc4
 - 1.0.3+up0.2.4-rc6
+rancher-vsphere-cpi:
+- 100.2.0+up1.1.100
+rancher-vsphere-csi:
+- 100.2.0+up2.4.1
```
aiyengar2, line 18: same comment about minor versions here

aiyengar2 (changes requested):

packages/rancher-vsphere/rancher-vsphere-csi/package.yaml
PennyScissors: Because this updates 2.3.0 to 2.4.1 and there's been times in the past where bumping the patch for upstream updates caused painful situations for introducing Rancher specific patches. So for vSphere specifically, it is safer if we just bump minor if upstream bumps minor as well.

On a side note, since vSphere charts are local, the +up version represents the 2.5 variant of the same vSphere chart as in `+up<2.5-version>`. Does that mean this should actually be 100.2.0 instead? 2.4.1 does not exist in 2.5, and I don't think there are plans to backport it since 2.5 is EOM. The way we treat the +up for vSphere is confusing because what's going to happen when we have 2.7 charts? Would it be then something like 200.x.x+up100.x.x?

PennyScissors:
There was a mistake in my previous comment regarding the CPI upstream. The correct upstream is the this:

Also, there is a new upstream 1.22.5 so we should update to that if possible. Looks like only the image tag changed so it should be a matter of adding the image to image-mirror and updating the tag in this PR.

diff between 1.21.1 and 1.22.5
```diff
➜ diff -uNr cloud-provider-vsphere-1.21/releases/v1.21/vsphere-cloud-controller-manager.yaml cloud-provider-vsphere/releases/v1.22/vsphere-cloud-controller-manager.yaml 
--- cloud-provider-vsphere-1.21/releases/v1.21/vsphere-cloud-controller-manager.yaml    2022-01-28 13:27:41.000000000 -0800
+++ cloud-provider-vsphere/releases/v1.22/vsphere-cloud-controller-manager.yaml 2022-01-28 13:27:52.000000000 -0800
@@ -234,7 +234,7 @@
       serviceAccountName: cloud-controller-manager
       containers:
         - name: vsphere-cloud-controller-manager
-          image: gcr.io/cloud-provider-vsphere/cpi/release/manager:v1.21.1
+          image: gcr.io/cloud-provider-vsphere/cpi/release/manager:v1.22.5
           args:
             - --cloud-provider=vsphere
             - --v=2
```

brandond:
@doflamingo721 is this still work in progress on your side, or is there a chance we could have this available for our next RKE2 release?

samkulkarni20:
> @doflamingo721 is this still work in progress on your side, or is there a chance we could have this available for our next RKE2 release?

@brandond It won't be ready for the RKE2 Feb release, but will be ready for March.

PR #1402, opened 2021-08-08
Disable pushprox-client internal metrics by default

dkeightley:
To avoid concerns during pen-testing and scans, adjust the listening port for pushprox client metrics, these metrics are not currently used, this PR addresses the clients listening on the host network namespace by disabling the internal metrics port by default:

Example of before/after:
```
# netstat -plant | grep -i 10014
tcp6       0      0 :::10014                :::*                    LISTEN      33815/pushprox-clie
```

```
# netstat -plant | grep -i 10014
#
```

PR #1689, opened 2022-01-13
Disable pushprox client metrics by default

aiyengar2:
By default, we don't define a serviceMonitor anyways that collects pushprox **client** metrics (e.g. the metrics pushprox posts about itself) so this PR disables publishing those metrics by default.

While it may be a later RFE in order to actually support a ServiceMonitor capable of scraping those client metrics, this is being done to prevent PushProx from unnecessarily binding a hostNetwork port on the nodes to host these metrics.

Followup to https://github.com/rancher/charts/pull/1402.

Related Issue: https://github.com/rancher/rancher/issues/36140


packages/rancher-monitoring/rancher-pushprox/charts/values.yaml
```diff
@@ -5,11 +5,6 @@
 # Default image containing both the proxy and the client was generated from the following Dockerfile
 # https://github.com/prometheus-community/PushProx/blob/eeadbe766641699129920ccfaaaa30a85c67fe81/Dockerfile#L1-L15
 
-# Note: if using a cloud provider, the nodes that contain the PushProxy client must allow the node(s) that contain(s)
```
aiyengar2, line 8: I don't think this note is accurate or necessary anymore. Might have been a remnant of the initial design of the chart.

dkeightley (dismissed):
LGTM

brendarearden (dismissed):
lgtm

dkeightley (dismissed):

pennyscissors (approved):

bkleiner:
It appears that this change does _not_ prevent the clients from opening a port, but makes it [fallback to the default port](https://github.com/prometheus-community/PushProx/blob/master/cmd/client/main.go#L50), exposing whichever pushprox client was started first.

PR #1690, opened 2022-01-13
[2.5.12] Update CSI documentation

pennyscissors:
Continuation of this other PR: https://github.com/rancher/charts/pull/1675
Issue: rancher/rancher#35102


packages/rancher-vsphere-csi/charts/README.md
```diff
@@ -71,3 +71,9 @@ More information on managing Secrets using kubectl [here](https://kubernetes.io/
 ## Migration
 
 The CSI migration feature is only available for vSphere 7.0 U1.
+Follow the steps [here](https://rancher.com/docs/rancher/v2.6/en/cluster-provisioning/rke-clusters/cloud-providers/vsphere/out-of-tree/vsphere-volume-migration/?#) for CSI Migration.
```
aiyengar2, line 74: hmm we are linking 2.6 docs here but 2.5 docs below. Should they be consistent?

packages/rancher-vsphere-csi/charts/app-readme.md
```diff
@@ -6,6 +6,15 @@ CSI allows volume plugins to be installed on Kubernetes clusters as extensions.
...
 The CSI driver for vSphere is `csi.vsphere.vmware.com`.
 
+## Support Matrix
+
+| Rancher vSphere CSI | Minimum Kubernetes Release | Maximum Kubernetes Release |
+| ------------------- | -------------------------- | -------------------------- |
+| 2.3.1               | 1.19.x                     | 1.21.x                     |
+| 2.3.0               | 1.19.x                     | 1.21.x                     |
+| 2.1.1               |                            | 1.19.x                     |
+| 2.1.000             |                            | 1.19.x                     |
```
aiyengar2, line 16: should we just list 1.16.x for the blank minimums?

aiyengar2 (changes requested):

packages/rancher-vsphere-csi/charts/README.md
PennyScissors: good catch, updating now

packages/rancher-vsphere-csi/charts/app-readme.md
aiyengar2: I don’t think we should set a minimum on the chart, but we can indicate in the docs that before 1.16 is not officially supported

aiyengar2 (approved):

PR #1694, opened 2022-01-18
[dev-v2.6] Back to previous tolerations for vsphere charts

superseb:
rancher/rke2#2619


packages/rancher-vsphere/rancher-vsphere-cpi/charts/templates/daemonset.yaml
```diff
@@ -67,10 +67,10 @@ spec:
           value: "true"
         - key: node-role.kubernetes.io/control-plane
           effect: NoSchedule
-          value: "true"
+          operator: Exists
         - key: node-role.kubernetes.io/etcd
           effect: NoSchedule
```
PennyScissors, line 72: The issue says the effect for etcd should be NoExecute. I'm not sure whether that's true, but if you looked into that and decided it should remain as NoSchedule, may I know the reason?

packages/rancher-vsphere/rancher-vsphere-csi/package.yaml
```diff
@@ -1,2 +1,2 @@
 url: local
-version: 100.1.0+up2.3.0
```
PennyScissors, line 2: It might be better to do `100.1.1+up2.3.0` since it's a small change. We normally bump the major number when a  new Rancher version is released, the minor number when we update the chart based on the upstream or introduce significant/breaking changes, and the patch for everything else.

superseb: Fixed!

pennyscissors (dismissed):
LGTM, make sure you add the charts to the `New charts in 2.6.4` issue (not sure if we have one yet).

pennyscissors (approved):
@samkulkarni20 @superseb Looks like this PR is ready to merge, but if we merge it the charts in the rancher/vsphere-charts repo will not be in-sync with the ones here and we could potentially end up losing this change in the process of setting up the vSphere charts in a separate repo. Once rancher/vsphere-charts is ready, we need to either close this PR and make the changes from this PR in rancher/vsphere-charts, or merge this and then bring the changes to rancher/vsphere-charts. I'll leave it up to you to decide Sameer.

samkulkarni20:
@PennyScissors @superseb I was thinking we'll let all changes to charts happen in the new upstream i.e. `rancher/vsphere-charts` repo now. We currently have 2 PRs that need to be migrated to `vshpere-charts` repo. One of them is this PR from @superseb &  the [other PR](https://github.com/rancher/charts/pull/1685) is from @doflamingo721 to upgrade the underlying CPI & CSI chart versions. @doflamingo721 can club both of those changes & create a single PR in the `vsphere-charts` repo. @superseb is that ok with you?

PR #1700, opened 2022-01-21
Add a version of the rancher-webhook chart for the Rancher v2.6.3-patch1 release

maxsokolovsky:
Related to https://github.com/rancher/rancher/issues/36177.
This makes a chart for rancher-webhook `v0.2.4-rc7`.
It's to be used only for the Rancher release `v2.6.3-patch1`.


maxsokolovsky:
Looks like there is no CI - is it configured on a per-branch basis?

thedadams (changes requested):
Per "offline" discussion, version `v0.2.3-patch1` is not what we should go with here.

maxsokolovsky:
This makes a chart for rancher-webhook `v0.2.4-rc7`.

thedadams (approved):

cmurphy (approved):
lgtm but I think @aiyengar2 should rubber-stamp it

packages/rancher-webhook/package.yaml
```diff
@@ -0,0 +1,2 @@
+url: https://github.com/rancher/webhook/releases/download/v0.2.4-rc7/rancher-webhook-0.2.4-rc7.tgz
+version: 1.0.3
```
cbron, line 2: Can you cut a 0.2.5-rc1 with actual new code in master, then update this in charts dev-v2.6 to be 1.0.4+0.2.5-rc1, then announce somewhere so people don't accidentally use 0.2.4.

release.yaml
```diff
@@ -0,0 +1,2 @@
+rancher-webhook:
+- 1.0.3+up0.2.4-rc7
```
cbron, line 2: Just a reminder to cut final tag before we go live.

cbron (approved):

packages/rancher-webhook/package.yaml
maxsokolovsky: Will do, do you want the `v0.2.5-rc1` tag to be on the latest commit in master?

maxsokolovsky: If yes, then the latest commit will have two tags: `v0.2.4-rc6` and `v0.2.5-rc1`, if that's OK...

cbron: Yes thats ok. We want 0.2.4-rc6 to be forgotten.

aiyengar2 (approved):
LGTM

PR #1701, opened 2022-01-21
Adding charts for supporting Windows GMSA

phillipsj:
Windows GMSA requires several steps for configuring it correctly. We are creating a single chart that will install all the required components and is configurable if you just need the base ones.


phillipsj:
Linked to: https://github.com/rancher/windows/issues/22

phillipsj:
@sirredbeard FYI

packages/rancher-windows-gmsa/windows-gmsa-admission-webhook/package.yaml
```diff
@@ -0,0 +1,2 @@
+url: local
+packageVersion: 1
```
aiyengar2, line 2: This concept no longer exists in rancher/charts in 2.6, we use `version` now.

aiyengar2 (changes requested):

phillipsj:
gmsa-chart-deployed

packages/rancher-windows-gmsa/package.yaml
```diff
@@ -0,0 +1,8 @@
+url: local
+version: 1.0.0
+additionalCharts:
+  - workingDir: charts-crd
+    crdOptions:
+      templateDirectory: crd-template
+      crdDirectory: templates
+      addCRDValidationToMainChart: true
```
aiyengar2, line 8: nit: newline

packages/rancher-windows-gmsa/charts/templates/validatingwebhook.yaml
```diff
@@ -0,0 +1,34 @@
...
+    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/{{ .Release.Name }}
+  {{- end }}
+  labels:
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+webhooks:
+  - name: admission-webhook.windows-gmsa.sigs.k8s.io
+    clientConfig:
+      service:
+        name: {{ .Release.Name }}
+        namespace: {{.Release.Namespace}}
```
aiyengar2, line 16:
```suggestion
        namespace: {{ .Release.Namespace }}

```

packages/rancher-windows-gmsa/charts/templates/validatingwebhook.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingWebhookConfiguration
+metadata:
+  name: {{ .Release.Name }}
+  {{- if .Values.certificates.certManager.enabled }}
+  annotations:
+    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/{{ .Release.Name }}
+  {{- end }}
+  labels:
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
```
aiyengar2, line 10: since this gets used everywhere, it might make more sense to put this in a helper function.

e.g. for Monitoring

https://github.com/rancher/charts/blob/73f1cd647751650190644b23d0e41221bdfed644/charts/rancher-monitoring/templates/_helpers.tpl#L105-L113

https://github.com/rancher/charts/blob/73f1cd647751650190644b23d0e41221bdfed644/charts/rancher-monitoring/templates/prometheus/prometheus.yaml#L9

packages/rancher-windows-gmsa/charts/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,8 @@
+# the service account for the webhook
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Release.Name }}
+  namespace: {{.Release.Namespace}}
```
aiyengar2, line 6:
```suggestion
  namespace: {{ .Release.Namespace }}

```

packages/rancher-windows-gmsa/templates/crd-template/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+version: 1.0.0
+description: Installs the CRDs for Windows GMSA.
+name: rancher-windows-gmsa-crd
+type: application
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/hidden: "true"
+  catalog.cattle.io/namespace: kube-system
```
aiyengar2, line 9: Is there a reason why this is in `kube-system`? Generally, we place feature charts in their own namespaces, e.g. `cattle-monitoring-system` for Monitoring or `cattle-windows-gmsa-system` for example

aiyengar2: this is also the pattern for wins-upgrader: https://github.com/rancher/charts/blob/502035b9ac9763e2ebbdde5c05f2137862bb2678/charts/rancher-wins-upgrader/100.0.0%2Bup0.0.1/Chart.yaml#L3

packages/rancher-windows-gmsa/charts/templates/validatingwebhook.yaml
aiyengar2: why isn't this label also used on the deployment?

packages/rancher-windows-gmsa/charts/templates/mutatingwebhook.yaml
```diff
@@ -0,0 +1,32 @@
...
+  {{- if .Values.certificates.certManager.enabled }}
+  annotations:
+    cert-manager.io/inject-ca-from: {{ .Release.Namespace }}/{{ .Release.Name }}
+  {{- end }}
+webhooks:
+  - name: admission-webhook.windows-gmsa.sigs.k8s.io
+    clientConfig:
+      service:
+        name: {{ .Release.Name }}
+        namespace: {{.Release.Namespace}}
```
aiyengar2, line 14:
```suggestion
        namespace: {{ .Release.Namespace }}

```

packages/rancher-windows-gmsa/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,60 @@
...
+        - name: {{ .Release.Name }}
+          image: '{{ template "system_default_registry" . }}{{ .Values.image.repository }}:{{ .Values.image.tag }}'
+          imagePullPolicy: {{ .Values.image.imagePullPolicy }}
+          readinessProbe:
+            httpGet:
+              scheme: HTTPS
+              path: /health
+              port: 443
+          ports:
+            - containerPort: 443
```
aiyengar2, line 42: should the port to run this on be parametrized in the values.yaml? There's no direct benefit, but I think it might make sense over hard-coding it

packages/rancher-windows-gmsa/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,60 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ .Release.Name}}
+  namespace: {{.Release.Namespace}}
```
aiyengar2, line 5:
```suggestion
  namespace: {{ .Release.Namespace }}

```

packages/rancher-windows-gmsa/charts/templates/credentialspec.yaml
```diff
@@ -0,0 +1,24 @@
...
+        Scope: {{ .Values.credential.domainJoinConfig.dnsName }} #DNS Domain Name
+  CmsPlugins:
+    - ActiveDirectory
+  DomainJoinConfig:
+    DnsName: {{ .Values.credential.domainJoinConfig.dnsName }} #DNS Domain Name
+    DnsTreeName:  {{ .Values.credential.domainJoinConfig.dnsName }} #DNS Domain Name Root
+    Guid:  {{ .Values.credential.domainJoinConfig.guid }} #GUID
+    MachineAccountName:  {{ .Values.credential.domainJoinConfig.machineAccountName }} #Username of the GMSA account
+    NetBiosName:  {{ .Values.credential.domainJoinConfig.netBiosName }} #NETBIOS Domain Name
+    Sid:  {{ .Values.credential.domainJoinConfig.sid }} #SID of GMSA
```
aiyengar2, line 23: nit: I don't think we need these comments here if it exists on the values.yaml

packages/rancher-windows-gmsa/charts/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,15 @@
...
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{ .Release.Name }}
+  labels:
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+subjects:
+  - kind: ServiceAccount
+    name: {{ .Release.Name }}
+    namespace: {{.Release.Namespace}}
```
aiyengar2, line 11:
```suggestion
    namespace: {{ .Release.Namespace }}

```

packages/rancher-windows-gmsa/charts/templates/_helpers.tpl
```diff
@@ -0,0 +1,7 @@
+# Rancher
+
+{{- define "system_default_registry" -}}
+{{- if .Values.global.cattle.systemDefaultRegistry -}}
+{{- printf "%s/" .Values.global.cattle.systemDefaultRegistry -}}
+{{- end -}}
+{{- end -}}
```
aiyengar2, line 7: nit: newline

packages/rancher-windows-gmsa/charts/Chart.yaml
```diff
@@ -0,0 +1,24 @@
+annotations:
+  catalog.cattle.io/auto-install: rancher-windows-gmsa-crd=match
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/display-name: Windows GMSA
+  catalog.cattle.io/os: windows
+  catalog.cattle.io/provides-gvr: windows.k8s.io.gmsacredentialspecs/v1
+  catalog.cattle.io/release-name: rancher-windows-gmsa
```
aiyengar2, line 7: Missing annotations:
- `catalog.cattle.io/namespace`
- `catalog.cattle.io/rancher-version`
- `catalog.cattle.io/kube-version`

packages/rancher-windows-gmsa/charts/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+apiVersion: v2
+appVersion: 0.3.0
+description: Windows GMSA Configuration
+icon: https://charts.rancher.io/assets/logos/windows-gmsa.svg
```
aiyengar2, line 11: Note: QA should know that this icon will not work until we ship this chart for the first time

packages/rancher-windows-gmsa/charts/Chart.yaml
aiyengar2: especially since we only support 1.21+

packages/rancher-windows-gmsa/charts/questions.yaml
```diff
@@ -0,0 +1,59 @@
...
+        label: DNS Tree Domain
+        description: Root name of the domain in DNS
+        type: string
+        required: true
+      - variable: credential.domainJoinConfig.netBiosName
+        label: NETBIOS Name
+        description: NETBIOS Name for the domain.
+        type: string
+        required: true
+
```
aiyengar2, line 40: awkward newline in between?

packages/rancher-windows-gmsa/charts/questions.yaml
```diff
@@ -0,0 +1,59 @@
+questions:
+  - variable: credential.enabled
+    default: true
+    description: Create a GMSA Credential Spec when installing GMSA Webhook
```
aiyengar2, line 4:
```suggestion
    description: Whether to create a GMSA Credential Spec when installing GMSA Webhook
```

packages/rancher-windows-gmsa/charts/questions.yaml
```diff
@@ -0,0 +1,59 @@
+questions:
+  - variable: credential.enabled
+    default: true
+    description: Create a GMSA Credential Spec when installing GMSA Webhook
+    label: Create a GMSA Credential Spec
```
aiyengar2, line 5:
```suggestion
    label: Whether to create a GMSA Credential Spec
```

packages/rancher-windows-gmsa/charts/questions.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - variable: credential.domainJoinConfig.netBiosName
+        label: NETBIOS Name
+        description: NETBIOS Name for the domain.
+        type: string
+        required: true
+
+  - variable: certificates.certManager.enabled
+    default: true
+    description: Enable cert manager integration. Cert manager should be already installed
+    label: Enable Cert Manager integration
```
aiyengar2, line 44:
```suggestion
    label: Generate certificate through cert-manager
```

packages/rancher-windows-gmsa/charts/questions.yaml
```diff
@@ -0,0 +1,59 @@
...
+        required: true
+      - variable: credential.domainJoinConfig.netBiosName
+        label: NETBIOS Name
+        description: NETBIOS Name for the domain.
+        type: string
+        required: true
+
+  - variable: certificates.certManager.enabled
+    default: true
+    description: Enable cert manager integration. Cert manager should be already installed
```
aiyengar2, line 43:
```suggestion
    description: Use cert-manager to generate certificates for the webhook
```

packages/rancher-windows-gmsa/charts/questions.yaml
```diff
@@ -0,0 +1,59 @@
...
+    show_subquestion_if: false
+    subquestions:
+      - variable: certificates.secretName
+        default: webhook-server-cert
+        description: Use certificates from secret. Secret should exists in the app namespace, with certs data (ca.crt, tls.crt & tls.key)
+        label: Secret name
+        type: string
+        required: true
+      - variable: certificates.caBundle
+        description: Use self signed CA Bundle. It should be provided in base64 format
```
aiyengar2, line 56:
```suggestion
        description: Provide self-signed CA Bundle as a base64-encoded string
```

packages/rancher-windows-gmsa/charts/templates/_helpers.tpl
aiyengar2: maybe we should also include these? For RKE1 / RKE2 windows support

https://github.com/rancher/charts/blob/502035b9ac9763e2ebbdde5c05f2137862bb2678/charts/rancher-monitoring/100.1.0%2Bup19.0.3/templates/_helpers.tpl#L106-L124

packages/rancher-windows-gmsa/charts/templates/credentialspec.yaml
```diff
@@ -0,0 +1,24 @@
+{{- if .Values.credential.enabled -}}
+apiVersion: windows.k8s.io/v1
+kind: GMSACredentialSpec
+metadata:
+  name: {{ lower .Values.credential.domainJoinConfig.machineAccountName }}  #This is an arbitrary name but it will be used as a reference
```
aiyengar2, line 5: nit:
```suggestion
  name: {{ .Values.credential.domainJoinConfig.machineAccountName | lower }}  #This is an arbitrary name but it will be used as a reference

```

packages/rancher-windows-gmsa/charts/templates/credentialspec.yaml
```diff
@@ -0,0 +1,24 @@
...
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+credspec:
+  ActiveDirectoryConfig:
+    GroupManagedServiceAccounts:
+      - Name: {{ .Values.credential.domainJoinConfig.machineAccountNamename }}   #Username of the GMSA account
+        Scope: {{ .Values.credential.domainJoinConfig.netBiosName }}  #NETBIOS Domain Name
+      - Name: {{ .Values.credential.domainJoinConfig.machineAccountNamename }}   #Username of the GMSA account
+        Scope: {{ .Values.credential.domainJoinConfig.dnsName }} #DNS Domain Name
+  CmsPlugins:
+    - ActiveDirectory
```
aiyengar2, line 16: Should this also be parametrized?

packages/rancher-windows-gmsa/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,60 @@
...
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: {{ .Release.Name }}
+  template:
+    metadata:
+      labels:
+        app: {{ .Release.Name }}
```
aiyengar2, line 16: Why don't these also have `chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}`?

Or alternatively, why don't all your other resources have `app: {{ .Release.Name }}`?

As mentioned in another comment somewhere in this review, I think we should standardize on annotations and toss them into a helper function, so that way you can use that by default and add more if you need specific, component-specific annotations later.

aiyengar2: If there's a specific reason why we want the labels on this to be different than anything else, feel free to comment and skip changes for this though!

packages/rancher-windows-gmsa/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,60 @@
...
+      serviceAccountName: {{ .Release.Name }}
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+              - matchExpressions:
+                  - key: kubernetes.io/os
+                    operator: In
+                    values:
+                      - linux
```
aiyengar2, line 27: We should either replace this with the node-selector helper template listed above or create a new helper function for `linux-node-affinity` that takes in overrides and use that here. That way we can standardize on templates to get stuff working on Windows.

packages/rancher-windows-gmsa/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,60 @@
...
+      {{- with .Values.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      containers:
+        - name: {{ .Release.Name }}
+          image: '{{ template "system_default_registry" . }}{{ .Values.image.repository }}:{{ .Values.image.tag }}'
```
aiyengar2, line 34: nit: no need for single quotes
```suggestion
          image: {{ template "system_default_registry" . }}{{ .Values.image.repository }}:{{ .Values.image.tag }}

```

packages/rancher-windows-gmsa/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,60 @@
...
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+              - matchExpressions:
+                  - key: kubernetes.io/os
+                    operator: In
+                    values:
+                      - linux
+      {{- with .Values.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
```
aiyengar2, line 30: need to add in RKE1 windows toleration

packages/rancher-windows-gmsa/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,60 @@
...
+              value: /tls/crt
+      volumes:
+        - name: tls
+          secret:
+            secretName: {{ .Values.certificates.secretName }}
+            items:
+              - key: tls.key
+                path: key
+              - key: tls.crt
+                path: crt
```
aiyengar2, line 60: nit: I prefer keeping file extensions if possible. What do you think?

```suggestion
            - name: TLS_KEY
              value: /etc/ssl/rancher-windows-gmsa-webhook/tls.key
            - name: TLS_CRT
              value: /etc/ssl/rancher-windows-gmsa-webhook/tls.cert
      volumes:
        - name: /etc/ssl/rancher-windows-gmsa-webhook
          secret:
            secretName: {{ .Values.certificates.secretName }}
            items:
              - key: tls.key
                path: tls.key
              - key: tls.crt
                path: tls.cert
```

packages/rancher-windows-gmsa/charts/templates/issuer.yaml
```diff
@@ -0,0 +1,50 @@
+{{- if .Values.certificates.certManager.enabled -}}
+  {{- $certmanagerVer :=  split "." .Values.certificates.certManager.version -}}
+  {{- if or (.Capabilities.APIVersions.Has "cert-manager.io/v1") (and (gt (len $certmanagerVer._0) 0) (eq (int $certmanagerVer._0) 1) (ge (int $certmanagerVer._1) 0)) }}
+apiVersion: cert-manager.io/v1
+  {{- else if or (.Capabilities.APIVersions.Has "cert-manager.io/v1beta1") (and (gt (len $certmanagerVer._0) 0) (eq (int $certmanagerVer._0) 0) (ge (int $certmanagerVer._1) 16)) }}
+apiVersion: cert-manager.io/v1beta1
+  {{- else if or (.Capabilities.APIVersions.Has "cert-manager.io/v1alpha2") (and (gt (len $certmanagerVer._0) 0) (eq (int $certmanagerVer._0) 0) (ge (int $certmanagerVer._1) 11)) }}
+apiVersion: cert-manager.io/v1alpha2
+  {{- else if or (.Capabilities.APIVersions.Has "certmanager.k8s.io/v1alpha1") (and (gt (len $certmanagerVer._0) 0) (eq (int $certmanagerVer._0) 0) (lt (int $certmanagerVer._1) 11)) }}
+apiVersion: certmanager.k8s.io/v1alpha1
```
aiyengar2, line 10: This should definitely be in a helper function since it is called twice and has complex logic

IMO you should be able to call it like:

```yaml
apiVersion: {{ include "cert-manager.apiVersion" }}
```

aiyengar2: This logic also seems way to complicated; why do we need to support multiple cert-manager versions? Is this specifically requested or added to future-proof?

I think it's sufficient enough that we should just allow the user to pass in the apiVersion as a string to override if they want, otherwise, we just always do `cert-manager.io/v1`

packages/rancher-windows-gmsa/charts/questions.yaml
aiyengar2: I'm also concerned whether we should allow users to pass in a caBundle this way, since it's a security risk right? Anyone with permissions to view the values provided to the Helm deployment can `base64 -d` it and grab the underlying certs used.

I think the existing two options of a) use generated or b) put it in a secret are sufficient and this can be dropped as a Helm chart option. What do you think?

packages/rancher-windows-gmsa/charts/templates/issuer.yaml
phillipsj: @aiyengar2 I pulled this logic straight from the externalip-webhook chart we already have.

packages/rancher-windows-gmsa/charts/templates/_helpers.tpl
sirredbeard: We are not targeting RKE1 with GMSA support.

packages/rancher-windows-gmsa/charts/questions.yaml
phillipsj: This is how this is handled in the ExternalIP chart.

packages/rancher-windows-gmsa/charts/templates/credentialspec.yaml
phillipsj: I don't believe there are currently any other plugins available.

packages/rancher-windows-gmsa/charts/Chart.yaml
phillipsj: @aiyengar2 anytime I include these annotations my chart isn't visible in the UI.

aiyengar2: @phillipsj that might have to do with your local dev setup or it could be a bug with Rancher (specifically when parsing these annotations in Windows clusters maybe?). Nevertheless, we still need these annotations set on the chart.

@PennyScissors might be able to help you look into why these annotations aren't working since he's familiar with some of this code.

release.yaml
```diff
@@ -1,18 +1,24 @@
...
-- 1.0.3+up0.2.4-rc2
-- 1.0.3+up0.2.4-rc3
-- 1.0.3+up0.2.4-rc4
-- 1.0.3+up0.2.4-rc6
-- 1.0.4+up0.2.5-rc1
+  - 1.0.3+up0.2.4-rc2
+  - 1.0.3+up0.2.4-rc3
+  - 1.0.3+up0.2.4-rc4
+  - 1.0.3+up0.2.4-rc6
+  - 1.0.4+up0.2.5-rc1
```
aiyengar2, line 18: remove spaces?

index.yaml
```diff
@@ -6005,6 +6005,50 @@ entries:
     urls:
     - assets/rancher-windows-exporter/rancher-windows-exporter-0.1.000.tgz
     version: 0.1.000
+  rancher-windows-gmsa:
```
aiyengar2, line 6008: index.yaml changes should be in the make charts commit

packages/rancher-windows-gmsa/charts/questions.yaml
```diff
@@ -0,0 +1,58 @@
+questions:
+  - variable: credential.enabled
+    default: true
+    description: Whether to create a GMSA Credential Spec when installing GMSA Webhook
+    label: Whether to create a GMSA Credential Spec
```
aiyengar2, line 5: Whoops sorry. I missed this in my review 😅 

This should be `Whether to create a GMSA Credential`, not creating the spec right?

packages/rancher-windows-gmsa/charts/questions.yaml
aiyengar2: this wasn't resolved or addressed? If you like the original wording better, we should still change `cert manager` to `cert-manager` to keep the style consistent with what upstream uses in https://cert-manager.io/

packages/rancher-windows-gmsa/charts/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,16 @@
+# bind that role to the webhook's service account
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{ .Release.Name }}
+  labels:
+    chart: {{ template "gmsa.chartref" . }}
```
aiyengar2, line 7: Why not just include the key `chart` in the template as well? So you can just call this as:

```yaml
  labels: {{ template "gmsa.chartref" . | nindent 4 }}
```

packages/rancher-windows-gmsa/charts/templates/credentialspec.yaml
```diff
@@ -0,0 +1,25 @@
+{{- if .Values.credential.enabled -}}
+apiVersion: windows.k8s.io/v1
+kind: GMSACredentialSpec
+metadata:
+  name: {{ .Values.credential.domainJoinConfig.machineAccountName | lower }}
```
aiyengar2, line 5: what if the machineAccountName is not provided? Since these are all required values, we should use the [`required`](https://austindewey.com/2018/12/28/helm-tricks-input-validation-with-required-and-fail/) functionality in Helm to ensure this fails on running `helm template` if these values are not provided.

This can be done within some files (like here) or done as a call within a dummy file, just like we do for `validate-install-crd` here: https://github.com/rancher/charts/blob/73f1cd647751650190644b23d0e41221bdfed644/charts/rancher-monitoring/templates/validate-install-crd.yaml#L16

packages/rancher-windows-gmsa/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,65 @@
...
+  selector:
+    matchLabels:
+      app: {{ .Release.Name }}
+  template:
+    metadata:
+      labels:
+        app: {{ .Release.Name }}
+    spec:
+      securityContext:
+        runAsUser: 1000
```
aiyengar2, line 21: should be overridable and exposed to the end usr in case they want to change the user that runs this.

Example: https://github.com/rancher/charts/blob/502035b9ac9763e2ebbdde5c05f2137862bb2678/charts/rancher-logging/100.0.1%2Bup3.15.0/values.yaml#L81-L93

https://github.com/rancher/charts/blob/502035b9ac9763e2ebbdde5c05f2137862bb2678/charts/rancher-logging/100.0.1%2Bup3.15.0/templates/deployment.yaml#L48-L53

You probably only need containerSecurityContext, not podSecurityContext

packages/rancher-windows-gmsa/charts/templates/issuer.yaml
aiyengar2: Hmm, in that case using the logic is fine but it should definitely be in a helper function, not repeated twice in the code.

packages/rancher-windows-gmsa/charts/templates/issuer.yaml
```diff
@@ -0,0 +1,51 @@
...
+  {{- else }}
+# Setting latest version as default
+apiVersion: cert-manager.io/v1
+  {{- end }}
+kind: Issuer
+metadata:
+  name: {{ .Release.Name }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
```
aiyengar2, line 47: nit: use template

aiyengar2 (changes requested):

packages/rancher-windows-gmsa/charts/templates/credentialspec.yaml
phillipsj: @aiyengar2 are we able to make something required that could be disabled?

aiyengar2: yes, for example in the link above, we have an if statement gating the required block:

https://github.com/rancher/charts/blob/73f1cd647751650190644b23d0e41221bdfed644/charts/rancher-monitoring/templates/validate-install-crd.yaml#L1

This is what ensures that a `helm template` (which finds 0 cluster roles since it's not running in a real cluster) does not fail on not finding the CRDs installed.

packages/rancher-windows-gmsa/charts/templates/_helpers.tpl
```diff
@@ -0,0 +1,28 @@
...
+{{/* Create chart name and version as used by the chart label. */}}
+{{- define "gmsa.chartref" -}}
+{{- replace "+" "_" .Chart.Version | printf "chart: %s-%s" .Chart.Name -}}
+{{- end }}
+
+{{/* Determine ApiVersion for Cert-manager */}}
+{{- define "cert-manager.apiversion" -}}
+  {{- $certmanagerVer :=  split "." .Values.certificates.certManager.version -}}
+  {{- if or (.Capabilities.APIVersions.Has "cert-manager.io/v1") (and (gt (len $certmanagerVer._0) 0) (eq (int $certmanagerVer._0) 1) (ge (int $certmanagerVer._1) 0)) }}
+{{- printf "apiVersion: cert-manager.io/v1" -}}
```
aiyengar2, line 18: you don't need to printf, you could just put `apiVersion: cert-manager.i/v1` directly with no brackets

packages/rancher-windows-gmsa/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,68 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    seccomp.security.alpha.kubernetes.io/pod: runtime/default
+  name: {{ .Release.Name}}
```
aiyengar2, line 6: nit: space at the end

packages/rancher-windows-gmsa/charts/values.yaml
```diff
@@ -0,0 +1,49 @@
...
+  repository: rancher/mirrored-sigwindowstools-k8s-gmsa-webhook
+  tag: v0.3.0
+  imagePullPolicy: IfNotPresent
+
+global:
+  cattle:
+    systemDefaultRegistry: ""
+  kubectl:
+    repository: rancher/kubectl
+    tag: v1.18.6
```
aiyengar2, line 31: can this be bumped?

aiyengar2 (changes requested):

packages/rancher-windows-gmsa/charts/templates/credentialspec.yaml
phillipsj: I don't think it should be required. If they are enabling it they would set the values. If the values aren't set, then the Credential just wouldn't work, but wouldn't prevent the whole from being available.

packages/rancher-windows-gmsa/charts/questions.yaml
phillipsj: I missed that.

packages/rancher-windows-gmsa/charts/templates/deployment.yaml
phillipsj: We had discussed this yesterday. I would like to hold off and making this change until the final approach for how to handle this is implemented. I will create a GitHub issue to track.

packages/rancher-windows-gmsa/charts/app-readme.md
```diff
@@ -0,0 +1,9 @@
+# Windows GMSA Admission Webhook
+
+This chart creates the GMSA CRD, Credential, and Admission Webhook. The official documentation and tutorials can be found [here](https://github.com/kubernetes-sigs/windows-gmsa).
+
+## Prerequisites
+
+- Active Directory that support Group Managed Service Accounts
```
aiyengar2, line 7: *supports

packages/rancher-windows-gmsa/charts/questions.yaml
```diff
@@ -0,0 +1,58 @@
...
+        default: webhook-server-cert
+        description: Use certificates from secret. Secret should exists in the app namespace, with certs data (ca.crt, tls.crt & tls.key)
+        label: Secret name
+        type: string
+        required: true
+      - variable: certificates.caBundle
+        description: Provide self-signed CA Bundle as a base64-encoded string
+        label: CA Bundle
+        type: string
+        required: true
```
aiyengar2, line 58: should no longer exist?

packages/rancher-windows-gmsa/charts/templates/_helpers.tpl
```diff
@@ -0,0 +1,33 @@
...
+apiVersion: cert-manager.io/v1alpha2
+  {{- else if or (.Capabilities.APIVersions.Has "certmanager.k8s.io/v1alpha1") (and (gt (len $certmanagerVer._0) 0) (eq (int $certmanagerVer._0) 0) (lt (int $certmanagerVer._1) 11)) }}
+apiVersion: cert-manager.io/v1alpha1
+  {{- else }}
+apiVersion: cert-manager.io/v1
+  {{- end }}
+{{- end }}
+
+{{- define "certificates.cabundle"}}
+{{- (lookup "v1" "Secret" "" "{{ .Release.Namespace }}" "{{ .Values.certificates.secretName }}").data.ca.crt }}
```
aiyengar2, line 31: what happens if the Secret does not exist? We should emit an error on install saying we couldn't grab it.

To ensure that this error only gets called on `helm install` and not on `helm template`, we could use the hack we use in `validate-install-crd.yaml` to only try to execute the lookup if we were able to grab at least one clusterrole (which is always gonna exist by default in any new kubernetes cluster, but will never exist on a helm template that points to no cluster):
https://github.com/rancher/charts/blob/73f1cd647751650190644b23d0e41221bdfed644/charts/rancher-monitoring/templates/validate-install-crd.yaml#L1

aiyengar2 (changes requested):
Woo thank you for your changes! Just have some nits and we're good to go.

packages/rancher-windows-gmsa/charts/Chart.yaml
```diff
@@ -0,0 +1,28 @@
+annotations:
+  catalog.cattle.io/auto-install: rancher-windows-gmsa-crd=match
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/display-name: Windows GMSA
+  catalog.cattle.io/experimental: "true"
+  catalog.cattle.io/kube-version: '>= 1.21.0'
+  catalog.cattle.io/namespace: cattle-windows-gmsa-system
+  catalog.cattle.io/os: windows
+  catalog.cattle.io/provides-gvr: windows.k8s.io.gmsacredentialspecs/v1
+  catalog.cattle.io/rancher-version: '>= 2.6.4 < 2.6.99'
```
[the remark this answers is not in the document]
aiyengar2: resolved.

packages/rancher-windows-gmsa/charts/Chart.yaml
```diff
@@ -0,0 +1,28 @@
+annotations:
+  catalog.cattle.io/auto-install: rancher-windows-gmsa-crd=match
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/display-name: Windows GMSA
+  catalog.cattle.io/experimental: "true"
+  catalog.cattle.io/kube-version: '>= 1.21.0'
```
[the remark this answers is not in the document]
aiyengar2: resolved.

phillipsj:
@aiyengar2 it works.
itworks

charts/rancher-windows-gmsa/1.0.0/templates/_helpers.tpl
```diff
@@ -0,0 +1,35 @@
...
+  {{- else }}
+apiVersion: cert-manager.io/v1
+  {{- end }}
+{{- end }}
+
+{{- define "certificates.cabundle"}}
+{{- if gt (len (lookup "rbac.authorization.k8s.io/v1" "ClusterRole" "" "")) 0 -}}
+{{- (lookup "v1" "Secret" "" "{{ .Release.Namespace }}" "{{ .Values.certificates.secretName }}").data.ca.crt }}
+{{- end -}}
+{{- end }}
```
aiyengar2, line 34: I don't think this works, this needs to be tested. The lookup function only takes 4 arguments, you've listed 5 here.

Here's a fixed version you can test with checks on the contents of the secret as well:

```yaml
{{- define "certificates.cabundle"}}
{{- $secret := (lookup "v1" "Secret" .Release.Namespace .Values.certificates.secretName) -}}
{{- if lt (len $secret) 1 -}}
{{- required (printf "CA Bundle  secret '%s' in namespace '%s' must exist" .Values.certificates.secretName .Release.Namespace) "" -}}
{{- else -}}
{{- required (printf "CA Bundle secret '%s' in namespace '%s' is empty" .Values.certificates.secretName .Release.Namespace) $secret.data -}}
{{- if or (not (hasKey $secret.data "ca.crt")) (not (hasKey $secret.data "tls.crt")) (not (hasKey $secret.data "tls.key")) -}}
{{- required (printf "CA Bundle secret '%s' in namespace '%s' must contain ca.crt, tls.key, and tls.cert; found the following keys in the secret: %s" .Values.certificates.secretName .Release.Namespace $secret.data) "" -}}
{{- end -}}
{{- end -}}
{{- get $secret.data "ca.crt" }}
{{- end }}
```

charts/rancher-windows-gmsa/1.0.0/templates/deployment.yaml
```diff
@@ -0,0 +1,68 @@
...
+              value: /etc/ssl/rancher-windows-gmsa-webhook/tls.cert
+      volumes:
+        - name: tls
+          secret:
+            secretName: {{ .Values.certificates.secretName }}
+            items:
+              - key: tls.key
+                path: tls.key
+              - key: tls.crt
+                path: tls.cert
```
aiyengar2, line 67: nit: extension should be consistent (`.cert` -> `.crt`)

```suggestion
          env:
            - name: TLS_KEY
              value: /etc/ssl/rancher-windows-gmsa-webhook/tls.key
            - name: TLS_CRT
              value: /etc/ssl/rancher-windows-gmsa-webhook/tls.crt
      volumes:
        - name: tls
          secret:
            secretName: {{ .Values.certificates.secretName }}
            items:
              - key: tls.key
                path: tls.key
              - key: tls.crt
                path: tls.crt

```

charts/rancher-windows-gmsa/1.0.0/values.yaml
```diff
@@ -0,0 +1,42 @@
...
+    tag: v1.22.6
+    pullPolicy: IfNotPresent
+
+## SecurityContext holds pod-level security attributes and common container settings.
+## This defaults to non root user with uid 1000 and gid 2000.	*v1.PodSecurityContext	false
+## ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
+##
+podSecurityContext:
+  runAsNonRoot: false
+  runAsUser: 0
```
aiyengar2, line 38: can you add a note about why we run as root here?

aiyengar2 (changes requested):
almost good to go! lookup function seems to need a bit of work, try the suggested changes though.

charts/rancher-windows-gmsa/1.0.0/templates/_helpers.tpl
aiyengar2: Ran test cases locally:

```bash
arvindiyengar: ~/Rancher/charts/src/github.com/rancher/charts 
$ helm upgrade --install rancher-windows-gmsa packages/rancher-windows-gmsa/charts
Error: UPGRADE FAILED: execution error at (rancher-windows-gmsa/templates/_helpers.tpl:34:4): CA Bundle secret 'gmsa-server-cert' in namespace 'default' must exist
arvindiyengar: ~/Rancher/charts/src/github.com/rancher/charts 
$ k create secret generic gmsa-server-cert                                        
secret/gmsa-server-cert created
arvindiyengar: ~/Rancher/charts/src/github.com/rancher/charts 
$ helm upgrade --install rancher-windows-gmsa packages/rancher-windows-gmsa/charts
Error: UPGRADE FAILED: execution error at (rancher-windows-gmsa/templates/_helpers.tpl:37:4): CA Bundle secret 'gmsa-server-cert' in namespace 'default' is empty
arvindiyengar: ~/Rancher/charts/src/github.com/rancher/charts 
$ k edit secret gmsa-server-cert                                                  
secret/gmsa-server-cert edited
arvindiyengar: ~/Rancher/charts/src/github.com/rancher/charts 
$ helm upgrade --install rancher-windows-gmsa packages/rancher-windows-gmsa/charts
Error: UPGRADE FAILED: execution error at (rancher-windows-gmsa/templates/_helpers.tpl:40:4): CA Bundle secret 'gmsa-server-cert' in namespace 'default' must contain ca.crt, tls.key, and tls.cert; found the following keys in the secret: map[hi:aGVsbG8K]
arvindiyengar: ~/Rancher/charts/src/github.com/rancher/charts 
$ k edit secret gmsa-server-cert                                                  
secret/gmsa-server-cert edited
arvindiyengar: ~/Rancher/charts/src/github.com/rancher/charts 
$ helm upgrade --install rancher-windows-gmsa packages/rancher-windows-gmsa/charts
Error: UPGRADE FAILED: execution error at (rancher-windows-gmsa/templates/_helpers.tpl:40:4): CA Bundle secret 'gmsa-server-cert' in namespace 'default' must contain ca.crt, tls.key, and tls.cert; found the following keys in the secret: map[ca.crt:aGVsbG8K]
arvindiyengar: ~/Rancher/charts/src/github.com/rancher/charts 
$ k edit secret gmsa-server-cert                                                  
secret/gmsa-server-cert edited
arvindiyengar: ~/Rancher/charts/src/github.com/rancher/charts 
$ helm upgrade --install rancher-windows-gmsa packages/rancher-windows-gmsa/charts
Release "rancher-windows-gmsa" has been upgraded. Happy Helming!
NAME: rancher-windows-gmsa
LAST DEPLOYED: Thu Feb 10 14:29:06 2022
NAMESPACE: default
STATUS: deployed
REVISION: 5
TEST SUITE: None
```

aiyengar2 (approved):
LGTM 🥳

rosskirkpat (approved):
🚀

luthermonson (approved):

PR #1705, opened 2022-01-21
Add rancher-version annotation back to monitoring and webhook

aiyengar2:
Was probably removed since rancher-version annotations were added in an OOB release and the package was never updated to contain it again.


brendarearden (approved):
LGTM

PennyScissors (approved):

PR #1706, opened 2022-01-24
[2.6] Forward Port CSI 2.3.1

doflamingo721:
New 2.5 charts that were forward ported to 2.6 (hidden in 2.6):
CSI 2.3.1


aiyengar2 (changes requested):
packages changes are missing? 

Also, the first 3 commits should just be the last commit with the message `"make charts"`. The current last commit should be your first commit, and also include `packages/` changes.

samkulkarni20:
@aiyengar2 I don't think package changes are required or in fact should be made here. This is a forward port of 2.5 chart which will be hidden in Rancher 2.6. The only motive of forward porting it, is to make sure a Rancher upgrade containing the CSI 2.3.1 chart works. Users won't be able to do a fresh install of CSI 2.3.1 chart on Rancher 2.6.
If we made package changes in 2.6 similar to what we have in 2.5, then it will mess up any further 2.6 chart updates in future, won't it?
CC: @PennyScissors

aiyengar2:
Ah I misunderstood the intent of this PR @samkulkarni20 @doflamingo721! Sorry for the confusion.

However, if we need to forward port charts from `release-v2.5`, we might as well do them all at once, so I wrote https://github.com/rancher/charts/pull/1727 instead which uses `make validate` and other scripts to automatically copy over all the charts from `dev-v2.5`.

doflamingo721:
Closing this in favour of https://github.com/rancher/charts/pull/1727

PR #1711, opened 2022-01-28
Fix vSphere CSI controller ClusterRole

PhilipSchmid:
The `vsphere-csi-controller-role` ClusterRole currently misses permissions for `cnsvolumeoperationrequests` CRs. This causes volume mounts to fail.  In the ["upstream" Helm chart](https://github.com/kubernetes-sigs/vsphere-csi-driver/blame/master/manifests/vanilla/vsphere-csi-driver.yaml#L53-L55), this (& many more) permission(s) are already present since multiple months.

Thanks & regards,
Philip


Dan-Coleman:
Also an issue for us, be good to get this fixed

Twiiiiin:
If I may add, this as it is won't work anyway, the driver requests also the binding to be in the namespace "vmware-system-csi".
As it is now, the chart forces to install in kube-system, and even with this change PV creation will not work. 

A change has to be added also to the Charts.yaml, specifying the different namespace

PhilipSchmid:
Closing because Rancher doesn't seem to be interested. Use the official deployment: https://github.com/kubernetes-sigs/vsphere-csi-driver/tree/master/manifests/vanilla

PR #1722, opened 2022-02-08
[DRAFT] Istio volume mount support

brendarearden:
### Issue 
https://github.com/rancher/rancher/issues/34858


brendarearden:
These changes still result in an access denied situation for the data on the volume mount.

PR #1725, opened 2022-02-09
Kev2 cluster import fix (integrate changes)

annablender:
Address https://github.com/rancher/rancher/issues/36128.

Update AKS, EKS, and GKE operator chart versions in rancher/charts

Fixes
* https://github.com/rancher/eks-operator/pull/59
* https://github.com/rancher/aks-operator/pull/52
* https://github.com/rancher/gke-operator/pull/73


packages/rancher-aks-operator/rancher-aks-operator-crd/package.yaml
```diff
@@ -1,2 +1,2 @@
 url: https://github.com/rancher/aks-operator/releases/download/v1.0.3/rancher-aks-operator-crd-1.0.3.tgz
```
kinarashah, line 1: We'd need to change this to `v1.0.4-rc1`, https://github.com/rancher/charts/pull/1625 for reference.

packages/rancher-aks-operator/rancher-aks-operator/package.yaml
```diff
@@ -1,2 +1,2 @@
 url: https://github.com/rancher/aks-operator/releases/download/v1.0.3/rancher-aks-operator-1.0.3.tgz
-version: 100.0.2
+version: 100.0.3
```
thedadams, line 2: Bumping these versions is correct, but the versions in the above URL also need to be bumped.

Same for all three operators.

thedadams (changes requested):

thedadams (changes requested):
Sorry, this part is new and I forgot about it. The new versions of the operators and their CRDs also need to be added to [release.yaml](https://github.com/raulcabello/charts/blob/dev-v2.6/release.yaml).

thedadams (dismissed):

thedadams (dismissed):

paynejacob (approved):

annablender:
Fixed merge conflict. Changes are still the same

paynejacob (approved):

thedadams (changes requested):
There is an extra commit in this PR that should be removed.

paynejacob (approved):

thedadams (approved):

PR #1728, opened 2022-02-10
change rancher backup version to 2.1.1

raulcabello:
I don't see` Rancher Backup v2.1.1-rc1` in the app catalog. It looks like it's not displayed because of the `rc1` in the app version in the `package.yaml`
That's why I changed the app version to `v2.1.1`  in the `package.yaml`, then run make charts and now I can see it in the app catalog


sirredbeard (approved):
LGTM

sirredbeard (commented):
LGTM

sirredbeard (approved):
LGTM

PR #1732, opened 2022-02-15
[2.6.5] Add neuvector chart

selvamt94:
* Chart deploys Neuvector components with tag 5.0.0-preview.1 from docker.io registry by default. Images are publicly available, so no need to provide pull secret. Please check for docker rate limit issue if image pull fails.
* Neuvector components are deployed on neuvector-system namespace with default service account. Please create new service account and configure it on values.yaml if neuvector components need to use different service account other than default.
* Set right container runtime from container runtime section. For example, docker runtime for RKE, containerd runtime for RKE2+k8s, k3s for K3s


cbron:
This shouldn't merge until after 2.6.4 is released. We can get someone from team1 to review if needed.

charts/neuvector/1.8.901+up1.8.9/templates/crd.yaml
```diff
@@ -0,0 +1,926 @@
+{{- if .Values.crdwebhook.enabled -}}
```
cbron, line 1: The CRDs should be split into a secondary chart. It sounds like there is already a CRD-subchart separately, so that can be moved into the second chart. I'll link more info.

packages/neuvector/generated-changes/patch/values.yaml.patch
```diff
@@ -0,0 +1,58 @@
+--- charts-original/values.yaml
++++ charts/values.yaml
+@@ -4,8 +4,8 @@
+ 
+ openshift: false
+ 
+-registry: registry.neuvector.com
+-tag: 4.4.3
++registry: docker.io
```
cbron, line 9: Need to use system_default_registry: https://github.com/rancher/charts/blob/b55e6ec01907964723942fdbd9f63b9afef7a1ed/packages/rancher-cis-benchmark/charts/templates/_helpers.tpl#L6-L12

packages/neuvector/generated-changes/patch/values.yaml.patch
```diff
@@ -0,0 +1,58 @@
...
++tag: 5.0.0-preview.1
+ oem:
+ imagePullSecrets:
+ psp: false
+@@ -20,7 +20,7 @@
+       maxSurge: 1
+       maxUnavailable: 0
+   image:
+-    repository: neuvector/controller
++    repository: neuvector/controller.preview
```
cbron, line 19: All of these need to be in format of image: repository/tag
https://github.com/rancher/system-charts/blob/cb84791246fad83dd4bfe909fc7245ea23f69fa1/charts/rancher-monitoring/v0.1.3/charts/prometheus/templates/prometheus.yaml#L26

packages/neuvector/generated-changes/patch/values.yaml.patch
```diff
@@ -0,0 +1,58 @@
...
+   replicas: 3
+   disruptionbudget: 0
+   schedulerName:
+@@ -154,7 +154,7 @@
+   # If false, enforcer will not be installed
+   enabled: true
+   image:
+-    repository: neuvector/enforcer
++    repository: neuvector/enforcer.preview
```
cbron, line 28: All of these also need to be either in rancher dockerhub or mirrored to rancher dockerhub. I believe we said the latter, so we need to make those repo's and get them mirrored.

charts/neuvector/1.8.901+up1.8.9/values.yaml
```diff
@@ -0,0 +1,287 @@
...
+# Declare variables to be passed into the templates.
+
+openshift: false
+
+registry: docker.io
+tag: 5.0.0-preview.1
+oem:
+imagePullSecrets:
+psp: false
+serviceAccount: default
```
cbron, line 12: Does this mean its using the default service account ? It needs to use a custom one or it will break CIS benchmarks.

charts/neuvector/1.8.901+up1.8.9/Chart.yaml
```diff
@@ -0,0 +1,28 @@
...
+  catalog.cattle.io/ui-component: neuvector
+  catalog.cattle.io/upstream-version: 1.8.9
+apiVersion: v1
+appVersion: 4.4.3
+description: Helm chart for NeuVector's core services
+home: https://neuvector.com
+icon: https://avatars2.githubusercontent.com/u/19367275?s=200&v=4
+keywords:
+- security
+kubeVersion: '>=1.18.0-0'
```
cbron, line 19: We should remove kubeVersion and only use `catalog.cattle.io/kube-version`. 
The reason is because if user has NV installed, and upgrades their chart but not K8s, the standard helm kubeVersion will block edits of the chart, but ours only blocks install.

charts/neuvector/1.8.901+up1.8.9/Chart.yaml
```diff
@@ -0,0 +1,28 @@
...
+maintainers:
+- email: support@neuvector.com
+  name: becitsthere
+name: neuvector
+sources:
+- https://github.com/neuvector/neuvector
+- https://github.com/neuvector/manager
+- https://github.com/neuvector/scanner
+version: 1.8.901+up1.8.9
```
cbron, line 28: @aiyengar2 should these be 100.0.0+up1.8.9 ? Its a brand new chart, so we could also start at 1.0.0. But I think it should be 100 to be normal.

charts/neuvector/1.8.901+up1.8.9/templates/crd.yaml
cbron: I believe you can use `generateCRDChart` to do this still. Check other charts package.yaml.

cbron (changes requested):

charts/neuvector/1.8.901+up1.8.9/templates/crd.yaml
garyduan: There is a separate CRD chart, the purpose of that chart is to allow users to install CRD before installing NeuVector. The core chart has everything, so they can do the installation in one shot.

packages/neuvector/generated-changes/patch/values.yaml.patch
garyduan: In the value.yaml, we intentionally make the tag and repo in separate values. This is to prevent users from giving different release versions to controller/enforcer/manager accidentally.

packages/neuvector/generated-changes/overlay/app-readme.md
```diff
@@ -0,0 +1,14 @@
...
+NeuVector integrates tightly with Rancher and Kubernetes to extend the built-in security features for applications that require defense in depth. Security features include:
+
++ Build phase vulnerability scanning with Jenkins plug-in and registry scanning
++ Admission control to prevent vulnerable or unauthorized image deployments using Kubernetes admission control webhooks
++ Complete run-time scanning with network, process, and file system monitoring and protection
++ The industry's only layer 7 container firewall for multi-protocol threat detection and automated segmentation
++ Advanced network controls including DLP detection, service mesh integration, connection blocking and packet captures
++ Run-time vulnerability scanning and CIS benchmarks
+
+Please Note: Before installing this chart, you will need to get an image pull secret and license key from NeuVector. Without this data supplied, the chart will not work. Configure correct container runtime and runtime path.
```
cbron, line 14: Is this true ? I thought the open-source version worked out of the box. If so, where can they get this from ? 

Also the image pull secret would be invalid because images moved to system_default_registry. Looks like it was removed below.

packages/neuvector/generated-changes/overlay/questions.yaml
```diff
@@ -0,0 +1,224 @@
...
+  label: Manager Image Tag
+  group: "Container Images"
+- variable: enforcer.image.repository
+  default: "neuvector/enforcer.preview"
+  description: enforcer image repository
+  type: string
+  label: Enforcer image path
+  group: "Container Images"
+- variable: enforcer.image.tag
+  default: "5.0.0-preview.3"
```
cbron, line 40: Preview 3 ?

packages/neuvector/generated-changes/overlay/crds/crd.yaml
```diff
@@ -0,0 +1,1104 @@
+{{- if .Values.crdwebhook.enabled -}}
```
cbron, line 1: Is this file actually how the crd chart is being generated ? I was thinking it was because of the crd folder in upstream: https://github.com/neuvector/neuvector-helm/tree/master/charts/crd

cbron (commented):
Few more questions then lgtm

packages/neuvector/generated-changes/patch/Chart.yaml.patch
```diff
@@ -0,0 +1,32 @@
...
++annotations:
++  catalog.cattle.io/auto-install: neuvector-crd=match
++  catalog.cattle.io/certified: rancher
++  catalog.cattle.io/display-name: NeuVector
++  catalog.cattle.io/kube-version: '>=1.18.0-0'
++  catalog.cattle.io/namespace: cattle-neuvector-system
++  catalog.cattle.io/os: linux
++  catalog.cattle.io/permit-os: linux
++  catalog.cattle.io/provides-gvr: neuvector.com/v1
++  catalog.cattle.io/rancher-version: '>= 2.6.5-0'
```
PennyScissors, line 13: this should be `catalog.cattle.io/rancher-version: '>= 2.6.5-0 <= 2.6.100-0'`

packages/neuvector/generated-changes/patch/Chart.yaml.patch
```diff
@@ -0,0 +1,32 @@
...
++++ charts/Chart.yaml
+@@ -1,10 +1,28 @@
++annotations:
++  catalog.cattle.io/auto-install: neuvector-crd=match
++  catalog.cattle.io/certified: rancher
++  catalog.cattle.io/display-name: NeuVector
++  catalog.cattle.io/kube-version: '>=1.18.0-0'
++  catalog.cattle.io/namespace: cattle-neuvector-system
++  catalog.cattle.io/os: linux
++  catalog.cattle.io/permit-os: linux
```
PennyScissors, line 11: this should be `catalog.cattle.io/permits-os: linux,windows` and you will need to make sure the appropriate node-selectors/tolerations are set so the neuvector pods are not scheduled on Windoes nodes.

packages/neuvector/generated-changes/patch/Chart.yaml.patch
```diff
@@ -0,0 +1,32 @@
+--- charts-original/Chart.yaml
++++ charts/Chart.yaml
+@@ -1,10 +1,28 @@
++annotations:
++  catalog.cattle.io/auto-install: neuvector-crd=match
++  catalog.cattle.io/certified: rancher
++  catalog.cattle.io/display-name: NeuVector
++  catalog.cattle.io/kube-version: '>=1.18.0-0'
```
PennyScissors, line 8: Does neuvector support k8s 1.22? if it doesn't make sure to set this to `catalog.cattle.io/kube-version: '>=1.18.0-0 < 1.22.0-0`

packages/neuvector/generated-changes/overlay/app-readme.md
selvamt94: Yes, mistake.  we will remove the pull secret requirement line from the app-readme.

packages/neuvector/generated-changes/overlay/questions.yaml
selvamt94: Somehow question.yaml is not matching the tag value from values.yaml. The good part is Rancher UI is showing tag value from values.yaml (5.0.0-b1) even though question.yaml has a different value(5.0.0-preview.3).

We will correct it to 5.0.0-b1

packages/neuvector/generated-changes/overlay/crds/crd.yaml
selvamt94: This crd is the same one that is available in the core chart itself under template. We are using crd.yaml from the core chart for building the feature chart.

This separate crd chart is useful for the customer who wants to deploy CRs from the CRD before deploying core components.

packages/neuvector/generated-changes/patch/Chart.yaml.patch
selvamt94: we will update it.

packages/neuvector/generated-changes/patch/Chart.yaml.patch
selvamt94: we agreed with @cbron to use only linux for permit-os now because Neuvector supports only Linux nodes.

packages/neuvector/generated-changes/patch/Chart.yaml.patch
selvamt94: Neuvector supports 1.22 as well 1.23.

packages/neuvector/generated-changes/overlay/app-readme.md
cbron: The license key too ?

packages/neuvector/generated-changes/overlay/questions.yaml
cbron: Default may be overriden by values.yaml, you could just remove it to avoid confusion.

packages/neuvector/generated-changes/patch/Chart.yaml.patch
cbron: Main question here is how NV reacts to a cluster with Windows nodes, since unlike most apps we have, it needs to interact with those nodes. Since we don't know for sure, I think its best to hold off for now on allowing that use case.

charts/neuvector/100.0.0+up2.2.0-b1/values.yaml
```diff
@@ -0,0 +1,295 @@
...
+    replicas: 3
+    dockerPath: ""
+    strategy:
+      type: RollingUpdate
+      rollingUpdate:
+        maxSurge: 1
+        maxUnavailable: 0
+    image:
+      repository: rancher/mirrored-neuvector-scanner
+      tag: latest
```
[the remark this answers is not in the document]
cbron: Yes its fine, as long as it works with images.txt generator.

packages/neuvector/generated-changes/overlay/crds/crd.yaml
cbron: Ah ok, so it actually comes from the overlay then. Which means every time that file changes you'll have to update the overlay.

packages/neuvector/generated-changes/overlay/app-readme.md
selvamt94: Yes, license key requirements will also be removed. Shall I update the PR with all these changes now or wait for more reviews from others too?

packages/neuvector/generated-changes/patch/Chart.yaml.patch
PennyScissors: @cbron We should still add windows to it in order to make neuvector deployable in a hybrid RKE2 cluster. Take for example the CIS benchmark chart which does not support Windows nodes, but you can still deploy it in a hybrid cluster due to these node-selectors/tolerations:

https://github.com/rancher/charts/blob/eaba8696d0e569a64c45ef7f610b501528daae9d/charts/rancher-cis-benchmark/2.0.3/templates/deployment.yaml#L44-L51

https://github.com/rancher/charts/blob/eaba8696d0e569a64c45ef7f610b501528daae9d/charts/rancher-cis-benchmark/2.0.3/templates/_helpers.tpl#L14-L27

packages/neuvector/generated-changes/overlay/app-readme.md
cbron: Either works, typically people just push updates consistently

selvamt94: I have updated the changes except for this comment https://github.com/rancher/charts/pull/1732#discussion_r848901197

packages/neuvector/generated-changes/patch/Chart.yaml.patch
cbron: We could try it. My main point above was that those other apps just deal with the linux nodes they get scheduled on. Where as NeuVector might interact with the traffic of all nodes in the hybrid cluster, and thats the real unknown part, how NeuVector reacts in a hybrid cluster at all, regardless of which node in the cluster it gets scheduled on.

cbron (approved):
lgtm, there is open conversation on mixed windows clusters, but I'd like to address that in follow-up PR when we move 5.0.0-b1 to 5.0.0

PennyScissors (approved):
Approving since everything else looks good to me, but there is still this question that need to be addressed before merging.
https://github.com/rancher/charts/pull/1732#discussion_r851048894

cbron:
Issue: https://github.com/rancher/rancher/issues/37377

packages/neuvector/generated-changes/patch/Chart.yaml.patch
garyduan: We do have partial deployment cases, where not all nodes have enforcers running. The network rules are learned/defined and enforced accordingly. Having said that, the network function works the best in full deployment. I suggest we disable neuvector in mixed environment for now, and update the chart later.

PR #1734, opened 2022-02-15
Add commonLabels to pushProxy subcharts

atsai1220:
This should solve: https://github.com/rancher/rancher/issues/39427

This will ensure push proxy subcharts have the option to use `commonLabels`. This allows teams to scope the ServiceMonitors Prometheus is targeting. Otherwise using `prometheus.prometheusSpec.serviceMonitorSelector` and `commonLabels` in the main chart is not viable as ServiceMonitors by subcharts are not included.


recena:
Closing since we don't plan more changes in `2.6.x`.

PR #1736, opened 2022-02-16
Adding permits-os annotation to SRIOV

luthermonson:
This chart needs no node selectors as the operator already has a `node-role.kubernetes.io/control-plane: "true"` assignment from the chart itself.

https://github.com/rancher/dashboard/issues/5137 https://github.com/rancher/rancher/issues/36401


aiyengar2 (approved):

luthermonson:
superceded by https://github.com/rancher/charts/pull/1779

PR #1737, opened 2022-02-16
Add permits-os to System Upgrade Controller

luthermonson:
This chart already has a node affinity for the controller to run on any other OS but windows so the only patch needed is the annotation.

https://github.com/rancher/dashboard/issues/5137 https://github.com/rancher/rancher/issues/36401


aiyengar2 (approved):
LGTM

luthermonson:
superceded by https://github.com/rancher/charts/pull/1779

PR #1745, opened 2022-02-17
Adding permits-os annotations to Longhorn charts

phillipsj:
This adds the permits-os annotations.

* https://github.com/rancher/rancher/issues/36401
* https://github.com/rancher/dashboard/issues/5137


aiyengar2 (changes requested):
where were the changes made to allow Longhorn to be deployed onto a Windows cluster? I don't see nodeSelectors or tolerations

innobead:
Please refer to the previous discussion, https://github.com/rancher/charts/pull/1144#discussion_r622620753. 

If allowing Longhorn display on the mixed cluster, users need to set up node selector/tolerations by themselves, so probably this will cause some confusion because actually Longhorn only works for Linux.

cc @shuo-wu @khushboo-rancher @kaxing

innobead:
BTW, one question: will `catalog.cattle.io/os` be deprecated?

phillipsj:
Replaced by https://github.com/rancher/charts/pull/1783

PR #1758, opened 2022-02-18
Eks security group fix

annablender:
Issue: https://github.com/rancher/rancher/issues/34749

Fixed issue in `eks-operator` https://github.com/rancher/eks-operator/pull/60. Tagged a new rc and updated the operator and crd versions in `rancher/charts`.

This also makes updates to the chart for https://github.com/rancher/dashboard/issues/5137 and https://github.com/rancher/rancher/issues/36401


thedadams (changes requested):
When upgrading RC versions (not adding a new version with an RC), the old RCs need to be removed from `charts` and `assets` as a part of the first commit.

thedadams (approved):

paynejacob (approved):

PR #1712, opened 2022-01-28
Add rancher metrics

rmweir:
**Description:**
Added service monitor for rancher. This should provide access to prometheus-client metrics in rancher.

**Issue:**
https://github.com/rancher/rancher/issues/36234

PR #1766, opened 2022-02-24
Support monitoring on k8s 1.22 RKE1 and RKE2 clusters and support pushprox in SELinux

aiyengar2:
### Problem 1
Starting in kubernetes 1.22, insecure ports are no longer supported for kube-controller-manager, so secure port support was added to the latest RKE 1.22.x. 

This is causes issues because rancher-monitoring needs to support several versions of k8s but previous versions of RKE do not support connection on secure ports for these resources. 

### Solution 1
Add a `kubeVersionOverrides` field that allows certain `.Values` to be overridden depending on the k8s version. 

The kubeVersionOverride has been set to override with insecure ports for previous versions of rke / rke2 kube-controller-manager.

Related Issues: https://github.com/rancher/rancher/issues/35571, https://github.com/rancher/rancher/issues/35751, https://github.com/rancher/rancher/issues/36522

Followup to https://github.com/rancher/charts/pull/1647

### Problem 2
rancher-pushprox uses a hostPath volume but does not define an seLinux type for the container process that allows the container to read the files in `/etc/kubernetes/ssl`.

### Solution 2
Add a new type in https://github.com/rancher/rancher-selinux/pull/11 and allow it to be passed into pushprox.

Relies on https://github.com/rancher/rancher-selinux/pull/11

Related Issue: https://github.com/rancher/rancher/issues/36105


packages/rancher-monitoring/rancher-service-metrics/charts/templates/servicemonitor.yaml
```diff
@@ -1,6 +1,9 @@
 {{- $chartLabel := .Values.chartLabel }}
 {{- if eq $chartLabel "" }}
-{{- $chartLabel = (index (lookup "apps/v1" "Deployment" "cattle-system" "rancher").metadata.labels "chart") }}
+{{- $rancher := (lookup "apps/v1" "Deployment" "cattle-system" "rancher") }}
+{{- if $rancher -}}
+{{- $chartLabel = (index $rancher.metadata.labels "chart") }}
+{{- end }}
```
aiyengar2, line 6: cc: @rmweir, I caught this after trying to deploy Monitoring. Seems to be a bug we missed in https://github.com/rancher/charts/pull/1712

packages/rancher-monitoring/rancher-monitoring/generated-changes/patch/values.yaml.patch
```diff
@@ -231,12 +248,25 @@
...
++  - constraint: "< 1.22"
++    values:
++      metricsPort: 10252 # default to insecure port in k8s < 1.22
++      clients:
++        https:
++          enabled: false
++          insecureSkipVerify: false
++          useServiceAccountCredentials: false
 +
 +rke2Scheduler:
```
aiyengar2, line 261: Q: why are we modifying only kubeScheduler in RKE2 and not RKE1?

A: RKE1 has not moved to using the secure port yet for scheduler, only for controller manager. When RKE1 moves to using the scheduler port in a future k8s release (probably 1.23), we will need to fix this.

prachidamle:
@aiyengar2 Will the limitation of using insecure ports affect any other charts other than monitoring?

aiyengar2:
> @aiyengar2 Will the limitation of using insecure ports affect any other charts other than monitoring?

I doubt so @prachidamle, monitoring is the only chart that I'm aware of that needs to hit the endpoints for internal Kubernetes components. Other than Monitoring, only provisioning / rke / rke2 should be affected afaik.

PennyScissors (dismissed):
lgtm, good job on the values override, cool stuff.

PennyScissors (dismissed):

packages/rancher-monitoring/rancher-service-metrics/charts/templates/servicemonitor.yaml
aiyengar2: Fixed in https://github.com/rancher/charts/pull/1768/commits/4f04899ceec6190abe5abbe8081a8f2d6bf88c8f, dropped from this PR.

prachidamle (dismissed):

PennyScissors (approved):

PR #1768, opened 2022-02-24
Add performance dashboards

rmweir:


jiaqiluo (approved):
Both the changes for checking rancher deployment and adding performance dashboards LGTM.
Need to resolve the conflicts.

charts/rancher-monitoring/100.1.1+up19.0.3/templates/rancher-monitoring/dashboards/rancher/performance-dashboards.yaml
```diff
@@ -0,0 +1,17 @@
...
+{{ toYaml .Values.grafana.sidecar.dashboards.annotations | indent 4 }}
+  labels:
+    {{- if $.Values.grafana.sidecar.dashboards.label }}
+    {{ $.Values.grafana.sidecar.dashboards.label }}: "1"
+    {{- end }}
+    app: {{ template "kube-prometheus-stack.name" $ }}-grafana
+{{ include "kube-prometheus-stack.labels" $ | indent 4 }}
+data:
+{{ (.Files.Glob "files/rancher/performance/*").AsConfig | indent 2 }}
+{{- end }}
```
cbron, line 17: This means we install the dashboard even if the performance system is not turned on. Which I suppose makes sense, because you'd turn it on/off with the env var only.

cbron (dismissed):
May need some fine-tuning of graph titles eventually. Seems to follow standard practices.

charts/rancher-monitoring/100.1.1+up19.0.3/templates/rancher-monitoring/dashboards/rancher/performance-dashboards.yaml
rmweir: @cbron yeah, it's not a super big deal if the data isn't there. The graphs are just empty in that case.

cbron (approved):

packages/rancher-monitoring/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/dashboards/rancher/performance-dashboards.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if and .Values.grafana.enabled .Values.grafana.defaultDashboardsEnabled }}
```
aiyengar2, line 1: this should only be added if rancherServiceMetrics is enabled right?

aiyengar2 (changes requested):
Can you also paste a screenshot of the dashboards that are produced onto this PR?

packages/rancher-monitoring/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/dashboards/rancher/performance-dashboards.yaml
rmweir: @aiyengar2 I don't think it really matters, worse case you can look at the dashboard, it is just empty, but I can add it. Although there is a value I think most people would toggle the feature using the environment variable.

cbron: People would follow this guide to turn it on: https://github.com/rancher/rancher/wiki/How-to-setup-prometheus-dashboard-for-controllers

We def don't want them turning it on/off by upgrading the chart. However it could be default off, then an upgrade to on for initial capture, then use env var from there on out. Majority of people, I would assume 95% or more, would not use this graph. 

The new check is on `.Values.rancherServiceMetrics.enabled` which we just need to test

rmweir:
@aiyengar2 graphs:
image
image
image
image
image
image
image
image

packages/rancher-monitoring/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/dashboards/rancher/performance-dashboards.yaml
rmweir: Added the `.Values.rancherServiceMetrics.enabled` check, tested, and appears to work.

aiyengar2: @rmweir thanks, sounds good.

@cbron to be clear (since I'm not sure if that was your impression), turning on or off the Grafana dashboard won't cause you to lose any data. The data is persisted in Prometheus; Grafana is just making a call to Prometheus and presenting it in a fancy dashboard. So upgrading the chart and no longer seeing the dashboard followed by upgrading the chart to enable it again would only cause you to lose data for the time that you turned it off and then turned it back on (because the ServiceMonitor disappeared, not because the dashboard did).

aiyengar2 (approved):
LGTM

PR #1774, opened 2022-02-24
bump rancher-webhook to v0.2.5-rc2

kinarashah:
https://github.com/rancher/rancher/pull/36525


packages/rancher-webhook/package.yaml
```diff
@@ -1,2 +1,2 @@
-url: https://github.com/rancher/webhook/releases/download/v0.2.5-rc1/rancher-webhook-0.2.5-rc1.tgz
-version: 1.0.4
+url: https://github.com/rancher/webhook/releases/download/v0.2.5-rc2/rancher-webhook-0.2.5-rc2.tgz
+version: 1.0.5
```
jiaqiluo, line 2: Just curious: I noticed the latest released version is rancher-webhook-1.0.2+up0.2.2, so is there any reason we skip 1.0.3 and 1.0.4?
https://github.com/rancher/charts/tree/release-v2.6/assets/rancher-webhook

jiaqiluo (changes requested):

packages/rancher-webhook/package.yaml
kinarashah: Oh, I am not sure - I only bumped the existing version. 
https://github.com/rancher/charts/pull/1704/files removed v1.0.3 and added v1.0.4, do we need to remove version with each update for RC?

jiaqiluo: > do we need to remove the version with each update for RC?
yes, we removed the old RC each time when adding a new RC. 

I think there  is something out of sync: 
 `1.0.3+up0.2.4` was released in the branch `release-v2.6.3-patch1`, but `release-v2.6.3-patch1` is not merged back to `release-v2.6`: https://github.com/rancher/charts/compare/release-v2.6...release-v2.6.3-patch1
 
I am not sure if it is just forgotten or there is some specific reason for it.

aiyengar2: I don't think we intend for 1.0.3+up0.2.4 to ever be accessible in our normal release line for Rancher 2.6 since it was built specifically for 2.6.3-patch1, which does not follow our normal release cycle.

We could bring it into release-v2.6, but I'm not sure if there's a benefit. cc: @cbron what are your thoughts?

kinarashah: @jiaqiluo synced offline with @aiyengar2, we're keeping `1.0.3` separate for `2.6.3-patch1` and we'll decide if needs to be ported to `release-v2.6`. I pushed changes to now use `1.0.4+up0.2.5-rc2` for Rancher v2.6.4.

aiyengar2 (approved):

jiaqiluo (approved):

PR #1777, opened 2022-02-28
Bump aks operator and crd versions

annablender:
Issue: https://github.com/rancher/rancher/issues/36425

Bump AKS operator and crd versions to [v1.0.4-rc3](https://github.com/rancher/aks-operator/releases/tag/v1.0.4-rc3) to get oms agent panic fix.


release.yaml
```diff
@@ -59,6 +59,7 @@ rancher-eks-operator-crd:
 - 100.0.2+up1.1.3-rc2
 rancher-aks-operator:
 - 100.0.3+up1.0.4-rc1
```
paynejacob, line 61: why is rc1 still in here?

paynejacob (changes requested):
This pr should be in 2 commits, the first removes the old version and updates the package.  The second runs make charts.

release.yaml
thedadams: It is correct for the old RC versions to be in the release.yaml file.

annablender: @thedadams Ok thanks!

thedadams (approved):

paynejacob (approved):

PR #1779, opened 2022-03-01
Super PR for Permits OS Annotions

luthermonson:
https://github.com/rancher/rancher/issues/36401 https://github.com/rancher/dashboard/issues/5137 https://github.com/rancher/dashboard/issues/5172


phillipsj (dismissed):
LGTM!

release.yaml
```diff
@@ -78,6 +78,10 @@ rancher-cis-benchmark-crd:
 system-upgrade-controller:
 - 100.0.2+up0.3.2
 rancher-cis-benchmark:
-- 2.0.3
+- 2.0.3-rc1
 rancher-cis-benchmark-crd:
-- 2.0.3
+- 2.0.3-rc1
```
aiyengar2, line 102: should be dropped? why is this in this commit?

release.yaml
```diff
@@ -78,6 +78,10 @@ rancher-cis-benchmark-crd:
...
+rancher-gatekeeper:
+- 100.0.1+up3.6.0
+rancher-gatekeeper-crd:
+- 100.0.1+up3.6.0
```
aiyengar2, line 106: these versions have already been released, by adding them to the release.yaml you're basically saying that you want to modify an already released chart (which is not what you are intending; you want to create a new release).

package.yaml needs to be bumped to 100.0.2

release.yaml
```diff
@@ -12,6 +12,7 @@ longhorn-crd:
 - 1.2.300+up1.2.3
 rancher-alerting-drivers:
 - 1.0.301
+- 100.0.1
```
aiyengar2, line 19: already been released, this should be 100.0.2

release.yaml
```diff
@@ -23,6 +23,14 @@ rancher-backup:
...
 - 1.2.100+up1.2.1
 - 2.1.1+up2.1.1-rc1
+harvester-cloud-provider:
+- 100.0.1+up0.1.8
+- 100.0.0+up0.1.8
+- 100.0.0+up0.1.8
+harvester-csi-driver:
+- 100.0.1+up0.1.9
+- 100.0.0+up0.1.9
+- 100.0.0+up0.1.9
```
aiyengar2, line 33: why are there so many? There should just be `100.0.1+up0.1.9`

packages/rancher-webhook/generated-changes/patch/Chart.yaml.patch
```diff
@@ -0,0 +1,10 @@
+--- charts-original/Chart.yaml
++++ charts/Chart.yaml
+@@ -3,6 +3,7 @@
+   catalog.cattle.io/hidden: "true"
+   catalog.cattle.io/namespace: cattle-system
+   catalog.cattle.io/os: linux
++  catalog.cattle.io/permits-os: linux,windows
+   catalog.cattle.io/release-name: rancher-webhook
+ apiVersion: v2
+ appVersion: 0.2.5-rc1
```
aiyengar2, line 10: this might need to be redone? I see we already have an rc2 https://github.com/rancher/charts/tree/dev-v2.6/charts/rancher-webhook/1.0.4%2Bup0.2.5-rc2

release.yaml
```diff
@@ -106,3 +106,8 @@ rancher-gatekeeper-crd:
 - 100.0.1+up3.6.0
 system-upgrade-controller:
 - 100.0.2+up0.3.1
+- 100.0.2+up0.3.2
+sriov:
+- 100.0.2+up0.1.0
+sriov-crd:
+- 100.0.2+up0.1.0
```
aiyengar2, line 113: was this supposed to contain changes for sriov? confused about this commit, seems like it should be merged into the next one

aiyengar2 (changes requested):
@luthermonson it might just be easier to drop all the release.yaml changes from each of the commits and group them up into 1 commit with all release.yaml changes at the end (before the `make charts` commit at the end).

Spotted some errors along the way, commented below.

release.yaml
luthermonson: Someone else did it, it's in Dev-2.6 now

release.yaml
luthermonson: ya merged bad, ill fix

PR #1781, opened 2022-03-01
Bump backup restore 2.1.1 rc2

thedadams:
Issue:
https://github.com/rancher/rancher/issues/36062


release.yaml
```diff
@@ -15,9 +15,11 @@ rancher-alerting-drivers:
 rancher-backup:
 - 1.2.100+up1.2.1
 - 2.1.1+up2.1.1-rc1
+- 2.1.1+up2.1.1-rc2
 rancher-backup-crd:
 - 1.2.100+up1.2.1
 - 2.1.1+up2.1.1-rc1
+- 2.1.1+up2.1.1-rc2
```
jiaqiluo, line 22: Since `2.1.1+up2.1.1-rc1` is dropped in this PR, I guess we should remove it from the list?

jiaqiluo (changes requested):

release.yaml
thedadams: No, we can't. Validation fails. All old RCs must appear here.

jiaqiluo: OK. 
LGTM if that is how chart-builder works today.

jiaqiluo (approved):

paynejacob (approved):

PR #1794, opened 2022-03-08
Add longhorn 1.2.4 chart

innobead:

cc @khushboo-rancher @kaxing @meldafrawi @cchien816 @yangchiu

ref: https://github.com/rancher/rancher/issues/36796


PennyScissors (commented):
LGTM except for the chart's version and image tags which I assume will be updated to non-rc versions once v1.2.4 is released.

innobead:
> LGTM except for the chart's version and image tags which I assume will be updated to non-rc versions once v1.2.4 is released.

Correct, it will be updated when the upstream 1.2.4 is released.

cchien816:
V2 Chart Testing - v2.6.x Rancher Explorer
Validate result : **Pass**

Tested with https://github.com/innobead/charts-1.git branch longhorn-v1.2.4-rancher-2.6

- Fresh installation
- Uninstallation
- Upgrade from v1.2.3 - > v1.2.4-rc1
- Upgrade from v1.1.3 → v1.2.4-rc1

innobead:
@PennyScissors Please help review again, then merge after @khushboo-rancher verify the final update. Thank you.

packages/longhorn/longhorn-1.2/package.yaml
```diff
@@ -1,7 +1,7 @@
 url: https://github.com/longhorn/charts.git
 subdirectory: charts/longhorn
-commit: d699089b1a62011c9cdeb6e8dacba203f30655a8
-version: 100.1.2
+commit: d3ac3c2ee47cd9bd420b36cdef0d4e4fb24f9622
+version: 100.1.3
```
PennyScissors, line 4: Looks like the longhorn chart version we last released is `100.1.1+up1.2.3` for the 1.2 version line, so you don't need to bump this again if it was bumped already. The next to-be-released version should be one step up from the previous, so that would be `100.1.2+up1.2.4`. You need to remove the entries in release.yaml and assets associated with `100.1.2+up1.2.3` and `100.1.3+up1.2.4`, and generate new ones using `100.1.2` instead to get `100.1.2+up1.2.4`.

PennyScissors (changes requested):

khushboo-rancher:
Verified with mirrored image v1.2.4 - https://github.com/rancher/image-mirror/pull/221 with the repo `longhorn-v1.2.4-rancher-2.6` branch - `longhorn-v1.2.4-rancher-2.6`

Functionality wise - **Pass** , I have the same doubt about the version as @PennyScissors mentioned.
Screen Shot 2022-03-17 at 11 57 20 AM

Validated installation, uninstallation, upgrade and P0 cases.

packages/longhorn/longhorn-1.2/package.yaml
innobead: It seems there is an intermediate change (https://github.com/rancher/charts/commit/41c2d8dd4fde86e477f877fc0df57fb865197130) to cause another `100.1.2+up1.2.3`. Fixed.

khushboo-rancher:
Version naming of Longhorn v1.2.4 is fixed now.
Screen Shot 2022-03-17 at 5 24 55 PM

PennyScissors (approved):

shuo-wu (commented):
I have no idea of why we need to remove `100.1.2+up1.2.3` for Longhorn v1.2.4 and why we add it besides `100.1.1+up1.2.3` before...

innobead:
> I have no idea of why we need to remove `100.1.2+up1.2.3` for Longhorn v1.2.4 and why we add it besides `100.1.1+up1.2.3` before...

https://github.com/rancher/charts/pull/1794#discussion_r829607099, but this is more like an intermediate in between, so get people confused.

jehof:
Will longhorn 1.2.4 also be released for rancher 2.5. please make it possible.

PR #1721, opened 2022-02-07
[2.6] Update `istio-install-job.yaml` to match new nginx user id and group id.

rayandas:
related: https://github.com/rancher/istio-installer/pull/45
SLE BCI istio installer image has a different nginx user id and group id (not 101). So we were getting permission issues as on the istio-installer pod, it was running as uid 101 and not nginx(uid 499 on SLE BCI image). This PR will fix that.

- [x] update the new istio-installer image tag in `values.yaml` when [it's](https://github.com/rancher/istio-installer/pull/45) merged.

PR #1795, opened 2022-03-08
[2.5] Update `istio-install-job.yaml` to match new nginx user id and group id.

rayandas:
SLE BCI istio installer image has a different Nginx user id and group id (not 101). It was set to 101 and because of this there were errors in istio-installer job.
This PR updates the Nginx uid and gid to 499 and 487 [similar to this](https://github.com/rancher/charts/pull/1721). 
Fixes: https://github.com/rancher/rancher/issues/36767


jiaqiluo (approved):
LGTM

PR #1798, opened 2022-03-09
fixes the path to Harvester's cloud provider configuration

belgaied2:
As mentioned in this issue: https://github.com/rancher/rancher/issues/36309 or this one: https://github.com/rancher/rancher/issues/36833 , the Harvester Cloud Provider configuration in the HostPath scenario is at `/var/lib/rancher/rke2/etc/config-files/cloud-provider-config` not `/etc/kubernetes/cloud-config`.


belgaied2:
Got information from Harvester team that the issue is already fixed somewhere else and that this change might break something in RKE. Thus closing this pull request.

PR #1802, opened 2022-03-11
[dev-v2.6] Bump backup-restore-operator to v2.1.1-rc3

superseb:
https://github.com/rancher/backup-restore-operator/issues/165


pennyscissors (dismissed):

PennyScissors:
Oops, I approved and then thought of this. Can you remove the rc2 entries for the chart and crd in the release.yaml file [here](https://github.com/superseb/charts/blob/03fd59a96f2907d472a6e6c45445dc2e12a72ded/release.yaml#L20-L27) since rc3 would replace it?

superseb:
@PennyScissors This was based on https://github.com/rancher/charts/pull/1781#discussion_r817192394 and the fact it doesn't validate when I remove it. Also the `rancher-*-operator` list multiple RC's so something is going on here.

PennyScissors:
@superseb I was not aware of that, you can leave it in there for now then since it won't really cause any harm.

aiyengar2:
@superseb I should have fixed that issue already so it should be safe to remove RCs now. The issue we saw originally was caused by two issues:
1) Our configuration.yaml was wrong: https://github.com/rancher/charts/commit/3cce697dd01467c3ac0d91ae924588ba42063064
2) Our validation process was broken in small ways, which was addressed in the v0.3.2 and v0.3.3 releases

pennyscissors (approved):

PR #1811, opened 2022-03-17
updated k8s restrictions on suc chart

paynejacob:


thedadams (dismissed):

jakefhyde (approved):

kinarashah:
@paynejacob what issue is this change for? It doesn't need to run for 1.23?

paynejacob:
@kinarashah there is no issue @aiyengar2 reached out to me about this.  We do not want any k8s restrictions on the helm chart because they are too restrictive in the future.

kinarashah (approved):

thedadams (approved):

paynejacob:
this chart is already released

PR #1822, opened 2022-03-24
Fix Istio install/upgrade fails on an air-gapped setup

rayandas:
Related: https://github.com/rancher/rancher/issues/35742

Read more about why this PR is required: https://github.com/rancher/rancher/issues/35742#issuecomment-994801502


packages/rancher-istio/rancher-istio/charts/values.yaml
```diff
@@ -66,6 +66,9 @@ global:
     enabled: true
   rbac:
     pspEnabled: true
+  # Set it false before installing/upgrading istio in an air-gapped setup. It's set to true by default.
+  configValidation:
+    enabled: true
```
aiyengar2, line 71: I don't think this change makes sense. Where is this value being used by the Helm chart? Has this change been tested?

I'd suspect you need to add this value into the global config of the IstioOperator resource, which is found in https://github.com/rancher/charts/blob/eaba8696d0e569a64c45ef7f610b501528daae9d/charts/rancher-istio/100.1.2%2Bup1.11.7/configs/istio-base.yaml#L102-L110.

But there's no field there actually referencing this value that was set. Also, I'm not sure if this field should belong in `.values.global` of the Helm chart itself.

aiyengar2 (changes requested):
Pretty sure the approach on this PR is incorrect, but please correct me if I missed something!

Also, to review this change I need to understand the context of this field and why we are providing it. I perused through the comments on Istio docs in https://istio.io/latest/docs/ops/common-problems/validation/ and through Brenda's comment in https://github.com/rancher/rancher/issues/35742#issuecomment-994801502, but what doesn't make sense to me is this comment:

> This causes problems if you attempt an uninstall and reinstall at the newest version >=1.11.x in Airgapped environments because the Istio Operator attempts to reconcile the resources that are installed and is unable to apply the necessary patches it needs to without internet access and errors.

Why do we need internet access to make the patches? I think there needs to be investigation on why Istio might have this requirement before we can pursue this path.

FYI if Istio doesn't have any specific requirements that you can find that are causing this constraint, it's also possible that what you're running into here is simply the issue we found in https://github.com/rancher/rancher/issues/32416#issuecomment-828881726 where the istiod service and deployment needs to be up before the helm install can progress.

packages/rancher-istio/rancher-istio/charts/values.yaml
rayandas: Hey, yeah I think this should be set in [charts/charts/rancher-istio/100.1.2+up1.11.7/configs/istio-base.yaml](https://github.com/rancher/charts/blob/eaba8696d0e569a64c45ef7f610b501528daae9d/charts/rancher-istio/100.1.2%2Bup1.11.7/configs/istio-base.yaml#L102-L110) as well which will use the value from values.yaml while installing istio. 
We can't actually add any new repo in apps and marketplace on an air-gapped setup, so I tested it this way: while installing Istio, in the yaml, I added the below config (set it as false because, it's set to true by default.) inside global and tried to install istio. Then again uninstalled and reinstalled/upgraded to test if it's working. Istio installation was getting installed/upgraded successfully. 
```
configValidation:
    enabled: false
```
Without setting the `configValidation` to false, only on airgapped setup it's breaking. If we set it to false while installing or upgrading it's getting installed successfully.

rayandas:
> Why do we need internet access to make the patches?

I am not sure about this, I went through istio docs and other resources to know more about this but couldn't find anything. 
This install/upgrade issue happens only on air-gapped setup, as it's a bit different than normal Rancher setup, I am suspecting that's the reason.

rayandas:
@aiyengar2 can you please review it whenever you get time?

packages/rancher-istio/rancher-istio/charts/values.yaml
aiyengar2: I see changes were added to the istio-base.yaml, resolving this comment

aiyengar2:
> I am not sure about this, I went through istio docs and other resources to know more about this but couldn't find anything.
This install/upgrade issue happens only on air-gapped setup, as it's a bit different than normal Rancher setup, I am suspecting that's the reason.

If we want to pursue a fix for this, I think we need to be sure on why the change is necessary, which might involve debugging why the EnvoyFilter created requires network access; for example, if we discover that it needs certain resources packaged upstream, the solution might be to point the EnvoyFilter at a local deployment that serves the required assets from a ConfigMap that contains it, which is generally the approach for these types of requirements in Helm charts deployed in air-gapped setups.

I don't like the approach of this current PR of just allowing a user to shut off validation, since that would mean that we're not offering validation as a feature in airgapped-setups. If there's a time-crunch to get this issue in, that would be okay as a workaround but if we can delay this issue to investigate a proper, larger fix to support validation in airgapped setups that would be better.

Talked to @SheilaghM and @MKlimuszka about this and seems like we can push out this issue to 2.6.6 to perform that investigation.

rayandas:
Alright! I will start working on it and debug as you suggested.

rayandas:
@aiyengar2 Did you go through the upstream issue that Brenda raised? https://github.com/istio/istio/issues/36526 In the logs, it seems like API calls are failing. 
I tried installing istio 1.10.4 in an airgapped setup and after uninstalling I also saw that `validatingwebhookconfiguration` `istiod-istio-system` is not getting cleaned up. From Istio 1.11+ versions a new validatingwebhookconfiguration is getting created which is `istio-validator-istio-system`.

rayandas:
Closing the issue in favour of https://github.com/rancher/istio-installer/pull/56 (which is merged and tested and released. which fixes the issue.)

PR #1835, opened 2022-03-30
Bump rancher-webhook to 1.0.4+up0.2.5

aiyengar2:
`1.0.3+up0.2.5` was incorrectly released to the release-v2.6 branch yesterday since the version expected by Rancher is `1.0.4+up0.2.5` (since `1.0.3` was already released in release-v2.6.3-patch1).

To prevent installations, `1.0.3+up0.2.5` is being hidden via chart annotation and `1.0.4+up0.2.5`, which matches it exactly with one semver version higher, is being released.


thedadams:
Issue: https://github.com/rancher/rancher/issues/37113

packages/rancher-webhook/generated-changes/patch/Chart.yaml.patch
```diff
@@ -1,10 +1,11 @@
 --- charts-original/Chart.yaml
 +++ charts/Chart.yaml
-@@ -3,6 +3,7 @@
+@@ -3,6 +3,8 @@
    catalog.cattle.io/hidden: "true"
    catalog.cattle.io/namespace: cattle-system
    catalog.cattle.io/os: linux
 +  catalog.cattle.io/permits-os: linux,windows
++  catalog.cattle.io/rancher-version: <2.5.99-0
```
thedadams: If I understand correctly, this would mean that if someone booted a new Rancher v2.5.12, then they might get this version of rancher-webook because it satisfies the min-version in Rancher.

If I am correct about that, then I think this should be something else.

thedadams (commented):
One question, otherwise looks good.

packages/rancher-webhook/generated-changes/patch/Chart.yaml.patch
aiyengar2: They won't see that @thedadams since they are pointing to the release-v2.5 branch, not release-v2.6

aiyengar2: This is the same mechanism that we use to hide rancher-vsphere-cpi 100.0.0 btw https://github.com/rancher/charts/blob/77d77f2a278514384d0da4cf03fa1700d17baf2f/index.yaml#L6677

thedadams: Ah! That makes sense. Thanks!

PennyScissors (approved):

PR #1836, opened 2022-03-30
[dev-v2.6] Add logging-operator 100.1.1+up3.17.3

superseb:
Depends on https://github.com/rancher/image-mirror/pull/228

Previous introduction of logging-operator 3.17.3 was missing accompanying images (fluentd/fluent-bit), the released version was restricted to `<2.5.99` and the new version was added with the updated images

The new fluent-bit images should also resolve https://github.com/rancher/rancher/issues/35395 and similar issues to logging not resuming after some sort of interruption


aiyengar2 (changes requested):
@superseb can you restructure your PR so that your changes make it into the `packages/` directory before they go to charts (instead of making charts PRs directly)

Similar to https://github.com/rancher/charts/pull/1835 (skipping the first commit), it should be:
- packages changes to add the annotation + release.yaml addition of the released version
- PACKAGE=rancher-logging make charts
- packages changes to bump the version and make the changes you want + release.yaml addition of the new version
- PACKAGE=rancher-logging make charts

packages/rancher-logging/generated-changes/patch/Chart.yaml.patch
```diff
@@ -10,7 +10,7 @@
 +  catalog.cattle.io/namespace: cattle-logging-system
 +  catalog.cattle.io/permits-os: linux,windows
 +  catalog.cattle.io/provides-gvr: logging.banzaicloud.io.clusterflow/v1beta1
-+  catalog.cattle.io/rancher-version: <2.5.99-0
++  catalog.cattle.io/rancher-version: '>2.6.0-0'
```
aiyengar2, line 13: nit: `'> 2.6.0-0'`, not `'>2.6.0-0'`. Space is missing compared to previous logging chart

release.yaml
```diff
@@ -1,3 +1,61 @@
+fleet:
+- 100.0.3+up0.3.9
```
aiyengar2, line 2: additional charts should be deleted from release.yaml; these were removed in the webhook PR

aiyengar2 (changes requested):
Some more changes; also can you split the first and second commit into 4 commits, in pairs of packages/ + release.yaml changes and `make charts` changes (assets, index.yaml, charts)

PennyScissors (approved):

aiyengar2 (approved):

PR #1851, opened 2022-04-11
[2.6] Add OPA-Gatekeeper 3.7.1

doflamingo721:
Update opa-gatekeeper to 3.7.1

Linked issue:
https://github.com/rancher/rancher/issues/37229

Adds:
```
rancher-gatekeeper:100.1.0+up3.7.1
rancher-gatekeeper-crd:100.1.0+up3.7.1
```


packages/rancher-gatekeeper/generated-changes/patch/values.yaml.patch
```diff
@@ -40,6 +40,7 @@
 +    systemDefaultRegistry: ""
 +  kubectl:
 +    repository: rancher/kubectl
-+    tag: v1.20.2
++    tag: v1.20.
```
aiyengar2, line 43: seems to be missing patch version?

packages/rancher-gatekeeper/generated-changes/patch/values.yaml.patch
```diff
@@ -7,31 +7,31 @@
...
+-      tag: v3.7.1
 +      repository: rancher/kubectl
 +      tag: v1.20.2
        pullPolicy: IfNotPresent
        pullSecrets: []
 -image:
 -  repository: openpolicyagent/gatekeeper
 -  crdRepository: openpolicyagent/gatekeeper-crds
--  release: v3.6.0
+-  release: v3.7.1
```
aiyengar2, line 23: this is a breaking change, we should have logic for a user to fail a helm upgrade if these values are set or backwards compatibility with these old fields. Between the two, I'd prefer the former

aiyengar2 (changes requested):
concerned about breaking changes being made to the values.yaml spec, we should introduce helm checks that make sure that a user upgrading from an older version of gatekeeper with overrides is informed about what needs to be changed by a helm failure

packages/rancher-gatekeeper/generated-changes/patch/values.yaml.patch
doflamingo721: Added the missing patch version

packages/rancher-gatekeeper/generated-changes/patch/values.yaml.patch
doflamingo721: I've added the logic to fail the helm upgrade if the field `image` is overridden. Can you please review it once?

PennyScissors: Gatekeeper introduced the CRD image and the `crdRepository` logic with version 3.5.1, but this logic was never released in any of our rancher-gatekeeper charts because it doesn't follow the pattern the airgap exporter needs and therefore the images were not being picked up. It was quickly replaced by the new logic in this PR and it's what we've been using since then so it is not a breaking change. This also doesn't affect 2.5 because the latest gatekeeper chart in 2.5 is older than 3.5.1 so it doesn't have a CRD image.

aiyengar2: Sounds good, thanks for looking into this @PennyScissors. If it doesn't affect 2.5 and it was never released in a chart, I agree that validation is not necessary.

PennyScissors: @doflamingo721 can you revert the validation you added?

doflamingo721: @PennyScissors I reverted the validations, can you please review it again?

PennyScissors (approved):

aiyengar2 (approved):

PR #1855, opened 2022-04-14
pin v0.1.4 webhook

luthermonson:


.gitignore
```diff
@@ -2,3 +2,4 @@ bin
 *.DS_Store
 .idea
 *.iml
+.charts-build-scripts
```
PennyScissors, line 5: nit: can you revert this?

packages/rancher-webhook/package.yaml
```diff
@@ -1,2 +1,2 @@
-url: https://github.com/rancher/webhook/releases/download/v0.1.4-rc1/rancher-webhook-0.1.4-rc1.tgz
-packageVersion: 0
```
PennyScissors, line 2: Why was this removed?

luthermonson: accident i would suspect. putting it back

PennyScissors (approved):

PR #1860, opened 2022-04-19
[2.6] istio-installer version bump

doflamingo721:
The newer istio-installer version has a workaround for the error occurring during reinstalling istio as mentioned [here](https://github.com/rancher/rancher/issues/37218#issuecomment-1100244331). This was caused because after uninstalling istio, it leaves behind a MutatingWebhookConfiguration. We need to cleanup this webhook before reinstalling istio to fix the error. 

Linked issue:
https://github.com/rancher/rancher/issues/37218

Linked istio-installer PR:
https://github.com/rancher/istio-installer/pull/53


PennyScissors (approved):
LGTM. For future reference, if you are modifying the same version of a chart (in this case 100.2.0+up1.12.6) you don't need to manually delete it and commit. When you run make charts with the new change/patch, the existing assets will be replaced by the new ones with your change so the delete commit is redundant.

PR #1867, opened 2022-04-21
Add logging-operator 3.17.4

doflamingo721:
Depends on: https://github.com/rancher/image-mirror/pull/237

This PR adds `rancher-logging:100.1.2+up3.17.4`.
Also fixes https://github.com/rancher/rancher/issues/37394, `fluentd.bufferStorageVolume` can now be configured from Values spec.


superseb (commented):
This doesn't include the fluentd image bump as discussed in Slack, please include why so the reason is known.

packages/rancher-logging/generated-changes/overlay/templates/_generic_logging.yaml
```diff
@@ -65,11 +65,9 @@ spec:
     configReloaderImage:
       repository: {{ template "system_default_registry" . }}{{ .Values.images.config_reloader.repository }}
       tag: {{ .Values.images.config_reloader.tag }}
-    {{- if not .Values.disablePvc }}
```
doflamingo721, line 68: @superseb Upgrading fluentd won't have any effect on this issue.  The reason `fluentd.bufferStorageVolume` was not getting configured from the chart values was because of this check. So I used the images which upstream logging-operator uses. https://github.com/banzaicloud/logging-operator/blob/3.17.4/pkg/sdk/logging/api/v1beta1/logging_types.go#L115

samkulkarni20:
> This doesn't include the fluentd image bump as discussed in Slack, please include why so the reason is known.

@superseb @SheilaghM I don't think we should upgrade to an image with which the operator hasn't been tested yet. In case it brings any issues with it, we'll need to take care of them on our own. This also goes against the principle that we usually try to follow with Rancher charts, which is not to customize them more than needed, since it adds to the future chart management efforts.
In addition to this, the new image is not that far off from where we're currently at. So I don't think we would lose a lot by staying with the current image tag.

packages/rancher-logging/generated-changes/overlay/templates/_generic_logging.yaml
superseb, line 68: The original implementer of this logic was Dan Bason afaics, did you ask him why it was implemented this way?

doflamingo721: Hey @dbason, Because of the `if` check you see above we're not able to configure the `fluentd.bufferStorageVolume` from chart values since we need `disablePvc`  set to true. I'm not aware of the reason why this check was implemented at that time. Can you please let us know why it was implemented that way?

charts/rancher-logging/100.1.2+up3.17.4/templates/_generic_logging.yaml
```diff
@@ -0,0 +1,116 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+{{ include "logging-operator.labels" . | indent 4 }}
+spec:
+  controlNamespace: {{ .Release.Namespace }}
+  fluentbit:
+    image:
+      repository: {{ template "logging-operator.fluentbitImageRepository" . }}
+      tag: {{ template "logging-operator.fluentbitImageTag" . }}
+    {{- if not .Values.disablePvc }}
```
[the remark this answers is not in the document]
doflamingo721: Yes. We won't need to change the logic for fluentbit since it does not require `disablePvc` to be set to `true` unlike fluentd. Referred to this documentation for [fluentd](https://banzaicloud.com/docs/one-eye/logging-operator/configuration/fluentd/#custom-fluentd-hostpath-volume-for-buffers) and [fluentbit](https://banzaicloud.com/docs/one-eye/logging-operator/configuration/fluentbit/#hostpath-volumes-for-buffers-and-positions).

packages/rancher-logging/generated-changes/overlay/templates/_generic_logging.yaml
dbason: It was some time ago, so off the top of my head I don't remember the specifics sorry.  Out of interest how are you creating a volume for the buffers without a PVC?

doflamingo721: @superseb Since we have a requirement to add the ability to configure `fluentd.bufferStorageVolume` block via chart values on the logging resource and as @dbason confirmed it would be fine to remove the checks which are preventing us to do so, Should we go ahead with this change?

superseb (approved):

dbason (approved):

PR #1872, opened 2022-04-26
Gatekeeper: Add node selectors and tolerations for windows

eliyamlevy:
Issue: https://github.com/rancher/rancher/issues/37029


superseb (changes requested):
From my limited charts knowledge:

* this seems to change an existing version, while I think it should be a new version with the fix?
* The related issue (https://github.com/rancher/rancher/issues/37029) shows certain jobs being scheduled to windows nodes but I dont see changes in `templates/jobs.yaml`

PennyScissors:
@superseb you are right about the version. This used to be in a new version, but then it was released in 2.6.5 so the version needs to be bumped again. As for jobs, you are also right. This PR should include this change to add the linux tolerations/node-selectors to the jobs for CRD uninstall https://github.com/rancher/charts/pull/1823/commits/fb996bdae04fec2c3812a6087bcb5fad4f443364

superseb (dismissed):

superseb (approved):

PR #1874, opened 2022-04-27
Remove controllers dashboard

rmweir:
**Problem:**
Norman metric's labels have too high of a cardinality as they include object key as a label. This is a problem because metrics produces a row for every combination of labels. This can have a large impact on performance and is cautioned against in the prometheus [docs](https://prometheus.io/docs/practices/naming/):
`CAUTION: Remember that every unique combination of key-value label pairs represents a new time series, which can dramatically increase the amount of data stored. Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values.`

**Solution:**
Norman metrics will be removed as aside from using the key lasso metrics have the same data since norman now uses lasso. The controllers dashboard which uses the norman metrics can be removed and data not captured by the performance debugging dashboard has been added.

**Issue:**
https://github.com/rancher/rancher/issues/37274


aiyengar2 (approved):
LGTM

cbron (approved):

PR #1880, opened 2022-05-01
Update Neuvector 5.0.0-b2

selvamt94:


release.yaml
```diff
@@ -1,7 +1,9 @@
 neuvector:
 - 100.0.0+up2.2.0-b1
+- 100.0.1+up2.2.0-b2
```
cbron, line 3: Code changes look fine, but I believe we should remove b1 right ? When we release we will only release `2.2.0` and not `2.2.0-b1` or `2.2.0-b2` , but we need b2 for testing now.

PennyScissors: @cbron @selvamt94 The expectation is that these are used for testing in the cases a pre-release version is necessary, each new pre-release chart replaces the previous pre-release chart's assets (tgz, charts/ entry, index entrym, and release.yaml entry), and then once QA has signed off on all issues related to said chart the pre-release suffix needs to be removed by the team/SME so the chart has the final to-be-released version on it.

That said, if `-bX` is the equivalent of our `-rcX`, then the new chart needs to be `100.0.0+up2.2.0-b2`, not `100.0.1+up2.2.0-b2` (No need to bump the chart's version to 100.0.1, just the pre-release). Also, you need to run `make remove` for both neuvector and neuvector-crd, and the `100.0.0+up2.2.0-b1` entries in release.yaml need to be replaced by `100.0.0+up2.2.0-b2`.

To make this easier to review, the commits should go like this:
1st commit: Run make remove for neuvector and neuvector-crd, commit output
2nd commit: Changes made to packages/neuvector to update to `-b2` and updated release.yaml entries
3rd commit: Run make charts, commit output

packages/neuvector/package.yaml
```diff
@@ -1,5 +1,5 @@
-url: https://neuvector.github.io/neuvector-helm/core-2.2.0-b1.tgz
-version: 100.0.0
+url: https://neuvector.github.io/neuvector-helm/core-2.2.0-b2.tgz
+version: 100.0.1
```
PennyScissors, line 2: No need to bump this, if `-bX` is a pre-release version, only the pre-release is bumped. the 100.x.x version is bumped for each Rancher chart's release (100.0.1 would be the version to release in 2.6.6).

release.yaml
selvamt94: I will make changes and resubmit as suggested. Thanks.

cbron: Thanks Steven, I missed not bumping the patch version, that cleared up the diff.

cbron (commented):
lgtm

cbron:
issue: https://github.com/rancher/rancher/issues/37420

PennyScissors (approved):

PR #1887, opened 2022-05-10
Add Prometheus Federator chart

aiyengar2:
Currently adding the RC version of the chart that does not have the Rancher version patched.

Before the final release, we need to ensure that the patch is removed and that the chart has the right image tag.

We also need to ensure that the final version that is cut is `0.1.0`, where the version matches what will be committed in the prometheus-federator repository.


packages/rancher-monitoring/prometheus-federator/generated-changes/patch/Chart.yaml.patch
```diff
@@ -0,0 +1,11 @@
+--- charts-original/Chart.yaml
++++ charts/Chart.yaml
+@@ -6,7 +6,7 @@
+   catalog.cattle.io/os: linux,windows
+   catalog.cattle.io/permits-os: linux,windows
+   catalog.cattle.io/provides-gvr: helm.cattle.io.projecthelmchart/v1alpha1
+-  catalog.cattle.io/rancher-version: '>= 2.6.0-0 <=2.6.99-0'
++  catalog.cattle.io/rancher-version: '>= 2.6.5-0 <=2.6.99-0'
```
PennyScissors, line 8: Can you make this `'>= 2.6.5-0 <= 2.6.100-0'`?

release.yaml
```diff
@@ -2,6 +2,8 @@ neuvector:
 - 100.0.0+up2.2.0-b2
 neuvector-crd:
 - 100.0.0+up2.2.0-b2
+prometheus-federator:
+- 0.1.0+up0.0.1
```
PennyScissors, line 6: Do we not want to follow the 100.x.x in this case?

aiyengar2: This falls under a chart that we own ourselves, so we don't need to add 100.x.x to it; examples of other charts like this are Backup Restore Operator and CIS Benchmark, which are both 1.x.x in 2.5, 2.x.x in 2.6.

The reason why we start at 0.x.x instead of 1.x.x is because CIS / Backup needed to differentiate from the system chart version, which was 0.x.x; this is not the case for Prometheus Federator since there is no system chart version of it.

The reason why we do 0.1.x instead of 0.0.x is because 0.0.x marks experimental versions released in the Prometheus Federator repository **only** that are not limited to Rancher 2.6.5+; it's deployable in earlier 2.6 versions of Rancher but would require a user to manually add a chart repository to do so. 

For Rancher, the first version we are officially supporting is 0.1.x+, which is scoped to 2.6.5+.

packages/rancher-monitoring/prometheus-federator/generated-changes/patch/Chart.yaml.patch
aiyengar2: Sure

aiyengar2:
https://github.com/rancher/prometheus-federator/pull/6 will address migration to 0.1.0.

aiyengar2:
> > Before the final release, we need to ensure that the patch is removed and that the chart has the right image tag
> 
> Should we use `-patch` or `-rc` on the chart's version to make it more obvious and help prevent this in future releases?

yes we should. But that's a change we need to make in the release process of the Prometheus Federator repository; as of now, I've only been modifying the image tags, not the chart version. Will do it for next time.

PennyScissors (approved):

PR #1905, opened 2022-05-18
Add kube-version annotations to istio 1.10

doflamingo721:
Linked issue:
https://github.com/rancher/rancher/issues/37697


release.yaml
```diff
@@ -17,6 +17,9 @@ rancher-gatekeeper:
 rancher-gatekeeper-crd:
 - 100.1.0+up3.7.1
 rancher-istio:
```
aiyengar2, line 19: release.yaml should be deleted and recreated only containing these changes since it's an OOB release; all the other charts should be removed.

aiyengar2 (changes requested):
Need to remove other entries from the release.yaml for a new OOB release, otherwise LGTM

aiyengar2:
Assigning @PennyScissors as a second reviewer since this is an OOB release

charts/rancher-istio/100.0.0+up1.10.4/Chart.yaml
```diff
@@ -2,6 +2,7 @@ annotations:
   catalog.cattle.io/auto-install: rancher-kiali-server-crd=100.0.0+up1.35.0
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/display-name: Istio
+  catalog.cattle.io/kube-version: '>= 1.18.0-0 < 1.22.0-0'
```
aiyengar2, line 5: one more thing; this currently indicates we should not be able to install it on a 1.22 or less cluster, but the issue seems to indicate we only need to drop support for 1.23+? https://github.com/rancher/rancher/issues/37697

Also, why are we limiting to 1.18+ / where does that requirement come from?

release.yaml
doflamingo721: Updated the release.yaml

aiyengar2 (approved):
LGTM! Waiting on @PennyScissors to approve before merge

pennyscissors (approved):

PR #1911, opened 2022-05-23
Adding rancher-csp-adapter chart

MbolotSuse:
Adds the chart for the [csp-adapter](https://github.com/rancher/csp-adapter)

PR #1914, opened 2022-05-25
Adding rancher-csp-adapter chart

cbron:
Takes over from: https://github.com/rancher/charts/pull/1911
See comments there

Issue: https://github.com/rancher/rancher/issues/37495


charts/rancher-csp-adapter/0.1.0+up0.1.0-rc5/templates/configmap.yaml
```diff
@@ -0,0 +1,6 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: {{ .Chart.Name }}-status
+  namespace: cattle-csp-adapter-system
+data: {}
```
aiyengar2, line 6: nit: newline

Also, do you need to include the data field if it’s empty anyways?

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc5/templates/configmap.yaml
```diff
@@ -0,0 +1,6 @@
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: {{ .Chart.Name }}-status
+  namespace: cattle-csp-adapter-system
```
aiyengar2, line 5: Should be `{{ .Release.Namespace }}`

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc5/templates/configmap.yaml
```diff
@@ -0,0 +1,6 @@
+kind: ConfigMap
```
aiyengar2, line 1: Order is flipped for apiVersion and kind?

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc5/templates/deployment.yaml
```diff
@@ -0,0 +1,32 @@
...
+  selector:
+    matchLabels:
+      app: {{ .Chart.Name }}
+  template:
+    metadata:
+      labels:
+        app: {{ .Chart.Name }}
+    spec:
+      containers:
+      - env:
```
aiyengar2, line 16: Needs node selectors and tolerations for windows support

Needs service account that has permission to use a defined PSP + a default securityPolicy for hardened cluster support. Also needs additional job to patch default service account and add network policy, see internal Confluence doc on Helm Chart Requirements for more instructions / code that can be copied and pasted to support this.

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc5/templates/rbac.yaml
```diff
@@ -0,0 +1,114 @@
...
+subjects:
+  - kind: ServiceAccount
+    name: {{ .Chart.Name }}
+    namespace: cattle-csp-adapter-system
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ .Chart.Name }}-role
+  namespace: cattle-csp-adapter-system
```
aiyengar2, line 71: Should be helm release namespace

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc5/templates/rbac.yaml
```diff
@@ -0,0 +1,114 @@
...
+  resources:
+  - configmaps
+  verbs:
+  - create
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: {{ .Chart.Name }}-binding
+  namespace: cattle-csp-adapter-system
```
aiyengar2, line 106: Should be helm release namespace

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc5/templates/rbac.yaml
```diff
@@ -0,0 +1,114 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: {{ .Chart.Name }}-role
+subjects:
+  - kind: ServiceAccount
+    name: {{ .Chart.Name }}
+    namespace: cattle-csp-adapter-system
```
aiyengar2, line 114: Should be helm release namespace

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc5/templates/serviceAccount.yaml
```diff
@@ -0,0 +1,17 @@
...
+  {{- fail "If the aws provider is enabled, account number and role name must be provided" -}}
+  {{- end -}}
+{{- else -}}
+{{- fail "One cloud provider must be enabled" -}}
+{{- end -}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Chart.Name }}
+  namespace: cattle-csp-adapter-system
```
aiyengar2, line 13: Should be helm release namespace

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc5/values.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+image:
+  repository: rancher/rancher-csp-adapter
+  tag: v0.1.0-rc5
+  imagePullPolicy: IfNotPresent
+
+global:
+  cattle:
+    systemDefaultRegistry: ""
+  hostNetwork: false
```
aiyengar2, line 11: Why is this a global setting?

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc5/values.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+tolerations: []
```
aiyengar2, line 13: Tolerations are not used? Also we should have nodeSelector here too and add it to the deployment

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc5/values.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+# at least one csp must be enabled like below
+#aws:
```
aiyengar2, line 16: Uncomment this and set enabled to false by default?

aiyengar2 (changes requested):

aiyengar2 (commented):
I’ll be out tomorrow for the long weekend, so I’m going to remove my request changes review in favor of a comment and let @PennyScissors approve this. If I’m back and this hasn’t been merged, I can provide a re-review

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc5/Chart.yaml
```diff
@@ -0,0 +1,15 @@
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/hidden: "true"
+  catalog.cattle.io/kube-version: '>= 1.22.0'
+  catalog.cattle.io/namespace: cattle-csp-adapter-system
+  catalog.cattle.io/os: linux
+  catalog.cattle.io/permits-os: linux
+  catalog.cattle.io/rancher-version: '>= 2.6.6'
+  catalog.cattle.io/release-name: rancher-csp-adapter
```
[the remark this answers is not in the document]
cbron: It's a hidden chart. But I'll add anyway in case it's public in future

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc5/templates/configmap.yaml
cbron: I brought up same thing, but this is hardcoded in Rancher now: https://github.com/rancher/rancher/pull/37828/files

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc5/Chart.yaml
```diff
@@ -0,0 +1,15 @@
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/hidden: "true"
+  catalog.cattle.io/kube-version: '>= 1.22.0'
```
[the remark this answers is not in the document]
cbron: What's the change here then ? This was a business decision we made with this specific operator, that we wouldn't support clusters less than 1.22. Did you mean it shouldn't be set here ?

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc5/templates/configmap.yaml
cbron: Tested, and turns out no.

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc5/values.yaml
cbron: This would be local cluster only app. Won't support windows clusters.

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc5/values.yaml
cbron: I think this was copy paste error. I'll let @MbolotSuse chime in. Deleting for now.

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc5/templates/deployment.yaml
cbron: It's already using a non-default service account. We'll add hardened cluster and psp support separately here: https://github.com/rancher/rancher/issues/37841

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc5/Chart.yaml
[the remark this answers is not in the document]
MbolotSuse: This chart is only supported on EKS clusters. [Support Matrix for 2.6.5](https://www.suse.com/assets/EN-2.6.5SupportMatrix-300422-0116-26.pdf) says that the lowest supported version is 1.22.6 for EKS local clusters (which is where this is only able to be deployed). I don't think that it is necessary to specify patch version here - that seems to be excessive.

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc5/templates/configmap.yaml
MbolotSuse: Rancher needs the output produced by this deployment (which is a configmap). This needs to be in a specific namespace so that rancher can actually retrieve it. There are multiple ways of signaling this to rancher, but in order to ensure that the namespace is labeled as a part of the system project, rancher needs to know the namespace beforehand.

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc5/values.yaml
MbolotSuse: Log says that this was there when the chart was handed over to me. I'll double check to make sure that we aren't using this, but if it's not being used should be safe to remove.

MbolotSuse:
Closing in favor of #1918

PR #1918, opened 2022-05-27
Adding rancher-csp-adapter chart

MbolotSuse:
Takes over from: https://github.com/rancher/charts/pull/1914
See comments there

Issue: https://github.com/rancher/rancher/issues/37495

I'm back from PTO now, so this should serve as the PR for any future changes to the charts. Apologies for the PR whiplash.


rmweir (dismissed):
LGTM

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc6/Chart.yaml
```diff
@@ -0,0 +1,17 @@
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/display-name: Rancher CSP Adapter
+  catalog.cattle.io/hidden: "true"
+  catalog.cattle.io/kube-version: '>= 1.22.0'
+  catalog.cattle.io/namespace: cattle-csp-adapter-system
```
cbron, line 6: FYI this is hardcoded in rancher, we don't support this being anything else

cbron (dismissed):
lgtm, still needs a charts team approval

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc6/Chart.yaml
```diff
@@ -0,0 +1,17 @@
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/display-name: Rancher CSP Adapter
+  catalog.cattle.io/hidden: "true"
+  catalog.cattle.io/kube-version: '>= 1.22.0'
```
PennyScissors, line 5: Can you make this a closed range and follow this format? `'>= 1.22.0-0 < 1.XX.0-0'` where XX is the max Kubernetes version the chart supports plus 1.

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc6/Chart.yaml
```diff
@@ -0,0 +1,17 @@
...
+  catalog.cattle.io/namespace: cattle-csp-adapter-system
+  catalog.cattle.io/os: linux
+  catalog.cattle.io/permits-os: linux
```
PennyScissors, line 8: Does this chart support hybrid RKE2 clusters with both linux and windows nodes? If it does, this should be `linux,windows` and node-selectors/tolerations will need to be added to make sure resources are schedules on linux nodes only (We do this for nearly all of our other charts). Just verifying this was purposely set to linux only and not missed.

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc6/Chart.yaml
```diff
@@ -0,0 +1,17 @@
...
+  catalog.cattle.io/rancher-version: '>= 2.6.6-0 <= 2.6.99-0'
```
PennyScissors, line 9: Change this to `'>= 2.6.6-0 < 2.7.0-0'`

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc6/Chart.yaml
MbolotSuse: Never mind, I'm told this will eventually have to support psps so I'll have to upper bound this at 1.25

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc6/Chart.yaml
cbron: No, it's local cluster only, which does not support windows. It could support RKE2 linux local cluster.

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc6/Chart.yaml
PennyScissors: Yes, but I will submit a PR soon to fix these and stick to one convention. Some charts use < 2.6.99-0, others use <= 2.6.99-0 and this is problematic for cases where we set the Rancher dev version to 2.6.99 and need to do comparisons like we do for airgap image exporting (See [here](https://github.com/rancher/rancher/blob/68c6c1377b6d1dec9846193fd093a3b6bba8476e/pkg/image/charts.go#L205-L211)). So the < 2.7.0-0 would take care of this and also help stick to the same convention we use for the kube-version.

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc6/Chart.yaml
MbolotSuse: Fixed this - let me know if this is not what you were thinking.

charts/rancher-csp-adapter/0.1.0+up0.1.0-rc6/Chart.yaml
MbolotSuse: Fixed - let me know if this is not what you were thinking.

pennyscissors (approved):

cbron (approved):

PR #1924, opened 2022-06-02
[release-v2.6] Infracloud - Fix rancher-version and kube-version annotations

PennyScissors:
Issues:


PennyScissors:
@samkulkarni20 Can you help review this? I need confirmation the ranges for the kube-version annotations I updated in this PR as correct. In some cases the charts had a kubeVersion already which I just matched in the kube-version annotation, but in others I had to use the default and I want to make sure any of the charts where I put < 1.22 actually supports 1.22 and over and needs to be updated.

samkulkarni20 (approved):
All annotations, including the kube-version are correct. Approving the PR.
For gatekeeper support matrix, I've started a discussion [here](https://github.com/orgs/open-policy-agent/discussions/213)

charts/rancher-gatekeeper/100.0.0+up3.5.1/Chart.yaml
```diff
@@ -2,10 +2,11 @@ annotations:
   catalog.cattle.io/auto-install: rancher-gatekeeper-crd=match
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/display-name: OPA Gatekeeper
+  catalog.cattle.io/kube-version: '>= 1.16.0-0 < 1.22.0-0'
   catalog.cattle.io/namespace: cattle-gatekeeper-system
   catalog.cattle.io/os: linux
   catalog.cattle.io/provides-gvr: config.gatekeeper.sh.config/v1alpha1
-  catalog.cattle.io/rancher-version: < 2.6.1-0
```
aiyengar2, line 8: This seems like an intentional change to hide the chart after 2.6.1 right? We should look into why this was necessary, but based on this we should be releasing:

```
'>= 2.6.0-0 < 2.6.1-0'
```

packages/rancher-gatekeeper/generated-changes/patch/Chart.yaml.patch
```diff
@@ -5,11 +5,11 @@
 +  catalog.cattle.io/auto-install: rancher-gatekeeper-crd=match
 +  catalog.cattle.io/certified: rancher
 +  catalog.cattle.io/display-name: OPA Gatekeeper
-+  catalog.cattle.io/kube-version: '>= 1.16.0-0'
++  catalog.cattle.io/kube-version: '>= 1.16.0-0 < 1.22.0-0'
```
aiyengar2, line 8: let's just mark this chart doNotRelease in release-v2.6 and not modify the `packages/` changes. Same for istio.

aiyengar2 (changes requested):

aiyengar2 (changes requested):
Can you also double check the concern from https://github.com/rancher/charts/pull/1928#pullrequestreview-1017839314? We should make sure releasing a chart does not lessen the kubeVersion we expect it to be supported on on release

charts/rancher-gatekeeper/100.0.0+up3.5.1/Chart.yaml
PennyScissors: Good catch, definitely missed that < in this one!

PennyScissors:
@aiyengar2 Update gatekeeper to match Rancher's k8s support matrix. As for Istio I verified it against our matrix, but the Istio upstream versions support slightly older k8s versions so what this PR has seems correct based on this table https://istio.io/latest/docs/releases/supported-releases/#support-status-of-istio-releases

charts/rancher-gatekeeper/100.0.1+up3.6.0/Chart.yaml
```diff
@@ -2,9 +2,11 @@ annotations:
   catalog.cattle.io/auto-install: rancher-gatekeeper-crd=match
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/display-name: OPA Gatekeeper
+  catalog.cattle.io/kube-version: '>= 1.16.0-0 < 1.24.0-0'
```
aiyengar2, line 5: This was released in 2.6.1 https://github.com/rancher/rancher/issues/34589

According to https://www.suse.com/suse-rancher/support-matrix/all-supported-versions/rancher-v2-6-1, this should have only been tested up to k8s 1.21; are we sure 1.22 and 1.23 are supported?

charts/rancher-gatekeeper/100.1.0+up3.7.1/Chart.yaml
```diff
@@ -2,11 +2,11 @@ annotations:
   catalog.cattle.io/auto-install: rancher-gatekeeper-crd=match
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/display-name: OPA Gatekeeper
-  catalog.cattle.io/kube-version: '>= 1.16.0-0'
+  catalog.cattle.io/kube-version: '>= 1.16.0-0 < 1.25.0-0'
```
aiyengar2, line 5: This was released in Rancher 2.6, which supports up to 1.23

https://github.com/rancher/rancher/issues/37376

https://www.suse.com/assets/EN-2.6.5SupportMatrix-300422-0116-26.pdf

Are we sure this works on 1.24?

aiyengar2 (changes requested):

aiyengar2 (approved):
LGTM

PennyScissors:

PR #1925, opened 2022-06-02
[release-v2.6] Harvester - Fix rancher-version and kube-version annotations

PennyScissors:
Issues:


PennyScissors:
@guangbochen can you help find reviewers for this?

doflamingo721 (approved):

doflamingo721 (approved):
Approving the PR. Had an offline discussion on Harvester community slack and there isn't a Harvester <> Kubernetes support matrix documented yet. These kube-version constraints looks good as of now.

charts/harvester-cloud-provider/100.0.0+up0.1.8/Chart.yaml
```diff
@@ -1,10 +1,10 @@
 annotations:
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/display-name: Harvester Cloud Provider
-  catalog.cattle.io/kube-version: '>= 1.18'
+  catalog.cattle.io/kube-version: '>= 1.18.0-0 < 1.22.0-0'
```
guangbochen, line 4: is this just for the support matrix doc? currently, both Harvester CSI and CCM chart are verified with k8s version from v1.21.x to v1.24.x, so should we change this one accordingly?

guangbochen (commented):
we have another pending PR https://github.com/rancher/charts/pull/1931, may need to take this into account, can we merge PR #1931 first and patch all the versions accordingly if possible, thanks.

PennyScissors:
@guangbochen This PR is only modifying released versions of the harvester charts in the release-v2.6 branch so it won't affect the other PR you linked. I just need verification that the kube-version constraints I added to those older versions are correct.

charts/harvester-cloud-provider/100.0.0+up0.1.8/Chart.yaml
PennyScissors: @guangbochen This PR is only modifying released versions of the harvester charts in the release-v2.6 branch so it won't affect the other PR you linked. I just need verification that the kube-version constraints I added to those older versions are correct. Would it make sense to set all those older version to use `< 1.24.0-0'` instead instead of `< 1.22.0-0'`?

guangbochen: Yes, using kube-version with `< 1.24.0-0` will be more appropriate, thanks. cc @yaocw2020

PennyScissors: @guangbochen yes, we are enforcing a closed range so that we can eventually automate generating a support matrix. This is only for older version of the charts as this change is made to the release-v2.6 branch and not dev, so it won't affect the PR you linked. Should I update the constraint of all these old charts to < 1.25 instead of < 1.22?

PennyScissors: @guangbochen Just updated, can you re-review?

guangbochen (approved):
LGTM, thanks.

PennyScissors:

PR #1926, opened 2022-06-02
[release-v2.6] Team 1 - Fix rancher-version and kube-version annotations

pennyscissors:
Issues:
https://github.com/rancher/rancher/issues/37476


selvamt94:
Neuvector chart 2.2.0 supports k8s version 1.23.x-x. We will update here about supporting k8s version 1.24 soon.

PennyScissors:
@selvamt94 Thanks, I will update neuvector 100.0.0+up2.2.0 to support 1.23. However, no need to update this PR if the new neuvector version supports 1.24 since this is specifically for charts that have already been released. You can update the kube-version in the dev-v2.6 branch once the new neuvector chart is ready for review.

charts/neuvector/100.0.0+up2.2.0/Chart.yaml
```diff
@@ -2,12 +2,12 @@ annotations:
   catalog.cattle.io/auto-install: neuvector-crd=match
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/display-name: NeuVector
-  catalog.cattle.io/kube-version: '>=1.18.0-0'
+  catalog.cattle.io/kube-version: '>= 1.18.0-0 < 1.22.0-0'
```
cbron, line 5: Neuvector works with 1.22+, I think all the way up to 1.24 and beyond.

charts/rancher-external-ip-webhook/100.0.0+up1.0.0/Chart.yaml
```diff
@@ -1,8 +1,10 @@
 annotations:
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/display-name: External IP Webhook
+  catalog.cattle.io/kube-version: '>= 1.16.0-0 < 1.22.0-0'
```
cbron, line 4: This is fine, and was added in later versions of the chart.

charts/rancher-webhook/1.0.0+up0.2.0/Chart.yaml
```diff
@@ -1,8 +1,10 @@
 annotations:
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/hidden: "true"
+  catalog.cattle.io/kube-version: '>= 1.16.0-0 < 1.22.0-0'
```
cbron, line 4: Webhook has to run every time Rancher starts, so this should be not maxed out. It supports 1.24, but in the future has to support every version rancher supports. As far as specific versions go, it will get upgraded as soon as rancher is upgraded, so old versions don't matter.

cbron (changes requested):

charts/neuvector/100.0.0+up2.2.0/Chart.yaml
PennyScissors: This version specifically @selvamt94 confirmed it supports 1.23, so I'm waiting on other reviewers to update it the PR for everything together.

charts/rancher-webhook/1.0.0+up0.2.0/Chart.yaml
PennyScissors: @cbron We are trying to enforce a closed range, so in this case I think what would make sense is to bump it to `< 1.24.0-0` as that covers what Rancher supports right now, and then update the constraint as new K8s versions are introduced into Rancher. Does that sound good?

cbron: For this it doesn't matter at all, I guess it's sort of annoying but we can do it. Moving forward we would need to add one in the main repo.

charts/rancher-webhook/1.0.2+up0.2.2/Chart.yaml
```diff
@@ -1,8 +1,10 @@
 annotations:
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/hidden: "true"
+  catalog.cattle.io/kube-version: '>= 1.16.0-0 < 1.22.0-0'
```
cbron, line 4: Actually now that I think about it, we have to be careful about what we put here. Whatever versions of Rancher used this webhook, we have to allow all the supported versions of k8s. 

2.6.0-2.6.1 - 1.18-1.21 - 1.0.0+up0.2.0
2.6.3 - 1.18-1.22 - 1.0.2+up0.2.2
2.6.4 - 1.18-1.23 - 1.0.4+up0.2.5

So that means:
1.0.0 and 1.0.1 need to support < 1.22
1.0.2 needs to support < 1.23 (this file)
1.0.3 and 1.0.4 need to support <1.24 (1.0.3 was introduced in a patch release so being on safe side here)

Can't we just put <1.25 for all of the ones 1.0.2+ though ? Since that's when the next breaking changes are going to be.

cbron (changes requested):

PennyScissors:
Updated all webhook's constraints to < 1.25, and neuvector's to <1.24.

charts/rancher-webhook/1.0.3+up0.2.5/Chart.yaml
```diff
@@ -1,6 +1,7 @@
 annotations:
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/hidden: "true"
+  catalog.cattle.io/kube-version: '>= 1.16.0-0 < 1.25.0-0'
   catalog.cattle.io/namespace: cattle-system
   catalog.cattle.io/os: linux
   catalog.cattle.io/permits-os: linux,windows
```
cbron, line 7: Hmm, I can't figure out where this chart is used. In 2.6.4 we use 1.0.4, and in 2.6.3-patch3 we used `1.0.3+up0.2.4` 
https://github.com/rancher/rancher/blob/v2.6.3-patch3/package/Dockerfile#L49
Either way this one shouldn't have rancher-version of <2.5.99-0, I don't know how that got added. 
Thats not in the upstream: https://github.com/rancher/webhook/blob/v0.2.5/charts/rancher-webhook/Chart.yaml
I think the installer for webhook just ignores this annotation.

cbron (approved):
Some oddities, otherwise lgtm

rmweir (approved):
LGTM, hopefully we don't forget to add a new webhook for v2.7 though.

PennyScissors:
Closed in favor of this PR to consilidate PRs from all teams: https://github.com/rancher/charts/pull/1964

PR #1927, opened 2022-06-02
[release-v2.6] Team 2 - Fix rancher-version and kube-version annotations

PennyScissors:


jakefhyde (approved):

jiaqiluo (changes requested):
The aks/eks/gke-operator charts are installed in the local cluster automatically by Rancher, which means the `kube-version` limit should match the one in rancher's chart: https://github.com/rancher/rancher/blob/release/v2.6/chart/Chart.yaml#L6

PennyScissors:
@jiaqiluo I agree, but what about older versions? Should I use <1.25 for everything instead since that's the next version when we expect things to break?

jiaqiluo (changes requested):
I suggest to make the kube-vesion range `'>= 1.18.0-0 < 1.25.0-0'` for  `system-upgrade-controller`  for the same reason as those aks/eks/gke-operator charts

jiaqiluo (approved):

PennyScissors:
Closed in favor of this PR to consilidate PRs from all teams: https://github.com/rancher/charts/pull/1964

PR #1929, opened 2022-06-02
[release-v2.6] Longhorn - Fix rancher-version and kube-version annotations

PennyScissors:
Issues:


innobead (approved):
LGTM

innobead (changes requested):
After verification, the max support should be 1.25 because deprecated PSP will be removed and it's actually still used in the current Longhorn versions.

cc @PhanLe1010

PhanLe1010:
@innobead 

Acutally, Longhorn 1.3.0 doesnot support kubernetes v1.25. It supports Kubernetes [1.8;1.24].
Ref: https://github.com/longhorn/longhorn-manager/pull/1342#issuecomment-1133434517

innobead:
> @innobead
> 
> Acutally, Longhorn 1.3.0 doesnot support kubernetes v1.25. It supports Kubernetes [1.8;1.24]. Ref: [longhorn/longhorn-manager#1342 (comment)](https://github.com/longhorn/longhorn-manager/pull/1342#issuecomment-1133434517)

Exactly, I mean `< 1.25.0-0`.

PennyScissors:
@innobead This PR is for charts older than 1.3.0, not 1.3.0, so looks like this is good to approve and merge correct?

charts/longhorn/100.1.1+up1.2.3/Chart.yaml
```diff
@@ -2,11 +2,11 @@ annotations:
   catalog.cattle.io/auto-install: longhorn-crd=match
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/display-name: Longhorn
-  catalog.cattle.io/kube-version: '>=1.18.0-0'
+  catalog.cattle.io/kube-version: '>= 1.18.0-0 < 1.23.0-0'
```
innobead, line 5: This should be '>= 1.18.0-0 < 1.25.0-0', the same as 1.3.0, and also for other 1.2.x charts. 

cc @PhanLe1010

packages/longhorn/longhorn-1.2/generated-changes/patch/Chart.yaml.patch
```diff
@@ -5,11 +5,11 @@
 +  catalog.cattle.io/auto-install: longhorn-crd=match
 +  catalog.cattle.io/certified: rancher
 +  catalog.cattle.io/display-name: Longhorn
-+  catalog.cattle.io/kube-version: '>=1.18.0-0'
++  catalog.cattle.io/kube-version: '>= 1.18.0-0 < 1.23.0-0'
```
innobead, line 8: This should be '>= 1.18.0-0 < 1.25.0-0', the same as 1.3.0.

innobead (changes requested):
LGTM except 1.2.x. 

I need @PhanLe1010 to review this as well.

shuo-wu (changes requested):
Same issue. For Longhorn v1.2.x, the Kubernetes version should be `< 1.25.0-0` rather than `< 1.23.0-0`

charts/longhorn/100.0.0+up1.1.2/Chart.yaml
```diff
@@ -2,10 +2,11 @@ annotations:
   catalog.cattle.io/auto-install: longhorn-crd=match
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/display-name: Longhorn
+  catalog.cattle.io/kube-version: '>= 1.16.0-0 < 1.22.0-0'
   catalog.cattle.io/namespace: longhorn-system
   catalog.cattle.io/os: linux
   catalog.cattle.io/provides-gvr: longhorn.io/v1beta1
-  catalog.cattle.io/rancher-version: '>= 2.6.0-0'
+  catalog.cattle.io/rancher-version: '>= 2.6.0-0 < 2.7.0-0'
```
PhanLe1010, line 9: Does it mean that we will update this field again when rancher release v2.7.0? Because this chart will still be compatible with the new rancher release

PennyScissors: Yes, because in 2.7 we will most likely cut a new release-v2.7 branch for all charts, bump the versions to 200.x.x where it applies, and bump this to `'>= 2.7.0-0 < 2.8.0-0'`.

charts/longhorn/100.1.1+up1.2.3/Chart.yaml
PennyScissors: Perfect, I'll update all of Longhorn's 1.2.x to < 1.25 instead.

PennyScissors:
Updated all longhorn's 1.2.x to use constraint < 1.25 instead of < 1.23.

PhanLe1010 (approved):
LGTM

innobead (approved):
LGTM

PennyScissors:

PR #1932, opened 2022-06-06
Set SRIOV chart as linux-only

manuelbuil:
Signed-off-by: Manuel Buil <mbuil@suse.com>


manuelbuil:
This is described by https://github.com/rancher/rancher/issues/37469

PR #1952, opened 2022-06-23
Bump rancher-gke-operator to 1.1.4

macedogm:
Bump executed to refresh base image and update its packages. No modifications were done to the chart itself.


macedogm:
@annablender can you please review this PR again?

annablender:
Pending question to @kinarashah about whether tagging a release version vs an rc is correct

annablender (changes requested):
The base image change here is not a functional change, so it makes more sense to tag an rc instead of a release version. Tag `1.1.4-rc1`

macedogm:
The bump was done https://github.com/rancher/charts/pull/1965.

PR #1953, opened 2022-06-23
Bump rancher-eks-operator to 1.1.4

macedogm:
Bump executed to refresh base image and update its packages. No modifications were done to the chart itself.


annablender (approved):
LGTM, long as there are no merge conflicts

annablender (changes requested):
Asset generation should be in a separate 'make charts' commit

annablender (approved):

annablender (changes requested):
The base image change here is not a functional change, so it makes more sense to tag an rc instead of a release version. Tag `1.1.4-rc1`

macedogm:
The bump was done in https://github.com/rancher/charts/pull/1965.

PR #1959, opened 2022-06-28
Update Fleet packages to v0.3.10-rc1

prachidamle:
Creating PR to update Fleet to v0.3.10-rc1

https://github.com/rancher/fleet/issues/676


index.yaml
```diff
@@ -218,7 +259,7 @@ entries:
       catalog.cattle.io/release-name: fleet-agent
     apiVersion: v2
     appVersion: 0.3.9
-    created: "2022-03-28T14:25:40.763319-07:00"
```
aiyengar2, line 221: created timestamp changes shouldn't exist? I believe this is because you made changes to 100.0.3, removed those changes, and then added them back, which resulted in the old digest but a newer creation timestamp.

index.yaml
```diff
@@ -380,7 +438,7 @@ entries:
       catalog.cattle.io/release-name: fleet-crd
     apiVersion: v2
     appVersion: 0.3.9
-    created: "2022-03-28T14:25:41.966705-07:00"
```
aiyengar2, line 383: same issue as the timestamp above

aiyengar2 (changes requested):
should probably remove the last two commits and re-run make charts; currently the make charts commit generates a new 100.0.3 + a 100.0.4 and then removes 100.0.3 in the last commit, which causes the 100.0.3 index.yaml's timestamp to be modified unnecessarily!

index.yaml
prachidamle: removed the timestamp update by deleting these changes and re-running make charts

aiyengar2 (approved):

PR #1964, opened 2022-06-29
Fix rancher-version and kube-version annotations

PennyScissors:
This PR cherry picks and squashes the commits from the PRs below (all reviewed and approved by their respective teams) to consolidate them and make the release easier:
team1 https://github.com/rancher/charts/pull/1926
team2 https://github.com/rancher/charts/pull/1927
team3 https://github.com/rancher/charts/pull/1928
team4 https://github.com/rancher/charts/pull/1923
infracloud https://github.com/rancher/charts/pull/1924
longhorn https://github.com/rancher/charts/pull/1929
harvester https://github.com/rancher/charts/pull/1925

PR #1968, opened 2022-07-05
Bring kube-version and rancher-version changes to dev

PennyScissors:
https://github.com/rancher/charts/pull/1964


release.yaml
```diff
@@ -1,32 +1,149 @@
 rancher-monitoring:
+- 100.0.0+up16.6.0
```
aiyengar2, line 2: we need to remove all the extra entries from the release.yaml here

packages/rancher-wins-upgrader/package.yaml
```diff
@@ -2,3 +2,4 @@ url: https://github.com/rancher/wins.git
 subdirectory: charts/rancher-wins-upgrader
 commit: 4d46d8dd81cd741a51313923d459f28beea56800
 version: 100.0.1
+doNotRelease: true
```
aiyengar2, line 5: can we remove all the doNotRelease tags in favor of just making the modification to the latest package checked in?

aiyengar2 (changes requested):

release.yaml
```diff
@@ -146,4 +146,5 @@ harvester-cloud-provider:
 harvester-csi-driver:
 - 100.0.0+up0.1.9
 - 100.0.1+up0.1.9
-- 100.0.2+up0.1.11
\ No newline at end of file
+- 100.0.2+up0.1.11
+- 100.0.2+up0.1.14
```
PennyScissors, line 34: Found this issue, will fix in a separate PR once this is merged.

release.yaml
PennyScissors: updated

packages/rancher-wins-upgrader/package.yaml
PennyScissors: updated

PR #1983, opened 2022-07-19
Remove preceding slash from Grafana linkURLs to prevent broken drilldown links

aiyengar2:
This resolves a regression introduced to certain Grafana dashboards as part of a migration of the Rancher Monitoring chart to a new upstream.

`linkURL` values were expected to be changed from `./RELATIVE_LINK` to `RELATIVE_LINK`, but some / most were left as `/RELATIVE_LINK`, which resulted in Grafana ignoring the subpath of the external URL it was hosted in in favor of just trying to find the `RELATIVE_LINK` at the domain.

Related Issue: https://github.com/rancher/rancher/issues/37196


aiyengar2:
You can tell that this was the fix because the dashboard `Kubernetes / Networking / Cluster` still had working drill-down links in the `Current Status` table. This is because it was already using the right link format: https://github.com/rancher/charts/blob/253803ad397e64cf94aa58c8aa0fca8bc9eeb07e/charts/rancher-monitoring/100.1.3%2Bup19.0.3/templates/grafana/dashboards-1.14/cluster-total.yaml#L521

aiyengar2:
@geethub97 when you rebase Grafana to 8.x.x we should double check this is still fixed. I see some stuff in https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/659#issuecomment-901932330 that seems to indicate a move from Angular to React in Grafana may cause us to see different behavior, but based on the content of the comment I think it should still be fixed after this.

eliyamlevy (approved):

PR #1988, opened 2022-07-22
Update Webhook to 0.2.6-rc8

MbolotSuse:


KevinJoiner (dismissed):

charts/rancher-webhook/1.0.5+up0.2.6-rc7/Chart.yaml
```diff
@@ -1,18 +1,18 @@
 annotations:
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/hidden: "true"
-  catalog.cattle.io/kube-version: '>= 1.16.0-0 < 1.25.0-0'
+  catalog.cattle.io/kube-version: 0.2.6-rc7
```
superseb, line 4: this seems wrong, is this intentional?

MbolotSuse: Not intentional. I'm taking a look at it now. I think that when this annotation was originally added, it was added in this repo (rather than the source repo of rancher/webhook). I recently added this annotation to the webhook's repo, and as such was trying to get it to use what the webhook specifies rather than merging something from this repo. 

However, I think there's a part in the webhook's CI (sed command) which is replacing all version keys with the rc version. I'll get a better version of this pushed up, sorry for the inconvenience.

MbolotSuse: Alright - fixed the ci in the webhook repo and cut a new RC which doesn't have this issue. Let me know if you see any other issues - thanks for pointing this out.

superseb (approved):

KevinJoiner (approved):

PR #1989, opened 2022-07-22
Bump webhook for 2.5 to 0.1.5-rc1

MbolotSuse:


KevinJoiner (approved):

cbron (approved):
I was thinking this had to be in release.yaml but I guess that was only in 2.6. Is there a way to see what actual changes have gone in here ?  If not it should be fine. It has max rancher-version, I don't think we need a max kube-version because it's not 2.6+.

MbolotSuse:
@cbron I was a bit confused about the missing nature of release.yaml. However, I think that the index file is really what matters in terms of chart index searchability. 

As far as comparing changes goes, you can compare what is in the new directory with what was in the `0.1.4.00+up0.1.4` directory. I've ran a diff on those directories locally and pasted that here (along with the command ran) for you convenience:

Command (ran from `charts/rancher-webhook/rancher-webhook`):
`diff 0.1.400+up0.1.4 0.1.500-rc1+up0.1.5-rc1`

Output:

```
diff 0.1.400+up0.1.4/Chart.yaml 0.1.500-rc1+up0.1.5-rc1/Chart.yaml
9c9
< appVersion: 0.1.4
---
> appVersion: 0.1.5-rc1
12c12
< version: 0.1.400+up0.1.4
---
> version: 0.1.500-rc1+up0.1.5-rc1
Common subdirectories: 0.1.400+up0.1.4/templates and 0.1.500-rc1+up0.1.5-rc1/templates
diff 0.1.400+up0.1.4/values.yaml 0.1.500-rc1+up0.1.5-rc1/values.yaml
3c3
<   tag: v0.1.4
---
>   tag: v0.1.5-rc1
```
Let me know if you have any questions/concerns on the above.

PR #2001, opened 2022-07-28
vSphere CSI provider: allow getting csinodetopologies, as it is expected by node-driver-registrar

fpants:
PR for issue rancher/rancher#43439


nicholasSUSE:
Closed due to inactivity

PR #2009, opened 2022-08-10
Adapter final rc

MbolotSuse:
Bump rancher csp adapter to 1.0.0-rc7


charts/rancher-csp-adapter/1.0.0+up1.0.0-rc7/templates/_helpers.tpl
```diff
@@ -51,3 +43,15 @@ aws
 false
 {{- end -}}
 {{- end }}
+
+{{- define "system_default_registry" -}}
+{{- if .Values.global.cattle.systemDefaultRegistry -}}
+{{- printf "%s/" .Values.global.cattle.systemDefaultRegistry -}}
+{{- else -}}
+    {{- if eq (include "csp-adapter.csp" .) "aws" -}}
+    {{- "709825985650.dkr.ecr.us-east-1.amazonaws.com/suse/" -}}
```
cbron, line 52: So this would be registry for either EU or US then ? And if not it will pull from dockerhub

MbolotSuse: This is the US registry. The idea is that if they specify that the adpater is being installed for AWS, it defaults to US registry. It is acceptable to have customers use this for both listings IIRC.

cbron: Yes thats what I was asking, I forgot if it was ok for EU listing to pull from US registry.

cbron (approved):
lgtm

aiyengar2 (approved):
~Bumping the RC is fine, but have concerns about the contents of https://github.com/rancher/csp-adapter/pull/19!~

Addressed. LGTM

PR #2011, opened 2022-08-11
Bump rancher-webhook to 0.1.5-rc2

KevinJoiner:


MbolotSuse (approved):
LGTM - though this is pending review from team 3, I'm a bit less familiar with how 2.5 charts work.

PR #2018, opened 2022-08-12
[v2.5] Update rancher-operator to v0.1.5-rc1

cmurphy:
https://github.com/rancher/rancher/issues/38596


jakefhyde (approved):

MbolotSuse:
@cmurphy I might be wrong, but I think you also need to update the `rancher-operator-crd` chart. Especially because the crds changed due to the [go generate](https://github.com/rancher/rancher-operator/pull/50/commits/f0f7fbfdd0a61e38fec435e3b91d003e1b3ea474#diff-e00e0daed2d415e84de7b1f74881c084ed8143e59db779dee4deb30dd3559d1f). LMK if you disagree - not super familiar with this chart.

cmurphy:
@MbolotSuse yep good catch

jakefhyde (approved):

MbolotSuse (approved):
Good with all of this, but I'm not super familiar with the doNotRelease tag. Seems like the right thing to change it though.

@aiyengar2 might be good to get your input (though we have to make at least one more PR with this, so ok with merging it for now).

cmurphy:
The chart wouldn't generate without changing the doNotRelease field, so I think we need it.

cbron:
doNotRelease really means never release. It’s to allow changes in the generating code without re-releasing chart.

PR #2020, opened 2022-08-15
Bump gke operator v1.1.4

jakefhyde:


HarrisonWAffel (dismissed):

jiaqiluo (changes requested):
need to fix the CI failure first.  
 https://github.com/rancher/charts/blob/dev-v2.6/release.yaml  also needs to be updated.

jiaqiluo (approved):

HarrisonWAffel (approved):

PR #2039, opened 2022-08-19
Fix Longhorn CRD charts and hide latest Longhorn released chart with broken CRD install

aiyengar2:


derekbit (approved):
LGTM.

innobead:
This is good, thanks @aiyengar2 . 

@khushboo-rancher Let's plan to have another round test on this PR. Thanks. 

cc @yangchiu @chrisc1981 @Lichiehyao

innobead (commented):
LGTM. Thanks, @aiyengar2 👍 !

But, two things need to be adjusted and focus on the 1.3.1 update only.
- No need to update 1.1, because it will not be released anymore, EOM
- Postpone 1.2 update until the upcoming 1.2.5 is ready (ref: https://github.com/rancher/charts/pull/2037)

After the above fixes, @PhanLe1010  and @khushboo-rancher  will help verify before merging. 

P.S. If can't continue this PR, @PhanLe1010 can take over this.

PR #2062, opened 2022-09-02
[2.6.9] [sriov] Update images and chart and add node-feature-discovery as dependency

thomasferrandiz:
node-feature-discovery provides the ability to automatically label nodes as sriov capable.


manuelbuil (approved):

packages/rancher-sriov/templates/crd-template/Chart.yaml
```diff
@@ -9,4 +9,4 @@ annotations:
   catalog.cattle.io/namespace: cattle-sriov-system
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/hidden: "true"
-  catalog.cattle.io/permits-os: linux,windows
+  catalog.cattle.io/permits-os: linux
```
aiyengar2, line 12: did you remove support from tolerating a single Windows node in a cluster? So this can only be installed in a cluster with only Linux nodes?

`permits-os` == you will not fail with these OSs existing. i.e. you have nodeSelectors and tolerations

`os` == you deploy os specific components. Like rancher-monitoring deploying rancher-windows-exporters to specifically target Windows nodes.

aiyengar2 (changes requested):

packages/rancher-sriov/templates/crd-template/Chart.yaml
manuelbuil: sriov in Kubernetes is a linux-only technology. So this is more fixing a bug

release.yaml
```diff
@@ -4,8 +4,12 @@ fleet-agent:
 - 100.1.0+up0.4.0-rc3
 fleet-crd:
 - 100.1.0+up0.4.0-rc3
+sriov:
```
aiyengar2, line 7: Can you fix the release.yaml indentations so that they are all consistent?

aiyengar2 (changes requested):
Can you move the release.yaml changes to the first commit and rename the second commit “make charts”? We shouldn’t be having generated changes and manual changes in the same commit, the second commit is expected to be just generated code.

release.yaml
thomasferrandiz: It should be good now

aiyengar2 (approved):

PR #2083, opened 2022-09-13
Empty release.yaml after releasing longhorn chart 1.25

prachidamle:


aiyengar2:
Sorry for the miscommunication! This PR isn't necessary since the release.yaml already matches the charts we expect to release in the upcoming release.

PR #2084, opened 2022-09-14
Copy Longhorn v1.2.5 charts to Rancher chart release-v2.7

PhanLe1010:
This is so that Rancher 2.7 users can see Longhorn v1.2.5 chart. This is pure copying over. We don't use the `101.x.x` package to release this chart.

Starting from the next Longhorn release, we will use the `101.x.x` package to release new charts into Rancher release-v2.7

Signed-off-by: Phan Le <phan.le@suse.com>


aiyengar2 (changes requested):
~@PhanLe1010 we'll take care of the copy over! For pure copy-overs, we just create PRs from release to release.~

~See https://github.com/rancher/charts/pull/2108~

aiyengar2 (changes requested):
Apologies for the wrong communication with the previous comment!

We do expect feature teams to file PRs to forward port; however, instead of relying on manually copying we expect them to use https://github.com/rancher/charts/blob/dev-v2.7/scripts/forward-port

Can you file a PR with commits titled "Forward port ${CHART}-${VERSION} from release-v2.6" that runs the following instead of the current structure?

```
CHART=longhorn VERSION=100.1.3+up1.2.5 make forward-port
```

```
CHART=longhorn-crd VERSION=100.1.3+up1.2.5 make forward-port
```

PhanLe1010:
> Can you file a PR with commits titled "Forward port {VERSION} from release-v2.6" that runs the following instead of the current structure?
CHART=longhorn VERSION=100.1.3+up1.2.5 make forward-port
CHART=longhorn-crd VERSION=100.1.3+up1.2.5 make forward-port

@aiyengar2 Thanks for the instruction! Does it mean that we need to do this all existing Longhorn version (1.1.x, 1.2.x, and 1.3.x) because they all have `catalog.cattle.io/rancher-version '>= 2.6.0-0 < 2.7.0-0'` annotation?

nicholasSUSE:
Closed due to inactivity and lack of labels.

PR #2089, opened 2022-09-16
Add a new rancher-csp-adapter v2.0.0-rc1 chart

maxsokolovsky:


MbolotSuse:
Related to rancher/rancher#38943

MbolotSuse (dismissed):
LGTM. Do get an approval from team 3 before merging. 

We also need to talk about what this means for the listing before merging - not sure if/how we can maintain multiple versions.

release.yaml
```diff
@@ -6,6 +6,7 @@ rancher-cis-benchmark-crd:
 - 3.0.0-rc1
 rancher-csp-adapter:
```
eliyamlevy, line 7: Not sure if we can have two different versions on the release.yaml

eliyamlevy: Was this intended?

aiyengar2: As Eli mentioned above, do you plan on releasing 2 versions of this chart? If not, you should make remove the other one. If so, then this is fine

maxsokolovsky: The two versions of the rancher-csp-adapter? We initially wanted to actually remove the 1.0.1 (and 1.0.0, which is unreleased but presenting the repo) versions, but we can't due to backward compatibility reasons with dev-2.6, as far as I understand.
As @aiyengar2 explained offline, we want to ensure that users who upgrade to 2.7 can still continue to in-place upgrade their 2.6 charts until they are able to fully migrate over to 2.7. And the 1.0.0 and 1.0.1 will be hidden in v2.7 by an annotation.

maxsokolovsky: For v2.7, we plan to release only one version. Correct me if I am wrong, @MbolotSuse. But when I tried to remove 1.0.1, validation failed, even if I had removed its entry from release.yaml.

eliyamlevy: PR for versions 1.0.0 or 1.0.1 should go to dev-v2.6 branch.

maxsokolovsky: Only keeping one version (v2.0.0).

aiyengar2 (approved):

PR #2098, opened 2022-09-21
Set pod annotations in rancher-pushprox

### LINKED ISSUE #43487: Allow pod annotations in rancher-pushprox
For my use case I need to set pod annotations in the deployment of the client component of the rancher-pushprox Helm chart.

The current Helm chart doesn't allow such a configuration.

OK to do a PR about that?

sebastien-prudhomme:
Fix rancher/rancher#43487


nicholasSUSE:
Closing due to inactivity

PR #2102, opened 2022-09-22
Update rancher-alerting charts for Rancher v2.7.0

prachidamle:
https://github.com/rancher/rancher/issues/38982


release.yaml
```diff
@@ -24,3 +26,7 @@ rancher-monitoring-crd:
 - 101.0.0+up19.0.3
 rancher-project-monitoring:
 - 1.0.0+up0.1.0
+rancher-prom2teams:
+- 101.0.0+up0.2.0
+rancher-sachet:
+- 101.0.0
```
aiyengar2, line 32: these aren't being released!

packages/rancher-alerting/rancher-prom2teams/package.yaml
```diff
@@ -3,4 +3,3 @@ subdirectory: helm
 commit: a1691cce91855536be93311fe6b2a4b95b9ce97c # the commit points to the tag 3.2.3
 version: 101.0.0
 doNotRelease: true
-doNotRelease: true
```
aiyengar2, line 6: good catch, I didn't realize I added this key twice. Only one `doNotRelease` should exist (since the chart will already be released under [`charts/rancher-alerting-drivers/VERSION/charts/prom2teams`](https://github.com/rancher/charts/tree/dev-v2.7/charts/rancher-alerting-drivers/100.0.3/charts/prom2teams))

aiyengar2 (changes requested):

release.yaml
prachidamle: oh yeah!

aiyengar2 (approved):

PR #2104, opened 2022-09-22
Prepare rancher-webhook-external-ip 2.7

KevinJoiner:
## Issue https://github.com/rancher/rancher/issues/38954

This uses the same tag as rancher v2.6


cmurphy (commented):
Do we need to update the chart here first? https://github.com/rancher/externalip-webhook/tree/master/chart

KevinJoiner:
@cmurphy I don't think there is anything in https://github.com/rancher/externalip-webhook/tree/master/chart that needs to be updated.
For external-ip-webhook charts, we add the rancher version annotation in this repo via `packages/rancher-external-ip-webhook/generated-changes/patch/Chart.yaml.patch` in the `make charts` commit.

prachidamle:
@KevinJoiner we also need to change the following annotation in the Chart.yaml
` catalog.cattle.io/rancher-version: '>= 2.6.0-0 < 2.7.0-0'`

index.yaml
```diff
@@ -3966,6 +3966,39 @@ entries:
     - assets/rancher-eks-operator-crd/rancher-eks-operator-crd-100.0.0+up1.1.1.tgz
     version: 100.0.0+up1.1.1
   rancher-external-ip-webhook:
+  - annotations:
+      catalog.cattle.io/certified: rancher
+      catalog.cattle.io/display-name: External IP Webhook
+      catalog.cattle.io/kube-version: '>= 1.16.0-0 < 1.22.0-0'
```
prachidamle, line 3972: This will need to change since in 2.7.0 we support Kubernetes 1.23 and 1.24

index.yaml
```diff
@@ -3966,6 +3966,39 @@ entries:
...
+    description: |
+      Deploy the external-ip-webhook to mitigate k8s CVE-2020-8554
+    digest: 3c239ed2e9eb65c9e65fd4c4e915669469b1fdd08265e3ff7a4f92a22efab718
+    home: https://github.com/rancher/externalip-webhook
+    keywords:
+    - cve
+    - externalip
+    - webhook
+    - security
+    kubeVersion: < 1.22.0
```
prachidamle, line 3992: This will need to change since in 2.7.0 we support Kubernetes 1.23 and 1.24

index.yaml
cmurphy: No, for external-ip-webhook we specifically don't support using it on >=1.22 since this functionality is built into k8s in 1.22.

KevinJoiner: >Note: This chart is deprecated for kubernetes version 1.21 and unsupported starting with 1.22. To mitigate CVE-2020-8554, enable the [DenyServiceExternalIPs admission controller](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#denyserviceexternalips) on the cluster.

Based on this note should we just remove this chart from v2.7 charts since we no longer support 1.22

cmurphy (approved):

prachidamle (approved):

PR #2117, opened 2022-09-27
Add more restricted RBAC to cis benchmark chart

galal-hussein:
Add more restricted RBAC rules to cis-benchmark chart:

- Add clusterRole and bind it to cis-operator-serviceaccount which will be used by the controller pod
- Add role and bind it to cis-serviceaccount which will be used by the security scan pod

https://github.com/rancher/cis-operator/issues/165


brandond (dismissed):

brandond:
RBAC LGTM, I'm not sure what CI is failing on - do you need to `make charts` after changing things or something?

dereknola (dismissed):

prachidamle:
@galal-hussein you need to run `make charts` and add the output as second commit.

Also, can you please test run CIS scan permissive/hardened profiles on rke1/rke2/k3s  with these role changes to ensure the scan results are unchanged?

prachidamle:
@galal-hussein we should also test the upgrade cases to ensure that when a user with `rancher-cis-benchmark` chart already installed upgrades to this version of the chart, the scans continue to work

eliyamlevy:
@galal-hussein The build is failing on this one, please run `make validate` before pushing to ensure it will pass the build checks.

eliyamlevy (approved):

PR #2119, opened 2022-09-28
Remove Istio versions prior to 1.14.x

rayandas:
This PR removes `1.11` `1.12` and `1.13` from `packages/rancher-istio` dir.

Related: https://github.com/rancher/rancher/issues/39135


prachidamle:
@aiyengar2 do we need to remove all charts or keep them and just not create the 101 versions of them?

aiyengar2 (changes requested):
@rayandas all we need is the commit to remove the packages in https://github.com/rancher/charts/pull/2119/commits/85a0ad924527b74fd4b0014e3c29a2a64bd32187. The remaining charts should not be removed since `dev-v2.7` and `release-v2.7` are expected to be strict supersets of `release-v2.6`.

The 2.6 versions of Istio will only be hidden by Rancher annotation from a **fresh** install onto a 2.7 cluster; we still want to allow users to perform in-place upgrades of **existing** installs while they are completing the migration process

rayandas:
@aiyengar2 Updated the PR with your suggested changes.

aiyengar2 (approved):

PR #2121, opened 2022-09-28
Bump rancher-aks-operator v1.0.7-rc2

KevinJoiner:
Updates from rc1 to rc2 to include changes for https://github.com/rancher/rancher/issues/37927


cbron (approved):

charts/rancher-aks-operator-crd/101.0.0+up1.0.7-rc2/Chart.yaml
```diff
@@ -6,7 +6,7 @@ annotations:
   catalog.cattle.io/permits-os: linux,windows
   catalog.cattle.io/release-name: rancher-aks-operator-crd
 apiVersion: v2
-appVersion: 1.0.7-rc1
+appVersion: 1.0.7-rc2
```
cbron, line 9: Something weird is going on between 2.6 and 2.7: 
2.6 is at 1.0.7-rc1: https://github.com/rancher/charts/tree/dev-v2.6/charts/rancher-aks-operator
But so is 2.7: https://github.com/rancher/charts/tree/dev-v2.7/charts/rancher-aks-operator
So I think 1.0.7 needs to be removed from 2.7 and the minor or major version needs to get bumped. 
cc @annablender / @snasovich

cbron (changes requested):
Branching problems

cbron (approved):
I think this is all correct now.

prachidamle (approved):

PR #2133, opened 2022-10-04
Release longhorn v1.3.2 into Rancher 2.7

PhanLe1010:
https://github.com/rancher/rancher/issues/39224


packages/longhorn/longhorn-1.3/generated-changes/patch/Chart.yaml.patch
```diff
@@ -9,11 +9,11 @@
 +  catalog.cattle.io/namespace: longhorn-system
 +  catalog.cattle.io/os: linux
 +  catalog.cattle.io/provides-gvr: longhorn.io/v1beta1
-+  catalog.cattle.io/rancher-version: '>= 2.6.0-0 < 2.7.0-0'
++  catalog.cattle.io/rancher-version: '>= 2.7.0-0 < 2.8.0-0'
```
PhanLe1010, line 12: How do you think @aiyengar2 @innobead @derekbit ?

innobead (changes requested):
In general, LGTM, but the same question about the exclude in the crd chart. 

Also, suggest consolidate these two commits because there was one for 1.3.2-rc1.

PhanLe1010:
> In general, LGTM, but the same question about the exclude in the crd chart.

I responded in the other thread. Thanks 

> Also, suggest consolidate these two commits because there was one for 1.3.2-rc1.

Thank you. Fixed

innobead (dismissed):

khushboo-rancher:
Validated on Rancher 2.7.0-rc7, Installation, uninstallation and some P0 checks are working fine.

khushboo-rancher:
I verified the upgrade path as below, LGTM.
Longhorn v1.3.1 on Rancher 2.6.9 --> Rancher 2.7.0 --> Longhorn 1.3.2

PhanLe1010:
@aiyengar2 @PennyScissors Can we move this one forward?

PennyScissors:
Please split this into 2 separate commits. First should be changes made packages and release.yaml, and the second should be the output of running make charts.

pennyscissors (changes requested):
Please split this into 2 separate commits. First should be changes made packages and release.yaml, and the second should be the output of running make charts.

PhanLe1010:
> Please split this into 2 separate commits. First should be changes made packages and release.yaml, and the second should be the output of running make charts.

@PennyScissors Done. Thank you

yangchiu:
Test installallation and uninstallation on rancher v2.7.0: **Pass**

packages/longhorn/longhorn-1.3/generated-changes/patch/Chart.yaml.patch
pennyscissors: lgtm

pennyscissors (approved):

PR #2136, opened 2022-10-05
Modify the minimum kubernetes version for cis-1.23

galal-hussein:


prachidamle:
@galal-hussein To fix the CI failures you also need to add an entry in release.yaml for the CIS charts. Also you need to run the following and add the files as second commit 
```
export PACKAGE=rancher-cis-benchmark
make charts
```

prachidamle:
Also not sure if we have the approval to add this is yet. Pls confirm

galal-hussein:
Closing this PR, since we are reverting the PSP to PSA changes to the cis-1.23

PR #2168, opened 2022-10-18
Update ui-plugin-operator commit ref

pennyscissors:
Issue: https://github.com/rancher/rancher/issues/39323


release.yaml
```diff
@@ -89,25 +89,25 @@ rancher-webhook:
...
 - 0.0.100
 harvester-cloud-provider:
-  - 101.0.0+up0.1.13
+- 101.0.0+up0.1.13
 harvester-csi-driver:
-  - 101.0.0+up0.1.14
+- 101.0.0+up0.1.14
 system-upgrade-controller:
 - 101.0.0+up0.3.3
 ui-plugin-operator:
```
PennyScissors, line 107: Only change that matters below, the rest is just formatting

release.yaml
```diff
@@ -105,9 +105,9 @@ harvester-csi-driver:
...
-- 101.0.0+up0.1.0
+- 101.0.0+up0.1.0-rc3
 ui-plugin-operator-crd:
-- 101.0.0+up0.1.0
+- 101.0.0+up0.1.0-rc3
```
prachidamle, line 110: We are moving backwards from 0.1.0 to 0.1.0-rc3?

aiyengar2 (approved):

release.yaml
aiyengar2: We’re not moving backwards; previous updates of the chart never modified the rc version in the Chart.yaml itself

PR #2173, opened 2022-10-21
Sync up release-v2.6 with dev-v2.6 after Longhorn OOB release

PennyScissors:


aiyengar2 (approved):
LGTM. Seems to also be addressing the fact that we did not empty out the release.yaml after 2.6.9

PR #2187, opened 2022-11-01
Add NeuVector chart version 2.2.4

selvamt94:


release.yaml
```diff
@@ -18,8 +18,10 @@ longhorn-crd:
 - 100.2.3+up1.3.2
 neuvector:
 - 101.0.0+up2.2.3
+- 101.0.1+up2.2.4
 neuvector-crd:
 - 101.0.0+up2.2.3
+- 101.0.1+up2.2.4
```
cbron, line 24: Seems like release 2.7 is out of sync. We should only be releasing one at a time, but 2.2.3 is already live in release 2.6. So 2.7.0 is only "releasing" 2.2.4.

cmurphy (approved):

cbron (approved):

PR #2188, opened 2022-11-01
Add NeuVector chart version 2.2.4

selvamt94:


cmurphy (approved):

packages/neuvector/generated-changes/overlay/questions.yaml
```diff
@@ -1,11 +1,5 @@
 questions:
 #image configurations
-- variable: registry
-  default: "docker.io"
-  description: image registry
-  type: string
-  label: Image Registry
-  group: "Container Images"
```
cbron, line 8: Sort of surprised this got backported into 2.6. I know we made this switch in 2.7, just want to make sure we are doing this purposefully in 2.6.

cbron (approved):
one comment, otherwise lgtm

packages/neuvector/generated-changes/overlay/questions.yaml
selvamt94: @cbron In 2.6 also, NeuVector is not using registry from questions.yaml. It is a bug, so removed from questions.yaml.  Registry need to be set only by systemDefaultRegistry. 
https://github.com/selvamt94/charts/blob/dev-v2.6/charts/neuvector/100.0.2%2Bup2.2.3/templates/controller-deployment.yaml#L59

cbron: Ok makes sense

cbron:
Needs rebase, then we can merge

PR #2190, opened 2022-11-02
[dev-v2.7] Update vSphere CSI chart to 101.1.0+up2.6.1-rancher1

rayandas:
Related issue: https://github.com/rancher/rancher/issues/39462 and https://github.com/rancher/rancher/issues/38188
For CPI the version restriction is:
`catalog.cattle.io/kube-version: '>= 1.18.0-0 < 1.25.0-0'`
For CSI the version restriction is:
`catalog.cattle.io/kube-version: '>= 1.20.0-0 < 1.24.0-0'`

While creating RKE2 cluster, it defaults to K8s 1.24, as the current CSI chart doesn't support for K8s 1.24, that's why the CSI configuration is not visible. This PR changes might fix it.


jiaqiluo:
@rayandas   
> For CSI the version restriction is:
catalog.cattle.io/kube-version: '>= 1.20.0-0 < 1.24.0-0'

This is because the upstream CSI v2.5.1 does not support 1.24. The support is added in upstream CSI v2.6.0 https://github.com/kubernetes-sigs/vsphere-csi-driver/releases/tag/v2.6.0

Our CSI chart is maintained at  https://github.com/rancher/vsphere-charts/tree/main/charts/rancher-vsphere-csi.
The good news is that we already update our CSI chart to match the latest version in the upstream which supports 1.24.
So instead of making this PR, we should update the package.yaml to point to a new commit in rancher/vsphere-chart.

rayandas:
@jiaqiluo Yeah I updated the chart.

packages/rancher-vsphere/rancher-vsphere-csi/generated-changes/patch/Chart.yaml.patch
```diff
@@ -4,8 +4,8 @@
    catalog.cattle.io/namespace: kube-system
    catalog.cattle.io/os: linux,windows
    catalog.cattle.io/permits-os: linux,windows
--  catalog.cattle.io/rancher-version: '>= 2.6.0-0 <= 2.6.100-0'
+-  catalog.cattle.io/rancher-version: '>= 2.6.0-0'
 +  catalog.cattle.io/rancher-version: '>= 2.7.0-0 < 2.8.0-0'
    catalog.cattle.io/release-name: vsphere-csi
  apiVersion: v1
- appVersion: 2.5.1-rancher1
+ appVersion: 2.6.1-rancher1
```
annablender, line 11: @brandond recently discovered the appVersion needs to be updated when updates are done to CSI/CPI charts. For CSI, this should be `2.6.2-rancher1.` Could we also update CPI in this PR as well?

release.yaml
```diff
@@ -65,7 +65,7 @@ rancher-project-monitoring:
 rancher-vsphere-cpi:
 - 101.0.0+up1.2.2
 rancher-vsphere-csi:
-- 101.0.0+up2.5.1-rancher1
+- 101.1.0+up2.6.1-rancher1
```
annablender, line 68: Should be `2.6.2-rancher1`

annablender (changes requested):

a-blender (approved):
LGTM. I'm going to put in a PR that will bump CSI to `v2.6.2` as per the current chart changes but this is fine for the UI bug fix

snasovich:
@annablender , I don't believe we're proceeding with this PR, let's just bump it straight to 2.6.2 or even 2.7.0 that you've introduced in `vsphere-charts` supercharts per https://github.com/rancher/vsphere-charts/pull/35 (more context here - https://github.com/rancher/rancher/issues/39462#issuecomment-1306433102).

PR #2204, opened 2022-11-09
[v2.7.1] Update vSphere CSI and CPI charts

a-blender:
Issue: 

https://github.com/rancher/rancher/issues/38188
https://github.com/rancher/rancher/issues/38187

Update vsphere CSI/CPI charts

```
rancher-vsphere-cpi: 101.1.0+up1.4.1
rancher-vsphere-csi: 101.1.0+up2.6.2-rancher1
```

We have a confusing versioning process for CSI/CPI in rancher charts. When doing updates, I had to guess what `appVersion` to bump to because it's not clear, and not aligned with k8s versions or anything else. Based on [this thread](https://suse.slack.com/archives/C0383T4RAFP/p1667946216668639), I think @brandond and I should spent some time next release cycle hammering this out so it's not so messy. `appVersion` is optional and informational in the [helm docs](https://helm.sh/docs/topics/charts/#the-appversion-field). The vsphere chart doesn't really bundle "an app" so we could remove `appVersion` and use the latest CPI version as the `appVersion,` eg: 1.24.0.

Next release cycle, we should
* Bump the CSI chart version to 2.7.0-rancher1 (or 2.7.x)
* Possibly remove the default images from `values.yaml.` The `versionOverrides` appear to always be used based on the Kubernetes version, instead of the default.
* Possibly remove `appVersion` from chart metadata


prachidamle:
@annablender could you please add in the release this PR is intended for in the title? I think this is for 2.7.1?

brandond (changes requested):
The presence of .orig and .rej files indicates that the patch didn't apply cleanly. You'll need to rework the patches, or drop them if they're no longer necessary. And remove the files created by the failed patch, of course,

packages/rancher-vsphere/rancher-vsphere-cpi/generated-changes/patch/questions.yaml.patch
```diff
@@ -0,0 +1,26 @@
+--- charts-original/questions.yaml
++++ charts/questions.yaml
+@@ -40,23 +40,3 @@
+     type: string
+     group: Configuration
+     show_if: "vCenter.credentialsSecret.generate=false"
+-
+-  - variable: vCenter.labels.generate
```
brandond, line 8: Seeing a bunch of new patches removing bits from the chart; is that intentional?

snasovich (changes requested):
I don't believe it's expected that previous chart version be removed (see deleted `tgz`s in assets folder). See on of previous PRs that bumped versions as example: https://github.com/rancher/charts/pull/1842/files.

That would make sense if were updating not-yet-released chart, but `rancher-vsphere-cpi-101.0.0+up1.2.2` will be released with 2.7.0 so we can't simply replace it with `rancher-vsphere-cpi-101.0.0+up1.4.1.tgz`, we need to introduce new `rancher-vsphere-cpi-101.1.0+up1.4.1.tgz`.

packages/rancher-vsphere/rancher-vsphere-cpi/generated-changes/patch/questions.yaml.patch
a-blender: No, looks like the patch undid a lot of the changes I need... after talking to @snasovich I apparently also need to bump to supercharge version 100.1, because 100.0 is already being released for 2.7.0. I'll make those changes and re-push to the branch.

release.yaml
```diff
@@ -0,0 +1,6 @@
+rancher-vsphere-cpi:
+  - 101.0.0+up1.2.2
```
snasovich: I think we only want to list newly introduced charts. `rancher-vsphere-cpi:101.0.0+up1.2.2` and `rancher-vsphere-csi:101.0.0+up2.5.1-rancher1` were released with 2.7.0.

charts/rancher-vsphere-csi/101.1.0+up2.6.2-rancher1/Chart.yaml
```diff
@@ -0,0 +1,22 @@
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/display-name: vSphere CSI
+  catalog.cattle.io/kube-version: '>= 1.20.0-0 < 1.24.0-0'
```
snasovich: This is not what we expect to see here. The version of chart we're pulling has different constraint (and a lot of other stuff is different):
https://github.com/rancher/vsphere-charts/blob/ba41db5e0814f719722954441edf409c12c45dd4/charts/rancher-vsphere-csi/Chart.yaml#L4

This is probably due to all these `generated-changes/patch` files in `packages` folder that apply updates to charts pulled from `rancher/vsphere-charts` and result in final charts that look incorrect.

I'm not sure how you ended up with `generated-changes/patch` files but the end result is clearly wrong. See another relevant comment here: https://github.com/rancher/charts/pull/2204#discussion_r1019661139.

snasovich (changes requested):

charts/rancher-vsphere-csi/101.1.0+up2.6.2-rancher1/Chart.yaml
a-blender: This resulted from adding a patch because I was receiving other errors with `make charts` and thought I needed to patch. A `make patch` patches the chart with updates in the package/generated-changes folder, so I needed to remove that dir and regenerate it _after_ my updates to fix it. I checked the new Chart.yaml and deployment files and they are in line with upstream

brandond (approved):
I don't know what the `release.yaml` is for, but other than that it lgtm.

a-blender:
@brandond `release.yaml` is for tracking unreleased versions of all charts. We just released 2.7, so I emptied release.yaml and it's only tracking the newest versions of csi/cpi right now, which have not been released yet. Thank you!

snasovich (approved):

PR #2233, opened 2022-11-23
Adding clusterid to alerts #2167

geethub97:
Official PR for https://github.com/rancher/rancher/issues/37128

Draft PR: https://github.com/rancher/charts/pull/2167


hoerup:
Will it be backported to 2.6 as well ?

packages/rancher-monitoring/rancher-monitoring/generated-changes/patch/templates/prometheus/rules-1.14/alertmanager.rules.yaml.patch
```diff
@@ -0,0 +1,66 @@
+--- charts-original/templates/prometheus/rules-1.14/alertmanager.rules.yaml
++++ charts/templates/prometheus/rules-1.14/alertmanager.rules.yaml
+@@ -42,6 +42,7 @@
+       for: 10m
+       labels:
+         severity: critical
++        cluster_id: {{.Values.global.cattle.clusterId}}
```
aiyengar2, line 7: Can we make all of these references `{{ .Values.global.cattle.clusterId }}` instead of `{{.Values.global.cattle.clusterId}}`

Also, what happens if `global.cattle.clusterId` is empty (for CI, for example)? Can we conditionally remove this if it's not provided? like `{{ if .Vales.global.cattle.clusterId }}` or similar

aiyengar2 (changes requested):
While this is a valid solution to proceed forward, we should investigate whether we can add the clusterID on **scrape** instead of adding it to every single rule, since this is a lot harder to maintain.

I'd suspect you can do this by setting the `.spec.endpoints.metricRelabelings[]` on each ServiceMonitor, which should apply the labels before **ingestion** .

See https://albersdevelopment.net/2019/08/28/prometheus-adding-a-label-to-a-target/ for more information, you can basically target the `__address__` sourceLabel and define a brand new `target_label` (`cluster_id`) with `{{ .Values.global.cattle.clusterId }}` under `replacement`.

PR #2243, opened 2022-11-30
Bump rancher-webhook to 0.2.8-rc1

MbolotSuse:
Bumping rancher webhook to v0.2.8-rc1


packages/rancher-webhook/package.yaml
```diff
@@ -1,2 +1,2 @@
-url: https://github.com/rancher/webhook/releases/download/v0.2.7/rancher-webhook-0.2.7.tgz
-version: 1.0.6
+url: https://github.com/rancher/webhook/releases/download/v0.2.8-rc1/rancher-webhook-0.2.8-rc1.tgz
+version: 1.0.7
```
maxsokolovsky, line 2: Is this bumped because the webhook's patch version also was changed?

MbolotSuse: No. We bump this value to indicate compatibility with a new version of rancher. Since we intend the next 2.6 release to have this version of the webhook (but still want the charts repo to have record of current setups so that 2.6.9 doesn't break), we bump this version.

maxsokolovsky (approved):

KevinJoiner (approved):

PR #2259, opened 2022-12-09
[dev v2.7] updated alerting drivers chart

vardhaman22:
related to issue: https://github.com/rancher/rancher/issues/39784
also contains psp related fixes for sachet, alerting drivers and prom2teams chart:

https://github.com/rancher/rancher/issues/39849 psp prom2teams
https://github.com/rancher/rancher/issues/39848 psp alerting drivers

Adds:
rancher-alerting-drivers: 101.0.1

dependency:
needs image mirror pr: https://github.com/rancher/image-mirror/pull/327 to be merged first

The fixes for alerting drivers ticket need to be tested after Monitoring PSP changes are merged in https://github.com/rancher/rancher/issues/39846


packages/rancher-alerting/rancher-alerting-drivers/charts/Chart.yaml
```diff
@@ -1,13 +1,13 @@
 annotations:
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/display-name: Alerting Drivers
-  catalog.cattle.io/kube-version: '>= 1.16.0-0 < 1.25.0-0'
+  catalog.cattle.io/kube-version: '>= 1.16.0-0 < 1.26.0-0'
```
prachidamle, line 4: Have we tested this chart on k8s 1.25 cluster? You could test it by provisioning an rke1 k8s 1.25 cluster on v2.7-head setup 

We will need to make changes is PSP's are used as mentioned in this ticket for supporting k8s 1.25 https://github.com/rancher/rancher/issues/39848

vardhaman22: tested installing the chart on 1.25 and applied the psp related changes as well. @prachidamle

stormqueen1990 (commented):
LGTM!

I'm just a bit concerned as to whether these deployments will work if they're using the `default` service account for the namespace.

@aiyengar2 do you have more context on the service account rules?

aiyengar2:
> LGTM!
> 
> I'm just a bit concerned as to whether these deployments will work if they're using the `default` service account for the namespace.
> 
> @aiyengar2 do you have more context on the service account rules?

As far as I know, I think that you need a ServiceAccount to be attached to workloads in hardened setups since hardened clusters turn off automounting the default service account in a namespace, but I’m not sure if that guidance will change in 1.25.

To be safe, we should probably keep the ServiceAccounts for all the workloads. But we’d need to test this in a hardened setup to be sure.

vardhaman22:
> > LGTM!
> > I'm just a bit concerned as to whether these deployments will work if they're using the `default` service account for the namespace.
> > @aiyengar2 do you have more context on the service account rules?
> 
> As far as I know, I think that you need a ServiceAccount to be attached to workloads in hardened setups since hardened clusters turn off automounting the default service account in a namespace, but I’m not sure if that guidance will change in 1.25.
> 
> To be safe, we should probably keep the ServiceAccounts for all the workloads. But we’d need to test this in a hardened setup to be sure.

added back the service account in the deployments and removed the condition from service account files. 
@aiyengar2

stormqueen1990 (dismissed):
LGTM!

prachidamle (dismissed):

stormqueen1990 (dismissed):

mitulshah-suse (approved):

PR #2263, opened 2022-12-12
Using upstream queries to resolve latency for Rancher Pods view and Grafana Rancher / Pod and Rancher / Pod (Containers) dashboards

geethub97:
https://github.com/rancher/rancher/issues/36107


aiyengar2 (approved):
LGTM

PR #2265, opened 2022-12-13
KEv2 branch operators

a-blender:
Issue: https://github.com/rancher/rancher/issues/39860

rancher/charts:dev-v2.7
* **AKS operator bump to v1.1.0-rc1**
* **EKS operator bump to 1.2.0-rc2**
* GKE operator - nothing, we are using v1.1.5

AKS operator has a new branch `release/v1.0` and EKS operator has a new branch `release/v1.1` that tracks changes in both v2.6 and v2.7. Rancher 2.7 needs to use the latest bug fixes and features from both operators.

I am following semvr for this PR. Both charts were at version 101.1.0. I am bumping the **patch** version of the AKS operator chart to 101.1.1+up[1.1.0-rc1](https://github.com/rancher/aks-operator/releases/tag/v1.1.0-rc1) because we are only including bug fixes in the tag after branching. I am bumping the **minor** version of the EKS operator chart to 101.2.0+up[1.2.0-rc2](https://github.com/rancher/eks-operator/releases/tag/v1.2.0-rc2) because we are including both bug fixes and the custom [node group node role](https://github.com/rancher/eks-operator/commit/13ad27e6563532ed16ee9a97615cde62f6e419e4) RFE which is added functionality.


snasovich (approved):
LGTM, but I would like to hear @aiyengar2 's input regarding minor/patch versions bumps - not sure how exactly we version charts.

aiyengar2 (approved):
re: 

> LGTM, but I would like to hear @aiyengar2 's input regarding minor/patch versions bumps - not sure how exactly we version charts.

Looks good to me. The general guidance with bumping the minor version is that it's up to a chart developer; you generally want a cut a minor when there's a possibility you might want to retain multiple release lines for a chart, if you're adding major functionality or refactoring to it (for example, the Monitoring chart bumped a minor version cause we rebased to 30+ major versions up from the base).

> I am bumping the minor version of the EKS operator chart to 101.2.0+up[1.2.0-rc2](https://github.com/rancher/eks-operator/releases/tag/v1.2.0-rc2) because we are including both bug fixes and the custom [node group node role](https://github.com/rancher/eks-operator/commit/13ad27e6563532ed16ee9a97615cde62f6e419e4) RFE which is added functionality.

I'm not familiar with this added functionality and its impact, but if it is a major change that might merit maintaining the ability to cut a fix to the older version of the chart, this sounds reasonable to me. Always safer to cut a new minor than not to.

PR #2270, opened 2022-12-15
[dev-v2.7 CIS] Add condition to check PSP compatibility 

rayandas:
Related issue: https://github.com/rancher/rancher/issues/39851


stormqueen1990:
I believe we will need to hold off on this change as it seems that even if we remove PSPs this chart may not work on Kubernetes v1.25.

cc @prachidamle

prachidamle:
yes, lets hold off on merging this PR for now since we dont have a CIS scan profile supported for k8s 1.25 yet. 
@rayandas is it correct that there will be no scan profile available on rancher dashboard for a user on k8s 1.25 since we dont have a CIS profile for k8s 1.25 supported? On a k8s 1,25 cluster, will the user be able to run any CIS scan?

rayandas:
@stormqueen1990 @prachidamle There's some progress here.
Earlier we were using `{{- if and .Values.global.rbac.pspEnabled (.Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy") }}` in `charts/rancher-cis-benchmark/3.0.1-rc2/templates/rbac.yaml` but that was incorrect also the crd was not getting installed because of the psp compatibility (but this was fixed in https://github.com/rancher/charts/pull/2274)

We had to use `{{- if .Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy" }}` instead to check the compatibility.

I installed `rancher-cis-benchmark-3.0.1-rc3` (from https://github.com/rayandas/charts/tree/cis-psp-compatibility branch) on Rancher v2.7-head (a2ef50b) with K8s `v1.25.2`. CIS Benchmark app got installed successfully. 
Also I ran 3 scans, only `rke-profile-permissive-1.23` was running and the tests were successful. 1.20 and 1.6 scans are incompatible. 

Screenshot from 2022-12-26 17-43-49

stormqueen1990:
These changes LGTM!

prachidamle (dismissed):

stormqueen1990 (dismissed):

prachidamle:
@doflamingo721 The commit message says "add condition to check for PSP capability in rancher-gatekeeper-crd" - needs to be modified to reflect the rancher-cis-benchmark changes

doflamingo721:
@prachidamle Updated the commit message.

prachidamle (approved):

PR #2279, opened 2022-12-20
Add Kubernetes 1.25 support to KEv2 operator charts

a-blender:
Issue: https://github.com/rancher/rancher/issues/39929

I bumped the kube annotation `catalog.cattle.io/kube-version` in the kev2 operator charts to `>= 1.20.0-0 < 1.26.0-0` so they support Kubernetes 1.25. This is only being done for v2.7 so I only updated the operator tags that are being used by Rancher 2.7 so aks-operator v1.1.0-rc1 eks-operator v1.2.0-rc2 and gke-operator v1.1.5. Branch is also only being merged into charts/dev-v2.7 so no backport.

The gke-operator chart has been bumped one patch version to `101.0.1` with the updated constraints because v1.1.5 was already released for Rancher 2.7.0.


release.yaml
```diff
@@ -56,3 +56,7 @@ sriov:
 - 101.0.2+up0.1.0
 sriov-crd:
 - 101.0.2+up0.1.0
+rancher-gke-operator:
+- 101.0.0+up1.1.5
```
snasovich: `rancher-gke-operator` and `rancher-gke-operator-crd` were released with version `101.0.0+up1.1.5` for 2.7.0 (see https://github.com/rancher/charts/tree/release-v2.7/charts/rancher-gke-operator) so I don't think we can update this chart any longer.
We need to introduce a new one (I guess 101.0.1+up1.1.5) with these constraints updated.

aks and eks operators are fine as we introduced new versions for upcoming 2.7 release already so we can continue modifying them until released.

jiaqiluo (dismissed):
LGTM

snasovich (changes requested):

release.yaml
a-blender: @snasovich Ahh I see. I would think `101.1.0` since introducing 1.25 support is pulling in new functionality, not a bug fix but it's also a small change code wise. Thoughts?

Update: I think 101.0.1 is fine

jiaqiluo (changes requested):
Can we do a rebase, instead of merge, to fix the conflicts? 
Otherwise LGTM

snasovich (dismissed):

jiaqiluo (approved):

sbstp (approved):

snasovich (approved):

PR #2280, opened 2022-12-20
Bump to fleet 0.6.0 rc.1

thardeck:


packages/fleet/fleet-agent/package.yaml
```diff
@@ -1,2 +1,2 @@
-url: https://github.com/rancher/fleet/releases/download/v0.5.0/fleet-agent-0.5.0.tgz
+url: https://github.com/rancher/fleet/releases/download/v0.6.0-rc.1/fleet-agent-0.6.0-rc.1.tgz
 version: 101.0.0
```
mattfarina: The version in all 4 charts needs to be incremented to `101.1.0`

release.yaml
```diff
@@ -56,3 +56,18 @@ sriov:
...
+- 101.1.0+up0.6.0-rc.1
+fleet:
+- 101.0.0+up0.5.0
+- 101.1.0+up0.6.0-rc.1
+fleet-agent:
+- 101.0.0+up0.5.0
+- 101.1.0+up0.6.0-rc.1
+fleet-crd:
+- 101.0.0+up0.5.0
+- 101.1.0+up0.6.0-rc.1
```
mattfarina: There's duplication of keys here.

mattfarina (changes requested):

release.yaml
thardeck: Good catch, it seems one of the scripts extended it.

mattfarina (approved):

PR #2281, opened 2022-12-20
Bump rancher-webhook v0.3.2 rc1

crobby:


charts/rancher-webhook/2.0.0+up0.3.2-rc1/Chart.yaml
```diff
@@ -0,0 +1,18 @@
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/hidden: "true"
+  catalog.cattle.io/kube-version: '>= 1.16.0-0 < 1.25.0-0'
```
maxsokolovsky: Does this need to be updated somehow?

KevinJoiner: This RC should still function for 1.16.0 since we are only adding namespace validation.
Though it will eventually need to be updated here https://github.com/rancher/webhook/blob/release/v0.3/charts/rancher-webhook/Chart.yaml

maxsokolovsky: I was thinking about the high version. This version of the webhook still should not be deployed on a local or remote cluster that's >=1.25?

KevinJoiner: Technically not this RC since go.mod and go.sum still point to 1.24

KevinJoiner (approved):

maxsokolovsky (approved):

PR #2282, opened 2022-12-20
Wrap PSPs in capabilities conditional for k8s 1.25 support

geethub97:
https://github.com/rancher/rancher/issues/39846


packages/rancher-monitoring/rancher-pushprox/charts/templates/pushprox-clients-rbac.yaml
```diff
@@ -78,4 +81,5 @@ spec:
   - pathPrefix: {{ required "Need access to volume on host with the SSL cert files to use HTTPs" .Values.clients.https.certDir }}
     readOnly: true
 {{- end }}
-{{- end }}{{- end }}
\ No newline at end of file
+{{- end }}{{- end }}
+{{- end }}
```
aiyengar2: nit:

this should probably be swapped to:

```
{{- end }}
{{- end }}
{{- end }}{{- end }}
```

Since that's the way they get called in the `if` blocks, the only reason why the last two blocks are together is because this file starts with two `if` blocks in the same line, re: `{{- if .Values.clients }}{{- if .Values.clients.enabled }}`

Not a large enough issue to block approval though

aiyengar2 (commented):
LGTM, merge all the `make charts` commits into one final commit and have a small recommendation for one of the files in my comments below.

aiyengar2 (dismissed):

aiyengar2 (approved):

PR #2283, opened 2022-12-20
Bump webhook and csp-adapter

cmurphy:


release.yaml
```diff
@@ -16,6 +16,8 @@ rancher-aks-operator:
 rancher-aks-operator-crd:
 - 101.1.0+up1.0.8-rc1
 - 101.1.1+up1.1.0-rc1
+rancher-csp-adapter:
+- 2.0.1+up2.0.1-rc1
```
cbron, line 20: I thought csp-adapter didn't have the `+up` reference, was that just put on by charts scripts ?

cmurphy: `make validate` fails if it doesn't look like this. I don't know the history of this chart's versioning scheme.

cbron: @MbolotSuse can you confirm this when you are back, approving for now to get things moving

cbron (approved):

PR #2298, opened 2022-12-28
Istio components hpa and pdb config

vardhaman22:
workaround for issue: https://github.com/rancher/rancher/issues/30082

added replica and podDisruptionBudget configuration for istio components
1. egressgateway
2. ingressgateway
3. pilot

these will be useful in case user needs to have a HA istio and  PDB configuration will allow users to drain nodes for non HA istio installation.


aiyengar2 (dismissed):

charts/rancher-istio/101.1.0+up1.15.3/configs/istio-base.yaml
```diff
@@ -19,6 +19,10 @@ spec:
     - enabled: {{ .Values.egressGateways.enabled }}
       name: istio-egressgateway
       k8s:
+        hpaSpec:
+          minReplicas: {{ .Values.egressGateways.minReplicas }}
+        podDisruptionBudget:
+          minAvailable: {{ .Values.egressGateways.podDisruptionBudget.minAvailable }}
```
aiyengar2: Is there any reason why we want to expose only these specific fields? IMO, we should just allow users to provide the blanket `.Values.*.hpaSpec` and `.Values.*.podDistruptionBudget` they desire:

```suggestion
{{- if .Values.egressGateways.hpaSpec }}
        hpaSpec: {{ toYaml .Values.egressGateways.hpaSpec | nindent 10 }}
{{- end }}
{{- if .Values.egressGateways.podDisruptionBudget }}
        podDisruptionBudget: {{ toYaml .Values.egressGateways.podDisruptionBudget | nindent 10 }}
{{- end }}
```

aiyengar2 (commented):
Have a question

charts/rancher-istio/101.1.0+up1.15.3/configs/istio-base.yaml
aiyengar2: @Tejeev thoughts here?

vardhaman22: this suggestion seems good to me
updated the PR with suggested changes.

aiyengar2 (approved):

prachidamle (approved):

charts/rancher-istio/101.1.0+up1.15.3/configs/istio-base.yaml
Tejeev: Discussed with Fred as well, and none of us can think of any reason not to open up these HA management configurations to the users. I know I've seen at least one customer (deploying Istio independently) that managed their Istio replica counts among other related configs, which I didn't have the opportunity to dig into, so you just never know what they may want to configure, and I think it makes sense to give them the options (but with good defaults) since there isn't anything inherently insecure about opening up this normal k8s configuration for the users.

Worst case, someone shoots themselves with a missed footgun here .. and we tell them to go back to the tested defaults.

PR #2366, opened 2023-01-27
Updating metricRelabelings to include cluster ids for all service monitors in Rancher Monitoring

geethub97:
https://github.com/rancher/rancher/issues/37128


geethub97:
image

geethub97:
image

charts/rancher-monitoring/101.1.0+up40.1.2/charts/grafana/templates/servicemonitor.yaml
```diff
@@ -30,6 +30,22 @@ spec:
...
+      replacement: {{ .Values.global.cattle.clusterId }}
+    {{- end }}
+    {{ else }}
+    {{ if .Values.global.cattle.clusterId }}
+    metricRelabelings:
+    - sourceLabels: [__address__]
+      targetLabel: cluster_id
+      replacement: {{ .Values.global.cattle.clusterId }}
+    {{- end }}
+    {{- end }}
```
aiyengar2, line 48:
```suggestion
    metricRelabelings:
    {{- if .Values.serviceMonitor.metricRelabelings }}
    {{- toYaml .Values.serviceMonitor.metricRelabelings | nindent 6 }}
    {{- end }}
    {{ if .Values.global.cattle.clusterId }}
    - sourceLabels: [__address__]
      targetLabel: cluster_id
      replacement: {{ .Values.global.cattle.clusterId }}
    {{- end }}
```

aiyengar2: If we're always going to add `metricRelabelings`, we should just remove it from the if block.

charts/rancher-monitoring/101.1.0+up40.1.2/charts/hardenedKubelet/templates/pushprox-servicemonitor.yaml
```diff
@@ -8,6 +8,13 @@ metadata:
   labels: {{ include "pushProxy.serviceMonitor.labels" . | nindent 4 }}
 spec:
   endpoints: {{include "pushProxy.serviceMonitor.endpoints" . | nindent 4 }}
+    {{ if .Values.global.cattle.clusterId }}
+    - metricRelabelings:
+      - action: replace
+        sourceLabels: [__address__]
+        targetLabel: cluster_id
+        replacement: {{ .Values.global.cattle.clusterId }}
+    {{- end }}
```
aiyengar2: I don't think this works. We'd need to add it to **each** endpoint, not define a separate endpoint that does metricRelabelings.

The fix here should be put into the template within `_helpers.tpl`.

charts/rancher-monitoring/101.1.0+up40.1.2/charts/kube-state-metrics/templates/servicemonitor.yaml
```diff
@@ -35,6 +35,18 @@ spec:
...
+          targetLabel: cluster_id
+          replacement: {{ .Values.global.cattle.clusterId }}
+         {{- end }}
+     {{ else }}
+       {{ if .Values.global.cattle.clusterId }}
+      metricRelabelings:
+        - sourceLabels: [__address__]
+          targetLabel: cluster_id
+          replacement: {{ .Values.global.cattle.clusterId }}
+       {{- end }}
```
aiyengar2, line 49: same comment as above; move the `metricsRelabeling` outside the `if` statement and you can avoid repeating code.

charts/rancher-monitoring/101.1.0+up40.1.2/charts/prometheus-node-exporter/templates/servicemonitor.yaml
```diff
@@ -47,5 +47,17 @@ spec:
     {{- with .Values.prometheus.monitor.metricRelabelings }}
       metricRelabelings:
         {{- toYaml . | nindent 8 }}
+        {{ if .Values.global.cattle.clusterId }}
```
aiyengar2, line 50: same comment as above regarding moving `metrics-relabeling` outside the `if` statement.

charts/rancher-monitoring/101.1.0+up40.1.2/templates/alertmanager/servicemonitor.yaml
```diff
@@ -37,6 +37,18 @@ spec:
 {{- if .Values.alertmanager.serviceMonitor.metricRelabelings }}
     metricRelabelings:
 {{ tpl (toYaml .Values.alertmanager.serviceMonitor.metricRelabelings | indent 6) . }}
+    {{ if .Values.global.cattle.clusterId }}
```
aiyengar2, line 40: same comment as above regarding moving `metrics-relabeling` outside the `if` statement.

charts/rancher-monitoring/101.1.0+up40.1.2/templates/exporters/core-dns/servicemonitor.yaml
```diff
@@ -31,6 +31,18 @@ spec:
 {{- if .Values.coreDns.serviceMonitor.metricRelabelings }}
     metricRelabelings:
 {{ tpl (toYaml .Values.coreDns.serviceMonitor.metricRelabelings | indent 4) . }}
+    {{ if .Values.global.cattle.clusterId }}
```
aiyengar2, line 34: same comment as above regarding moving `metrics-relabeling` outside the `if` statement.

aiyengar2 (changes requested):

charts/rancher-monitoring/101.1.0+up40.1.2/charts/grafana/templates/servicemonitor.yaml
aiyengar2: This should be fixed in all non-pushprox charts.

geethub97: some of the charts have values for metricRelabelings in the values.yaml file, but not all of them do so that was the primary reason for the check 

the other reason is in the comments here: 
https://github.com/rancher/charts/pull/2233#discussion_r1048843829

> Also, what happens if `global.cattle.clusterId` is empty (for CI, for example)? Can we conditionally remove this if it's not provided? like `{{ if .Vales.global.cattle.clusterId }}` or similar

aiyengar2 (approved):

PR #2367, opened 2023-01-27
added additional permissions to cluster roles used by gatekeeper chart jobs

vardhaman22:
this is needed so that post install and post upgrade jobs can add psa related labels to namespace in which the chart is installed.
related issue: https://github.com/rancher/rancher/issues/40108


charts/rancher-gatekeeper-crd/102.0.0+up3.10.0/templates/jobs.yaml
```diff
@@ -40,6 +40,13 @@ spec:
             - name: crd-manifest
               readOnly: true
               mountPath: /etc/config
+          securityContext:
+            seccompProfile:
```
aiyengar2, line 45: shouldn't all of these be gated by `.Values.enableRuntimeDefaultSeccompProfile`?

charts/rancher-gatekeeper/102.0.0+up3.10.0/templates/namespace-post-upgrade.yaml
```diff
@@ -95,6 +95,12 @@ rules:
       {{- range .Values.postUpgrade.labelNamespace.extraNamespaces }}
       - {{ . }}
       {{- end }}
+  - apiGroups:
+      - management.cattle.io
+    resources:
+      - projects
+    verbs:
+      - updatepsa
```
aiyengar2, line 103: tied to broader issue with this PR referenced in https://github.com/rancher/rancher/issues/40108#issuecomment-1419569066

aiyengar2 (changes requested):

charts/rancher-gatekeeper/102.0.0+up3.10.0/templates/namespace-post-upgrade.yaml
aiyengar2: addressed in https://github.com/rancher/rancher/issues/40108#issuecomment-1419597434. No more concerns with the overall approach of this PR.

charts/rancher-gatekeeper-crd/102.0.0+up3.10.0/templates/jobs.yaml
```diff
@@ -40,6 +40,13 @@ spec:
...
+                type: RuntimeDefault
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+              - ALL
```
aiyengar2: For these values outside of the seccomp profile stuff, should we just allow users to provide it as a general `.Values.securityContext` or similar field?

charts/rancher-gatekeeper-crd/102.0.0+up3.10.0/templates/jobs.yaml
aiyengar2: I think a broader issue here might also be that our UI does not provide the capability to update values for a CRD chart on install. We might need to open a UI ticket for this.

But in general, if you were to deploy this on a cluster whose nodes do not use `seccomp`, wouldn't you run into issues? Has this been tested?

aiyengar2 (changes requested):

prachidamle:
@vardhaman22 Can we run these tests and post results for them https://github.com/rancher/rancher/issues/40108#issuecomment-1420051736

Also, what namespace does the `updatepsa` happen for?

charts/rancher-gatekeeper-crd/102.0.0+up3.10.0/templates/jobs.yaml
vardhaman22: updated the values for crd chart as per your suggestion but as you mentioned UI related work will be needed so that user can modify these values for CRD chart.

charts/rancher-gatekeeper-crd/102.0.0+up3.10.0/templates/jobs.yaml
vardhaman22: done

aiyengar2 (approved):

vardhaman22:
> @vardhaman22 Can we run these tests and post results for them [rancher/rancher#40108 (comment)](https://github.com/rancher/rancher/issues/40108#issuecomment-1420051736)
> 
> Also, what namespace does the `updatepsa` happen for?

update namespace labels happens for the namespace in which the chart is installed. 
@doflamingo721 has updated the test results on the [issue](https://github.com/rancher/rancher/issues/40108).

charts/rancher-gatekeeper-crd/102.0.0+up3.10.0/templates/jobs.yaml
pjbgf: For context, k8s 1.25+ has the seccomp `RuntimeDefault` profile set as default to all pods. So this may not be needed in the feature - although it would make linters happy.

> But in general, if you were to deploy this on a cluster whose nodes do not use seccomp, wouldn't you run into issues?

AFAIK, seccomp would fail open - if it is not supported (e.g. Windows nodes) it will just be ignored.

PR #2380, opened 2023-01-30
[rancher-cis-benchmark] Update chart for new PSP approach

stormqueen1990:
### Issue: rancher/rancher#40345
---
### Changes
* Include validation to check if the PSP API is present in the target cluster.
* Standardize the PSP switch.
* Bump the chart version to 4.0.0-rc1.
* Remove the 3.0.1-rc8 version of the chart.
---
### Testing
- [x] Ensure a `helm template <release-name> <chart-path>` never fails
- [x] Ensure a `helm template --set global.cattle.psp.enabled=true <release-name> <chart-path>` never fails

#### Kubernetes < v1.25 specific tests

- [x] Ensure a `helm install --dry-run <release-name> <chart-path>` does not fail when your `KUBECONFIG` is pointing to a 1.24 cluster
- [x] Ensure a `helm install --dry-run --set global.cattle.psp.enabled=true <release-name> <chart-path>` does not fail when your `KUBECONFIG` is pointing to a 1.24 cluster
- [x] Ensure a `helm install <release-name> <chart-path>` does not fail when your `KUBECONFIG` is pointing to a 1.24 cluster
- [x] Ensure a `helm install --set global.cattle.psp.enabled=true <release-name> <chart-path>` does not fail when your `KUBECONFIG` is pointing to a 1.24 cluster

#### Kubernetes >= v1.25 specific tests

- [x] Ensure a `helm install --dry-run <release-name> <chart-path>` does not fail when your `KUBECONFIG` is pointing to a 1.25 cluster
- [x] Ensure a `helm install --dry-run --set global.cattle.psp.enabled=true <release-name> <chart-path>` **does fail** when your `KUBECONFIG` is pointing to a 1.25 cluster
- [x] Ensure a `helm install <release-name> <chart-path>` does not fail when your `KUBECONFIG` is pointing to a 1.25 cluster
- [x] Ensure a `helm install --set global.cattle.psp.enabled=true <release-name> <chart-path>` **does fail** when your `KUBECONFIG` is pointing to a 1.25 cluster, with the error from `validate-psp-install.yaml` showing up


packages/rancher-cis-benchmark/charts/Chart.yaml
```diff
@@ -2,7 +2,7 @@ annotations:
   catalog.cattle.io/auto-install: rancher-cis-benchmark-crd=match
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/display-name: CIS Benchmark
-  catalog.cattle.io/kube-version: '>= 1.21.0-0 < 1.25.0-0'
+  catalog.cattle.io/kube-version: '>= 1.21.0-0 < 1.26.0-0'
```
deniseschannon, line 5: Is there a reason why we are hardcoding to not work on 1.26? It seems like an extra hassle for us to maintain? Note: I didn't check this across all the charts, but wanted to ask if there was a specific reason. I noticed it in the SUSE projects, so was curious.

stormqueen1990: No specific reason! I just didn't check if this chart has any deprecations that need to be addressed for Kubernetes v1.26. More around erring on the side of caution.

release.yaml
```diff
@@ -30,8 +30,10 @@ rancher-backup-crd:
 - 102.0.0+up3.1.0-rc1
 rancher-cis-benchmark:
 - 3.0.1-rc6
+- 3.1.0-rc1
```
aiyengar2: since the last released version is `3.0.0` (ref: https://github.com/rancher/charts/tree/release-v2.7/charts/rancher-cis-benchmark), following our pattern of bumping the major version for all charts with PSP changes we should be releasing `4.0.0` with this release.

Also `3.0.1-rc6` should have been `make remove`d as well; can we remove that chart in this PR as well from the `release.yaml` and `charts`/`assets`/`index.yaml`?

packages/rancher-cis-benchmark/charts/Chart.yaml
aiyengar2: @deniseschannon generally we set the Rancher version annotation to versions that QA has explicitly tested during a release, so charts should always be setting an upper bound here as a best practice. 

Otherwise, if a chart wouldn't work in a new Kubernetes version, we'd have to go back and OOB release previous charts to no longer be deployable on the new cluster version to inform users.

release.yaml
stormqueen1990: Sure thing, will make the changes!

stormqueen1990: This is updated!

release.yaml
```diff
@@ -26,8 +30,10 @@ rancher-backup-crd:
 - 102.0.0+up3.1.0-rc1
 rancher-cis-benchmark:
 - 3.0.1-rc8
+- 4.0.0-rc1
 rancher-cis-benchmark-crd:
 - 3.0.1-rc8
```
prachidamle: release.yaml should not have the 3.0.1-rc8 version, will we be removing this in a separate commit?

stormqueen1990: I was kind of curious if we should have `3.0.1-rc8` regardless, since it contains the fix to the RBAC issue that prevented upgrades.

stormqueen1990: `make remove`d as requested! ✔️

prachidamle (dismissed):
LGTM

prachidamle (approved):

PR #2385, opened 2023-02-02
Dev v2.7: Update rancher-charts/longhorn chart to support updated approach for PSPs

PhanLe1010:
https://github.com/rancher/rancher/issues/40342


PhanLe1010:
Test plan is at https://github.com/rancher/rancher/issues/40342#issuecomment-1413094823

packages/longhorn/longhorn-1.4/package.yaml
```diff
@@ -1,4 +1,4 @@
 url: https://github.com/longhorn/charts.git
 subdirectory: charts/longhorn
 commit: 82be2970c8ffb8b50d6fcad77a3326d8e2434012
-version: 101.2.0
+version: 102.2.0
```
pennyscissors, line 4: This should be 102.0.0

packages/longhorn/longhorn-1.4/generated-changes/patch/questions.yaml.patch
```diff
@@ -117,3 +117,23 @@
...
+ - variable: privateRegistry.createSecret
+-  default: "true"
++  default: "false"
+   description: "Create a new private registry secret"
+   type: boolean
+   group: "Private Registry Settings"
+@@ -824,8 +824,8 @@
+     show_if: "service.ui.type=NodePort||service.ui.type=LoadBalancer"
+     label: UI Service NodePort number
+ - variable: enablePSP
```
pennyscissors, line 132: Is Longhorn going to stick to using `enablePSP` rather than `global.cattle.psp.enabled` like our other charts? Considering we are bumping all charts to 102.x.x to communicate the 1.25+ and PSPs breaking change, this would be the perfect time to change the name of this variable to be consistent across our charts. However, I'm not familiar with the implications of changing this variable and any impact it may have on Longhorn upgrades and other scenarios so please take this comment as a suggestion and not a change request.

packages/longhorn/longhorn-1.4/package.yaml
PhanLe1010: We will have 3 release lines as:
* 102.0.0 for Longhorn 1.2.x
* 102.1.0 for Longhorn 1.3.x
* 102.2.0 for Longhorn 1.4.x (this chart PR)

So, this one should be `102.2.0`. What do you think? @PennyScissors

packages/longhorn/longhorn-1.4/generated-changes/patch/questions.yaml.patch
PhanLe1010: `enablePSP` exists in Longhorn upstream chart. 
* In that upstream chart, there is no concept Rancher `cattle` system so it makes less sense to rename the variable to `global.cattle.psp.enabled` in upstream chart
* The other approach is to keep the upstream the same but modify the variable name when building the rancher charts. This has the disadvantage that the upstream and rancher chart will be diverted. Additional work to verify upgrade will also needed as you mentioned.

packages/longhorn/longhorn-1.4/package.yaml
innobead: @PhanLe1010 1.2 is EOL. I think we can remove it from the release.

cc @rebeccazzzz

packages/longhorn/longhorn-1.4/generated-changes/patch/questions.yaml.patch
innobead: Agreed with Phan, suggest just keeping the original setting as is because longhorn chart has been released. Need to keep the compatibility if could. Also, don't feel any meaningful change required except for the convention or any other reason beyond this?

BTW, we do have global.cattle in values.yaml but they are used for Rancher specific behavior like mixed cluster. Enabling PSP is general not specific for Rancher.

packages/longhorn/longhorn-1.4/package.yaml
pennyscissors: Right, I forgot about Longhorn's multiple release lines. It makes sense now.

packages/longhorn/longhorn-1.4/generated-changes/patch/questions.yaml.patch
pennyscissors: If the cattle concept in Longhorn is different than the cattle concept in the rest of our charts, and the variable name cannot be changed upstream because of this difference in concept, I guess it makes sense to stick to the original variable name. The benefit would be convention and better user experience since all charts would have PSP enablement done in the same way, but I would rather stick to the original name than have the Longhorn chart here deviate from uptream as that will add additional work when bringing the Longhorn chart from upstream into the rancher/charts repository. Thanks for the detailed clarification!

packages/longhorn/longhorn-1.4/package.yaml
PhanLe1010: @innobead I see. Is it ok to remove it in the next major release bump (to 103.x.x) just in case we need to add a hot fix in Longhorn 1.2.x? If not, I will remove the 1.2.x release line and make the versions:
* 102.0.0 for Longhorn 1.3.x
* 102.1.0 for Longhorn 1.4.x

innobead: Yeah, we don't really have a clear process about when to discontinue or remove the Rancher chart release if it reaches EOL. Probably we can do this when there is a major version (Rancher chart) bump. 

Let's start from 103 and keep 1.2.x as is first to have some buffer for any exception. However, we should not have any hotfix for 1.2.x any more. 

cc @rebeccazzzz to understand this new policy. 

@PhanLe1010 Could you help document this on confluence to mention we will review if remove the old release chart when a new major version bump of Rancher chart? Thanks.

PhanLe1010: @innobead Updated the doc https://confluence.suse.com/display/LON/How+to+create+Rancher+Chart+PRs
Thank you

innobead (dismissed):
LGTM

weizhe0422 (dismissed):
LGTM. I also installed Longhorn v1.4.0 on Rancher v2.7.1 using the repository and did a few things to make sure it worked, including creating volumes and creating snapshots/backup checks. Thank you.

截圖 2023-02-07 下午7 26 19

release.yaml
```diff
@@ -86,5 +86,7 @@ ui-plugin-operator-crd:
 - 102.0.0+up0.2.0-rc1
 longhorn:
   - 101.2.0+up1.4.0
```
prachidamle: Shouldn't we be removing 101.2.0+up1.4.0 version for longhorn and longhorn-crd, since we are adding 102.2.0+up1.4.0? cc @PennyScissors

PhanLe1010: Yes, we should now. At the time that I created the PR, 101.2.0+up1.4.0 wasn't in the release branch yet so I have to add it here otherwise the PR verification would fail. Thank you for the reminder

PhanLe1010: I fixed it. Thanks @prachidamle

innobead: I thought 101 was released already, wasn't it?

PhanLe1010: Yes, 101.2.0+up1.4.0 was released already now. So I fixed the PR to remove it. 

At the time that I created the PR, 101.2.0+up1.4.0 was not release so I had to add it here to pass the PR validation 


Context: 
* If the chart has not been released to `release-xxx` branch, the PR needs to include the chart in `release.yaml` to indicate that the chart is being added. 
* If the chart has been released to `release-xxx` branch, the PR needs NOT to include the chart in `release.yaml`

prachidamle (approved):

prachidamle:
Merging based on validation passed by Longhorn QA here https://github.com/rancher/rancher/issues/40342#issuecomment-1434424396

PR #2389, opened 2023-02-03
[rancher-cis-benchmark] Remove unreleased cis-benchmark version 3.0.1-rc7

doflamingo721:
Depends on: https://github.com/rancher/charts/pull/2380
The newer cis-benchmark version 3.1.0-rc1 has the latest PSP related changes. This 3.0.1-rc7 version in unreleased and hence needs to be deleted.


prachidamle:
This can be closed since the version is removed with another PR.

PR #2394, opened 2023-02-07
fix cis rbac

vardhaman22:
related to issue: https://github.com/rancher/rancher/issues/40405

removed the delete clusterrolebinding hook. Which was not necessary since now the clusterrolebinding is renamed and older ones will be deleted by helm automatically in case of upgrade. 
earlier there was a conflict because roleref of the same clusterrolebinding was updated, now since it it renamed update scenario will not be there i.e. new binding resource will be created and older one will be removed by helm, and hence the delete rolebinding hook is no longer needed.


stormqueen1990 (commented):
LGTM!

galal-hussein (approved):

prachidamle (approved):

PR #2395, opened 2023-02-07
Add NeuVector chart version 2.4.2

selvamt94:


prachidamle:
The existing unreleased `101.0.2+up2.4.0` version for neuvector and neuvector-crd chart should be removed using `make remove`

release.yaml
```diff
@@ -16,8 +16,10 @@ fleet-crd:
 - 101.1.0+up0.6.0-rc.2
 neuvector:
 - 101.0.2+up2.4.0
+- 102.0.0+up2.4.2
```
prachidamle, line 12: We should remove `101.0.2+up2.4.0` since we are increasing the major version to 102.x.y

cbron: agree, we only release 1 app version per rancher version

cbron (commented):
Looks good besides the single version. Appears all references to `values.psp` were updated.

release.yaml
pennyscissors: To remove a chart you need to run `CHART=<chart-name> VERSION=<chart-version> make remove`, remove the entry from release.yaml, and commit everything that was modified by the command. Then do the same steps to remove the crd chart in a separate commit.

selvamt94: After make remove, do I need to do rebase as per this link. https://github.com/rancher/charts/blob/dev-v2.7/docs/developing.md#rebasing-an-existing-package

If so what version should I specify for the below line. I will contact you tomorrow via slack to take your help to fix this. 

3. Modify the package.yaml to point to your new upstream.

selvamt94: @prachidamle @PennyScissors Removed the unreleased version 101.0.2+up2.4.0 and resubmitted the PR.

cbron (dismissed):

release.yaml
```diff
@@ -10,8 +10,10 @@ fleet-crd:
...
 neuvector-crd:
 - 101.0.2+up2.4.0
+- 102.0.0+up2.4.2
```
prachidamle, line 14: We need another commit to remove 101.0.2+up2.4.0 from neuvector and neuvector-crd chart

release.yaml
prachidamle: @selvamt94 101.0.2+up2.4.0 should be removed from release.yaml file too

release.yaml
selvamt94: Removed the unreleased version from release.yaml

prachidamle (approved):

PR #2418, opened 2023-02-14
Bump vSphere CSI/CPI major version to indicate k8s 1.25 support

a-blender:
Issue: https://github.com/rancher/rancher/issues/40464


packages/rancher-vsphere/rancher-vsphere-cpi/package.yaml
```diff
@@ -1,4 +1,4 @@
 url: https://github.com/rancher/vsphere-charts.git
 subdirectory: charts/rancher-vsphere-cpi
 commit: 96687762c6642efc3650626a0738c8caa45ae7c3
-version: 101.1.0
```
HarrisonWAffel, line 4: I think if we're bumping a major version we would want to reset the minor version, right? So it would be `102.0.0`

Other reviewers can correct me if I'm wrong, but that was my understanding of the process

HarrisonWAffel (changes requested):
Sorry to be a pain, but did you run `make remove`? 

It's helpful to have three commits when doing these changes, `make remove`, `change version`, `make charts`. Otherwise its a bit hard to determine whats been changed.

jiaqiluo (changes requested):
Can you split the 1st commit into two: one for the cleanup and another for the change? 
This is the preferred pattern.

packages/rancher-vsphere/rancher-vsphere-cpi/package.yaml
a-blender: Thanks! I'll use `make remove` from now on, that seems a lot faster. I ran these

```
make remove CHART=rancher-vsphere-csi VERSION=101.1.0+up2.6.2-rancher1
make remove CHART=rancher-vsphere-cpi VERSION=101.1.0+up1.4.1
```

HarrisonWAffel (approved):

jiaqiluo (approved):

PR #2419, opened 2023-02-15
Update sriov chart to 102.0.0

thomasferrandiz:
Issue: https://github.com/rancher/rancher/issues/40464


release.yaml
```diff
@@ -73,9 +73,11 @@ rancher-webhook:
 sriov:
 - 101.0.2+up0.1.0
 - 101.0.3+up0.1.0
+- 102.0.0+up0.1.0
```
prachidamle, line 74: Do we intend to support multiple versions of this chart in Rancher v2.7.2? If not, we should only have the 102.x.y chart in the release.yaml and remove the earlier versions using `make remove`

thomasferrandiz: I don't think multiple versions will be needed, I removed them.

snasovich (approved):

snasovich:
@prachidamle , thank you for reviewing this. Could you please take another look if it's good to merge?

a-blender (approved):

stormqueen1990 (approved):

PR #2429, opened 2023-02-21
[dev-2.7] fix double cpu and memory metrics in workload dashboards

konih:
Fix for issue rancher/rancher#43468

Cadvisor exports cpu and memory metrics once for each container as well as the parent cgroup for each pod. The faulty queries sum the parents metrics with the individual container metrics resulting in a doubling of the memory and cpu usage.


nicholasSUSE:
Closed due to inactivity

PR #2431, opened 2023-02-22
Fixing issue 39972 Neuvector chart

selvamt94:


samjustus:
[#39972](https://github.com/rancher/rancher/issues/39972)

release.yaml
```diff
@@ -2,3 +2,5 @@ longhorn:
   - 101.2.0+up1.4.0
 longhorn-crd:
   - 101.2.0+up1.4.0
+neuvector:
```
aiyengar2, line 1: if we’re doing this as an OOB release to modify an existing chart (which makes sense given the issue, but cc: @cbron in case that wasn’t the intention?), the release.yaml should only contain the one chart that is intended to be OOB released. all other charts should be removed in the first commit of this PR

aiyengar2 (commented):
Do we need this as an OOB release straight to release-v2.7, or can we just put this in dev-v2.7 and allow this to be released in the next 2.7 release?

release.yaml
cbron: I thought this had to go straight to release branch because the file is autogenerated ? Or if you change old asset it would auto-generate correctly without the dot. But then you would have 2 different NV charts going out in the next release.

pennyscissors (changes requested):
This PR is skipping the current process we have in place for OOB chart releases. You need to create an issue in rancher/rancher to track the OOB release (include the chart version that needs to be released and any relevant issues), then make this PR to dev-v2.7 and link the OOB request issue. Once it is merged someone in Team 3 will take care of releasing the chart OOB and notifying the Neuvector team so they can close the OOB request issue. Alternatively, if it doesn't need to be released OOB, you can put it in dev-v2.7 and it will go out with 2.7.2 in April.

pennyscissors (approved):
lgtm, just needs the release.yaml updated to contain neuvector charts only and it should be good to release. I discussed this with @prachidamle offline and we decided it would be okay to make an exception for this fix as it is a minor change to an annotation and the functionality of the chart itself remain unchanged.

release.yaml
prachidamle: @selvamt94 Can you add this fix to release.yaml file?

prachidamle (approved):
lgtm

pennyscissors (approved):

stormqueen1990 (approved):

PR #2439, opened 2023-02-23
Support additional matches to be federated from cluster Prometheus

tamalsaha:


wrkode (approved):

tamalsaha:
To add some context to this PR, we have a project that exposes Prometheus metrics for our CRDs to the cluster scoped Prometheus. We want to federate those metrics into the rancher project specific Prometheus. To do that, we need to add `matches` to the federated scrape config. This pr gives us a way to do that via the Helm chart values.

bashofmann:
Unfortunately, this PR won't work like this.

The charts directory in this repo is auto-generated by rancher/charts-build-scripts . See https://github.com/rancher/charts/blob/dev-v2.7/docs/developing.md.

You also can’t just change something in an existing chart version, but you have to create a new one.

The source for the rancher-project-monitoring chart in this repo is here https://github.com/rancher/charts/tree/dev-v2.7/packages/rancher-monitoring/rancher-project-monitoring. And this references https://github.com/rancher/prometheus-federator, which also uses rancher/charts-build-script.

So the process to add something to this chart would be:
* Add this to https://github.com/rancher/prometheus-federator/tree/main/packages/rancher-project-monitoring, in this case https://github.com/rancher/prometheus-federator/blob/main/packages/rancher-project-monitoring/generated-changes/overlay/files/federate/federate-scrape-config.yaml and https://github.com/rancher/prometheus-federator/blob/main/packages/rancher-project-monitoring/generated-changes/patch/values.yaml.patch (make sure to read up on how the charts-build-script works to make editing this patches easier (in the end `make prepare`, edit generated `values.yaml`, `make patch`, `make clean`)
* Bump the version in https://github.com/rancher/prometheus-federator/blob/main/packages/rancher-project-monitoring/package.yaml
* Run `make charts`
* Create a PR
* Once this is merged, update https://github.com/rancher/charts/blob/dev-v2.7/packages/rancher-monitoring/rancher-project-monitoring/package.yaml
* Run `make charts` there
* Create PR

tamalsaha:
Thanks, @bashofmann !

@wrkode, the pr process is going well beyond my expertise. I wonder if you could work on your side to get this patch merged following necessary steps.

bashofmann:
A colleague and I looked further into this. Unfortunately, we won't be able to add it like this. Adding this setting would create a security vulnerability. Project owners could add federation matches that federate metrics for workloads that are in other projects. This would break the project isolation which project monitoring is meant to provide.

tamalsaha:
@bashofmann I understand what you are saying. May be the way to allow for more federated jobs is to have the cluster scoped prometheus chart to define what other jobs can be federated. This was project owners won't be able to bypass project boundary.

What do you think?

bashofmann:
That could be an option, but probably a larger thing. Can you work together with your consultant/support to raise an internal feature request for this?

tamalsaha:
any updates here, folks?

recena:
Closing due to lack of activity.

PR #2449, opened 2023-02-28
Add release validation scripts

pennyscissors:


scripts/release-validation/diff-with-released.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
+ROOT_DIR=../..
```
stormqueen1990, line 3: Do we want this to be something like:
```
SCRIPT_DIR="$(cd "$(dirname "$0")"; pwd)"
ROOT_DIR="${SCRIPT_DIR}/../.."
```
so that it can be ran from anywhere, instead of requiring the user to navigate to the scripts directory?

scripts/release-validation/validate-index.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# Check if every URL in index.yaml has a corresponding assets/ entry
+for url in $(echo $urls | xargs); do
+  if printf '%s\n' "${exclude[@]}" | grep -F -x ${url} 1>/dev/null; then
+    echo "skipping ${url}"
+    continue
+  fi
+  if [[ -f ${url} ]]; then
+    echo "found ${url}"
+    continue
```
stormqueen1990: This continue is superfluous, since this `if` statement has a corresponding `else` and no other statements follow.

stormqueen1990:
Something I just remembered: it would be nice to have an accompanying `README.md` on how to use these scripts.

stormqueen1990 (approved):

PR #2450, opened 2023-03-01
Bump harvester-csi-driver to v0.1.16 and harvester-cloud-provider to v0.1.14

Vicente-Cheng:
Issue: https://github.com/harvester/harvester/issues/3571

This PR includes the two version Bump
1. Bump harvester-csi-driver to v0.1.16
2. Bump harvester-cloud-provider to v0.1.14


guangbochen (dismissed):
LGTM, thanks.

jiaqiluo (dismissed):

charts/harvester-cloud-provider/102.0.0+up0.1.14/Chart.yaml
```diff
@@ -0,0 +1,21 @@
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/display-name: Harvester Cloud Provider
+  catalog.cattle.io/kube-version: '>= 1.18.0-0 < 1.26.0-0'
+  catalog.cattle.io/namespace: kube-system
+  catalog.cattle.io/os: linux
+  catalog.cattle.io/permits-os: linux
+  catalog.cattle.io/rancher-version: '>= 2.6.1-0 < 2.8.0-0'
```
snasovich: This range looks "more correct" in previous released version of charts - https://github.com/rancher/charts/blob/425972dfa60be22c850bdfe77fcd2eda0fd5d08c/charts/harvester-cloud-provider/101.0.0%2Bup0.1.13/Chart.yaml#L8.
Any specific reason we're extending range to include 2.6.1+ here?

This seems to be due to removal of `*.patch` files and has other effects like `permits-os` annotation no longer including `windows`.

Is this intentional?

snasovich (changes requested):

charts/harvester-cloud-provider/102.0.0+up0.1.14/Chart.yaml
guangbochen: @snasovich For the Harvester CSI and CCM chart, we are using exactly the same version for both Rancher v2.6 and v2.7 releases, and those charts are not used by the `Windows` so it should be fine to be removed. thanks.

snasovich: @guangbochen , I'm not _the_ expert on charts in Rancher, but I believe we generally want to maintain separate major versions for each Rancher minor versions (100.x.y for 2.6 line, 101.x.y for 2.7 line - and actually 102.x.y for 2.7.2+ to designate support for k8s 1.25 and handling of PSP removal).

So it will be best to retain `'>= 2.7.0-0 < 2.8.0-0'` range here. @aiyengar2 @PennyScissors , please chime in as _the_ charts experts.

Also, we're now missing `catalog.cattle.io/upstream-version` annotation that was previously added via `.patch` file.

guangbochen: if that's preferred we should be able to patch those values accordingly via the `*.patch` files.  cc @Vicente-Cheng WDUT?

Vicente-Cheng: Updated! Keep the `*.patch` files.
@snasovich, could you help to review it again? Thanks!

packages/harvester/harvester-cloud-provider/generated-changes/patch/Chart.yaml.patch
```diff
@@ -4,17 +4,17 @@
  annotations:
    catalog.cattle.io/certified: rancher
    catalog.cattle.io/display-name: Harvester Cloud Provider
--  catalog.cattle.io/kube-version: '>= 1.18.0-0 < 1.24.0-0'
+-  catalog.cattle.io/kube-version: '>= 1.18.0-0 < 1.26.0-0'
 +  catalog.cattle.io/kube-version: '>= 1.18.0-0 < 1.25.0-0'
```
jiaqiluo: Is this change expected?   
Rancher v2.7.2 will support k8s 1.25. We should drop this change if we want to show this chart in the downstream 1.25 cluster. 
I guess you might forget to drop it when adding this file back 😄 

(same for another patch file)

packages/harvester/harvester-cloud-provider/package.yaml
```diff
@@ -1,2 +1,2 @@
-url: https://github.com/harvester/charts/releases/download/harvester-cloud-provider-0.1.13/harvester-cloud-provider-0.1.13.tgz
-version: 101.0.0
\ No newline at end of file
+url: https://github.com/harvester/charts/releases/download/harvester-cloud-provider-0.1.14/harvester-cloud-provider-0.1.14.tgz
+version: 102.0.0
```
jiaqiluo, line 2: Sergey mentioned in his comment that 102.x.y is for 2.7.2+ to designate support for k8s 1.25 and handling of PSP removal. 
Can you confirm whether the new chart is for such purposes or contains such changes?

jiaqiluo (changes requested):

packages/harvester/harvester-cloud-provider/generated-changes/patch/Chart.yaml.patch
Vicente-Cheng: Thanks for the reminder! I would fix it.

packages/harvester/harvester-cloud-provider/package.yaml
Vicente-Cheng: Hi @jiaqiluo, I change the version to 101.0.1. Please help to review it again. Thanks!!

packages/harvester/harvester-csi-driver/generated-changes/patch/Chart.yaml.patch
```diff
@@ -1,20 +1,14 @@
...
-   catalog.cattle.io/display-name: Harvester CSI Driver
--  catalog.cattle.io/kube-version: '>= 1.18.0-0 < 1.24.0-0'
-+  catalog.cattle.io/kube-version: '>= 1.18.0-0 < 1.25.0-0'
+@@ -5,9 +5,10 @@
    catalog.cattle.io/namespace: kube-system
    catalog.cattle.io/os: linux
--  catalog.cattle.io/permits-os: linux
-+  catalog.cattle.io/permits-os: linux,windows
--  catalog.cattle.io/rancher-version: '>= 2.6.1-0 < 2.7.0-0'
+   catalog.cattle.io/permits-os: linux
```
jiaqiluo, line 6: `windows` was added to the label previously, but now is dropped. Is this an intended change? 
So the chart can not be installed in Windows anymore? 
(same for the chart harvester-cloud-driver)

jiaqiluo (changes requested):
I have one last question otherwise looks good to me.

packages/harvester/harvester-csi-driver/generated-changes/patch/Chart.yaml.patch
Vicente-Cheng: Hi @jiaqiluo, follow the above comment https://github.com/rancher/charts/pull/2450#discussion_r1133391434, so I remove `windows` from the permits-os.

Hi @guangbochen, do we need to support windows now? Thanks!

guangbochen: Yep, Windows node support is not in our scope yet, so we should be fine to remove it, thanks.

jiaqiluo (dismissed):

a-blender (commented):
Please correct me if I'm wrong, but if these are updates going into dev-v2.7 for the next release (2.7.2+) then shouldn't the harvester-csi-driver chart be updated to `120.0.0` and the `kube-version` be updated to `< 1.26.0-0` to support k8s 1.25? Is this chart compatible with 1.25 yet? I could've missed something here because I'm not a harvester expert but users who provision harvester clusters with k8s 1.25 will not have an option to install this chart with the way it is now

charts/harvester-cloud-provider/102.0.0+up0.1.14/Chart.yaml
pennyscissors: @snasovich Yes, that's correct. The range for these charts should be `'>= 2.7.0-0 < 2.8.0-0'`. Additionally, the chart needs to be updated to support PSPs and k8s 1.25, and the version updated to `102.x.y` to indicate that. Assuming the chart supports k8s 1.25, the kube-version annotation is correct (`catalog.cattle.io/kube-version: '>= 1.18.0-0 < 1.26.0-0'`), but if these charts use PSPs they need these changes https://github.com/rancher/rancher/issues/40331

Vicente-Cheng:
> Please correct me if I'm wrong, but if these are updates going into dev-v2.7 for the next release (2.7.2+) then shouldn't the harvester-csi-driver chart be updated to `120.0.0` and the `kube-version` be updated to `< 1.26.0-0` to support k8s 1.25? Is this chart compatible with 1.25 yet? I could've missed something here because I'm not a harvester expert but users who provision harvester clusters with k8s 1.25 will not have an option to install this chart with the way it is now

Hi @a-blender, thanks for your review.
About the version question, I am not sure about it. (Also, did you mean `102.0`)?
If yes, @jiaqiluo remind me on this link https://github.com/rancher/charts/pull/2450#discussion_r1135942884. For harvester-csi-driver we do not have specific handling for the PSP removal, so I change back to `101.0.x`.
Did I get some misunderstanding here?  cc @jiaqiluo @guangbochen 

For the `kube-version`, I remove the patch yaml for patching this parameter. For now the `kube-version` is `'>= 1.18.0-0 < 1.26.0-0'`.
Would it be Ok for the `kube-version`?

a-blender:
@Vicente-Cheng Yes, I meant `102.0` for that the harvester csi charts version

a-blender:
@jiaqiluo Still waiting on comments to be addressed.

snasovich:
> > Please correct me if I'm wrong, but if these are updates going into dev-v2.7 for the next release (2.7.2+) then shouldn't the harvester-csi-driver chart be updated to `120.0.0` and the `kube-version` be updated to `< 1.26.0-0` to support k8s 1.25? Is this chart compatible with 1.25 yet? I could've missed something here because I'm not a harvester expert but users who provision harvester clusters with k8s 1.25 will not have an option to install this chart with the way it is now
> 
> Hi @a-blender, thanks for your review. About the version question, I am not sure about it. (Also, did you mean `102.0`)? If yes, @jiaqiluo remind me on this link [#2450 (comment)](https://github.com/rancher/charts/pull/2450#discussion_r1135942884). For harvester-csi-driver we do not have specific handling for the PSP removal, so I change back to `101.0.x`. Did I get some misunderstanding here? cc @jiaqiluo @guangbochen
> 
> For the `kube-version`, I remove the patch yaml for patching this parameter. For now the `kube-version` is `'>= 1.18.0-0 < 1.26.0-0'`. Would it be Ok for the `kube-version`?

@Vicente-Cheng , if this chart declares support for k8s 1.25 support, we do want its version bumped to `102.0.0`. You state "For harvester-csi-driver we do not have specific handling for the PSP removal", but I'm not sure there is anything specific to handle here since I don't see this chart deploying PSPs. If there is nothing in `rancher/harvester-cloud-provider` image itself that deploys PSPs we should be good.

To sum up, if we're good from 1.25 support perspective here, the version should be changed to 102.0.0.

Vicente-Cheng:
Hi @snasovich, Thanks for your explanation.
I already update the version to `v102.0.0`. Please help to review it again.
Thanks! cc @guangbochen

a-blender (dismissed):

packages/harvester/harvester-cloud-provider/generated-changes/patch/Chart.yaml.patch
```diff
@@ -1,20 +1,14 @@
...
+@@ -5,9 +5,10 @@
    catalog.cattle.io/namespace: kube-system
    catalog.cattle.io/os: linux
--  catalog.cattle.io/permits-os: linux
-+  catalog.cattle.io/permits-os: linux,windows
--  catalog.cattle.io/rancher-version: '>= 2.6.1-0 < 2.7.0-0'
-+  catalog.cattle.io/rancher-version: '>= 2.7.0-0 < 2.8.0-0'
+   catalog.cattle.io/permits-os: linux
+-  catalog.cattle.io/rancher-version: '>= 2.6.1-0 < 2.8.0-0'
++  catalog.cattle.io/rancher-version: '>= 2.7-0 < 2.8.0-0'
```
snasovich: Just noticed version constraint has a typo here.
```suggestion
+  catalog.cattle.io/rancher-version: '>= 2.7.0-0 < 2.8.0-0'
```

snasovich (changes requested):
One comment to address, otherwise LGTM. Once it's fixed we can merge.

packages/harvester/harvester-cloud-provider/generated-changes/patch/Chart.yaml.patch
Vicente-Cheng: Hi @snasovich, nice catch. Thanks for your review!
Already fixed it. Please help to review it again.

jiaqiluo (approved):

snasovich (approved):
LGTM, but we want to merge only when we're certain Harvester 1.1.2 will be released prior to 4/10 which is planned date for Rancher 2.7 charts release

PR #2461, opened 2023-03-06
[Dev V2.6] Un-rc Rancher Backups

eliyamlevy:
Un-rced rancher backups for release.


nflynt (approved):
Looks good to me. v2.1.5 only has repo maintenance, so there shouldn't be any behavioral differences as far as the chart is concerned. :+1:

PR #2482, opened 2023-03-14
Add validate-psp-install.yaml to CRD charts 

stormqueen1990:
## Problem

Due to a missing validation during the CRD chart installation, sometimes when installing a chart where the main, non-CRD chart contains PSPs but not the CRD chart the validation is deferred until the main chart is installed. This causes the CRD chart to be installed and at least in the case of logging, for tabs to appear in the UI that cannot be used because the main chart is not present.

## Solution

Skip CRD installation if the install options are incompatible with the cluster.

### Manual Testing

1. Add this branch as a custom repository to Rancher.
2. Install the Monitoring or Logging charts with the PSP option enabled.
3. Check that CRD fails to install.
4. Change the switch to false.
5. Check that both charts now install successfully.


stormqueen1990:
Moving to separate PRs per chart.

PR #2493, opened 2023-03-20
Rancher backups: ignore fleet.cattle.io/cluster-registration-values during restore

nflynt:
## Problem
When performing a migration restore of an RKE2 cluster, the cluster nodes can become permanently stuck in "Updating" status. We believe this is due to the presence of stale `fleet.cattle.io/cluster-registration-values` tokens in the backup contents, so this change ignores these tokens as a hotfix to work around the issue.

## Testing
Repro steps in https://github.com/rancher/rancher/issues/40819 and also https://github.com/rancher/rancher/issues/40080 are provided. I've not been able to fully reproduce the issue on my end, but I suspect I am not recreating the customer's setup correctly due to my own familiarity. Requesting QA to perform more thorough testing.

### Manual Testing
I've run through a fairly standard backup and restore, with a default RKE2 custom cluster and restore steps laid out here: https://ranchermanager.docs.rancher.com/v2.6/how-to-guides/new-user-guides/backup-restore-and-disaster-recovery/migrate-rancher-to-new-cluster

I can confirm that token restorations are skipped and appear in the restore logs. My RKE2 clusters have all come up successfully with these changes, and switch to "Active" after a brief delay. However as I've so far been unable to reproduce the original issue, I am unsure of whether this change is sufficient to fix it.

This change is made to the restore operation only. Backups will still contain the ignored tokens. The goal is to correct the issue in **existing** backups that would break when restored. Later, the plan is to improve resource filtering capabilities so that the fix can be specified more generally in the **chart**, and the hard-coded hotfix removed from the operator.


eliyamlevy (approved):
LGTM

PR #2495, opened 2023-03-21
[v 2.7] rancher-tracing chart fix

vardhaman22:
## Issue: 

related to issue: https://github.com/rancher/rancher/issues/40913

## Problem

service account for deployment in rancher-tracing chart was not getting updated to default after disabling psp.

## Solution

moved the serviceaccount out of the psp condition i.e. now service account will be created everytime even if psp is disabled.

## Testing

as mentioned in the issue: https://github.com/rancher/rancher/issues/40913

### Manual Testing

installed the chart with psp enabled and then upgraded the chart with psp disabled. upgrade was working as expected.

### Automated Testing

N/A

## QA Testing Considerations

both should be tested

N/A

## Backporting considerations

Not needed


stormqueen1990 (approved):
LGTM!

PR #2497, opened 2023-03-22
vSphere CPI chart security fix

a-blender:
## Solution

Pull in vSphere CPI chart security fix and other tested updates from https://github.com/rancher/vsphere-charts/pull/45. CPI `100.6.0+up1.4.2` will be released in an OOB charts release.

Specifically check for
- chart updated to version `1.4.2`
- remove user/password fields in configmap
- add `vCenter.labels.region` and `vCenter.labels.zone` in questions.yaml and labels `default: false` instead of true
## Backporting considerations

Needs to be forward-ported to 2.7 (for Rancher 2.7.2 release)


charts/rancher-vsphere-cpi/100.6.0+up1.4.2/templates/configmap.yaml
```diff
@@ -0,0 +1,32 @@
...
+      insecureFlag: {{ .insecureFlag }}
+
+    # vcenter section
+    vcenter:
+      {{ .host | quote }}:
+        server: {{ .host | quote }}
+        datacenters:
+          - {{ .datacenters | quote }}
+    {{- if .labels.generate }}
+
```
stormqueen1990, line 26: Super minor but perhaps you might want to move the empty line up to create a logical block for the `vcenter` section?

charts/rancher-vsphere-cpi/100.6.0+up1.4.2/Chart.yaml
```diff
@@ -0,0 +1,22 @@
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/display-name: vSphere CPI
+  catalog.cattle.io/kube-version: '>= 1.18.0-0 < 1.26.0-0'
```
stormqueen1990, line 4: This chart is aimed at 2.6. Should the `kube-version` range here target only `< 1.25.0-0` instead?

charts/rancher-vsphere-cpi/100.6.0+up1.4.2/Chart.yaml
```diff
@@ -0,0 +1,22 @@
...
+  catalog.cattle.io/namespace: kube-system
+  catalog.cattle.io/os: linux
+  catalog.cattle.io/permits-os: linux,windows
+  catalog.cattle.io/rancher-version: '>= 2.6.0-0 < 2.7.0-0'
```
stormqueen1990, line 8: Good call adding an upper bound! 😃

stormqueen1990 (approved):

stormqueen1990 (changes requested):
I had accidentally approved instead of commenting before.

charts/rancher-vsphere-cpi/100.6.0+up1.4.2/Chart.yaml
a-blender: Thank you for catching this. Looks like there have been 2 patch bumps in the history https://github.com/rancher/charts/commits/dev-v2.6/packages/rancher-vsphere/rancher-vsphere-cpi/generated-changes/patch/Chart.yaml.patch but the latest is incorrect. Looks like < 1.26 was pulled in from vsphere-charts. I'll manually update that to < 1.25.

charts/rancher-vsphere-cpi/100.6.0+up1.4.2/templates/configmap.yaml
a-blender: IMO I think we can leave this. The charts works correctly and `labels.generate` field does not show up in the rendered configmap because it's a conditional value

charts/rancher-vsphere-cpi/100.6.0+up1.4.2/Chart.yaml
snasovich: @stormqueen1990 @a-blender , this was raised in an internal Slack discussion.
TLDR is it's `< 1.26` in both CPI and CSI charts so if we fixing it for CPI for consistency we want it fixed for CSI as well.

However, it's _very_ minor as the worst that would happen it technically allows to install this chart for Rancher 2.6.x running on k8s 1.25 which is not a supported configuration and may not actually work (if someone manages to do that, being able to install this chart will be least of their worries).

So, all things considered I think it's best to leave it "as is".

charts/rancher-vsphere-cpi/100.6.0+up1.4.2/templates/configmap.yaml
snasovich: Valid point, should have handled it as part of https://github.com/rancher/vsphere-charts/pull/45/files :)
Let's leave this "as is".

stormqueen1990: Yup, this is a readability comment for future maintenance 🙂

jiaqiluo (approved):

stormqueen1990 (approved):

snasovich (approved):

PR #2499, opened 2023-03-22
Add priorityclasses for fleet in a list of resources to backup/restore

nflynt:
## Problem
Fleet controller deployment won't come up active with priority class not found error on rancher migration as the priority class resource is not added in the resource set of backup/restore operator.

## Solution
Add priorityclasses for fleet in a list of resources to backup/restore

## Testing
Repro steps from the github issue appear to still be valid.


eliyamlevy (approved):
LGTM

PR #2507, opened 2023-03-28
[release-v2.7] OOB Charts release for Longhorn v1.4.1

### LINKED ISSUE #40019: Feature Charts: Add Longhorn 1.4.1 Chart in 2.7.x
Longhorn has an upcoming release 1.4.1 which will be released in mid-2023 so we need to add this new release in the feature charts along with Rancher 2.7.x. We will probably need to make this OOB.

Add 1.4.1 LH chart in feature chart 2.7.x - Cluster Manager/Rancher Marketplace
Cc: @cchien816 @innobead @khushboo-rancher @MKlimuszka @prachidamle @yangchiu @yasker

Related issue for 2.6.x: https://github.com/rancher/rancher/issues/40018

rohitsakala:
Fixes https://github.com/rancher/rancher/issues/40019


rohitsakala:
## Charts Checklist (built for v0.3.x charts-build-scripts)

### Checkpoint 0: Validate `release.yaml`

Validation steps:
- [x] Each chart version in `release.yaml` DOES NOT modify an already released chart. If so, stop and modify the versions so that it releases a net-new chart.
- [x] Each chart version in `release.yaml` IS exactly 1 more patch version than the last released chart version. If not, stop and modify the versions so that it releases a net-new chart.

Validation steps:
- [x] Running `make unzip` to regenerate the `charts/` from scratch, then `git diff` to check differences between `assets/` and `charts/` yields NO differences or innocuous differences.

Validation steps:
- [x] The `index.yaml` file has an entry for each chart version.
  - kube-version annotation
  - rancher-version annotation

### Checkpoint 3: Logical Checks

Validation steps (requires manual inspection of differences between previous and latest chart version):
- [x] Ensure no RC images are included.
- [x] Ensure all mirrored images include the right prefix.
- [x] Ensure any new images have `repository` and `tag` on the same level to ensure the rancher-images.txt is generated correctly for it for airgap scenarios.

pennyscissors (commented):
Why is there a commit removing the package contents?

release.yaml
```diff
@@ -1,4 +1,6 @@
 longhorn:
   - 101.2.0+up1.4.0
+  - 102.2.0+up1.4.1
 longhorn-crd:
   - 101.2.0+up1.4.0
+  - 102.2.0+up1.4.1
```
pennyscissors, line 4: This should only contain the charts that are being released

rohitsuse: Resolved.

rohitsakala:
> Member

I deleted the old package folder which had the patches which will be replaced with the new commit coming in. So I made a new commit deleting the old package. 

The reason for deletion was while cherry-picking, I faced merge conflicts. Since we know that we need the package content from the cherry-picked commit and need to delete that is already in the package's content. So I deleted the folder and then cherry-picked the commit. But now, I squashed them into one.

PR #2540, opened 2023-04-11
Fix eks operator v2.7.2 release

a-blender:


a-blender:
This was an attempt to fix a chart issue for 2.7.0-2.7.1 but turned out to not be needed for the release.

PR #2590, opened 2023-05-01
[v2.7] Update Windows GMSA Chart

HarrisonWAffel:
## Problem
The current windows GMSA webhook chart defines Kubernetes version constraints which prevent users from using the chart  on version 1.24 and above. Additionally, the upstream chart has been updated several times, the latest version being 0.6.1. 

## Solution
Update the GMSA chart in several ways. 
1. Bump the image used in the chart to v0.6.1, the latest upstream version
2. Add a `PodDisruptionBudget` to the chart to ensure the web-hook is always available, this CRD was added to the upstream chart recently and helps prevent potential resource blockage via this web-hook.
3. Fix an issue with the `credentialspec.yaml` file which prevented certain fields in the `credspec.ActiveDirectoryConfig.GroupManagedServiceAccounts` from rendering properly (name specified twice)
4. Update chart version constraints, this chart will function on 1.23, 1.24, and 1.25 without issue.
5. update other chart metadata

This PR has an accompanying PR in the image-mirror repo, as we have yet to mirror the upstream images 0.6.1 tag: https://github.com/rancher/image-mirror/pull/387

I've also updated the chart version to align with the versioning seen in other charts, this PR would introduce version `102.0.0+up0.6.1`. If we want to also provide support for this chart in 2.6, I can create an identical release of version `101.0.0+up0.6.1`, so we conform to the versioning differences for charts in 2.6 and 2.7.

## Testing:
+ Deploy this helm chart onto a 1.23, 1.24, or 1.25 cluster (Ideally, check all of them). 
+ During the installation process, create a credential spec object and a pod disruption budget
+ Setup a Windows Active Directory Domain Controller & Join the a worker node into a windows domain
    + Documentation on how to do this can be found in the team 2 confluence page under 'Features Knowledge Base'
    + recommend doing this on Azure, since you can use the `Windows Server 2019 Datacenter With Containers - x64 Gen2` image which skips a few prerequisite steps needed to run containers. 
+ Register the joined windows node into your cluster
+ Create a deployment which meets the following requirements
    + uses the image `harrisonwaffel/gmsa-debug:latest` (this image is 5gb+, will take time to pull - windows base images are large, but this option gives you more flexibility to run commands within the pod using `kubectl exec`.)
    + is created in the `rancher-windows-gmsa-system` namespace
    + uses the `rancher-windows-gmsa` service account 
    + is only scheduled onto windows workers
    + Has the following yaml added to the security options section of the spec 
```yaml
windowsOptions:
   gmsaCredentialSpecName: <your credential spec name>
```
+ After creating the deployment, confirm that the web-hook has properly expanded the credential CRD onto the pod spec by inspecting the YAML of a pod created by the deployment.
+ Ensure that pods enter the 'running' state properly and that the following error is not seen: `windows failed to create shim task: failed to generate container credential guard instance`. 
+ If the pod deploys correctly then we know that the web-hook is functioning as excepted and that the container runtime was able to communicate with the Domain Controller properly. 

If the above shim error is encountered, double check that the active directory domain controller is setup properly and that the worker node is properly joined to the domain. This error will also be seen if the contents of the credential specified in the spec are invalid, so ensuring that the GUID, SID, NETBIOS, and hostname values are valid should also be done. 

### Manual Testing
I've deployed this chart onto clusters with the following Kubernetes versions, `1.23`, `1.24`, and `1.25`. In all cases the chart deployed correctly and was able to update workloads with the given gMSA credentials. I did not test on `1.22` as that is no longer an option in Rancher 2.7, though running this chart on that version should not encounter issues either.

I've done the above testing on windows nodes which are joined properly, as well as ones which are not joined properly. I confirmed that unjoined nodes will encounter the shim error described previously, and that that error can be used as an indication of a bad active directory setup. 

### Automated Testing
Will look into adding hull coverage

Because this chart pulls an image from upstream these changes will not cause any regressions to the web-hook functionality itself. The main changes in this PR are

1. Use image version 0.6.1
2. Update Kubernetes version constraints as the chart only required 1.21+ 
3. Add new PodDisruptionBudget crd, update questions.yaml to fill that CRD in
4. Fix the default credential created during chart install 

The only potential regression I can think of is the credential no longer being created properly during the helm install process, or the PodDisruptionBudget not being created properly. However, I've tested both of those changes myself and did not see any issue.

## Backporting considerations
I'm not sure we want to back port this chart, since 2.6 is entering EOM


packages/rancher-windows-gmsa/charts/templates/credentialspec.yaml
```diff
@@ -7,9 +7,9 @@ metadata:
 credspec:
   ActiveDirectoryConfig:
     GroupManagedServiceAccounts:
-      - Name: {{ .Values.credential.domainJoinConfig.machineAccountNamename }}
+      - Name: {{ .Values.credential.domainJoinConfig.machineAccountName }}
```
HarrisonWAffel, line 10: This fixes an issue wherein the initial credential created during the helm install process would actually be unusable due to these fields being empty

charts/rancher-windows-gmsa/102.0.0+up0.6.1/templates/validate-install-crd.yaml
```diff
@@ -0,0 +1,14 @@
...
+# {{- range .Capabilities.APIVersions -}}
+# {{- if hasKey $found (toString .) -}}
+# 	{{- set $found (toString .) true -}}
+# {{- end -}}
+# {{- end -}}
+# {{- range $_, $exists := $found -}}
+# {{- if (eq $exists false) -}}
+# 	{{- required "Required CRDs are missing. Please install the corresponding CRD chart before installing this chart." "" -}}
+# {{- end -}}
+# {{- end -}}
```
bfbachmann, line 13: How come this is all commented out?

HarrisonWAffel: This is an autogenerated file, I think its because CRD validation isn't set for this chart. So while the file is generated, is commented out since we don't want the logic. You can see the same file in a few other charts used by Rancher too, also commented out

nicholasSUSE:
Hello @HarrisonWAffel can I close this PR?

nicholasSUSE:
Closed due to inactivity

PR #2592, opened 2023-05-02
[dev v2.7] Cut new RC for Backup Restore Chart

eliyamlevy:
## Issue: 
https://github.com/rancher/backup-restore-operator/issues/306 Updated Wrangler and a few other go packages.

## Testing

Since this is just updating packages/modules/images the general functionality should be tested. However there should be a specific round of testing focused on the encryption feature. Since this wrangler update led to having to update how the encryption transformer overrides are loaded.


crobby (approved):
Looks reasonable to me, passes validation.

PR #2606, opened 2023-05-05
Fix Epinio logo

fgiudici:
Just found that the Epinio logo in the assets is not a logo. Load the right logo.


thehejik (approved):
Tested in my fork by changing `index.yaml` which is now pointing to the updated logo from this PR `icon: https://raw.githubusercontent.com/fgiudici/rancher-charts/fix_epinio_logo/assets/logos/epinio.svg` and finally: 
image

brandond (approved):

brandond:
cc @MbolotSuse @PennyScissors

PR #2610, opened 2023-05-11
41517 epinio k1.26 support

andreas-kupries:
## Issue: 

Ref https://github.com/rancher/rancher/issues/41517
Ref https://github.com/epinio/epinio/issues/2295

## Problem

Update chart to changes in kube versions supported by Rancher.

## Testing

To be done by Epinio QA, will report here.

## Backporting considerations

No backporting required AFAIK.


andreas-kupries:
The attempt to remove old version of the chart failed in validation. The reason looks to be that the old versions have assets and are referenced by the index.yml, but are not listed in release.yml. While the `make remove` works and properly updates assets and index, the validation then breaks because they were not in the release and are thus not properly removed there. I do not know why/how/when these epinio versions were removed from release.yml but not their assets. And how this partial removal passed validation.

I am unclear how to bring assets, index, and release into a state where the old releases are gone and validation passes.

thehejik (approved):
### QA report
#### Testing environment
Rancher `v2.7.5-rc1` on top of a single-node `v1.26.4+k3s1` cluster with incorporated traefik lb. Helm chart git repository for Epinio Rancher App installation https://github.com/epinio/charts.git and branch `41517-epinio-k1.26-support`

(rancher deployed with  `--version v2.7.5-rc1  --set global.cattle.psp.enabled=false`)

#### Tested
* Adding the custom repository for this PR
* Installation of `Epinio (102.0.2+up1.6.2)` App helm chart in Rancher (local k3s cluster) by using default values (+entered base epinio domain) - `epinio` and `epinio-crd` apps were successfully deployed in `cattle-epinio-system` namespace
* All the input boxed for values were ok (Ingress class for the server, dex and for apps were added)
* No k8s API Deprecation messages present in the helm install log
* All the relevant resources like pods are in `Ready` state
* Accessed `epinio-ui` over ingress, login by using default credentials
* Created epinio `workspace` namespace for its applications
* Deployment of sample-app phpinfo() app stored in local folder
* Accessing the sample-app over provided URL in epinio-ui
* All resources still ok
* Uninstallation (`epinio` and then `epinio-crd`) ok

thehejik (approved):

mmartin24:
+1 on the report here: https://github.com/rancher/charts/pull/2610#pullrequestreview-1422514849
Just to add that we also:

-  tested deploying ranchar`v1.25.5+k3s2` and then upgraded to `v1.26.4+k3s1` directly using:

```
curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION="v1.26.4+k3s1" INSTALL_K3S_EXEC="--write-kubeconfig-mode \"0644\"" sh -
``` 

- Checked that ongoing apps were still accessible
- New apps could be deployed successfully

olblak (approved):

PR #2647, opened 2023-05-18
Update Chart.yaml to include compatibility with k8s 1.26

geethub97:


packages/rancher-monitoring/rancher-kube-state-metrics/generated-changes/patch/Chart.yaml.patch
```diff
@@ -3,7 +3,7 @@
 @@ -1,3 +1,10 @@
 +annotations:
 +  catalog.cattle.io/hidden: "true"
-+  catalog.cattle.io/kube-version: '>= 1.16.0-0 < 1.25.0-0'
++  catalog.cattle.io/kube-version: '>= 1.16.0-0 <= 1.26.0-0'
```
git-ival: Is there a reason this is strictly checking against `<= 1.26.0-0` instead of `< 1.27.0-0` like the other patch files in this PR?

nflynt (approved):
Time to test **all the things**

PR #2657, opened 2023-05-21
Monitoring Chart fixed endpoints definition

Sebastian-Prokesch:
additional hypen at metricRelabeling
started new list item on some ServiceMonitor Definitions.

moved metricRelabeling key into if structure

## Problem
After installing rancher-monitoring some Prometheus Target are not found. For example the Kube-API target.
The Reason for this is a wrong ServiceMonitor definition for some Services breaking the endpoints List.

## Solution
Fixed typo on ServiceMonitoring Ressources 

### Manual Testing
Fixed ServiceMonitors on Cluster and Target was available and could be scraped


Sebastian-Prokesch:
Hope this PR helps. If more Info is required please feel free to ask. 
(sry it's my first PR in a big project like this, hope i've done this right :) )

PR #2658, opened 2023-05-22
Add Elemental Chart

fgiudici:
## Issue: 

https://github.com/rancher/elemental-operator/issues/492  - add Elemental helm charts in rancher/charts repo


packages/elemental/elemental/generated-changes/patch/values.yaml.patch
```diff
@@ -0,0 +1,46 @@
...
++
+ seedImage:
+-  repository: "quay.io/costoolkit/elemental-seedimage"
+-  tag: v1.2.5
++  repository: "registry.opensuse.org/isv/rancher/elemental/stable/teal53/15.4/rancher/seedimage-builder/5.3"
++  tag: "1.2.5"
+   imagePullPolicy: IfNotPresent
++
++channel:
++  repository: "registry.opensuse.org/isv/rancher/elemental/stable/teal53/15.4/rancher/elemental-teal-channel/5.3"
```
rohitsuse: This should be `rancher` not `registry.opensuse.org`. You can add your docker image to https://github.com/rancher/image-mirror before submitting any new chart.

This goes for all other docker images. It should be `rancher` docker repo

packages/elemental/elemental/generated-changes/patch/values.yaml.patch
```diff
@@ -0,0 +1,46 @@
...
+ seedImage:
+-  repository: "quay.io/costoolkit/elemental-seedimage"
+-  tag: v1.2.5
++  repository: "registry.opensuse.org/isv/rancher/elemental/stable/teal53/15.4/rancher/seedimage-builder/5.3"
++  tag: "1.2.5"
+   imagePullPolicy: IfNotPresent
++
++channel:
++  repository: "registry.opensuse.org/isv/rancher/elemental/stable/teal53/15.4/rancher/elemental-teal-channel/5.3" 
++  tag: latest
```
rohitsuse: Also, generally `latest` is not recommended. Can you please put the specific tag ?

packages/elemental/elemental/generated-changes/patch/Chart.yaml.patch
```diff
@@ -0,0 +1,25 @@
...
++++ charts/Chart.yaml
+@@ -1,14 +1,18 @@
+ annotations:
++  catalog.cattle.io/auto-install: elemental-crd=match
+   catalog.cattle.io/certified: rancher
+   catalog.cattle.io/display-name: Elemental
+-  catalog.cattle.io/experimental: "true"
+   catalog.cattle.io/namespace: cattle-elemental-system
+   catalog.cattle.io/os: linux
+   catalog.cattle.io/provides-gvr: elemental.cattle.io/v1beta1
```
rohitsuse, line 13: You need to add the annotation `kube-version`  like this https://github.com/rancher/charts/blob/dev-v2.7/charts/epinio/102.0.3%2Bup1.8.1/Chart.yaml#L7

rohitsuse: You need to add `permist-os` annotation like this https://github.com/rancher/charts/blob/dev-v2.7/charts/epinio/102.0.3%2Bup1.8.1/Chart.yaml#L9

rohitsakala:
Other than the comments, everything else looks good to me.

rohitsakala (changes requested):
Other than the comments, everything else looks good to me.

packages/elemental/elemental/generated-changes/patch/values.yaml.patch
fgiudici: ~Hey @rohitsakala , thanks for the hint!
We build the images in OBS and now IBS: this means we end-up with our container images being in the registry.suse.com (just updated the link from opensuse to suse registry).
In particular, Elemental brings OS images as containers built on top of SLE OSes, so both the OS images and the channel (here the channel just carries the updated list of SLE based OS images) are build in IBS.
So all then is published on registry.suse.com.
IIUC, regular container images for the Rancher marketplace are uploaded from all the sources and mirrored on dockerhub. Then, for Rancher Prime, those images are mirrored on registry.suse.com from dockerhub.
Since the elemental container images are already built on IBS and automatically published on registry.suse.com, our plan would be to leave them there for everyone. How would that sound?~

Updated with mirrored image links from https://github.com/rancher/image-mirror/pull/428 (not there yet).
Added templating.

packages/elemental/elemental/generated-changes/patch/values.yaml.patch
fgiudici: The channel here is just a container image carrying a list of URL with updated Elemental OSes, so we wanted it to keep updating... but your comment raised some discussion inside the team, and you are likely right: we need a specific tag here, but then we need to rethink how to provide OS updates 🤔 
We'll need some time to address this

packages/elemental/elemental/generated-changes/patch/Chart.yaml.patch
fgiudici: Thanks, done!

packages/elemental/elemental/generated-changes/patch/values.yaml.patch
rohitsuse: You could update the image to the new tag and release a new Chart in `rancher/charts` by upgrading the version in `package.yaml`

fgiudici: We will check the same image tag regularly and we may required to update it... we anyway need a fixed tag to match the specific Chart version (and so Elemental release). Updated to fixed tag 👍🏼

fgiudici:
We need to wait for https://github.com/rancher/image-mirror/pull/428 before merging this.

fgiudici:
Target is now 2.8.0, moving to #3104 .

PR #2691, opened 2023-05-31
Bump rancher webhook to v0.3.5-rc7

MbolotSuse:
## Issue: 
Bumps rancher-webhook to v0.3.5-rc7


maxsokolovsky (approved):

pjbgf (approved):

pjbgf:
Do we also need a similar PR for `v0.2.x`?

MbolotSuse:
@pjbgf I don't think so.`Release/v2.6` seems to [already point to the right version of the webhook](https://github.com/rancher/rancher/blob/04456f99123416e5895dd28b2ca10468c78b9c03/package/Dockerfile#L58)

PR #2701, opened 2023-06-05
Fix/neuvector double command issue

pna-nca:
## Problem
See the referenced issue.

## Solution
The patch to NeuVector chart which introduces the bug is fixed by avoiding duplicated "command" instruction.

## Testing
Done.

### Manual Testing
Tested on internally deployed clusters.

### Automated Testing
N/A

## QA Testing Considerations
N/A

N/A

## Backporting considerations
Can be backported to 102.0.0+up2.4.2.


pna-nca:
The issue which this pull request fixes was silently fixed in 10bf48c7a6f4a4c461604327748581a242504a07.

PR #2719, opened 2023-06-13
[1.27] New gatekeeper version with updated kube-version annotation

diogoasouza:
[rancher#4178](https://github.com/rancher/rancher/issues/41781)


packages/rancher-gatekeeper/generated-changes/patch/Chart.yaml.patch
```diff
@@ -5,7 +5,7 @@
 +  catalog.cattle.io/auto-install: rancher-gatekeeper-crd=match
 +  catalog.cattle.io/certified: rancher
 +  catalog.cattle.io/display-name: OPA Gatekeeper
-+  catalog.cattle.io/kube-version: '>= 1.20.0-0 < 1.27.0-0'
++  catalog.cattle.io/kube-version: '>= 1.20.0-0 < 1.28.0-0'
```
rohitsuse, line 8: can we only put only the min version ?

PR #2747, opened 2023-06-27
Update aks-operator to v1.1.2-rc2

mjura:


mjura:
https://github.com/rancher/charts/pull/2748

PR #2753, opened 2023-06-29
[rancher-monitoring] Correct list permission with get

vatsalparekh:
## Issue: 

[SURE-6299](https://jira.suse.com/browse/SURE-6299)
## Problem

When accessing the `endpoints` object called `rancher-monitoring-alertmanager`, either from HTTP url or from kubectl, for a user which has the `view monitoring` permission only, it resulted in 403 error as followed

```sh
Error from server (Forbidden): endpoints is forbidden: User "u-rb45j" cannot list resource "endpoints" in API group "" in the namespace "cattle-monitoring-system"
``` 
## Solution

This error occurs because the clusterrole `monitoring-ui-view` gives `list` permission with `resouceName` mentioned. Which means a user with this permission cannot list all the `endpoints` or even `get` the specified `resourceName` due to lack of `get` permission. I think just giving it `get` permission will fix this.

```sh
> k get endpoints rancher-monitoring-alertmanager  -n cattle-monitoring-system
Error from server (Forbidden): endpoints "rancher-monitoring-alertmanager" is forbidden: User "u-rb45j" cannot get resource "endpoints" in API group "" in the namespace "cattle-monitoring-system" 
```
## Testing

Manually checked after applying this fix
```
> k get endpoints rancher-monitoring-alertmanager  -n cattle-monitoring-system
NAME                              ENDPOINTS           AGE
rancher-monitoring-alertmanager   10.42.136.29:9093   3h39m
> k get endpoints -n cattle-monitoring-system
Error from server (Forbidden): endpoints is forbidden: User "u-26md8" cannot list resource "endpoints" in API group "" in the namespace "cattle-monitoring-system"
```

## Backporting considerations

This might be happening in previous versions also


joshmeranda:
This fix looks ok to me.

@vatsalparekh can you update the commits here to follow the steps in https://github.com/rancher/charts/blob/dev-v2.7/docs/developing.md#making-changes-to-packages?

@geethub97 can you take a quick look?

geethub97 (changes requested):
The change itself looks fine, but can you please split the commit into two and also bump the version for monitoring and add that new version to the release yaml? 
In the end you should have 3 commits: 
- one with changes to the packages directory
- one for make charts
- one to update the release yaml

geethub97 (approved):

PR #2805, opened 2023-07-25
Add a new chart: rancher-provisioning-capi

Oats87:
## Issue: 
https://github.com/rancher/rancher/issues/41094

## Problem
There is no chart to install cluster-api core controllers into a Rancher cluster.

## Solution
Add a cluster-api CAPI chart.

## Testing
The v2prov integration tests (and corresponding QA integration tests that use the CAPR planner) will verify functionality.


jakefhyde (dismissed):

index.yaml
```diff
@@ -13544,6 +13544,32 @@ entries:
...
+  rancher-provisioning-capi:
+  - annotations:
+      catalog.cattle.io/certified: rancher
+      catalog.cattle.io/display-name: Rancher Provisioning CAPI Controller Manager
+      catalog.cattle.io/kube-version: '>=1.23.0-0'
+      catalog.cattle.io/namespace: cattle-provisioning-capi-system
+      catalog.cattle.io/os: linux
+      catalog.cattle.io/permits-os: linux,windows
+      catalog.cattle.io/provides-gvr: apps.deployment/v1
+      catalog.cattle.io/rancher-version: '>= 2.7.0-0'
```
jiaqiluo, line 13557: 1. Should it be at least 2.7.7? 
2. Should the chart be hidden in Rancher UI? If so we need `catalog.cattle.io/hidden: "true"`

jiaqiluo (changes requested):

index.yaml
Oats87: No, we don't want to lock in a specific version at this point given we don't actually know the next .x version. It's fine to leave it as `2.7.0`

Good catch on the hidden. It was originally there, but dropped when addressing Arvind's previous comments. PR incoming to fix the chart on the `provisioning` repo.

Oats87: PR to fix this is here: https://github.com/rancher/provisioning/pull/2 once this is merged I will merge-train it back into this PR.

Oats87: Change pushed to add the hidden annotation

jiaqiluo (approved):

jakefhyde (approved):

HarrisonWAffel (approved):

PR #2835, opened 2023-08-09
Add fleet 0.7.1 rc.2 to Rancher 2.7.6

thardeck:
Update Fleet to v0.7.1-rc.2

Changelog: https://github.com/rancher/fleet/releases/tag/v0.7.1-rc.2

Neuvector and Longhorn needed to be added to the `release.yaml` otherwise the validate script would have failed.

Supersedes #2832


release.yaml
```diff
@@ -2,12 +2,24 @@ prometheus-federator:
 - 3.0.0+up0.3.3
 rancher-project-monitoring:
 - 3.0.0+up0.3.3
+longhorn:
```
manno, line 5: I'm not sure why we would add this, seems like that should be a separate PR? 🤷

manno (approved):

release.yaml
rohitsuse: @manno The dev-v2.7.6 was not rebased with release-v2.7. So an error in the CI

PR #2851, opened 2023-08-14
Fixing issues using rancher-pushprox to hit authenticated kubernetes endpoints

rheyser:
## Issue: 

Users cannot use authenticated endpoints directly without modifying the servicemonitors created by this chart if using the upstream kube-prometheus-stack chart. 

Additionally, users cannot use imagePullSecrets for private repositories.

## Problem

In its current configuration, without using the proxies, which do not work with the upstream chart, users cannot authenticate to kubernetes api endpoints that require authentication to access them. Additionally, since the same endpoints do not expose certificates to any ip but `127.0.0.1`, I've added the capability to honor the insecureSkipVerify flag if it is set. While I was there, I also fixed a bug in metricRelabelings where it would generate an empty list if not configured, it would generate an invalid list if not configured. 

## Solution

This adds the ability to use authentication on the clients to the kubernetes components. Added the ability to correctly set the http(s) scheme if not using the proxy feature. Fixed a bug so service monitors will honor the insecureSkipVerify flag if set by the client. Fixed a bug in metricRelabelings where it would generate an empty list if not configured. Added imagepullsecrets as a global field to be used whenever an image is pulled. 

### Manual Testing

I've tested this using our current configuration yamls and helm templating the changes.

### Automated Testing

None.

## QA Testing Considerations

Test all the things. 

The primary focus of changes here were done to the rancher-pushprox client. The main focus of testing should be there and specifically on the service monitor. 

## Backporting considerations

Yes, needs to be backported to Rancher 2.6.


rheyser:
This should resolve https://github.com/rancher/rancher/issues/43414

joshmeranda (approved):

rohitsakala (changes requested):
This PR doesn't follow the guidelines of updating a charts. 

@joshmeranda Lets connect to make sure this PR follows them and then merge this.

joshmeranda:
closing in favor of https://github.com/rancher/charts/pull/2962

MKlimuszka:
Actually closing, based on Josh's comment.

PR #2881, opened 2023-08-17
Adding image check to pull request workflow and bumping scripts version

lucasmlp:
- Adds a new step in the pull request workflow to check all the container images used in the project
- The step will run only if the pull request is opened against a release branch
- If the step fails, the workflow doesn't fail. The reason for this is that we have two tags from the fluent-bit repository that doesn't exist in Docker Hub. Until we figure that out, the new step would fail all the PRs if it was required to run successfully. After sorting the tags out we'll set the step success to be required.


diogoasouza (approved):

.github/workflows/pull-request.yaml
```diff
@@ -26,4 +26,12 @@ jobs:
         run: sudo make validate
 
       - name: Run Hull tests
-        run: cd tests && go test -v ./...
\ No newline at end of file
+        run: cd tests && go test -v ./...
+
+      - name: Check container images 
+        run: make check-images
+        if: startsWith(github.ref, 'refs/heads/upstream/release-v')
```
rohitsuse, line 33: This needs to run for every PR so we catch the errors beforehand and not until the release date of charts.

.github/workflows/pull-request.yaml
```diff
@@ -26,4 +26,12 @@ jobs:
...
+        continue-on-error: true
```
rohitsuse, line 34: This field can be removed since this is the last step in the process.

rohitsakala:
Also, for the missing images, can we create a PR for adding the missing images in `image-mirror` github repo https://github.com/rancher/image-mirror before merging this PR ? so we can remove that continue-on-error field ?

PR #2884, opened 2023-08-18
Bump webhook to v0.4.0-rc2

cmurphy:


diogoasouza:
You need to add the new chart version to the `releases.yaml` file

diogoasouza (dismissed):

packages/rancher-webhook/package.yaml
```diff
@@ -1,3 +1,3 @@
-url: https://github.com/rancher/webhook/releases/download/v0.3.5/rancher-webhook-0.3.5.tgz
+url: https://github.com/rancher/webhook/releases/download/v0.4.0-rc1/rancher-webhook-0.4.0-rc1.tgz
 version: 103.0.0
```
maxsokolovsky, line 2: Aren't we also bumping the `version` to 104.0.0?

maxsokolovsky: Never mind, just saw a thread offline.

cmurphy:
Updated to use webhook rc2 which aligns the Chart.yaml rancher version

diogoasouza (approved):

cbron (approved):

PR #2901, opened 2023-08-22
[1.27] New gatekeeper version with updated kube-version annotation

nicholasSUSE:
(https://github.com/rancher/rancher/issues/41781)


packages/rancher-gatekeeper/generated-changes/patch/Chart.yaml.patch
```diff
@@ -5,7 +5,7 @@
 +  catalog.cattle.io/auto-install: rancher-gatekeeper-crd=match
 +  catalog.cattle.io/certified: rancher
 +  catalog.cattle.io/display-name: OPA Gatekeeper
-+  catalog.cattle.io/kube-version: '>= 1.20.0-0 < 1.27.0-0'
++  catalog.cattle.io/kube-version: '>= 1.20.0-0'
```
rohitsuse, line 8: Can we do this like '>= 1.20.0-0 < 1.28.0-0'. I can see a case where this can be easily missed. 

- Suppose release Rancher 2.8.1 and later another version 2.8.4 which supports k8s 1.28 but no new chart is being introduced. then updating this chart would fail by the customer since this chart doesn't support 1.28.

rohitsuse: I have also udpated the branching document

release.yaml
```diff
@@ -22,3 +22,7 @@ rancher-eks-operator:
 - 103.0.0+up1.3.0-rc1
 rancher-eks-operator-crd:
 - 103.0.0+up1.3.0-rc1
+rancher-gatekeeper:
+  - 103.0.1
```
rohitsuse, line 26: We need to put the full version here.

rohitsakala (changes requested):
The first commit message is not appropriate. Follow this https://cbea.ms/git-commit/

PR #2902, opened 2023-08-23
updating rancher-gatekeeper support for k8s 1.27

nicholasSUSE:
Feature request for updating supported kubernetes version for several charts: https://github.com/rancher/rancher/issues/41720
Request to update to gatekeeper: https://github.com/rancher/rancher/issues/41781


packages/rancher-gatekeeper/generated-changes/patch/templates/_helpers.tpl.patch
rohitsuse, line 1: This file shouldn't be generated when you run make charts

PR #2919, opened 2023-08-31
Adding check RC step to pull request workflow running on release branches

lucasmlp:


rohitsakala (commented):
Can we use something like this https://github.com/marketplace/actions/task-completed-checker ?

to enforce the checkbox ?

PR #2937, opened 2023-09-05
[v2.8] Remove doNotRelease tag from Fleet and add Fleet 0.8.0 rc.9

thardeck:
following the branching strategy guide.

I am using `yq` version 4 to change the `release.yaml` that's why the indentation of it is changed a bit.

```
Sadly `yq` enforces indentation for arrays atm because of this issue: https://github.com/mikefarah/yq/issues/1309
That's why I have to fix the indentation with sed manually after using yq.
Maybe we should enforce `yq` default indentation for the release.yaml in the future so manual fixes are not required.
```


rohitsakala:
@thardeck Could you create an issue for yq ? and add `team/mapps` and `next up` label so that we can work on it later

release.yaml
```diff
@@ -1,13 +1,16 @@
 rancher-monitoring:
-  - 103.0.0+up40.1.2
+- 103.0.0+up40.1.2
 rancher-monitoring-crd:
-  - 103.0.0+up40.1.2
+- 103.0.0+up40.1.2
 fleet:
-  - 102.1.1+up0.7.1
+- 102.1.1+up0.7.1
```
rohitsakala, line 6: Can we remove 102.1.1+up0.7.1 ? since we are only adding 103.0.0+up0.8.0-rc.9 ? and same for fleet-agent ?

rohitsakala (approved):

thardeck:
> @thardeck Could you create an issue for yq ? and add `team/mapps` and `next up` label so that we can work on it later

I have created an issue: https://github.com/rancher/charts/issues/2949

Issue #2949, opened 2023-09-06
[RFE] Specify and enforce `yq` indentation of release.yaml


thardeck:
### Is your feature request related to a problem? Please describe.
The issue is that the `release.yaml` does not follow a style. Some indent the list elements, others do not. Most nowadays seem to not indent the list elements (which is also supported by yaml) but not supported when using `yq` version 4 as described in this issue: https://github.com/mikefarah/yq/issues/1309

So if I am using `yq` version 4 to change the file, every other entry gets indented differently even though I just want to add one entry.

### Describe the solution you'd like
We indent the `release.yaml` as `yq` version 4 suggests and also document that as the default style so everyone else can follow it.
On top everyone can use `yq` for automating release bumps without any additional special magic to get the same indentation.

To see the `yq` indentation just run `yq -i release.yaml`.

### Describe alternatives you've considered
At the moment our release script runs `sed -i "s/^  -/-/" release.yaml` to remove the indentation before list elements.

### Additional context
We could easily add a Github action that runs `yq -i release.yaml` on each new pull request and complains if running that command would change the file to make sure that everyone follows the same style.

PR #2955, opened 2023-09-07
checking release yaml identation

diogoasouza:
## Issue https://github.com/rancher/charts/issues/2949
Adds a step that runs `yq -i release.yaml` and returns an error if there's any changes

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -21,7 +21,10 @@ jobs:
 
       - name: Pull in all relevant branches
         run: git fetch origin release-v2.6
-      
+
+      - name: Check release.yaml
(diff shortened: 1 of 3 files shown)


rohitsakala (approved):

.github/workflows/pull-request.yaml
```diff
@@ -21,7 +21,10 @@ jobs:
 
       - name: Pull in all relevant branches
         run: git fetch origin release-v2.6
```
cwayne18, line 23: Should we update this to `release-v2.8`?

diogoasouza: We should. I'll open a different PR as that change is not related to this one

PR #2804, opened 2023-07-25
Epinio 1.9.0 integration

andreas-kupries:
Updated integrated Epinio chart to release 1.9.0.

See https://github.com/epinio/epinio/issues/2464

Depends on https://github.com/rancher/image-mirror/pull/423 getting merged before testing is possible.

PR #2969, opened 2023-09-12
Revert PR #2804 chart removal

andreas-kupries:
## Issue: 

PR #2804 removed a number of older charts.
The removal from release.yaml apparently failed.
This PR brings the removed charts and assets back.

:warning: :construction: __With the PR created I see a lot of conflicts in assets and charts.
Looks that manual reversal is better than automatic of the commit.__ :construction:  :warning: 

## Problem

See above.

## Solution

See above.

## Testing

All the charts and assets were tested at the time they were initially added.


andreas-kupries:
PR is against wrong dev branch.

PR #2977, opened 2023-09-12
Un-RC CIS Benchmark Chart 4.2.0

JonCrowther:
Got sign off to un-RC the chart. Made a version with the un-RCed security-scan image and deleted all the RC versions.


andypitcher (approved):
LGTM.

Priyashetty17 (approved):

mitulshah-suse (approved):
LGTM

Priyashetty17 (approved):

MbolotSuse (approved):
Looks similar to other un-rcs for this chart. A couple of pointers:
- Typically you remove an old rc when you add a new one (would have allowed you to not have to remove so many things in this PR)
- Typically you want to separate these into 3 commits - on to change the version, one to pull in the new chart, one to remove the old chart. See this PR: https://github.com/rancher/charts/pull/2711 for reference. 

I don't think any of this is blocking, but they are things that I would recommend doing differently in the future.

PR #2868, opened 2023-08-15
Hardened k3s network policy

geethub97:
## Problem
Missing network policy in hardened.yaml lead to an error with the target serviceMonitor/kube-system/rancher-monitoring-coredns. 

## Solution
Added missing network policy. 

## Testing
Steps in linked ticket are sufficient to reproduce.

PR #2987, opened 2023-09-14
Change the monitoring coredns network policy to an "allow all" policy. 

geethub97:
## Problem
kubectl get all (or any kubectl command) outputs the following error:
Error from server (InternalError): an error on the server ("dial tcp: lookup kubernetes. Default on 10.43.0.10:53: read udp 10.42.0.26:56849->10.43.0.10:53: i/o timeout") has prevented the request from succeeding

This bug also prevents the monitoring chart from being deleted as the Kubernetes API calls hang. 

## Solution
Change the monitoring-network-policy to a default allow all policy, to allow proper network traffic. 

## Testing
Install monitoring chart version 103.0.0+up40.1.2 on rancher version 2.8 and execute any kubectl command from the Rancher UI. 

### Manual Testing
Installed monitoring chart version 103.0.0+up40.1.2 on rancher version 2.8 and executed any `kubectl get all` from the Rancher UI kubectl shell and received the correct response. Successfully deleted, reinstalled, and upgraded monitoring chart. 

As this change touches the same monitoring policy for https://github.com/rancher/rancher/issues/38820, I tested this change in a hardened cluster to ensure that no targets are down. 

### Automated Testing
N/A

## QA Testing Considerations
Fresh install, upgrade, deletion of monitoring chart. 

N/A 

## Backporting considerations
N/A as bug was introduced by the fix for https://github.com/rancher/rancher/issues/38820 in this PR: https://github.com/rancher/charts/pull/2868.


prachidamle:
@aiyengar2 @brandond Can you please review this change to the network policy which was introduced for k3s hardened cluster in this PR https://github.com/rancher/charts/pull/2868

charts/rancher-monitoring/103.0.0+up40.1.2/templates/rancher-monitoring/hardened.yaml
```diff
@@ -131,18 +131,17 @@ spec:
 apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
-  name: monitoring-coredns-network-policy
+  name: rancher-monitoring-default-allow-all
```
brandond: The policy still only applies to the coredns pods (as the podSelector is still present) so the name should reflect this, instead of suggesting that it is a default that applies to all pods in the namespace.
```suggestion
  name: rancher-monitoring-coredns-allow-all
```

packages/rancher-monitoring/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/hardened.yaml
```diff
@@ -131,18 +131,17 @@ spec:
 apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
-  name: monitoring-coredns-network-policy
+  name: rancher-monitoring-default-allow-all
```
brandond: same comment as above

brandond (approved):

brandond (changes requested):
nit on policy names

brandond (approved):

PR #3002, opened 2023-09-21
Images and tags checkbox

lucasmlp:
- Adds a workflow to comment on the PR and add a checkbox to remind the user to check the images and tags hierarchy
- Adds a task checker workflow to check if the checkbox was checked
- Removes the check from the pull request template

- [x] Dummy checkbox

### DIFF
--- a/.github/workflows/tasks-check.yaml
+++ b/.github/workflows/tasks-check.yaml
@@ -1,0 +1,20 @@
+name: Tasks Check
+
+on:
+  pull_request_review:
+    types: 
+      - submitted
+      - approved
+    branches:
+      - dev-v*
+      - release-v*
+
+jobs:
+  checkbox-check:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check checkboxes
+      uses: mheap/require-checklist-action@v2
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/tasks-check.yaml
```diff
@@ -0,0 +1,20 @@
+name: Tasks Check
+
+on:
+  pull_request_review:
```
rohitsuse, line 4: Can we do this before giving a pull request review  somehow ?

rohitsakala (changes requested):
Commented

.github/workflows/tasks-check.yaml
lucasmlp: It could be done using the pull_request event, but using different events generates different checks on the PR. That's something that I didn't know beforehand and now I know. It's not one check per workflow. I thought using different events would rerun the same check but it creates one check per event per workflow. For example:

- Tasks Check (pull_request/opened)
- Tasks Check (pull_request_review/approved)

If I added the pull_request event, It would generate the Tasks Check (pull_request/opened) and this one would only be updated if the user pushed something into the branch. Which may not happen and the PR would be forever stuck.

That's why I decided to run it after the PR approval.

rohitsakala (approved):

PR #3039, opened 2023-09-25
Update sriov chart to 103.0.0

mgfritch:
## Issue: 

Also updates chart to use node-feature-discovery v0.14.1

Resolves:
https://github.com/rancher/rke2/issues/4654
https://github.com/rancher/rke2/issues/4489
https://github.com/rancher/rke2/issues/4499

## Problem

* Update SRIOV charts to dev-v2.8
* Provide support for multi-arch SRIOV images
* kube-version > v1.28 requires an updated node-feature-discovery > v0.13.4

### Manual Testing

Deploy charts via helm using a three node RKE2 v1.28.2+rke2r1 cluster


manuelbuil (approved):

brandond (approved):

packages/rancher-sriov/generated-changes/patch/Chart.yaml.patch
```diff
@@ -5,7 +5,7 @@
 +  catalog.cattle.io/auto-install: sriov-crd=match
 +  catalog.cattle.io/certified: rancher
 +  catalog.cattle.io/experimental: "true"
-+  catalog.cattle.io/kube-version: '>= 1.16.0-0 < 1.27.0-0'
++  catalog.cattle.io/kube-version: '>= 1.16.0-0 < 1.29.0-0'
```
rohitsuse: Confirming if this is not a mistake. We want to support 1.28 as well ?

packages/rancher-sriov/generated-changes/patch/values.yaml.patch
```diff
@@ -36,25 +36,25 @@
 -  webhook: ghcr.io/k8snetworkplumbingwg/sriov-network-operator-webhook
 +  operator:
 +    image: rancher/hardened-sriov-network-operator
```
rohitsuse: We should follow the annotation of `repository: {imagename}` and `tag: {tag}` to make sure that these images are included in rancher-images.txt so that airgap customers can use this chart

rohitsuse: Same goes for all other charts

rohitsuse: There is a checkpoint also that you need to tick for the CI to pass. https://github.com/rancher/charts/pull/3039#issuecomment-1735931175

packages/rancher-sriov/generated-changes/patch/Chart.yaml.patch
mgfritch: I got confused on the support matrix, it's fixed

packages/rancher-sriov/generated-changes/patch/values.yaml.patch
mgfritch: done!

rohitsakala (approved):

PR #3121, opened 2023-10-12
rancher-monitoring: Add compatibility to IPv6 only clusters

fayak:
## Problem
Our cluster is IPv6 only. Having nginx listening only on IPv6 breaks the svc. This allows both ipv4 and v6

## Solution
listen additionally on ipv6

## Testing
Tested on our cluster with success


nicholasSUSE:
Hello @fayak 
Is this Pull Request still relevant?
Can I close it?

fayak:
@nicholasSUSE Closed in favor of https://github.com/rancher/charts/pull/3465 which is still relevant :)

PR #3165, opened 2023-10-23
Remove Rancher Istio 1.18.2

rohitsakala:
Removing Istio 1.18.2 since we have 1.19 now and for every Rancher release we want to release only one version. 1.18,1.17 is already available via Rancher 2.7 line


lucasmlp (approved):
LGTM 🚀

diogoasouza (approved):
LGTM

PR #3225, opened 2023-11-16
Rancher-monitoring: handle recreation of existing Alertmanager secret

SamuZad:
## Problem
When trying to update certain parts of Alertmanager's config (specifically, the global options), I noticed that there is no convenient way of doing it via Helm. This is due to the fact that when the Helm installation creates the alertmanager secret as a pre-hook, it [skips over](https://github.com/rancher/charts/blob/dev-v2.8/charts/rancher-monitoring/103.0.0%2Bup45.31.1/templates/alertmanager/secret.yaml#L2-L4) trying to do anything if the secret already exists. 

However, when trying to update a chart that is already installed, it is safe to assume that the chart (and therefore the secret) already exists.. The current workaround is that prior to upgrading the helm chart, one can delete the alertmanager secret, forcing a recreation on update. However, this is a manual step and gets rather annoying, especially when the global configuration is updated frequently.

## Solution
Adding in a `recreateIfExists` flag under `alertmanager.secret`, which when set to true will recreate the secret. It is set to false by default to preserve previous behaviour.

## Reproducing the issue:
Install the rancher-monitoring helm chart. Make an update to the `alertmanager.config.global.resolve_timeout` to a non default value (such as 15m), and you will see the following:
1. the alertmanager configuration will not reflect the new value
2. the alertmanager secret will not update due to [the condition](https://github.com/rancher/charts/blob/dev-v2.8/charts/rancher-monitoring/103.0.0%2Bup45.31.1/templates/alertmanager/secret.yaml#L4) which instructs it to skip:
`{{- if (not (lookup "v1" "Secret" (include "kube-prometheus-stack.namespace" .) $secretName)) }}`

When installing with my modifications (with `alertmanager.secret.recreateIfExists` set to `true`), the chart will correctly overwrite the existing secret.

## Backporting considerations
I think it would be nice to have this feature in older versions of this chart as well (and would be happy to implement, as it's not major work). However, it is not essential.


packages/rancher-monitoring/rancher-monitoring/generated-changes/patch/README.md.patch
```diff
@@ -38,6 +38,7 @@
 +| `kubeAdmProxy.enabled` | Create a PushProx installation for monitoring kube-proxy metrics in kubeAdm clusters | `false` |
 +| `kubeAdmEtcd.enabled` | Create a PushProx installation for monitoring etcd metrics in kubeAdm clusters | `false` |
 +
- 
++
```
joshmeranda, line 41: nit: cleanup new whitespace

joshmeranda (changes requested):
Looks mostly good to me. Only change would be to bump the version -> `103.0.1+up45.31.1`

SamuZad:
> Looks mostly good to me. Only change would be to bump the version -> `103.0.1+up45.31.1`

I bumped the chart to `103.0.1+up45.31.1` as requested

packages/rancher-monitoring/rancher-monitoring/generated-changes/patch/README.md.patch
SamuZad: I can't for the life of me how to figure out how to do this without breaking all of the CI validation. Also, looking at the git history this had a whitespace in line 41 previously, which is now replaced by a +.

So line 41 modifies as such: " " -> "+"

joshmeranda: Ah ok, yeah then don't worry about it :+1:

SamuZad:
@joshmeranda any chance this could be merged?

joshmeranda (approved):

joshmeranda (changes requested):
LGTM just conflicts need merging. Might be worth waiting until code freeze is over just so you don't have to keep resolving, I can ping when done (I know its annoying sorry)

joshmeranda:
@SamuZad right now rancher is in a code freeze for the 2.8 release. Once the freeze is over I can merge.

SamuZad:
Ah, I wasn't aware, thank you for the context 🙂

joshmeranda:
@SamuZad Can you resolve these merge confliicts? Should just need to rebase off of `dev-v2.8` and rerun `make charts && make validate`. Dev is unfrozen and we can merge once you're ready!

SamuZad:
> @SamuZad Can you resolve these merge confliicts? Should just need to rebase off of `dev-v2.8` and rerun `make charts && make validate`. Dev is unfrozen and we can merge once you're ready!

Thanks for letting me know! I rebased and ran the make commands, all the CI passed successfully 😁

joshmeranda (approved):

PR #3230, opened 2023-11-20
istio 1.18 on rancher 2.8

diogoasouza:


packages/rancher-istio/1.18/rancher-istio/charts/values.yaml
```diff
@@ -0,0 +1,119 @@
...
+    tag: 1.18.2-distroless
+  defaultPodDisruptionBudget:
+    enabled: true
+
+# Kiali subchart from rancher-kiali-server
+kiali:
+  enabled: true
+  # If you wish to change the authentication you can check the options in the Kiali documentation https://kiali.io/docs/configuration/authentication/
+  auth:
+    strategy: anonymous
```
rohitsuse, line 93: Are we changing this back to anonymous ?

release.yaml
lucasmlp, line 1: Do we need to release the versions following 102 versions in case users upgrade from Rancher 2.7?
If that's the case, shouldn't these charts also be tested by QA?

lucasmlp (approved):
LGTM 🚀

rohitsakala (changes requested):
Commented

packages/rancher-istio/1.18/rancher-istio/charts/values.yaml
diogoasouza: It's the same chart as the one in 2.7, and that one does not have any Auth strategy

rohitsakala (approved):

PR #3233, opened 2023-11-22
Release istio 103.0.0+up1.18.2

diogoasouza:
### Checkpoint 0: Validate `release.yaml`

Validation steps:
- [x] Each chart version in `release.yaml` DOES NOT modify an already released chart. If so, stop and modify the versions so that it releases a net-new chart.
- [x] Each chart version in `release.yaml` IS exactly 1 more patch version than the last released chart version. If not, stop and modify the versions so that it releases a net-new chart.

Validation steps:
- [x] Running `make unzip` to regenerate the `charts/` from scratch, then `git diff` to check differences between `assets/` and `charts/` yields NO differences or innocuous differences.

Validation steps:
  - kube-version annotation
  - rancher-version annotation


release.yaml
lucasmlp, line 1: Please remove other versions of Istio other than 103.0.0+up1.18.2

lucasmlp (changes requested):

lucasmlp (approved):

PR #3294, opened 2024-01-02
updating istio for 2.9 branch

diogoasouza:
## Issue: https://github.com/rancher/rancher/issues/43833


rohitsakala (changes requested):
Are we going to add 1.19 and 1.20 to dev-v2.9 ? 

Rest all, LGTM.

Also, if we can have multiple commits, it will be easy for reviewing. Something for next time.

rohitsakala (approved):

PR #3296, opened 2024-01-03
[v2.9] added cis 1.8 benchmarks and profiles

vardhaman22:
## Issue: 

https://github.com/rancher/rancher/issues/43899


vardhaman22:
closing this. based on offline discussion, Will forward port the chart to v2.9 directly. and use https://github.com/rancher/rancher/issues/43899 as a tracking issue for the fix.

PR #3297, opened 2024-01-03
[dev-v2.7] Bumping charts-build-scripts version to v0.5.3-rc.1

lucasmlp:
This PR solves the compatibility issues with the newly introduced build-scripts features by this [PR](https://github.com/rancher/charts-build-scripts/pull/118). In short, this modification adds the dependency version either on `Chart.yaml` or `requirements.yaml` for each package that has dependencies.

These modifications affected charts with dependencies. The following modifications were introduced:
- Patched all package versions to generate a new asset.
- Ran `make prepare`, modified `Chart.yaml`, `make clean`, and `make charts`.
  - rancher-alerting-drivers;
  - rancher-istio;
  - rancher-monitoring;
  - epinio;
  - fleet;
  - rancher-sriov and
  - prometheus-federator.


packages/epinio/epinio/generated-changes/patch/Chart.yaml.patch
rohitsuse, line 1: Why can't we fix the epinio chart in this way https://github.com/rancher/charts/pull/3300/files#diff-e7f85ead2fb290ad9627cc1125a43ee387761acc19bf28bb2a409be65a9b213f  instead of creating a new version of the chart ?

rohitsakala (changes requested):

PR #3319, opened 2024-01-09
Remove Epinio from v2.8.x marketplace

andreas-kupries:
## Issue:

Link to originating issue is to come.

## Problem

While Epinio is in a stable state providing it in the Marketplace implies resources to maintain and extend it.
The Epinio team however is refocused on other work in the Rancher system.

## Solution

Remove the Epinio assets from the marketplace of the Rancher lines in active development (2.8, 2.9).
Do not touch the EOLed lines of Rancher (2.7).

This PR is for the Rancher 2.8.x series.
#3320 is the equivalent PR for the 2.9.x series.
Epinio will not be removed from the 2.7.x series.

## Testing

As Epinio is removed no direct testing of the removed chart is possible nor sensible.
It should be testable that the chart is properly not accessible/visible any longer.

### Manual Testing

n/a

### Automated Testing

n/a

## QA Testing Considerations

n/a

n/a

## Backporting considerations

The removal only touches the lines of Rancher in active development (2.8, 2.9) and does not touch the EOL-ed lines (2.7, and earlier).


rohitsakala (changes requested):
1. Once a chart is released, we cannot delete it. So, only packages folder needs to be removed from this PR. Assets and Charts folder epinio files stay. 
2. If we want to follow the deprecation process, we need to deprecate it in 2.9 and then epinio will be deprecated in 2.11. That means , we need to have atleast one version of the epinio chart in 2.9 and 2.10.
3. A UI issue needs to be created to show deprecation banner on UI for epinio chart.

nicholasSUSE:
Hello @andreas-kupries 
Is this Pull Request still relevant?
Can I close it?

nicholasSUSE:
Unresponsive, closing.

andreas-kupries:
I actually escalated to @pgonin and @olblak and they started a conversation about this in slack.

PR #3325, opened 2024-01-10
Update rancher-backup chart annotation

eliyamlevy:
Backup Restore Rancher v2.9 annotation update.


diogoasouza (changes requested):
I don't think the .orig and .rej files are needed, can you remove it?

diogoasouza (approved):

PR #3329, opened 2024-01-12
[dev-v2.8.1] Monitoring air gap image pull / registry fix

alexandreLamarre:
## Issue: 

https://github.com/rancher/rancher/issues/44006

## Problem

The newer upstream version of kube-prometheus-stack `45.31.1` introduced its own opnionated way to handle image registries, which during the rebase we defaulted to using - it does not honor the cattle system registry settings.

## Solution

In the chart templates we now handle the default registry as the cattle system registry, because the upstream chart expects its image registry to be defined, and not to break existing conventions I added a new template to rancher-monitoring's `_helpers.tmpl` to handle rancher's registry logic and the upstream's chart's registry logic as follows:

Look up each service's (e.g. prometheus, alertmanager, grafana, etc...) image repository value (initially set to empty), if not set, look at cattle system's repository value, if not set then we default to the global image repository which by default is always set to `docker.io` 

If none of the above registries are set, we expect the image pull to fail.

## Testing

Install / upgrade to rancher 2.8 in an air-gapped setting, then install rancher monitoring : images will not pull from the air gapped registry, but will default to `docker.io`

### Manual Testing

Verified when helm templating that the cattle system registry's values are passed in correctly to the charts.

Verified non-air gapped installations are not affected.

Sanity check : installation in local rancher installation with local container registry honors its set container registries

### Automated Testing

N/A

## QA Testing Considerations

Air-gapped installation & scenarios for rancher-monitoring.

Air-gapped upgrade from 2.7 to 2.8

N/A

## Backporting considerations

N/A, but this fix should be forward fit to 2.9 charts


nflynt (approved):
I could nit pick about `temp_registry` as a variable name, but I'm not going to block based on that. This looks sane given what it needs to do, let's get it to QA for testing.

PR #3332, opened 2024-01-15
[dev-v2.8.1] Monitoring image registry fix for some nested chart templates

alexandreLamarre:
## Issue: 

https://github.com/rancher/rancher/issues/44006

## Problem

A couple of the Chart dependencies of the upstream prometheus operator also use specific registry handling that does not honor rancher's system registry, 

## Solution

Handle image registries similarly to how the base monitoring chart now handles them: 

-> default to cattle system registry for monitoring image registries, if not available then the global image registry, by default `docker.io`, otherwise each image's specific registry value

## Testing

Air gap installation of rancher-monitoring

### Manual Testing

Helm template with dependencies enabled.

Local testing with rancher deployment & private registry

### Automated Testing

N/A

## QA Testing Considerations

All rancher-monitoring scenarios in an air-gapped environment.

N/A

## Backporting considerations

N/A, should be forward fit to 2.9


packages/rancher-monitoring/rancher-monitoring/package.yaml
```diff
@@ -1,7 +1,7 @@
 url: https://github.com/prometheus-community/helm-charts.git
 subdirectory: charts/kube-prometheus-stack
 commit: beaa76a4ac26584c6cc77c27bc0d40769fb98b26
-version: 103.0.3
+version: 103.0.4
```
joshmeranda: We probably don't need to bump this version since the last versions had not been released yet.

joshmeranda (changes requested):

nflynt (approved):
Nothing's jumping out to me, this looks good :+1:

joshmeranda (approved):

PR #3340, opened 2024-01-16
Bump rancher webhook v0.5.0 rc1

JonCrowther:


JonCrowther:
Going to remake with the correct annotations from here https://github.com/rancher/webhook/pull/321

PR #3368, opened 2024-01-23
[dev-v2.8] Sync-up development branch after v2.8.1 release

lucasmlp:
Normally a _sync-up PR_ only contains changes in `release.yaml` and maybe on `regsync.yaml`. This is not the case because the monitoring chart was released from `dev-v2.8.1` branch. This _sync-up PR_ brings the changes made in `dev-v2.8.1` branch to `dev-v2.8`.

Some changes had to be made to the `rancher-monitoring` chart so that the asset matches the one in `release-v2.8`. Here's a brief explanation of why this change was necessary:
- `dev-v2.8.1` was branched out of `release-v2.8` instead of `dev-v2.8` and the fixes delivered in the `v2.8.1` charts release were made on `dev-v2.8.1`
- `release-v2.8` had `rancher-monitoring` chart at `103.0.0` and `dev-v2.8` at `103.0.2`
- When `dev-v2.8.1` was branched out of `release-v2.8` it was outdated if compared to `dev-v2.8.1`
- When I merged `release-v2.8` into `dev-v2.8` after the release, `make validate` started to break the pipeline. This happened because the `rancher-monitoring` packages folder had more changes due to the `103.0.1` and `103.0.2` versions of `rancher-monitoring` existed there before `103.0.1`
- Then, I had to patch the `chart/values.yaml`, `crd/values.yaml`, and `templates/alertmanager/secret.yaml` files so that the asset generated by `make charts` would be the same as the one in `release-v2.8`.


index.yaml
```diff
@@ -13113,6 +13113,134 @@ entries:
...
+      name: rkeScheduler
+      repository: file://./charts/rkeScheduler
+    - condition: global.cattle.windows.enabled
+      name: windowsExporter
+      repository: file://./charts/windowsExporter
+    description: kube-prometheus-stack collects Kubernetes manifests, Grafana dashboards,
+      and Prometheus rules combined with documentation and scripts to provide easy
+      to operate end-to-end Kubernetes cluster monitoring with Prometheus using the
+      Prometheus Operator.
+    digest: 7dd3c0a965bfffd5019bca923668ff0f9727272be3b4564771056e1d742b9594
```
rohitsuse: This digest looks is not same as in release-v2.8 and dev-v2.8.1 4ce289a2bc56327093b8c59a76994988d368f09120f6e347c8fcdac69e894e39.
 Please recheck.

PR #3370, opened 2024-01-23
Bump rancher webhook to v0.5.0-rc3

tomleb:


tomleb:
Will bump to rc4 since rc3 CI failed to publish all container images.

PR #3380, opened 2024-01-29
Bump rancher-csp-adapter to v2.0.3-rc1

mbelur:
## Problem:
Support for k8s 1.27

## Testing:

 Built the rancher image locally with the changes that support k8s 1.27 and installed this on EKS K8s 1.27 cluster
 Install the CSP adapter with these changes on an EKS Kubernetes (K8s) 1.27 cluster.
 Created a downstream EKS cluster with 21 nodes (with 1 entitlement) to trigger out-of-compliance message. When the downstream cluster was destroyed, validate that the out-of-compliance message is no longer displayed.
 Validate that the support config


MbolotSuse (changes requested):
Looks like there's a merge conflict that you will need to resolve (likely just rebase on dev-v2.7 and fix the conflict in release.yaml). Other than that, typically we separate these into two commits: One for changing release.yaml and package.yaml to use the new version, and one that just has the output of `make charts`. This isn't strictly necessary, but I think it would be preferable to follow the pattern if we can.

mbelur:
> Looks like there's a merge conflict that you will need to resolve (likely just rebase on dev-v2.7 and fix the conflict in release.yaml). Other than that, typically we separate these into two commits: One for changing release.yaml and package.yaml to use the new version, and one that just has the output of `make charts`. This isn't strictly necessary, but I think it would be preferable to follow the pattern if we can.

Fixed the merge conflict and followed the 2-commit pattern for the commits.

MbolotSuse (approved):

PR #3388, opened 2024-01-30
Fix toleration label inconsistent with the node label

vatsalparekh:
## Issue: 
 SURE-7505

## Problem

The toleration label is inconsistent with the node label which is `node-role.kubernetes.io/control-plane` compared to `node-role.kubernetes.io/controlplane` here in the charts
## Solution

Fix the label here

## QA Testing Considerations

Tests should validate the logging pods scheduled as per the set toleration


alexandreLamarre (approved):

MKlimuszka:
@vatsalparekh , thank you so much for this PR. Can you fill out the testing section so QA knows where to start?

vatsalparekh:
@MKlimuszka Done

PR #3366, opened 2024-01-22
Define a CODEOWNER file

recena:
This PR aims to start using `CODEOWNER` GitHub feature to simplify how we perform some engineering practices as code review.

⚠️ This PR does not affect to the core of this repository, the Helm Charts.

PR #3396, opened 2024-01-31
Define a CODEOWNER file

recena:
This PR is related to https://github.com/rancher/charts/pull/3366 but in this case against `dev-v2.8` branch.


rohitsakala (approved):
Contents are same as main PR.

PR #3404, opened 2024-02-01
Bump backup to 3.1.3 rc6 for dev v2.7.11

ericpromislow:
## Issue: 

Updates the Backup and Restore charts for PR [#401](https://github.com/rancher/backup-restore-operator/pull/401)

## Problem

When the `resourceNameRegexp` was set to `.`, `excludeResourceNameRegexp` specs were ignored, so more resources were being backed up and could cause problems if restored.

## Solution
This PR just bumps the RC version.

Should be the same as PR #3378 for dev-2.7


recena:
According to the `CODEOWNER` file, only @prachidamle is a valid [reviewer](https://github.com/rancher/charts/blob/dev-v2.7.11/.github/CODEOWNER#L29), and @rancher/mapps (as global reviewer).

@prachidamle Should we review the ownership?

prachidamle:
yes @recena A new team is being formed in github - they should be reviewing the Backup-restore chart updates. I will send update once EIO creates the team.

ericpromislow:
Wrong tag: we should be using v3.1.3rc5, not rc6

PR #3411, opened 2024-02-02
[v2.7.11] rancher-webhook k8s 1.27 support for 2.7.11

tomleb:
## Issue: https://github.com/rancher/rancher/issues/44066

Adding k8s 1.27 support to rancher-webhook for 2.7.11 and upgraded go to 1.20.

**Note:** I'm not sure if I missed something, wasn't sure if rancher-webhook needed to be added to release.yaml either..

## Testing

Go test + integration test via CI in the webhook repo

## Backporting considerations

Will need to port to 2.7 head as well afaik.


alexandreLamarre (approved):

tomleb:
Rebased due to conflict

release.yaml
```diff
@@ -43,3 +43,5 @@ system-upgrade-controller:
   - 102.2.0+up0.6.0
 harvester-csi-driver:
   - 102.0.2+up0.1.17
+rancher-webhook:
```
maxsokolovsky, line 46: I never properly understood when a chart entry may be in release.yaml, especially an RC...

tomleb: Honestly I'm not sure myself either, @MbolotSuse sorry to pull you into this but, do you know?

MbolotSuse: AFAIK, you need to add the value to release.yaml when you are adding a new version of the chart. Since you are adding a new version of the webhook chart, you need to add it here. My understanding is that for all intents and purposes you need to add/update this value every time you make a new webhook RC (or release a version of the webhook).

recena: @rancher/mapps Can you clarify it?

adamkpickering: Note that I am new to the MAPPS team, so I may have gotten some of the below wrong.

`release.yaml` is used by `make validate`. If a chart has been added/updated and is *not* in `release.yaml`, `make validate` will return an error. I believe that the purpose of this is to keep us from accidentally making changes to charts. `release.yaml` also tells us which charts have been added/updated since the last merge from `dev-v2.x` to `release-v2.x`. When we do one of these merges, we delete the contents of `release.yaml`.

tomleb: Gotcha, so the way I did here seems okay considering it's a new version of the webhook chart. Thanks everyone.

maxsokolovsky (approved):

PR #3413, opened 2024-02-02
[dev-v2.8] Sync-up dev branch after v2.8.1 release

diogoasouza:
Normally a _sync-up PR_ only contains changes in `release.yaml` and maybe on `regsync.yaml`. This is not the case because the monitoring chart was released from `dev-v2.8.1` branch. This _sync-up PR_ brings the changes made in `dev-v2.8.1` branch to `dev-v2.8`.

Some changes had to be made to the `rancher-monitoring` chart so that the asset matches the one in `release-v2.8`. Here's a brief explanation of why this change was necessary:
- `dev-v2.8.1` was branched out of `release-v2.8` instead of `dev-v2.8` and the fixes delivered in the `v2.8.1` charts release were made on `dev-v2.8.1`
- `release-v2.8` had `rancher-monitoring` chart at `103.0.0` and `dev-v2.8` at `103.0.2`
- When `dev-v2.8.1` was branched out of `release-v2.8` it was outdated if compared to `dev-v2.8.1`
- When I merged `release-v2.8` into `dev-v2.8` after the release, `make validate` started to break the pipeline. This happened because the `rancher-monitoring` packages folder had more changes due to the `103.0.1` and `103.0.2` versions of `rancher-monitoring` existed there before `103.0.1`
- The charts that have already been released cannot be modified, so we can't include the changes made in `103.0.1` and `103.0.2` into `103.0.3`


joshmeranda (approved):

index.yaml
```diff
@@ -13233,6 +13233,134 @@ entries:
...
+      catalog.cattle.io/rancher-version: '>= 2.8.0-0 < 2.9.0-0'
+      catalog.cattle.io/release-name: rancher-monitoring
+      catalog.cattle.io/requests-cpu: 4500m
+      catalog.cattle.io/requests-memory: 4000Mi
+      catalog.cattle.io/type: cluster-tool
+      catalog.cattle.io/ui-component: monitoring
+      catalog.cattle.io/upstream-version: 45.31.1
+    apiVersion: v2
+    appVersion: v0.65.1
+    created: "2024-02-02T16:39:37.22521-03:00"
```
rohitsuse: I think the `created` field should also match with what is in `release-v2.8`. When do you `git merge release-v2.8`, there will be merge conflicts in index.yaml. You copy the monitoring entry version `103.0.3+up45.31.1` part of the index from `release-v2.8` and paste it and then you copy the version `103.0.2+up45.31.1` in `dev-v2.8` as the second entry of `rancher-monitoring` in index.yaml. Then save the file and `git add`. This should solve the issue here. Let me know if it works.

The reason it should match is because later in the future when we create a PR from `dev-v2.8` to `release-v2.8`, we shouldn't get any conflicts. but now due to the difference in `created` field we will get conflicts. Correct me if I am wrong.

diogoasouza: You're right, it makes sense. I have changed the timestamp to match the one in the release like you said

lucasmlp (approved):
LGTM 🚀

packages/rancher-monitoring/rancher-monitoring/templates/crd-template/values.yaml
```diff
@@ -10,7 +10,7 @@ global:
 
 image:
   repository: rancher/shell
-  tag: v0.1.22
+  tag: v0.1.18
```
rohitsakala, line 13: we are downgrading the version. This doesn't look right.

rohitsakala (approved):

PR #3418, opened 2024-02-05
[dev-v2.7] Sync-up dev branch after OOB release

lucasmlp:
Bringing the January OOB release changes back to the dev branch so that the `release-v2.7` branch is in sync with `dev-v2.7`.


index.yaml
```diff
@@ -4158,6 +4158,37 @@ entries:
     - assets/longhorn-crd/longhorn-crd-1.0.200.tgz
     version: 1.0.200
   neuvector:
+  - annotations:
+      catalog.cattle.io/auto-install: neuvector-crd=match
+      catalog.cattle.io/certified: rancher
+      catalog.cattle.io/display-name: NeuVector
+      catalog.cattle.io/kube-version: '>=1.18.0-0 < 1.29.0-0'
```
nicholasSUSE: < 1.29 k8s version, is this right?

packages/neuvector/templates/crd-template/Chart.yaml
```diff
@@ -4,13 +4,13 @@ annotations:
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/hidden: true
 apiVersion: v1
-appVersion: 5.2.4
+appVersion: 5.2.4-s1
```
nicholasSUSE: -s1
Is this right?

lucasmlp: Yes, it is. doesn't look pretty, but it's the feature chart owners responsibility to till the correct `appVersion`

index.yaml
lucasmlp: It is. The chart supports k8s versions up to 1.28.

nicholasSUSE (approved):

PR #3439, opened 2024-02-07
Adding check step for packages/ folder modifications

nicholasSUSE:
## Problem
The rancher/charts repository has some [CI pipelines](https://github.com/rancher/charts/tree/dev-v2.9/.github/workflows) defined. At this moment, when someone pushes a commit, there is a checker (Validation Comment) to remind of some manual actions.

This checker should be enabled only if the pull request modifies the `/packages` folder.

## Solution

The "Check if /packages has been modified" step uses git diff to get a list of files that have been modified in the pull request. It then uses grep to check if any of the modified files are in the /packages directory. If they are, it sets an output variable modified to true.


.github/workflows/validation-comment.yaml
```diff
@@ -1,7 +1,7 @@
 name: Validation Comment
 
 on:
-  pull_request_target:
+  pull_request:
```
nicholasSUSE: I don't know why git did this change here on this commit. 
But this re-written to the right value: `pull_request_target` in the next commits.

nicholasSUSE:
Failed

PR #3440, opened 2024-02-07
Restrict when to show the validation suggestion

nicholasSUSE:
## Problem
The rancher/charts repository has some [CI pipelines](https://github.com/rancher/charts/tree/dev-v2.9/.github/workflows) defined. At this moment, when someone pushes a commit, there is a checker (Validation Comment) to remind of some manual actions.

This checker should be enabled only if the pull request modifies the `/packages` folder.

## Solution

The "Check if /packages has been modified" step lists all the modified files in all commits.
If any of these files has the prefix (`packages`) that is because we are editing `packages` folder. 

### Extra
I added documentation to support the development of github actions, a folder and 2 files needed for it. 
It is hard to test the github actions by doing several Pull Requests and seeing the results after they are merged.
The best way is to use the `act` project as stated in the documentation.


nicholasSUSE:
Failed

PR #3441, opened 2024-02-07
Restrict when to show the validation suggestion

nicholasSUSE:
## Problem
The `rancher/charts` repository has some [CI pipelines](https://github.com/rancher/charts/tree/dev-v2.9/.github/workflows) defined. At this moment, when someone pushes a commit, there is a checker (Validation Comment) to remind of some manual actions.

This checker should be enabled only if the pull request modifies the `/packages`, `/assets` and `/charts` folder.

## Solution

The "Check if /packages has been modified" step lists all the modified files in all commits.
If any of these files has the prefix with the names of the targeted folders, that is because we are editing  `/packages`, `/assets` or `/charts`  folder. 

## Manual testing

I tested with simulated events using an `events.json` file and **act** package that is not defined in this PR.
These are the values that I tested:

`events.json for this PR`: 
- this should not create an automatic comment) (OK)
```json
{
    "pull_request": {
        "number": 3441,
        "head": {
            "ref": "cicd-restriction-validation-suggestion",
            "sha": "ac670b061d60e31fb56757f3f0d594588d667c5a"
        }
    },
    "repository": {
        "owner": {
            "login": "nicholasSUSE"
        },
        "name": "charts"
    }
}
```

`events.json for the following PR`: https://github.com/rancher/charts/pull/3453
- This should create an automatic comment (OK)
- The comment was deleted after the test was concluded.
```json
{
    "pull_request": {
        "number": 3453,
        "head": {
            "ref": "emergency-fleet-bump-for-qa",
            "sha": "4d535ff39364e8e964fa9a067aae26e4ccf84cb7"
        }
    },
    "repository": {
        "owner": {
            "login": "nicholasSUSE"
        },
        "name": "charts"
    }
}

### DIFF
--- a/.github/workflows/validation-comment.yaml
+++ b/.github/workflows/validation-comment.yaml
@@ -5,6 +5,10 @@ on:
     branches:
       - dev-v*
       - release-v*
+    paths:
+      - 'assets/**'
+      - 'charts/**'
+      - 'packages/**'
 
 jobs:
   validation-comment:
@@ -25,7 +29,7 @@ jobs:
               `## Validation steps
               - Ensure all container images have repository and tag on the same level to ensure that all container images are included in rancher-images.txt which are used by airgap customers.
               <pre>
-              Ex:- 
+              Ex:-
                 longhorn-controller:
                   repository: rancher/hardened-sriov-cni
                   tag: v2.6.3-build20230913


recena:
@nicholasSUSE Did you consider the use of [this](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-and-excluding-paths)?

adamkpickering:
Wouldn't it be better to do this check as a part of `charts-build-scripts`? We could add a command to `charts-build-scripts`, and add a step in the workflow that calls it. That way, we don't require `act` (or a lengthy dev cycle using github) to run this check locally.

lucasmlp:
@nicholasSUSE I think that changes in the `charts` and `assets` should also trigger this pipeline. I agree with @recena that including those three paths (`packages, `charts` and `assets`) is the cleanest way to do this

nicholasSUSE:
## Validation steps
  - Ensure all container images have repository and tag on the same level to ensure that all container images are included in rancher-images.txt which are used by airgap customers.
  <pre>
  Ex:-
    longhorn-controller:
      repository: rancher/hardened-sriov-cni
      tag: v2.6.3-build20230913
  </pre>
  - Add a 👍 (thumbs up) reaction to this comment once done. CI won't pass without this reaction to the github-action bot's latest validation comment.
  - Approve the PR to run the CI check.

nicholasSUSE:
## Validation steps
  - Ensure all container images have repository and tag on the same level to ensure that all container images are included in rancher-images.txt which are used by airgap customers.
  <pre>
  Ex:-
    longhorn-controller:
      repository: rancher/hardened-sriov-cni
      tag: v2.6.3-build20230913
  </pre>
  - Add a 👍 (thumbs up) reaction to this comment once done. CI won't pass without this reaction to the github-action bot's latest validation comment.
  - Approve the PR to run the CI check.

nicholasSUSE:
> @nicholasSUSE Did you consider the use of [this](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-and-excluding-paths)?

Hello @recena 

From the documentation you provided:

> When using the `push` and `pull_request` events, you can configure a workflow to run based on what file paths are changed. Path filters are not evaluated for pushes of tags.

We are currently using `pull_request_target`. 
The reason for using this type of event is documented on my other PR here: https://github.com/rancher/charts/pull/3444/commits/65dc3a082e687b4ad24e3fd96f80aa2b09f95466

> Both are events that can trigger a workflow whenever a pull request is opened, synchronized, or closed, but:
      - `pull_request`: If the PR is from a fork, it will run with the fork permissions, this means it will have **read-only permissions to the base repository and no acess to secrets**.
      - `pull_request_target`: Runs with the permissions of the base repository even if the PR comes from a forked repository. Since it has write-permissions and access to secrets, this is useful for workflows that need to write or change in some way the Pull Requests.


I tested this and it did not work, It is better to proceed with my approach however I must update the code for checking the 3 directories instead of 1.

recena:
It is curious, we use `pull_request_target` in [validation-comment.yaml](https://github.com/rancher/charts/blob/dev-v2.9/.github/workflows/validation-comment.yaml) and `pull_request` in [pull-request.yaml](https://github.com/rancher/charts/blob/dev-v2.9/.github/workflows/pull-request.yaml).

nicholasSUSE:
> It is curious, we use `pull_request_target` in [validation-comment.yaml](https://github.com/rancher/charts/blob/dev-v2.9/.github/workflows/validation-comment.yaml) and `pull_request` in [pull-request.yaml](https://github.com/rancher/charts/blob/dev-v2.9/.github/workflows/pull-request.yaml).

`pull_request` event type does not have permission to create a comment in the **PR** when the Pull Request comes from a fork. 

I remember @lucasmlp implemented this so he might be able to explain this better.

nicholasSUSE:
> Wouldn't it be better to do this check as a part of `charts-build-scripts`? We could add a command to `charts-build-scripts`, and add a step in the workflow that calls it. That way, we don't require `act` (or a lengthy dev cycle using github) to run this check locally.

Hey Adam normally I would agree but we will have to do more workflows so I think it is good we start having the right development workflow now. 

Charts build scripts is a good option but I think this goes beyond it's scope for this issue.

adamkpickering:
I'm just concerned that once this goes in, it sets a precedent. Once this is in, it becomes easier to add other validations that require `act` etc, so they get added. And in the end we are left with a choice between:
- making a big investment to improve the developer experience by rewriting automations so they are in `charts-build-scripts`, which doesn't happen in my experience, or
- just continuing with a development workflow the way it is, even if it makes our lives more difficult than they have to be.

We haven't discussed this as a team, nor do we have a coherent idea of where we want our development workflow to go (at least, not that I'm aware of). Acting without that in place is a recipe for chaos. @recena?

Or maybe I'm making a bigger deal out of this than it is. It's just one PR right?

nicholasSUSE:
> I'm just concerned that once this goes in, it sets a precedent. Once this is in, it becomes easier to add other validations that require `act` etc, so they get added. And in the end we are left with a choice between:
> 
> * making a big investment to improve the developer experience by rewriting automations so they are in `charts-build-scripts`, which doesn't happen in my experience, or
> * just continuing with a development workflow the way it is, even if it makes our lives more difficult than they have to be.
> 
> We haven't discussed this as a team, nor do we have a coherent idea of where we want our development workflow to go (at least, not that I'm aware of). Acting without that in place is a recipe for chaos. @recena?
> 
> Or maybe I'm making a bigger deal out of this than it is. It's just one PR right?

I like your point but let me add some stuff to it. 

- **act** is widely used as far as I know as a helping tool that is not that difficult, it was not necessary for getting this job done but helped a lot for testing it, I could still achieve the same goal but with a lot more work.  
- Github Actions is also widely used for the purpose of pipeline automation, so I think we should use this tool since it was built for it.
- **charts-build-script** (I think) that should be used only for building, updating etc... (everything that is related to the building charts process).


 but as you said, we have not discussed this as a team. 
 I am open for us to have that discussion, my main point is, that let's use the right tools for the right jobs. 
 
 I agree that doing things in **charts-build-scripts** could be easier especially because it is in Golang, although I have no idea how we would integrate it to make Pull Requests comments in the Git UI (custom http requests or is there a specific library for that?).

adamkpickering:
I have more thoughts, but I think it's better if we wait to discuss this until we can do so as a team :smile:

nicholasSUSE:
After a meeting with @lucasmlp he showed me that the `paths` filter work. 
My mistake, reverting the commit to include it.

lucasmlp:
> @nicholasSUSE Did you consider the use of [this](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-and-excluding-paths)?

@recena I've tested this approach with @nicholasSUSE and it works. When using `pull_request_target` the changes *must* in the PR base branch before testing. So we tested this in my forked repo and it worked.

lucasmlp (approved):
LGTM 🚀

recena:
This approach is very simple!

PR #3442, opened 2024-02-08
Update go version and go mod dependencies

rohitsakala:
There are a lot of CVEs in the Rancher final container image due to this repository being cloned into the Rancher container image. So, in this PR, I am updating the go modules so that the count is brought back to zero.

- Before this PR, the Rancher 2.8.0 image had **23** CVEs from the `tests` package : 
[CVE_Report_Rancher_2_8_0.txt](https://github.com/rancher/charts/files/14206488/CVE_Report_Rancher_2_8_0.txt)

- After this PR, the next Rancher 2.8.x image will have **zero** CVEs from the `tests` package : 
[CVE_Report_Rancher_2_8_x.txt](https://github.com/rancher/charts/files/14206493/CVE_Report_Rancher_2_8_x.txt)


recena:
@rohitsakala It would be great if we add a before / after report of CVEs. What about applying these changes to the rest of actives branches?

rohitsakala:
> @rohitsakala It would be great if we add a before / after report of CVEs. What about applying these changes to the rest of actives branches?

Done. Added the reports and made a PR for dev-v2.7. For dev-v2.9 it is not necessary, since it is already upto date with ZERO CVE from `tests`

nicholasSUSE (approved):

pjbgf (approved):

PR #3444, opened 2024-02-08
Github Actions Workflow Documentation Improvement

nicholasSUSE:
## Problem
It is a considerable effort to use `act` locally, it has some edge cases and some non-trivial configuration processes. 

## Solution
Added documentation and some support files to help all that may need to work with Github Actions Workflow. 
This is the best way to test things locally speeding up the development time.


lucasmlp (changes requested):
Looks good to me
I would ask you to add a simple disclaimer. Our workflows runs in Ubuntu machines while running in GitHub Actions, but the user might be using `act` in a different OS. This may cause different results. As a final test, the engineer should always use GitHub Actions to validate the results.

PR #3471, opened 2024-02-15
Wrong CODEOWNERS syntax

recena:
This is a follow-up of https://github.com/rancher/charts/pull/3470


recena:
The error in the CODEOWNERS file is expected:

```
Unknown owner on line 8: make sure @andreas-kupries exists and has write access to the repository
```

We are re-org some GH teams.

rohitsakala (approved):

PR #3474, opened 2024-02-16
Add NeuVector chart version 2.7.3

selvamt94:
## Problem
NeuVector has released app version 5.3.0 along with helm chart 2.7.3, which need to be added to Rancher feature charts version 2.7 (release-v2.7). We would like to include this chart in Rancher Manager release v2.7.11 otherwise this need to be released as OOB.

## Solution
New Chart release
### Manual Testing
Tested fresh install of the chart 102.0.8+up2.7.3 and l upgrade from previous version on RKE and RKE2 cluster.
## QA Testing Considerations
Tested fresh install of the chart 102.0.8+up2.7.3 and l upgrade from previous version on RKE and RKE2 cluster.


index.yaml
```diff
@@ -4274,6 +4274,37 @@ entries:
     - assets/longhorn-crd/longhorn-crd-1.0.200.tgz
     version: 1.0.200
   neuvector:
+  - annotations:
+      catalog.cattle.io/auto-install: neuvector-crd=match
+      catalog.cattle.io/certified: rancher
+      catalog.cattle.io/display-name: NeuVector
+      catalog.cattle.io/kube-version: '>=1.18.0-0 < 1.29.0-0'
```
nicholasSUSE, line 4281: I don't know much about neuvector so forgive me if I am wrong. 

According to the patterns we use on Mapps and that I see in other charts on Rancher 2.7 version I think we only support until 1.27 kubernetes version.

nicholasSUSE (changes requested):

index.yaml
selvamt94: @nicholasSUSE NeuVector supports k8s 1.28 so updated the annotation. I tested the chart on downstream cluster with k8s 1.28.6

lucasmlp: @nicholasSUSE This chart might be used outside Rancher Manager. We offer all the charts via [charts.rancher.io](charts.rancher.io) so users can use them via `helm CLI`. It's up to the feature chart teams to say in which k8s versions the charts might be installed

lucasmlp (approved):
LGTM 🚀

nicholasSUSE (approved):

PR #3476, opened 2024-02-16
[dev-v2.7.11] unrc CIS benchmark chart

vardhaman22:
## Issue: 

https://github.com/rancher/rancher/issues/44106


pdellamore (approved):

andypitcher (dismissed):
@vardhaman22 I'm not seeing any references to CIS-1.8 profiles, is this normal ?

vardhaman22:
> @vardhaman22 I'm not seeing any references to CIS-1.8 profiles, is this normal ?

for 2.7.11 release we have not introduced new changes and kept the chart same as [cis-5.0.0](https://github.com/rancher/charts/blob/dev-v2.8/charts/rancher-cis-benchmark/5.0.0/templates) that's why there are no 1.8 profiles. It will be introduced in the newer version.

PR #3487, opened 2024-02-16
[release-v2.9] Restrict when to show the validation suggestion 

nicholasSUSE:
This was just a `cherry-pick` from Pull Request: https://github.com/rancher/charts/pull/3441

- Realized an extra step was needed since we can't use the `paths` filter on `pull_request_review` events. 
- Added extra logic using JavaScript to check the modified files and halt job execution.

### DIFF
--- a/.github/workflows/validation-check.yaml
+++ b/.github/workflows/validation-check.yaml
@@ -10,7 +10,26 @@ jobs:
     if: startsWith(github.event.pull_request.base.ref, 'dev-v') || startsWith(github.event.pull_request.base.ref, 'release-v')
     runs-on: ubuntu-latest
     steps:
+      - name: Check for modifications in specified folders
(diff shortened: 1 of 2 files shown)


nicholasSUSE:
canceling release-v2.9 because of the build lacking maintenance.

recena (approved):

lucasmlp (changes requested):
We should add the same filters in the Validation Check workflow. Otherwise it'll block PRs because the CI would be checking for a comment that doesn't exist, like this:
https://github.com/rancher/charts/pull/3492#issuecomment-1956777009

lucasmlp (commented):
Testing filter with modification outside specified folders

lucasmlp (approved):
Great work @nicholasSUSE 
Thank you 🚀

PR #3491, opened 2024-02-20
[dev-v2.8] Add Elemental chart v1.4.2

fgiudici:
## Issue: 

Bump Elemental version in Rancher Marketplace: https://github.com/rancher/elemental/issues/1221

Related PR to load the updated images:  https://github.com/rancher/image-mirror/pull/579


charts/elemental/103.1.0+up1.4.2/values.yaml
```diff
@@ -0,0 +1,42 @@
...
+  tag: "1.4.2"
+  imagePullPolicy: IfNotPresent
+
+seedImage:
+  repository: "rancher/mirrored-elemental-seedimage-builder"
+  tag: "1.4.2"
+  imagePullPolicy: IfNotPresent
+
+channel:
+  repository: "registry.suse.com/rancher/elemental-channel"
```
fgiudici: The channel repository is a bit tricky: I know that in order to support Rancher airgap we need to allow the images to be prepended with the systemDefaultRegistry.
Anyway, this would not work for the Elemental channel image, which contains just a list of URL where OS images could be downloaded (the OS images have full URLs).
So, mirroring this image, would not work in airgap as the links in the image content will still refer to outside repositories.

The idea is that the channel image should carry the links of the OS available, and the image should be distributed by the OS provider. The user can freely change the image with something else here.
For Elemental airgap we have instructions to help generate a custom channel image to be passed during installation.
So, long story short: we need a static link here for the repo, and that should belong to the registry from which the OS images will be pulled.
This also explains why this one of the very few values we added in the questions.yaml file.

charts/elemental/103.1.0+up1.4.2/values.yaml
```diff
@@ -0,0 +1,43 @@
...
+  tag: "1.4.2"
+  imagePullPolicy: IfNotPresent
+
+seedImage:
+  repository: "rancher/mirrored-elemental-seedimage-builder"
+  tag: "1.4.2"
+  imagePullPolicy: IfNotPresent
+
+channel:
+  image: "registry.suse.com/rancher/elemental-channel"
```
fgiudici, line 13: note: changed the variable from "repository" to "image" to avoid Rancher scripts to save the image for the airgap

davidcassany (approved):
Looks good :+1:

PR #3492, opened 2024-02-20
Define a CODEOWNER file for rancher/charts repository

recena:
This change represents much better what we are doing at @rancher/mapps.

1. `/assets`
2. `/charts`
3. `index.yaml`

About paths are excluded because the content there is auto-generated.


diogoasouza:
A couple of few things:

- Who's the owner of `tests/common`?
- I feel like `assets` and `charts` should have the same codeowner as their respective `packages/*` folder. Even though it's auto generated, someone could run a wrong script and change something there that is not owned by them. I don't see a scenario where a change is made to `packages/foo` and changes are generated in any folder other than `assets/foo` and `charts/foo`

recena:
@diogoasouza 

> Who's the owner of tests/common?

We must discuss it.

> I feel like assets and charts should have the same codeowner as their respective packages/* folder. Even though it's auto generated, someone could run a wrong script and change something there that is not owned by them. I don't see a scenario where a change is made to packages/foo and changes are generated in any folder other than assets/foo and charts/foo

It is a good idea, but soon, those folders will only be able to be modified by a service account (_aka_ bot).

.github/CODEOWNERS
```diff
@@ -1,5 +1,13 @@
...
-* @rancher/mapps
+docs @rancher/mapps
+scripts @rancher/mapps
+CNAME @rancher/mapps
+Makefile @rancher/mapps
+README.md @rancher/mapps
+_config.yml @rancher/mapps
+configuration.yaml @rancher/mapps
+regsync.yaml @rancher/mapps
+index.html
```
lucasmlp: This also should be reviewed by the Mapps team

lucasmlp (changes requested):

.github/CODEOWNERS
recena: @lucasmlp Got catch!

lucasmlp (approved):
LGTM 🚀

lucasmlp:
@recena The check *Validation_ Check / Check for positive reaction on bot's latest validation comment (pull_request_review)* is failing because we need to apply the same changes we did in [this PR](https://github.com/rancher/charts/pull/3441) on the Validation Check workflow.

diogoasouza (approved):

PR #3497, opened 2024-02-21
[dev-v2.8] adding paths filtering code to validation-check 

nicholasSUSE:
Propagation from: https://github.com/rancher/charts/pull/3514

## Problem

Our previous solution created another problem for the `validation-check workflow`. Sometimes the validation was required but the previous `validation-comment workflow` was not run because of the filters.

## Solution

Add the same filters for (`assets/`, `packages/`, `charts/`) directories to `validation-check workflow`. This workflow uses a different type of `github event` so the `paths` directive was not enabled for use. The solution implements the same functionality with `javascript` injected code in the workflow. 

## Updated Scope of PR

As discussed below we updated:
- `actions/github-script@v4` → `actions/github-script@v7`
- `actions/checkout@v3` → `actions/checkout@v4`

The main reason:  Warnings about the deprecation of old Node.js versions being used since Github Actions is now using an updated version. 

### Manual Testing

All changes were manually tested by:

1. On Forked Repository merging a branch with the changes
2. Creating a new `release-v-test` branch from the recently created one. 
3. Creating a Pull Request with changes (branch: `cicd-test`) that will trigger the workflows and ensure it is working. 

One of the closed Pull Requests in the forked repository used for testing: https://github.com/nicholasSUSE/charts/pull/14

#### Some needed changes after the update:

- https://github.com/marketplace/actions/github-script
- https://octokit.github.io/rest.js/v20

Javascript code like the following example: 
```
            const files = await github.pulls.listFiles({
```

Was updated to:
```
            const files = await github.rest.pulls.listFiles({
```

Because of changes in the `octokit` stated above in the bullet points.


.github/workflows/validation-check.yaml
```diff
@@ -26,7 +45,7 @@ jobs:
             const sortedComments = comments.data.sort((a, b) => new Date(b.created_at) - new Date(a.created_at));
 
             // Find the latest validation comment by github-actions[bot]
-            const latestValidationComment = sortedComments.find(comment => comment.user.login === 'github-actions[bot]' && comment.body.startsWith("## Validation steps"));
+            const latestValidationComment = sortedComments.find(comment => comment.user.login === 'github-actions[bot]' && comment.body.sta>
```
adamkpickering: This change looks like a typo?

.github/workflows/validation-check.yaml
```diff
@@ -47,4 +66,4 @@ jobs:
               }
             } else {
               core.setFailed("No validation comments by github-actions[bot] found.");
-            }
+            }
```
adamkpickering: Nit: I imagine we want a newline at the end of each file?

adamkpickering (changes requested):

.github/workflows/validation-check.yaml
nicholasSUSE: This makes no difference. 
This is the result of auto-format from VSCode github actions official extension.

adamkpickering (approved):

PR #3498, opened 2024-02-21
[dev-v2.7.11] adding paths filtering code to validation-check

nicholasSUSE:
Following this PR and improving the solution: https://github.com/rancher/charts/pull/3484

## Problem
Our previous solution created another problem for the `validation-check workflow`. 
Sometimes the validation was required but the previous `validation-comment workflow` was not run because of the filters.

## Solution
Add the same filters for (`assets/ packages/ charts/`) directories to `validation-check workflow`. 
This workflow uses a different type of `github event` so the `paths` directive was not enabled for use. 
The solution implements the same functionality with `javascript` injected code in the workflow.


adamkpickering (approved):

nicholasSUSE:
We are syncing this branch to dev-v2.7 and discontinuing this branch

PR #3500, opened 2024-02-22
[release-v2.8] adding paths filtering code to validation-check

nicholasSUSE:
Propagation from: https://github.com/rancher/charts/pull/3514

## Problem

Our previous solution created another problem for the `validation-check workflow`. Sometimes the validation was required but the previous `validation-comment workflow` was not run because of the filters.

## Solution

Add the same filters for (`assets/`, `packages/`, `charts/`) directories to `validation-check workflow`. This workflow uses a different type of `github event` so the `paths` directive was not enabled for use. The solution implements the same functionality with `javascript` injected code in the workflow. 

## Updated Scope of PR

As discussed below we updated:
- `actions/github-script@v4` → `actions/github-script@v7`
- `actions/checkout@v3` → `actions/checkout@v4`

The main reason:  Warnings about the deprecation of old Node.js versions being used since Github Actions is now using an updated version. 

### Manual Testing

All changes were manually tested by:

1. On Forked Repository merging a branch with the changes
2. Creating a new `release-v-test` branch from the recently created one. 
3. Creating a Pull Request with changes (branch: `cicd-test`) that will trigger the workflows and ensure it is working. 

One of the closed Pull Requests in the forked repository used for testing: https://github.com/nicholasSUSE/charts/pull/14

#### Some needed changes after the update:

- https://github.com/marketplace/actions/github-script
- https://octokit.github.io/rest.js/v20

Javascript code like the following example: 
```
            const files = await github.pulls.listFiles({
```

Was updated to:
```
            const files = await github.rest.pulls.listFiles({
```

Because of changes in the `octokit` stated above in the bullet points.


.github/workflows/validation-check.yaml
```diff
@@ -47,4 +66,4 @@ jobs:
               }
             } else {
               core.setFailed("No validation comments by github-actions[bot] found.");
-            }
+            }
```
adamkpickering, line 69: Nit: do we want a newline here?

adamkpickering: Nit: do we want a newline at the end of this file?

adamkpickering (changes requested):

.github/workflows/validation-check.yaml
nicholasSUSE: I don't know if there is a convention for this. I know it does not break, we can discuss it.

adamkpickering: On previous teams this has been something that others mentioned on my PRs, which is what caused me to mention it. But I realized I didn't understand why people want this, so I did a little searching and found this: https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

TLDR it is a unix thing, and has to do with certain utilities not working as expected in some cases if the newline is not present.

I don't think it's likely that we will run into problems because of this, but since we're here, we might as well fix it.

nicholasSUSE: Ok, I was updating the other propagation PRs before I read this message. 
I will make this change on the other PRs as well.

adamkpickering (approved):

PR #3516, opened 2024-02-26
[dev-v2.9] Disabled hull tests

vardhaman22:
## Issue: 

due to helm bump to v3.14.1 there is an issue in rancher/hull repository.
## Problem

there is an issue with template parsing in rancher/hull because of bumping helm to v3.14.1 
## Solution

disabling psp tests for now so that we can bump helm in rancher/charts and then we can re-enable them when hull is fixed.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -65,17 +65,17 @@ jobs:
         run: make check-rc
         if: startsWith(github.ref, 'refs/heads/release-v')
 
-  hull-tests:
-    name: Run Hull tests
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout base branch
-        uses: actions/checkout@v4
+  # hull-tests:
+  #   name: Run Hull tests
+  #   runs-on: ubuntu-latest
+  #   steps:
+  #     - name: Checkout base branch
+  #       uses: actions/checkout@v4
 
-      - name: Checkout PR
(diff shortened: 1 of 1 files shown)


lucasmlp (changes requested):
I don't think we need to change this for now. The `go.mod` file was not modified so it's using a set of _helm_ and _hull_ dependencies that works.
**All workflows are running the hull tests normally**. 
I think this change would be needed if we upgraded _helm_ and _hull_ dependencies.

adamkpickering:
@lucasmlp agreed. Can we close this then?

lucasmlp:
@adamkpickering  I was looking into the issues related to this one. Seems like [Dependabot tried to update our helm library to v3.14.1](https://github.com/rancher/charts/pull/3472), but the hull tests broke because the hull version capable of working with helm v3.14.1 is broken. I believe that there will be a version of hull that works with helm v.3.14 once [rancher/hull#43](https://github.com/rancher/hull/pull/43) is merged as described in the hull PR by @vardhaman22.
We could wait for a week to see if [rancher/hull#43](https://github.com/rancher/hull/pull/43) is merged and if not, close this and the respective rancher/charts PRs

adamkpickering:
IMO this PR is unnecessary. Unless I'm missing something, we don't need to disable the hull tests until helm is bumped, because nothing is broken.

Closing for now. If I'm missing something @vardhaman22 @lucasmlp please reopen.

PR #3519, opened 2024-02-26
[dev-v2.8] Rancher Alerting kubernetes 1.28 support

alexandreLamarre:
## Issue: 

https://github.com/rancher/rancher/issues/43738

## Problem

Rancher alerting drivers charts are not usable for kubernetes 1.28

## Testing

See related issue

### Manual Testing

helm template & helm install

### Automated Testing

N/A


release.yaml
```diff
@@ -4,6 +4,10 @@ fleet-agent:
   - 103.1.1+up0.9.1-rc.5
 fleet-crd:
   - 103.1.1+up0.9.1-rc.5
+harvester-cloud-provider:
+  - 103.0.1+up0.2.3
+harvester-csi-driver:
+  - 103.0.1+up0.1.17
```
joshmeranda: This PR shouldn't be introducing these charts, you can probably remove them unless the validation fails without

release.yaml
```diff
@@ -28,12 +34,12 @@ rancher-csp-adapter:
   - 103.0.1+up3.0.1-rc2
 rancher-eks-operator:
   - 103.0.1+up1.3.0
-  - 103.1.0+up1.3.1-rc4
   - 103.1.0+up1.3.1-rc1
+  - 103.1.0+up1.3.1-rc4
```
joshmeranda: same as above

release.yaml
```diff
@@ -28,12 +34,12 @@ rancher-csp-adapter:
...
 rancher-eks-operator-crd:
   - 103.0.1+up1.3.0
-  - 103.1.0+up1.3.1-rc4
   - 103.1.0+up1.3.1-rc1
+  - 103.1.0+up1.3.1-rc4
```
joshmeranda: same as above

joshmeranda (changes requested):

release.yaml
alexandreLamarre: removed

release.yaml
alexandreLamarre: removed

release.yaml
alexandreLamarre: Looks like someone may have been pushing directly to dev-v2.8? because these versions have been mysteriously appearing and disappearing for me in some of my branches when I rebase. I appended them back to the end of the file since they are now in the upstream dev-v2.8

joshmeranda (approved):

PR #3532, opened 2024-02-28
Remove rancher-wins dependency from Rancher monitoring chart

HarrisonWAffel:
## Note
This PR is being raised as a **draft** to gather initial impressions and feedback on the changes. Additional work in https://github.com/rancher/windows_exporter-package/pull/8 is required before this PR is truly ready for review. Additionally the host process container reference must be added into Rancher, as it is a net-new container image. 

## Problem
Currently, The Rancher monitoring sub-chart `rancher-windows-exporter` relies on `rancher-wins` in order to deploy the required node exporter binary. This sub-chart relies on `rancher-wins` as the [node exporter binary](https://github.com/prometheus-community/windows_exporter) needs access to the [Windows Management Instrumentation (WMI) API](https://learn.microsoft.com/en-us/windows/win32/wmisdk/wmi-start-page), and exposing host level APIs to a Windows container has historically not been possible. `rancher-wins` allows the sub-chart to deploy the binary as a host level process and proxy its port to other containers.

Unfortunately,  the `rancher-wins` dependency limits the environments that rancher-monitoring can successfully be installed on. For instance, Rancher monitoring will not collect metrics for Windows nodes in the following environments, as `rancher-wins` is not installed on them by default: 
+ GKE 
+ EKS 
+ AKS
+ Imported Clusters
+ Nodes joined to the cluster the Rancher server is running on 
   + though, afaik, this is an unsupported configuration

## Solution
In Kubernetes v1.22 [Windows HostProcess pods were introduced](https://kubernetes.io/docs/tasks/configure-pod-container/create-hostprocess-pod/), and the feature went GA in v1.26. The HostProcess security context option allows workloads to be run directly on the host, allowing access to the WMI API without the use of `rancher-wins` and its proxies. By relying on a native Kubernetes feature, we can ensure that the sub-chart will properly deploy in any Kubernetes environment which supports HostProcess pods. 

Additional benefits of removing the dependency on `rancher-wins` is that the sub-chart becomes much less complicated, and we will no longer use any images with Foreign Layer references (which relates to https://github.com/rancher/windows/issues/222)

## **A note for discussion**
HostProcess containers on Windows entered Alpha in 1.22, but from what I can see this chart supports k8s versions as [old as 1.16](https://github.com/rancher/charts/blob/420f61bf4ec91d6f7e4e9d0a4f40e0e312c5556e/charts/rancher-monitoring/104.0.2%2Bup45.31.1/Chart.yaml#L13C1-L14C1). In my opinion, we should change the minimum required version to at least 1.22. Partly because that is the minimum version needed for Windows HostProcess containers, and because I don't think QA tests this chart against all k8s versions we claim to support

### Manual Testing
I have deployed the updated monitoring chart to custom RKE2 windows clusters and AKS clusters with windows node pools. In both environments these changes have worked as expected, and I did not see any errors or significant differences when upgrading or downgrading between this new chart version and already released versions. 

### Automated Testing
n/a

## QA Testing Considerations
While I was able to test on AKS clusters, I want to work closely with QA on getting this validated on EKS clusters as well. 

Because we are running the node exporter in a new environment (HostProcess pod) we should be extra careful during testing to ensure that no existing metrics change in any way. During my testing I did not see any difference in how metrics were reported between this new version and already released versions, but a double check would be good. 

## Backporting considerations
Ideally this change would be back ported to all release lines. I'm not aware of any significant differences within the windows exporter sub-chart between each releases line of Rancher which would overly complicate the back porting effort


packages/rancher-monitoring/rancher-windows-exporter/charts/values.yaml
```diff
@@ -19,15 +18,22 @@ serviceMonitor:
...
 ## Components scraping metrics from Windows nodes
 ##
 clients:
   enabled: true
 
+  maxRequests: 5
   port: 9796
   image:
     repository: rancher/windows_exporter-package
-    tag: v0.0.3
```
HarrisonWAffel, line 30: Note: This image isn't final, we will likely cut `v0.1.0` or directly use the upstream image - tbd

packages/rancher-monitoring/rancher-monitoring/package.yaml
```diff
@@ -1,7 +1,7 @@
 url: https://github.com/prometheus-community/helm-charts.git
 subdirectory: charts/kube-prometheus-stack
 commit: beaa76a4ac26584c6cc77c27bc0d40769fb98b26
-version: 104.0.2
```
HarrisonWAffel, line 4: Note: This version isn't final, it's been added for testing and can be changed based off of any other monitoring work I may not be aware of

packages/rancher-monitoring/rancher-windows-exporter/charts/templates/daemonset.yaml
```diff
@@ -13,65 +12,50 @@ spec:
...
+              mountPath: /scripts
+        - name: configure-firewall-rule
+          securityContext: {{ toYaml $.Values.securityContext | nindent 12 }}
+          image: {{ template "system_default_registry" . }}mcr.microsoft.com/oss/kubernetes/windows-host-process-containers-base-image:v1.0.0
+          command:
+            - C:\WINDOWS\System32\WindowsPowerShell\v1.0\powershell.exe
+          args: ["-f", "scripts/configure-firewall.ps1"]
+          volumeMounts:
+            -  name: exporter-scripts
+               mountPath: /scripts
```
HarrisonWAffel, line 37: Note: mount paths and command arguments do not use absolute paths as hosted providers may not mount host process containers in `C:\`. For example, aks uses `C:\C`

HarrisonWAffel:
After discussion the Windows team has agreed that these changes should be done in 2.8 only, closing this PR and will open a new one soon

PR #3537, opened 2024-02-28
[dev-v2.8] Remove rancher-wins dependency from Rancher monitoring chart

HarrisonWAffel:
## Problem
Currently, The Rancher monitoring sub-chart `rancher-windows-exporter` relies on `rancher-wins` in order to deploy the required node exporter binary. This sub-chart relies on `rancher-wins` as the [node exporter binary](https://github.com/prometheus-community/windows_exporter) needs access to the [Windows Management Instrumentation (WMI) API](https://learn.microsoft.com/en-us/windows/win32/wmisdk/wmi-start-page), and exposing host level APIs to a Windows container has historically not been possible. `rancher-wins` allows the sub-chart to deploy the binary as a host level process and proxy its port to other containers.

Unfortunately,  the `rancher-wins` dependency limits the environments that rancher-monitoring can successfully be installed on. For instance, Rancher monitoring will not collect metrics for Windows nodes in the following environments, as `rancher-wins` is not installed on them by default: 
+ GKE 
+ EKS 
+ AKS
+ Imported Clusters
+ Nodes joined to the cluster the Rancher server is running on 
   + though, afaik, this is an unsupported configuration

## Solution
In Kubernetes v1.22 [Windows HostProcess pods were introduced](https://kubernetes.io/docs/tasks/configure-pod-container/create-hostprocess-pod/), and the feature went GA in v1.26. The HostProcess security context option allows workloads to be run directly on the host, allowing access to the WMI API without the use of `rancher-wins` and its proxies. By relying on a native Kubernetes feature, we can ensure that the sub-chart will properly deploy in any Kubernetes environment which supports HostProcess pods. 

Additional benefits of removing the dependency on `rancher-wins` is that the sub-chart becomes much less complicated, and we will no longer use any images with Foreign Layer references (which relates to https://github.com/rancher/windows/issues/222)

### Manual Testing
I have deployed the updated monitoring chart to custom RKE2 windows clusters, AKS clusters, and EKS clusters with windows node pools. In all environments these changes have worked as expected, and I did not see any errors or significant differences when upgrading or downgrading between this new chart version and already released versions. 

### Automated Testing
n/a

## QA Testing Considerations
While I was able to test on AKS clusters, I want to work closely with QA on getting this validated on EKS clusters as well. 

Because we are running the node exporter in a new environment (HostProcess pod) we should be extra careful during testing to ensure that no existing metrics change in any way. During my testing I did not see any difference in how metrics were reported between this new version and already released versions, but a double check would be good. 

## Backporting considerations
This change will not be backported to 2.7, and I am going to raise a 2.9 PR which migrates this sub-chart such that we simply patch the existing upstream chart. In short - there are no backporting or forwardporting considerations for this PR


packages/rancher-monitoring/rancher-monitoring/package.yaml
```diff
@@ -1,7 +1,7 @@
 url: https://github.com/prometheus-community/helm-charts.git
 subdirectory: charts/kube-prometheus-stack
 commit: beaa76a4ac26584c6cc77c27bc0d40769fb98b26
-version: 103.0.4
+version: 103.0.5
```
adamkpickering: This change strikes me as worthy of a minor (rather than patch) version bump. @lucasmlp @nicholasSUSE what do you think?

adamkpickering (changes requested):
AFAICT looks good from the point of view of the MAPPS team, other than the version thing. But let's wait for @lucasmlp or @nicholasSUSE's opinion.

Also, heads up: there is a merge conflict.

packages/rancher-monitoring/rancher-monitoring/package.yaml
HarrisonWAffel: I've updated the version to `103.1.0+up45.31.1`

alexandreLamarre (changes requested):
Overall, LGTM, just a few notes from the observabilty & backup side:
- Our new development philosophy is to merge things to dev-v2.9 first, so would also like to see a PR there
- ~~In order to minize overhead during releases, we're trying our best to only release one version per rancher release, in this case we also support bumping the minor, so we would aim to only release `103.1.0` for the next rancher version ... That would involve `make remove`ing the current rc `103.0.5-rc1` and changing the version `103.0.1-rc1`. WDYT @lucasmlp @nicholasSUSE ?~~ We should change the release version for this to `103.1.0-rc1`
- @pankajsqe is the test plan here clear, or do you need additional details?



cc @joshmeranda

joshmeranda:
I'm in favor of removing the `103.0.5-rc`

alexandreLamarre:
Actually hold that thought, we only have PR's open to create 103.0.5-rc1 assets, there are none in the upstream branch yet, if we just change the version in this PR to `103.1.0-rc1` and don't merge any PRs for the 103.0.5-rc1 assets, we won't have to do any extra release jutsu to make things clean

HarrisonWAffel:
@alexandreLamarre The 2.9 PR is [here](https://github.com/rancher/charts/pull/3607), however it differs substantially from this one as from 2.9 onwards we will be patching the upstream chart instead of maintaining the reimplementation that we currently have. That PR is in draft waiting on a change to the image mirror scripts for mirroring windows only upstream images. I don't think we should block this PR from being merged due to the 2.9 PR, the two are independent of one another.

I've updated the chart version in this PR to `103.1.0-rc1+up45.31.1`

alexandreLamarre:
@HarrisonWAffel Perfect thanks

alexandreLamarre (approved):

adamkpickering (approved):

PR #3540, opened 2024-02-29
Update Fleet to v0.10.0-rc.3

thardeck:
Update Fleet to v0.10.0-rc.3

Changelog: https://github.com/rancher/fleet/releases/tag/v0.10.0-rc.3


adamkpickering (approved):

charts/fleet/104.0.0+up0.10.0-rc.3/Chart.yaml.orig
```diff
@@ -0,0 +1,16 @@
+annotations:
```
weyfonk, line 1: What is the use of this file?

release.yaml
```diff
@@ -1,9 +1,9 @@
 fleet:
-  - 102.2.1+up0.8.1
```
weyfonk, line 2: Curious not to see 0.10.0 replacing _0.9.0_ here :thinking:

charts/fleet-crd/104.0.0+up0.10.0-rc.3/README.md
```diff
@@ -0,0 +1,5 @@
+# Fleet CRD Helm Chart
```
weyfonk, line 1: I'm surprised not to see actual CRDs in this chart (from [here](https://github.com/rancher/fleet/tree/master/charts/fleet-crd/templates)) :thinking:

weyfonk (changes requested):
I think CRDs are missing here.

release.yaml
thardeck: It seems we did not submit Fleet 0.9.x yet to the `dev-2.9` branch and I suppose this was forked from an older Rancher version.

charts/fleet/104.0.0+up0.10.0-rc.3/Chart.yaml.orig
thardeck: Good catch, thanks.

charts/fleet-crd/104.0.0+up0.10.0-rc.3/README.md
thardeck: This seems to be the same structure as in our released tarball: https://github.com/rancher/fleet/releases/download/v0.10.0-rc.3/fleet-crd-0.10.0-rc.3.tgz

But you are right, previous fleet-crd releases are different.

manno: Since we switched to codegen, https://github.com/rancher/fleet/blob/master/scripts/package-helm#L39 is outdated and would need to use 
`bash ./cmd/codegen/hack/generate_and_sort_crds.sh ./build/charts/fleet-crd/templates/crds.yaml`.


However, I don't think the release scripts should generate CRDs and such. Let's just release what's checked into the repo. That's what we tested.

recena:
@thardeck For upcoming contributions, please, use a forked repository.

thardeck:
We did some restructuring for the new Fleet version.
We will create a new pr after releasing a new Fleet version which fixes the crd issues.

PR #3545, opened 2024-02-29
add missing annotation for harvester-cloud-provider

markhillgit:
## Problem
catalog.cattle.io/permits-os annotation is missing in harvester-cloud-provider 102.0.2+up0.2.3
## Solution
added it


markhillgit:
will resubmit

PR #3552, opened 2024-03-04
[dev-v2.9] rancher-monitoring: add ScrapeConfig CRD

Carlotronics:
## Problem

ScrapeConfig CRD is provided in prometheus charts but missing in rancher monitoring.
## Solution

Add ScrapeConfig CRD

### DIFF
--- a/charts/rancher-monitoring-crd/104.0.2+up45.31.1/crd-manifest/crd-scrapeconfigs.yaml
+++ b/charts/rancher-monitoring-crd/104.0.2+up45.31.1/crd-manifest/crd-scrapeconfigs.yaml
@@ -1,0 +1,2148 @@
+# https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.71.2/example/prometheus-operator-crd/monitoring.coreos.com_scrapeconfigs.yaml
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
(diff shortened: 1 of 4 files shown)


joshmeranda:
@Carlotronics can you give us some more context? Is this causing issues on your system, point us to the `ScrapeConfig` definition in [kube-prometheus-stack-45.31.1](https://github.com/prometheus-community/helm-charts/tree/beaa76a4ac26584c6cc77c27bc0d40769fb98b26/charts/kube-prometheus-stack), etc?

alexandreLamarre (commented):
From the [docs](https://prometheus-operator.dev/docs/user-guides/scrapeconfig/)

> ScrapeConfig CRD :
>
> Starting with prometheus-operator v0.65.x, one can use the ScrapeConfig CRD to > scrape targets external to the Kubernetes cluster or create scrape configurations that are not possible with the higher level ServiceMonitor/Probe/PodMonitor resources.

Which is supported, since we are using `prometheus-operator:v0.65.1`, but its absent from the upstream chart we use `kube-prometheus-stack-45.31.1`, taking a quick glance at the upstream chart commit history, these configs were added in around version `48.x` which contains breaking changes to how the chart manages CRDs. So likely these will be supported during our next upgrade of the operator : https://github.com/rancher/rancher/issues/44614

nicholasSUSE:
Close message test

nicholasSUSE:
- Closing Pull Request.
- Not merged until 3 months after creation.
- automatic message.

PR #3538, opened 2024-02-28
[dev-v2.9] Add rancher istio 1.19.6 dev v2.9

nicholasSUSE:
## Issue: 

https://github.com/rancher/rancher/issues/44138

## Manual Testing
- [x] Provision rancher on k8s (latest -1) version or lesser
- [x]  Install chart on this cluster
- [x]  Fresh install/uninstall/upgrade/downgrade of the charts
- [x]  Upgrade this cluster in rancher to k8s latest available version. Make sure the existing chart and the app is functional after cluster upgrade.
- [x] Execute P0 scenario tests

PR #3561, opened 2024-03-06
[dev-v2.9] Fix rancher istio 1.18.2

nicholasSUSE:
## Problem

Istio 1.18.2 was added without some needed annotations on deployment.yaml of Kiali. 
Adding this fix and re-generating this chart. 

This was blocking the following PR: https://github.com/rancher/charts/pull/3538

## Manual Testing

- [x] Provision rancher on k8s (latest -1) version or lesser
- [x]  Install chart on this cluster
- [x]  Upgrade this cluster in rancher to k8s latest available version. Make sure the existing chart and the app is functional after cluster upgrade.
- [x] Execute P0 scenario tests


lucasmlp (approved):
LGTM 🚀

PR #3562, opened 2024-03-06
Bump rancher-webhook to v0.4.3-rc1

tomleb:
## Issue: https://github.com/rancher/rancher/issues/44018

Also includes CAPI removal from https://github.com/rancher/rancher/issues/44619 to make bump to 1.28 k8s possible.

## Problem

We want to support 1.28 k8s.

## Solution

Bump relevant dependencies. cluster-api v1.5.5 conflicts with k8s.io/* 0.28. One option is to remove it entirely from webhook since we aren't using it anymore (since v2.7.7).

The changes are:
- Upgraded cluster-api to v1.5.5 to match rancher/rancher v2.8
- Upgraded controller-runtime to v0.15.3
- Upgraded github.com/rancher/rke to v1.5.7-rc2
- Upgraded lasso which supports 1.28
- Upgraded to wrangler v2 so that's why there is a LOT of files modified.
- Upgraded dynamiclistener to v0.4.0-rc2 because it contains 1.28 support + wrangler v2. Otherwise conflicts, can't compile. Here's the comparison of the previous version vs this one: https://github.com/rancher/dynamiclistener/compare/v0.3.5...v0.4.0-rc2. Some non-trivial changes here, but it seemed okay to do in rancher/rancher so should be safe here.
- Upgrade the k8s dependencies to v0.28.6

And also fully removes CAPI webhook from rancher/webhook.


packages/rancher-webhook/package.yaml
```diff
@@ -1,3 +1,3 @@
-url: https://github.com/rancher/webhook/releases/download/v0.4.2/rancher-webhook-0.4.2.tgz
-version: 103.0.1
```
tomleb, line 2: Bumped the chart version here to `103.0.2`, I think I need to do this since we updated the charts in rancher-webhook, but I'm not 100% sure.

lucasmlp (approved):
LGTM 🚀

packages/rancher-webhook/package.yaml
lucasmlp: If `103.0.1` was released, you need to bum the chart. Otherwise, it's fine if you want to keep using `103.0.1`

JonCrowther (approved):

ericpromislow (approved):
Looks like a straightforward new-version chart

PR #3607, opened 2024-03-11
[dev-v2.9] Use Upstream Windows Node Exporter In Rancher Monitoring

HarrisonWAffel:
## Problem
Historically Rancher has implemented a custom chart for the Windows node exporter component of Monitoring V2. This was due to a dependency on `rancher-wins`, which was required in order to start the node exporter process on the host and scrape metrics from WMI. Unfortunately, this custom chart not only results in additional maintenance burdens, the reliance on `rancher-wins` prevents the windows node exporter from being deployed in certain environments (such as imported clusters, hosted providers, EKS, AKS, etc.).

## Solution
With the introduction of [host process pods on windows](https://kubernetes.io/docs/tasks/configure-pod-container/create-hostprocess-pod/) we can now deploy processes directly on the host without the need of `rancher-wins`. Additionally, upstream has updated their chart which can now be used directly by Rancher. This PR completely removes the old windows exporter sub-chart in favor of pulling in the upstream chart and modifying it as needed. This not only eliminates the use of `rancher-wins`, allowing the node exporter to be deployed in all environments, but also reduces the maintenance burden of the sub-chart. 

This PR makes minor changes to the upstream chart to ensure that configuration options and settings continue to be applied in the same way that they are in older chart versions. Additionally, a dedicated Powershell script has been added which automatically configures the require firewall rules (upstream uses a simple init container, but in practice this errors out after uninstall / reinstall of the chart).

## Testing
I've deployed this chart onto custom clusters created on Azure, as well as AKS and EKS. In all tests the chart properly deploys and I can create Grafana charts detailing windows node metrics. I am able to upgrade / downgrade / reinstall the chart without issue. 

### Manual Testing
see above

### Automated Testing
n/a

## QA Testing Considerations

We will want to test this change specifically in hosted providers such as EKS, AKS, and GKE. 

The intention of this PR is to persist all of the settings custom configuration of the existing custom chart onto the upstream chart. However, there are likely some behavior changes between these two charts - this is simply due to the fact that the existing custom chart has not been updated for ~1 year (or more). Testing should be done to identify any gaps or missing features that result from this change, so that we can either address the issue or document their removal from the upstream chart / exporter binary. 

## Backporting considerations
This change will not be backported to 2.7 or 2.8. We will continue to maintain the custom chart for those release lines, but will only use the upstream chart for 2.9+.


packages/rancher-monitoring/rancher-windows-exporter/package.yaml
```diff
@@ -1,3 +1,5 @@
-url: local
-version: 104.0.0
+url: https://github.com/prometheus-community/helm-charts.git
+subdirectory: charts/prometheus-windows-exporter
+commit: 3f7bb0e012e81a6497643d855df584ad5fe8b87a
+version: 104.0.1
```
joshmeranda: `104.0.0` hasn't been released yet so no need to bump version

packages/rancher-monitoring/rancher-monitoring/package.yaml
```diff
@@ -1,7 +1,7 @@
 url: https://github.com/prometheus-community/helm-charts.git
 subdirectory: charts/kube-prometheus-stack
 commit: beaa76a4ac26584c6cc77c27bc0d40769fb98b26
-version: 104.0.2
+version: 104.0.3
```
joshmeranda: `104.0.2` hasn't been released yet so no need to bump

packages/rancher-monitoring/rancher-windows-exporter/generated-changes/patch/templates/_helpers.tpl.patch
joshmeranda, line 1: Doesn't look like this file is needed (no deletions or lines changed)?

joshmeranda (changes requested):
Just some versioning and patch fixes, but mostly lgtm

packages/rancher-monitoring/rancher-windows-exporter/package.yaml
HarrisonWAffel: Yeah, I haven't focused too specifically on the version listed in this PR since I'm not 100% aware of any other work going on in monitoring which might land before this gets merged. Once this is taken out of draft I'll make sure this is correct 👍

packages/rancher-monitoring/rancher-windows-exporter/generated-changes/patch/templates/_helpers.tpl.patch
HarrisonWAffel: I had to add helpers for the `system-default-registry`, `windowsExporter.renamedMetricsRelabeling`, and `windowsExporter.renamedMetrics` since they weren't included in upstream. A lot of the context around those changes does look like upstream, but I have added lines for this file

lucasmlp:
No need to flag the PR as a draft in the title. Just open it as a draft and all will be fine

HarrisonWAffel:
This PR is finally ready for review! After looking at `release/v2.9` I could not find a "released" `104` version of Rancher monitoring, so I've updated this PR to use version `104.0.0-rc2`. Let me know if I should change that to another version or not

packages/rancher-monitoring/rancher-windows-exporter/package.yaml
```diff
@@ -1,3 +1,5 @@
-url: local
-version: 104.0.0
+url: https://github.com/prometheus-community/helm-charts.git
+subdirectory: charts/prometheus-windows-exporter
+commit: 3f7bb0e012e81a6497643d855df584ad5fe8b87a
+version: 104.0.0-rc2
```
joshmeranda: should be `104.0.0-rc1` we haven't been dealing with different rc numbers (less than intuitive I know)

release.yaml
```diff
@@ -120,12 +120,12 @@ rancher-logging-crd:
   - 102.0.2+up3.17.10
   - 103.0.1+up3.17.10
 rancher-monitoring:
-  - 104.0.0-rc1+up45.31.1
+  - 104.0.0-rc2+up45.31.1
```
joshmeranda: see versioning comment for `package.yaml`

joshmeranda (changes requested):

packages/rancher-monitoring/rancher-windows-exporter/package.yaml
HarrisonWAffel: Alright, changed back to `104.0.0-rc1`

adamkpickering (approved):
LGTM from a MAPPS perspective, but I think https://github.com/rancher/charts/pull/3607/files#r1571081158 still needs to be addressed?

HarrisonWAffel:
@adamkpickering Good catch! forgot I had done the same versioning change for the sub-chart as well. Should be fixed now

snasovich:
@joshmeranda , could you please re-review?

@HarrisonWAffel , looks like there are merge conflicts including in `tgz` file so you may need to bump version?

joshmeranda (approved):
LGTM

PR #3659, opened 2024-03-18
[dev-v2.7.12] Un-rc rancher-backup 102.0.3+up3.1.3-rc6

ericpromislow:
The rancher-backup version didn't change going from the previous release (2.7.11) to this one, so the relevant lines need to be removed from `release.yaml`.

There's no need to actually `un-rc` anything, as the final versions were released in 2.7.11.

### DIFF
--- a/release.yaml
+++ b/release.yaml
@@ -10,10 +10,6 @@ rancher-gke-operator:
   - 102.1.0+up1.1.7-rc1
 rancher-gke-operator-crd:
   - 102.1.0+up1.1.7-rc1
-rancher-backup:
-  - 102.0.3+up3.1.3-rc6
-rancher-backup-crd:
-  - 102.0.3+up3.1.3-rc6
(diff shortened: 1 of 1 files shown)


alexandreLamarre (approved):

nicholasSUSE (changes requested):
Where is the following or previous `make charts` command for this change?

joshmeranda (approved):

ericpromislow:
> Where is the following or previous make charts command for this change?

I did run `make charts`, but there were no changes to commit. Here's why:

When `dev-v2.7.11` was merged into `dev-v2.7` there was a conflict: `dev-v2.7.11` referred to `-rc5` while `dev-v2.7` referred to `rc-6`. But the `rc-less` assets were merged in, and the final charts were referring to the `rc-less` versions.

nicholasSUSE (approved):

PR #3730, opened 2024-04-03
[dev-v2.7][re-port] Fix node control plane label in rancher-logging

alexandreLamarre:
## Issue: 
Original issues

SURE-7505 

## Problem

Some patches were reverted during the fast forward of release-v2.7 to dev-v2.7 due to versioning issues on our end 

https://github.com/rancher/charts/pull/3388

> The toleration label is inconsistent with the node label which is node-role.kubernetes.io/control-plane compared to node-role.kubernetes.io/controlplane here in the charts

## Testing

See original issues

## QA Testing Considerations

Tests should validate the logging pods scheduled as per the set toleration


mallardduck (approved):
LGTM

PR #3744, opened 2024-04-04
[dev-v2.9] Forward ports prometheus-federator 103.0.1+up0.4.0

lucasmlp:
This PR forward-ports the following `v2.8.3` related charts from `release-v2.8`:

- prometheus-federator 103.0.1+up0.4.0


index.yaml
```diff
@@ -5970,27 +5970,27 @@ entries:
...
     description: Prometheus Federator
-    digest: b78bd24740ed6de630e8db8366aada8928cb224eb580ee843b099a056ab33059
+    digest: 97740e89c6f45f34c08a6d9642b94cf80a49bc6cc402b57591e07a84fadc6fb5
     icon: https://raw.githubusercontent.com/rancher/prometheus-federator/main/assets/logos/prometheus-federator.svg
     name: prometheus-federator
     urls:
-    - assets/prometheus-federator/prometheus-federator-103.0.1+up0.4.1.tgz
-    version: 103.0.1+up0.4.1
+    - assets/prometheus-federator/prometheus-federator-103.0.1+up0.4.0.tgz
+    version: 103.0.1+up0.4.0
```
adamkpickering, line 6134: This change looks suspect to me. Same thing for the lines above it in this file. Changes of this pattern are not present in any of the other forward port PRs I have seen. Are these changes supposed to be made?

release.yaml
```diff
@@ -105,10 +106,12 @@ rancher-monitoring:
   - 104.0.0-rc1+up45.31.1
   - 103.0.3+up45.31.1
   - 102.0.3+up40.1.2
+  - 103.0.4+up45.31.1
 rancher-monitoring-crd:
   - 104.0.0-rc1+up45.31.1
   - 102.0.3+up40.1.2
   - 103.0.3+up45.31.1
+  - 103.0.4+up45.31.1
```
adamkpickering: I don't think we should be making changes to `rancher-monitoring` and `rancher-monitoring-crd` in this PR...

adamkpickering (changes requested):
I think this PR may have some errors. Please confirm that everything is correct.

index.yaml
lucasmlp: Agreed. Manually added the entry that was modified. Now, this only adds the entry for the new asset. Thanks for catching it

release.yaml
lucasmlp: Just rebased the branch and it's fixed.

adamkpickering (approved):

PR #3627, opened 2024-03-14
[dev-v2.8] Add active remotedialer connections dashboard to rancher-monitoring's "Performance debugging" dashboard

aruiz14:
## Problem

We normally use this graph when troubleshooting Rancher performance problems, and I recurrently find it annoying having to visually compare existing Add/Removed graphs to figure out if the number of active connections is abnormal or not, to the point I normally just go to the Prometheus console and run that query there directly.

## Solution

Add a new panel which directly presents the number of active connections directly.

Screenshot 2024-03-14 at 13 03 04

I structured the commits in this PR by different steps:
  - Updates the dashboard schema by running the Rancher and running the current chart version, then saving the JSON model for this dashboard.
  - Changed RemoteDialer added and removed panels to use a rate, as suggested by the query builder.
  - Added a new panel for Active connections, by showing the difference between added and removed.
  - Finally, run `make` commands to generate all the files.
 
## Testing

Chart is installed normally, Grafana includes a pre-built "Performance debugging" dashboard, and it has the desired panels mentioned above.

### Regressions Considerations

There are 2 dashboards that were changed, and they won't show absolute numbers anymore, but that raw data is still available in the Prometheus query console.

## Backporting considerations

- [dev-v2.9 PR](https://github.com/rancher/charts/pull/3762)

PR #3762, opened 2024-04-10
[dev-v2.9] rancher-monitoring update

aruiz14:
## Problem

We normally use this graph when troubleshooting Rancher performance problems, and I recurrently find it annoying having to visually compare existing Add/Removed graphs to figure out if the number of active connections is abnormal or not, to the point I normally just go to the Prometheus console and run that query there directly.

## Solution

Add a new panel which directly presents the number of active connections directly.

Screenshot 2024-03-14 at 13 03 04

I structured the commits in this PR by different steps:
  - Updates the dashboard schema by running the Rancher and running the current chart version, then saving the JSON model for this dashboard.
  - Changed RemoteDialer added and removed panels to use a rate, as suggested by the query builder.
  - Added a new panel for Active connections, by showing the difference between added and removed.
  - Finally, run `make` commands to generate all the files.
 
## Testing

Chart is installed normally, Grafana includes a pre-built "Performance debugging" dashboard, and it has the desired panels mentioned above.

There are 2 dashboards that were changed, and they won't show absolute numbers anymore, but that raw data is still available in the Prometheus query console.

## Backporting considerations
- [dev-v2.8 backport](https://github.com/rancher/charts/pull/3627)


lucasmlp (approved):
LGTM 🚀

moio (approved):
lgtm

alexandreLamarre (approved):

alexandreLamarre (approved):

PR #3765, opened 2024-04-10
[dev-v2.9] fleet 104.0.0+up0.10.0-rc.10 update

rancherbot:
Update Fleet to v0.10.0-rc.10

Changelog: https://github.com/rancher/fleet/releases/tag/v0.10.0-rc.10


manno (approved):

thardeck (approved):

index.yaml
```diff
@@ -6084,30 +6084,6 @@ entries:
     urls:
     - assets/prometheus-federator/prometheus-federator-104.0.0-rc1+up0.4.1.tgz
     version: 104.0.0-rc1+up0.4.1
-  - annotations:
```
adamkpickering, line 6087: I don't think this block deletion has anything to do with this PR?

adamkpickering (changes requested):

index.yaml
thardeck: It looks like it but the only thing that we run for the commit `Autogenerated changes for Fleet v0.10.0-rc.10`, which changes the `index.yaml`, is removing the previous Fleet 0.10.0-rc.4 charts with `make remove` and then running `PACKAGE=fleet make charts`.

To be sure I also just ran `PACKAGE=fleet make charts` without the removal of the previous Fleet rc version and it still resulted in these annotations being removed.

Running `make validate` is also green.

Do you know anybody who has insight in the `make charts` scripts who can give an insight why this is happening?

thardeck:
Superseded by #3782

index.yaml
thardeck: @adamkpickering https://github.com/rancher/charts/pull/3762/files for example has also the same issue.

Maybe the annotation is not correctly referenced somewhere?

PR #3774, opened 2024-04-11
[v2.9] Update rancher-vsphere-csi

HarrisonWAffel:
## Problem

The vSphere CSI chart has been updated to include a new field in the config map, https://github.com/rancher/vsphere-charts/pull/75

## Solution

Update the upstream commit for `rancher-vsphere-csi` in order to pull in the updated config map for the 2.9 release line


jiaqiluo (approved):

index.yaml
```diff
@@ -6084,30 +6084,6 @@ entries:
     urls:
     - assets/prometheus-federator/prometheus-federator-104.0.0-rc1+up0.4.1.tgz
     version: 104.0.0-rc1+up0.4.1
-  - annotations:
```
adamkpickering: Did you mean to delete this block? I don't see how this relates to this PR.

adamkpickering (changes requested):

index.yaml
HarrisonWAffel: Hmm, no I didn't. It looks like this was done automatically when using the charts build scripts. Let me see if I can avoid this from happening

HarrisonWAffel: Okay, most recent pushes have resolved this issue

adamkpickering (approved):

index.yaml
jiaqiluo: @adamkpickering  good catch!

PR #3787, opened 2024-04-15
[dev-v2.8] rancher-logging 103.1.0+up4.4.0 update

HarrisonWAffel:
## Problem

## Solution
+ Remove `rancher-logging 103.1.0-rc1+up4.4.0`
+ Adds `rancher-logging 103.1.0-rc2+up4.4.0`

## Testing
I've confirmed the chart can be installed and works as expected. I've confirmed that `v1.8.10` of rancher/fluent-bit does not include any foreign layer references.


joshmeranda (changes requested):
Just fix the merge conflicts, otherwise LGTM

joshmeranda (approved):

regsync.yaml
lucasmlp, line 1: This file should not be changed by contributors. We have a GitHub action that automatically detects new docker images and updates `regsync.yaml`

lucasmlp (changes requested):

regsync.yaml
HarrisonWAffel: I've removed the changes to `regsync.yaml`.

lucasmlp (approved):
LGTM 🚀

PR #3788, opened 2024-04-15
[v2.9] Bump Windows fluent-bit version to remove foreign layers

HarrisonWAffel:
## Problem

## Solution
+ Remove `rancher-logging 104.0.0-rc1+up4.4.0`
+ Adds `rancher-logging 104.0.0-rc2+up4.4.0`

## Testing
I've confirmed the chart can be installed and works as expected. I've confirmed that `v1.8.10` of rancher/fluent-bit does not include any foreign layer references.


packages/rancher-logging/package.yaml
```diff
@@ -1,5 +1,5 @@
 url: oci://ghcr.io/kube-logging/helm-charts/logging-operator:4.4.0
-version: 104.0.0-rc1
+version: 104.0.0-rc2
```
joshmeranda: We've been working with the same rc number before the release (a little counter intuitive I know). Can you move these changes to 104.0.0-rc1 instead please.

Otherwise looks good to me.

joshmeranda (changes requested):

packages/rancher-logging/package.yaml
HarrisonWAffel: Ah okay, will do

HarrisonWAffel: Should be good now. I've done the same change for the 2.8 PR

joshmeranda (approved):

adamkpickering (approved):

PR #3799, opened 2024-04-17
[dev-v2.7] rancher-logging 102.0.3-rc1+up3.17.10 update

HarrisonWAffel:
## Problem

## Solution

## Testing
I've confirmed the chart can be installed and work as expected. I've confirmed the image does not contain foreign layers.


joshmeranda (approved):

regsync.yaml
lucasmlp, line 1: This file should not be changed by contributors. We have a GitHub action that automatically detects new docker images and updates `regsync.yaml`

lucasmlp (changes requested):
Please, make changes to only one main chart per PR. Otherwise, if things goes south and we need to revert this PR, it becomes quite tricky.

regsync.yaml
HarrisonWAffel: Ah okay, good to know

HarrisonWAffel: I've removed the changes to `regsync.yaml` and taken out the monitoring chart changes, this PR is now only focused on the logging chart changes

lucasmlp (approved):
LGTM 🚀

PR #3812, opened 2024-04-18
[dev-v2.7] rancher-monitoring 102.0.4-rc1+up40.1.2 update

HarrisonWAffel:
## Problem

Various components of Rancher are using Windows images which contain Foreign Layer references. This makes mirroring these images more difficult and is a common source of user confusion when configuring windows nodes in air-gapped environments. Other PRs to rancher/rancher and rancher-monitoring on 2.8 and 2.9 have already resulted in many of these images being fixed, however monitoring on 2.7 has not yet been addressed. 

## Solution
+ Bump rancher/windows_exporter-package to `v0.0.5`
    + `v0.0.5` is simply a rebuilt version of `v0.0.3` which no longer contains foreign layer references. There is no functional difference between the two.

## Testing
I've confirmed the chart can be installed and work as expected. I've confirmed the image does not contain foreign layers.


lucasmlp (approved):
LGTM 🚀

joshmeranda (approved):

PR #3823, opened 2024-04-25
[dev-v2.8] Development branch sync-up 

lucasmlp:
## Release PRs that are being synced back to this branch:
- https://github.com/rancher/charts/pull/3814
- https://github.com/rancher/charts/pull/3822


release.yaml
```diff
@@ -1,20 +1,15 @@
 fleet:
```
lucasmlp, line 1: Also removed fleet because it was released previously as a forward-ported chart, but was not removed here.

rohitsakala (approved):

PR #3879, opened 2024-05-08
[dev-v2.9] elemental 104.0.0+up1.4.3 update

fgiudici:
## Issue: 

https://github.com/rancher/elemental-operator/issues/698

PR #3875, opened 2024-05-07
[dev-v2.9] Csp adapter release v4.0.0 rc1

mbelur:
##Issue: 
rancher/rancher#43856

##Problem:
Create new rancher-csp-adapter chart for 2.9 rancher

##Testing:

scenario 1 (fresh install):

    On an EKS cluster (k8s version 1.27), install locally built rancher with cspAdapterMinVersion: 104.0.0+up4.0.0-rc1
    Validated the support config and licence entitlement functionality. Created a downstream EKS cluster with 21 nodes (with 1 entitlement) to trigger out-of-compliance message. Scaled down the cluster to 17 nodes, validated that the out-of-compliance message is no longer displayed. Scaled back up to 21 nodes, validated the out-of-compliance message popped up.

scenario 2 (upgrade scenario):

    Start with an EKS cluster (k8s version 1.27), install rancher 2.8.3 with csp-adapter 3.0.1
    Upgrade rancher to locally built rancher with cspAdapterMinVersion: 104.0.0+up4.0.0-rc1
    Update charts url repo/branch and upgrade csp-adapter version to 104.0.0+up4.0.0-rc1
    Validated the support config and licence entitlement functionality. Created a downstream EKS cluster with 21 nodes (with 1 entitlement) to trigger out-of-compliance message. Scaled down the cluster to 17 nodes, validated that the out-of-compliance message is no longer displayed. Scaled back up to 21 nodes, validated the out-of-compliance message popped up.

PR #3883, opened 2024-05-08
[dev-v2.9] Adding slash prefix to root files

lucasmlp:
## Problem

The `CODEOWNERS` feature always adds @rancher/mapps team as reviewers. There's a missing slash in the root files that might be causing this. PRs with modifications to `README.md` files in subdirectories are triggering a review from the Mapps team. 

Examples:
- [dev-v2.9] PR with `README.md` modification: https://github.com/rancher/charts/pull/3879
- [dev-v2.9] PR without `README.md` modification: https://github.com/rancher/charts/pull/3875

## Solution

Added slash prefix to all files in the root directory.

## Backporting considerations

This PR should be backported to lower minor development branches.


.github/CODEOWNERS
```diff
@@ -4,7 +4,7 @@ docs @rancher/mapps
 scripts @rancher/mapps
 CNAME @rancher/mapps
 Makefile @rancher/mapps
-README.md @rancher/mapps
+/README.md @rancher/mapps
```
recena, line 7: Why not with the rest of file in the root?

lucasmlp: I'll add it to be safe, but I don't see PRs with files named like the rest of them.

.github/CODEOWNERS
```diff
@@ -1,14 +1,14 @@
...
-docs @rancher/mapps
-scripts @rancher/mapps
-CNAME @rancher/mapps
-Makefile @rancher/mapps
-README.md @rancher/mapps
-_config.yml @rancher/mapps
-configuration.yaml @rancher/mapps
-regsync.yaml @rancher/mapps
-index.html @rancher/mapps
+/.github @rancher/mapps
```
recena, line 2: What about `./`?

lucasmlp: Works the same as the roof files. With the slash first  `/.github`. This means that even if there were subdirectories containing a `.github` file or folder, it wouldn't trigger a review from us because we only own the one in the root directory `/.github`.

recena (approved):

PR #3898, opened 2024-05-13
[v2.9] Bump eks-operator to v1.9.0-rc.1

mjura:


mjura:
it will be done in https://github.com/rancher/charts/pull/3900

PR #3908, opened 2024-05-13
[dev-v2.9] rancher-logging k8s 1.29 & 1.30 support

mallardduck:
## Issue: 

https://github.com/rancher/rancher/issues/45356
https://github.com/rancher/rancher/issues/45335

## Problem

For Rancher 2.9.x we need to support K8S 1.29 and 1.30

## Solution

Adjust the kubeVersion of the chart to support the window of K8S versions Rancher 2.9.X will support.

## Testing

Logging should be able to install on 2.9.x build using 1.29 or 1.30 cluster.
Specifically after this is merged version `104.0.0-rc1` of logging will support 1.29 or 1.30 clusters.

### Manual Testing

Successfully installed on 1.29 and 1.30 downstream clusters.

## QA Testing Considerations

We found an existing open issue in r/r that matches the issue I've encountered in my testing. https://github.com/rancher/rancher/issues/38730

Because this is an existing issue and not technically an issue with installing the logging chart on 1.29 this is not going to be treated as a blocker for this PR. Once I rebase it and do a quick retest after that I will undraft it for review.

PSP related things are vestigial at this point - the minimum k8s version is 1.26.
However, unsure if we should begin removing them as they come from upstream.

## Backporting considerations

Nope - only for Rancher 2.9.x


mallardduck:
Currently investigating 2 (seemingly related) issues with the chart on 2.9 alpha2.

When trying to create a ClusterOutput:
> [ClusterOutput.logging.banzaicloud.io](http://clusteroutput.logging.banzaicloud.io/) “cluster-output-syslog” is invalid: spec.syslog.buffer: Invalid value: “string”: spec.syslog.buffer in body must be of type object: “string”

And in the same process on the "Output buffer" tab:
> Error loading schema for [logging.banzaicloud.io](http://logging.banzaicloud.io/).v1beta1.output.spec.awsElasticsearch.buffer

mallardduck:
We found an existing open issue in r/r that matches the issue I've encountered in my testing. https://github.com/rancher/rancher/issues/38730

Because this is an existing issue and not technically an issue with installing the logging chart on 1.29 this is not going to be treated as a blocker for this PR. Once I rebase it and do a quick retest after that I will undraft it for review.

@pankajsqe - You may want to keep this in mind for testing that UI creation of Output/ClusterOutput will have this issue until triaged and fixed by UI team.

alexandreLamarre (approved):

alexandreLamarre:
@mallardduck  @pankajsqe  UI issue might be related to open API schema bugs mentionned in https://github.com/rancher/rancher/issues/45157 or https://github.com/rancher/rancher/issues/45158.

PR #3933, opened 2024-05-16
[prod-v2.8] rancher-istio 103.3.0 (TEST)

nicholasSUSE:
This is only a test


nicholasSUSE:
This test worked, closing

PR #3934, opened 2024-05-16
Documenting usage through OOB strategy for new production branches

nicholasSUSE:
## Solution
Documenting the new workflow for releasing charts in the new branches. 
This new workflow is focused on Out-Of-Band releases which shall be performed by the chart owners and not the Mapps team anymore.

PR #3937, opened 2024-05-17
Documenting usage through OOB strategy for new production branches for dev-v2.7 

nicholasSUSE:
### Solution
Documenting the new workflow for releasing charts in the new branches.
This new workflow is focused on Out-Of-Band releases which shall be performed by the chart owners and not the Mapps team anymore.

Propagation from: [Documenting 2.9 PR](https://github.com/rancher/charts/pull/3934)

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,61 @@ This repository contains Helm charts served by Rancher Apps & Marketplace.
 
 ---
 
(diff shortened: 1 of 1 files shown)


lucasmlp (approved):
LGTM 🚀

PR #3940, opened 2024-05-17
Documenting new assets lifecycle - prod-v2.9

nicholasSUSE:
### Solution

As agreed on the long-term proposal for making the maintainability better for the Charts repo, the new assets lifecycle will only hold the most recent 3 versions of Rancher charts. 
This documentation shows which versions should be placed in each branch.


README.md
```diff
@@ -75,6 +78,36 @@ upstream	https://github.com/rancher/charts (push)
 
 ```
 
+### New Assets Lifecycle
+
+Starting on `17/May/2024` there is a new asset lifecycle in place.
+
+We will keep only the relevant assets in the corresponding branch.
+
+This means a cycle of always 3 active branches, with each branch always holding up to 3 previous versions before it.
```
diogoasouza: It's up to 2, no?

nicholasSUSE: True, fixed

lucasmlp (approved):

PR #3951, opened 2024-05-22
[dev-v2.9] system-upgrade-controller 104.0.0+up0.7.0 update

chiukapoor:
## Issue:
* https://github.com/rancher/rancher/issues/44922

## Updates:
* `system-upgrade-controller` is updated to `104.0.0+up0.7.0`
* Removed PSP reference
* Update `rancher_min_version` to `v2.9.0-alpha1`


packages/system-upgrade-controller/package.yaml
```diff
@@ -1,5 +1,5 @@
 url: https://github.com/rancher/system-charts.git
 subdirectory: charts/rancher-k3s-upgrader/0.7.0
-commit: c95f8395d45c828d6e92917524e3aa285d2299f2
-version: 104.0.0
+commit: 0e464d6f3afd709d8f3d9a528eb65973f696c5ae
+version: 104.0.1
```
jiaqiluo: We don't need to bump the version of an unreleased chart.

jiaqiluo (changes requested):

packages/system-upgrade-controller/package.yaml
chiukapoor: Sure, reverted the version bump.

jiaqiluo (approved):

jakefhyde (approved):

PR #3958, opened 2024-05-23
The template namespace post upgrade yaml fails when using system default registry

diogoasouza:
Updating the OPA gatekeeper chart due to a bug in the `namespace-post-upgrade.yaml` file where it was not using the `systemDefaultRegistry` value


nicholasSUSE (approved):
I am approving this, supposing that this altered version was never released before.

Please confirm it @diogoasouza

PR #3963, opened 2024-05-24
adding extra options to make lifecycle-assets and inserting functiona…

nicholasSUSE:
There were some lacking optional argument flags for the script execution. 
- debugFlag
- chart 

Also, there was a mistake that was fixed.
`make remove` needs to add the removed chart to the `release.yaml` otherwise the ci for the PR will break.


lucasmlp (commented):
LGTM 🚀

PR #4004, opened 2024-06-03
[dev-v2.8] updating charts to use charts-build-scripts 0.8.1

nicholasSUSE:
Propagation from: 
https://github.com/rancher/charts/pull/4002


rohitsakala (approved):

nicholasSUSE:
I am bypassing the Build job because it got stuck since the check was implemented before the fix.

PR #4012, opened 2024-06-04
changing build-scripts version to 0.5.2

nicholasSUSE:
### Problem 
https://github.com/rancher/charts/blob/dev-v2.9/docs/How-To-Guides/how_to_resolve_patch_failure.md

### Solution
Temporarily returning charts-build-scripts to 0.5.2 because we have a release in place today.

I will fix this problem after the release


lucasmlp (approved):
LGTM 🚀

PR #4023, opened 2024-06-05
[release-v2.7] downgrading build scripts for release - temporary

nicholasSUSE:
## Problem
We are having some problems with the latest build scripts that will be resolved after the release 

## Solution
Use an older version until this is solved.


lucasmlp (approved):
LGTM 🚀

PR #4066, opened 2024-06-12
[dev-v2.9] rancher-cis-benchmark 6.0.0-rc3 update

krunalhinguu:
### Updates

- Update rancher-cis-benchmark to `6.0.0-rc3`
- Update kubectl to `v1.29.2`
- Update sonobuoy to `v0.57.1`

### Related

- https://github.com/rancher/rancher/issues/45753


packages/rancher-cis-benchmark/charts/values.yaml
```diff
@@ -45,7 +45,7 @@ global:
     clusterName: ""
   kubectl:
     repository: rancher/kubectl
-    tag: v1.28.7
+    tag: v1.29.2
```
andypitcher:
```suggestion
    tag: v1.29.2
```
Not 100% sure about this version of kubectl, based on [kubelet version skew policy](https://kubernetes.io/releases/version-skew-policy/#kubelet), if we have to cover up to k8s 1.31 we might need kubectl 1.30 at least, since it covers version-1 | version | version+1.

mitulshah-suse: the minimum version supported by this chart is supposed to be 1.28. So based on that, it seems 1.29.x is a better fit at least until 1.31 support is added.

andypitcher (approved):

mitulshah-suse (approved):
LGTM

PR #4100, opened 2024-06-18
Update logging.yaml

GeertvanHorrik:


alexandreLamarre (changes requested):
Hi @GeertvanHorrik thanks for the patch, could you follow https://github.com/rancher/charts/blob/dev-v2.9/docs/developing.md#making-changes-to-packages to introduce the fix into the chart?

GeertvanHorrik:
@alexandreLamarre thanks for pointing me to this direction. All I wanted is to fix the typo, but the steps shown in the readme seems too much work (need to set up a dev environment, etc). I suggest we close this and a ticket is created instead? Then it can be done by someone that has experience and a dev environment already set up.

PR #4123, opened 2024-06-25
[dev-v2.9] Harvester CSI driver k8s version 1.30 support

Vicente-Cheng:
## Problem
Rancher v2.9 will support both k8s version v1.29 and v1.30

## Solution
Update the kube-version range


w13915984028 (approved):
LGTM from Harvester, thanks.

nicholasSUSE (approved):

PR #4132, opened 2024-06-26
[dev-v2.9] rancher-alerting-drivers 104.0.0-rc1 update

alexandreLamarre:
## Issue: 

Chart visibility issue in rancher catalog during QA of 1.29/1.30 support

## Problem

Alerting chart fails to show up in the rancher catalog UI despite having the correct support annotations.

## Solution

After reverting an accidental bump to the `kubeVersion` supported version it works as expected in the catalog

## Testing

See above


mallardduck:
Does this change need to be applied to other charts that had this bumped?

For instance: https://github.com/rancher/charts/blob/1770f2d3739e696803d3d88aa4877d9422e81213/packages/rancher-monitoring/rancher-monitoring/generated-changes/patch/Chart.yaml.patch#L147-L148

alexandreLamarre:
Temporary issue with rancher-2.9-alpha5, resolved in 2.9 head

PR #4165, opened 2024-07-02
[dev-v2.9] rancher-vsphere-csi 104.0.0+up3.3.0-rancher1 add

krunalhinguu:
### Updates

- Added new version `104.0.0+up3.3.0-rancher1` for `rancher-vsphere-csi` with commit hash `9a36f8af39b27508beab166996848a22df69ed05`
 ### Related
- https://github.com/rancher/rancher/issues/45747


snasovich (changes requested):
`104.0.0` `rancher-vsphere-csi` chart is not released yet so instead of adding `104.0.1+up3.3.0-rancher1` we should change `104.0.0+up3.1.2-rancher4` to `104.0.0+up3.3.0-rancher1`.

Also, we shouldn't have any `.rej` files. Please refer to https://github.com/rancher/charts/blob/dev-v2.9/docs/How-To-Guides/how_to_resolve_patch_failure.md - it seems to be relevant to this situation.

snasovich (changes requested):
@krunalhinguu , old chart (`104.0.0+up3.1.2-rancher4`) is still there (and in `release.yaml`). I see `104.0.0+up3.3.0-rancher1` was added instead of **replacing** `104.0.0+up3.1.2-rancher4`.

Please also update PR title and description accordingly.

snasovich (approved):

snasovich (approved):

PR #4254, opened 2024-07-18
[dev-v2.8]  fix toleration key

jbiers:
## Issue: 
[#7505](https://jira.suse.com/browse/SURE-7505)
## Problem
This fix had been previously applied to **rancher:dev-v2.7** in [this PR](https://github.com/rancher/charts/pull/3388/files). The fix however was lost and re-implemented later into version 2.7. However, **dev-v2.8** branched off in the meantime and ended up not having the fix included.


assets/rancher-logging/rancher-logging-103.1.1+up4.4.0.tgz
joshmeranda, line 1: Can you change the version to `103.1.2-rc1+up4.4.0`

joshmeranda (changes requested):

PR #4255, opened 2024-07-18
[dev-v2.8] fix toleration key

jbiers:
## Issue: 
[#7505](https://jira.suse.com/browse/SURE-7505)
## Problem
This fix had been previously applied to **rancher:dev-v2.7** in [this PR](https://github.com/rancher/charts/pull/3388/files). The fix however was lost and re-implemented later into version 2.7. However, **dev-v2.8** branched off in the meantime and ended up not having the fix included.


packages/rancher-logging/package.yaml
```diff
@@ -1,5 +1,5 @@
 url: oci://ghcr.io/kube-logging/helm-charts/logging-operator:4.4.0
-version: 103.1.1
+version: 103.1.2
```
joshmeranda: Can you add `-rc1` to the end? We've gone through some versioning issues during releases and the rc makes it all manageable.

joshmeranda (changes requested):

packages/rancher-logging/package.yaml
joshmeranda, line 1: This change should also be done in its own commit for ease-of-review

joshmeranda (approved):

PR #4278, opened 2024-07-22
[v2.8] Fix aks,eks and gke support for Kubernetes v1.30

mjura:


valaparthvi:
Is 1.30 supported on Rancher 2.8? Rancher 2.8 only supports until 1.28 as far as I know.

mjura:
> Is 1.30 supported on Rancher 2.8? Rancher 2.8 only supports until 1.28 as far as I know.

awesome, than it is not needed

PR #4281, opened 2024-07-22
[dev-v2.9] Update sriov version to 104.2.0

mgfritch:
## Problem

add support for k8s v1.30

## Testing

Deploy charts via helm using a two node RKE2 v1.30.2+rke2r1


nicholasSUSE (changes requested):
The right version bump is 104.1.1+up0.1.0

nicholasSUSE (approved):

mgfritch:
> The right version bump is 104.1.1+up0.1.0

prior major versions of this chart (`102.x`, `103.x`) are bumping the minor version instead of the patch version
let's maintain the same convention for the current `104.x` series

rbrtbnfgl (approved):

PR #4295, opened 2024-07-25
[dev-v2.9] rancher-logging 104.0.0 support resources on daemonsets

ricoapon:
## Issue
I did not report one (just created this PR) and couldn't find an existing issue.
EDIT: created https://github.com/rancher/rancher/issues/46499 for this.

## Problem
Rancher-logging instals DaemonSets, which did not have any way to limit `resources`. This gave problems on our Kubernetes cluster where the pods could use 50% of the node memory.

## Solution
Make it possible to configure the resources on DaemonSets. Default should stay the same (no limits configured).

### Manual Testing
I installed the Helm chart manually on our own cluster to check that the `rke2` code works.


charts/rancher-logging/104.0.0+up4.4.0/values.yaml
```diff
@@ -362,13 +362,16 @@ additionalLoggingSources:
...
+    daemonSetResources: {}
   rke2:
     enabled: false
     stripUnderscores: false
+    daemonSetResources: {}
   k3s:
     enabled: false
     container_engine: "systemd"
     stripUnderscores: false
+    daemonSetResources: {}
```
ricoapon, line 374: I was unsure what to call them. We could do `resources`, but that is maybe a bit vague. A more specific name (refering to `DaemonSet`) will make others aware of where the resources are actually used.

nicholasSUSE:
### Warning
- This pull request has been inactive for more than 30 days.
- It will be closed if it is not merged until 3 months after creation.

PR #4298, opened 2024-07-29
[2.9.1] Bump vsphere cpi

thatmidwesterncoder:
## Problem

Lots of images showing up rancher-images.txt with security vulnerabilities related to old versions of k8s

## Solution

Bumping rancher-vsphere-cpi in order to remove support for non-supported k8s versions in rancher.


jiaqiluo (approved):

charts/rancher-vsphere-cpi/104.0.1+up1.8.1/Chart.yaml
```diff
@@ -0,0 +1,24 @@
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/display-name: vSphere CPI
+  catalog.cattle.io/kube-version: '>= 1.18.0-0 < 1.31.0-0'
```
jiaqiluo: we need to correct the range.

jiaqiluo (changes requested):

charts/rancher-vsphere-cpi/104.0.1+up1.8.1/Chart.yaml
thatmidwesterncoder: I have updated the range to exclude invalid k8s versions.

thatmidwesterncoder: Pulled in changes the correct way!

jiaqiluo (approved):

HarrisonWAffel (approved):

PR #4299, opened 2024-07-29
[2.9.1] Bump vsphere csi

thatmidwesterncoder:
## Problem

Lots of images showing up rancher-images.txt with security vulnerabilities related to old versions of k8s

## Solution

Bumping rancher-vsphere-csi in order to remove support for non-supported k8s versions in rancher.


jiaqiluo (approved):

charts/rancher-vsphere-csi/104.0.1+up3.3.0-rancher2/Chart.yaml
```diff
@@ -0,0 +1,24 @@
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/display-name: vSphere CSI
+  catalog.cattle.io/kube-version: '>= 1.20.0-0 < 1.31.0-0'
```
jiaqiluo: ~not a blocker:~ because the upstream chart drops the support for old k8s version < 1.27,  it will be better if the range can match that.

jiaqiluo: I thought about it twice and think we should fix the range to avoid the chart showing up on clusters with unsupported versions. I think UI has some logic depending on this annotation.

thatmidwesterncoder: @jiaqiluo I updated the generated-changes to include the new version patch!

jiaqiluo: @thatmidwesterncoder,  I know it is a bit annoying but the proper way to fix the annotation is to fix it on the upstream chart (https://github.com/rancher/vsphere-charts). I can help review PRs.

thatmidwesterncoder: oh wow - i didn't even think of that lol, I'll get going on that.

thatmidwesterncoder: PR here: https://github.com/rancher/vsphere-charts/pull/85

thatmidwesterncoder: Pulled in changes the correct way!

jiaqiluo (approved):

HarrisonWAffel (approved):

PR #4300, opened 2024-07-30
Add network policy for RKE2 Ingress Nginx

dharmit:
## Issue: 

Addresses issue raised in https://github.com/rancher/rancher/issues/45603.

## Problem

Prometheus target failing to come up

## Solution

Using the modified chart in this PR enables the target to come up

## Testing

- Install rancher-monitoring-crd through this PR
- Install rancher-monitoring using following Helm command:
  ```sh
  helm install rancher-monitoring charts/rancher-monitoring/105.0.0-rc1+up57.0.3/ \
  --namespace=cattle-monitoring-system \
  --set alertmanager.alertmanagerSpec.configSecret=alertmanager-rancher-monitoring-alertmanager \
  --set alertmanager.alertmanagerSpec.useExistingSecret=true \
  --set global.cattle.clusterId=local \
  --set global.cattle.clusterName=local \
  --set global.cattle.rkePathPrefix="" \
  --set global.cattle.rkeWindowsPathPrefix="" \
  --set global.cattle.systemDefaultRegistry="" \
  --set global.cattle.systemProjectId=p-cwh9d \
  --set global.cattle.url=https://13.202.137.42.sslip.io \
  --set global.systemDefaultRegistry="" \
  --set prometheus.prometheusSpec.evaluationInterval=1m \
  --set prometheus.prometheusSpec.retentionSize=50GiB \
  --set prometheus.prometheusSpec.scrapeInterval=1m \
  --set rke2ControllerManager.enabled=true \
  --set rke2Etcd.enabled=true \
  --set rke2IngressNginx.enabled=true \
  --set rke2IngressNginx.networkPolicy.enabled=true \
  --set rke2Proxy.enabled=true \
  --set rke2Scheduler.enabled=true
  ````
- Go to Rancher UI; Monitoring; Prometheus Targets
- In the Prometheus UI search for the ServiceMonitor `serviceMonitor/kube-system/rancher-monitoring-ingress-nginx/0`. It should be `UP`.


dharmit:
I have honestly got no clue about why there are so many lines modified in this PR. Steps I followed are below:
```sh
$ export PACKAGE=rancher-monitoring/rancher-monitoring
$ make prepare
// made changes
$ make patch
$ make charts
```
Did I do something wrong here? 🤔

alexandreLamarre (dismissed):
Rancher monitoring version needs to be bumped to `104.1.1-rc1`. Recommend separating `make patch` into a separate commit from `make charts` for exactly this scenario, where you would to do something like rebase with drop.

Need to remove `packages/rancher-monitoring/rancher-monitoring/generated-changes/overlay/crds` as a bug in charts-build-scripts includes these by accident.

packages/rancher-monitoring/rancher-monitoring/generated-changes/overlay/crds/crds/crd-alertmanagerconfigs.yaml
joshmeranda, line 1: Deleting this `packages/rancher-monitoring/rancher-monitoring/generated-changes/overlay/crds` directory should clean up those excess lines. This is a known issue caused by how we need to pull in our crds. There is a fix pending: https://github.com/rancher/charts-build-scripts/pull/143

joshmeranda (changes requested):

dharmit:
@joshmeranda @alexandreLamarre thanks for your feedback, guys! I have modified as per your recommendations. Can you PTAL again? Thanks.

joshmeranda:
Looking better, but git histroy could use some housekeeping. Can you update the commits to look closer to the steps outline [here](https://github.com/rancher/charts/blob/dev-v2.9/docs/developing.md#making-changes-to-packages)? Generally this will look like this:

joshmeranda (changes requested):
Looking better, but git history could use some housekeeping. Can you update the commits to look closer to the steps outline [here](https://github.com/rancher/charts/blob/dev-v2.9/docs/developing.md#making-changes-to-packages)? Generally this will look like this:

1. `export PACKAGE=rancher-monitoring/rancher-monitoring`
2. `make prepare`
3. Make the changes you need to the prepare chart, in this case adding the network policy
4. `make patch`
5. Commit the changes to `packages/rancher-monitoring/rancher-monitoring/generated-changes` (don't forget to remove `generated-changes/overlay/crds` introduced by the bug)
6. Update the version in `package.yaml`
7. Commit changes to `package.yaml`
8. `make charts`
9. Commit changes made by `make charts`
10. Update and commit `release.yaml` with new version of chart

Following these steps will make it a good bit easier for the reviewer to see what was changed and why. Also make each commit smaller and prevents new chart versions from stalling our browsers when trying to load 100+ new files. Thanks!

dharmit:
@joshmeranda I referred to the [Making Changes To Packages](https://github.com/rancher/charts/blob/dev-v2.9/docs/developing.md#making-changes-to-packages) section which mentions making version related changes first. In context of your comment, I followed steps 6-9 first and 1-5 after that.

If I follow the steps you suggested, it'd make changes to 104.1.0. And, AFAIK, 104.1.0 has already released, right? Correct me if I'm wrong because there's a good chance I am. :)

joshmeranda:
@dharmit You're right that we don't want to make changes to `104.1.0` but since you won't run `make charts` until after you bump the the version in step 6/7 you shouldn't be updating any existing charts.

I keep forgetting that the docs recommend you update the version first. That's fine as long as its a separate commit. Generally I'll update `package.yaml` in its own commit since its easier to see the new version in a commit with 1 changed file rather than 100+ changes files, then run `make charts` at the end. This will still create the new charts at versions `104.1.1` since `package.yaml` will have the bumped version.

index.yaml
```diff
@@ -17430,6 +17430,136 @@ entries:
     - assets/rancher-logging-crd/rancher-logging-crd-3.6.000.tgz
     version: 3.6.000
   rancher-monitoring:
+  - annotations:
+      artifacthub.io/license: Apache-2.0
+      artifacthub.io/links: |
+        - name: Chart Source
+          url: https://github.com/prometheus-community/helm-charts
+        - name: Upstream Project
+           url: https://github.com/prometheus-operator/kube-prometheus
```
ericpromislow: There's an extra space before `url:`. Strictly speaking it's not a yaml error, as this block is a string, but it looks like it's intended to be input for some other yaml process, and therefore should be conforming yaml code

ericpromislow (changes requested):
Something in this process is generating broken yaml.

dharmit:
> I keep forgetting that the docs recommend you update the version first. That's fine as long as its a separate commit. Generally I'll update `package.yaml` in its own commit since its easier to see the new version in a commit with 1 changed file rather than 100+ changes files, then run `make charts` at the end. This will still create the new charts at versions `104.1.1` since `package.yaml` will have the bumped version.

I see your point. However, what I have done is again just as per the docs recommendation. It recommends updating the version in `package.yaml`, then doing `make charts`, and committing the changes.

I can totally do what you're recommending if that's the general practice followed by O&B team. Let me know.

joshmeranda:
@dharmit Yeah that's totally fine to do it that way, its just less familiar. It makes it easier to ensure that the changes you made were brought into the generated chart. Feel free to continue!

But you'll still want to have separate commits for updating `package.yaml`, building the new charts at that version, making the necessary changes, and rebuilding the charts. The commit history for something like this pr should probably look like this:

1. Commit 1: `bump rancher-monitoring/rancher-monitoring version to 104.1.1+up57.0.3`
2. Commit 2: `make charts`
3. Commit 3: `update release.yaml` (can go anywhere as long as its updated`
4. Commit 4: `Add network policy for RKE2 Nginx Ingress`
5. Commit 5: 'make charts'

I get that this feels somewhat nitpicky, but because of the sheer amount of files that are updated in charts PRs it makes it much easier for reviewers to give a review.

dharmit:
> I get that this feels somewhat nitpicky, but because of the sheer amount of files that are updated in charts PRs it makes it much easier for reviewers to give a review.

No, it doesn't sound nitpicky at all, TBH. Thanks a tonne for laying down the steps. I'll update the PR accordingly. 👍🏽

release.yaml
```diff
@@ -1,10 +1,14 @@
-fleet-agent:
-  - 104.0.1+up0.10.1-rc.1
 fleet:
   - 104.0.1+up0.10.1-rc.1
+fleet-agent:
```
joshmeranda: We should avoid adding references to charts that weren't updated in this PR. Remove the `fleet-agent` and `rancher-vsphere-csi` fields please. Otherwise LGTM!

joshmeranda (approved):

joshmeranda (changes requested):
see `release.yaml` comments above

release.yaml
dharmit: Well, those changes are based on `make check-release-yaml`.

joshmeranda (approved):
LGTM

ericpromislow:
I don't think you need my review here -- reviews from Josh and Alexandre are sufficient.

dharmit:
@joshmeranda @alexandreLamarre I have cherry-picked commits to be added on top of dev-v2.10 branch. Can you PTAL? I think the CI needs to be re-triggered to be run against 2.10 branch.

mallardduck:
Seeing this upon install:

> W0813 13:11:43.161908      44 warnings.go:70] spec.template.spec.containers[2].ports[0]: duplicate port definition with spec.template.spec.containers[1].ports[0]

packages/rancher-monitoring/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/exporters/ingress-nginx/network-policy.yaml
```diff
@@ -0,0 +1,19 @@
+{{- if .Values.rke2IngressNginx.networkPolicyEnabled }}
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  annotations:
+    np.rke2.io/ingress: resolved
+  name: rke2-ingress-network-policy
+  namespace: {{ .Values.rke2IngressNginx.namespaceOverride }}
```
mallardduck: Seems like this doesn't actually have a default value. Is there supposed to be some sort of helper or lookup with this as an override when set? Or is this required? If required, we should add it as an empty (null) value and put a note that it should be set or something.

packages/rancher-monitoring/rancher-monitoring/generated-changes/patch/values.yaml.patch
```diff
@@ -338,6 +338,7 @@
 +  enabled: false
 +  metricsPort: 10254
 +  component: ingress-nginx
++  networkPolicyEnabled: false
```
mallardduck: This is OK for now - because it seems like it should only need to be on or off. However, other `networkPolicy` related values in this file make it an object. So if we need more than one value for this I would conform to that example as well. I.e.

```yaml
networkPolicy:
  - enabled: false
  - namespace: null # user must set
  ```

**Review by mallardduck — CHANGES_REQUESTED** [review]
Not quite working out the box for me when I set to enabled. In the rancher installed apps section it thinks that it has network policies, but they 404 when I click their links.

**Comment by mallardduck** [comment]
> I don't think you need my review here -- reviews from Josh and Alexandre are sufficient.

Screenshot 2024-08-13 at 9 35 29 AM


@ericpromislow - I think that because you left a "Requested Changes" review you must provide an "Approved" review before this PR will be passable. Ignore mine since I added them today intentionally for changes I suggested. But notice your's and alex...I think big picture we need a new PR even after mine and yours are approved since Alex can review currently.

File: packages/rancher-monitoring/rancher-monitoring/generated-changes/patch/values.yaml.patch (hunk 4 above)
Comment (dharmit) [reply packages/rancher-monitoring/rancher-monitoring/generated-changes/patch/values.yaml.patch]: That was my initial thought when introducing this value. But decided against it as it's a single value.

**Comment by joshmeranda** [comment]
> Seeing this upon install:
> 
> > W0813 13:11:43.161908      44 warnings.go:70] spec.template.spec.containers[2].ports[0]: duplicate port definition with spec.template.spec.containers[1].ports[0]

@mallardduck This one is coming from upstream: https://github.com/grafana/helm-charts/blob/main/charts/grafana/templates/_pod.tpl#L1016-L1021

File: packages/rancher-monitoring/rancher-monitoring/generated-changes/patch/values.yaml.patch (hunk 4 above)
Comment (joshmeranda) [reply packages/rancher-monitoring/rancher-monitoring/generated-changes/patch/values.yaml.patch]: Its a single value, but if we need to make it an object later, that could affect upgrade paths / is not backward compatible. Unless there is a specific reason this needs to be a single value, it would be better as an object.

Good catch @mallardduck

File: packages/rancher-monitoring/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/exporters/ingress-nginx/network-policy.yaml (hunk 3 above)
Comment (dharmit) [reply packages/rancher-monitoring/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/exporters/ingress-nginx/network-policy.yaml]: It's there in `values.yaml`.
https://github.com/rancher/charts/blob/4411cc0ccaca933bb9a495413dd887d0922e9033/charts/rancher-monitoring/104.1.0%2Bup57.0.3/values.yaml#L335-L347

File: release.yaml [hunk 5]
```diff
@@ -1,13 +1,17 @@
-fleet-agent:
-  - 104.0.1+up0.10.1-rc.1
 fleet:
```
dharmit: The changes in this file are a result of doing `make validate` and `make check-release-yaml`. I have made zero manual changes to this.

dharmit:
> Not quite working out the box for me when I set to enabled. In the rancher installed apps section it thinks that it has network policies, but they 404 when I click their links.

@mallardduck IIUC (which is based on description in https://github.com/rancher/rancher/issues/45603), you have to open Prometheus Targets from the Monitoring view and check if `serviceMonitor/kube-system/rancher-monitoring-ingress-nginx` is up. I'm not sure where in the Installed Apps would you see a Network Policy because I see none on my RKE2 setup. Also, please set up an RKE2 with CIS enabled. Below are the configuration files I used:

```sh
$ cat /etc/rancher/rke2/config.yaml
cni: calico
write-kubeconfig-mode: "0644"
profile: cis
pod-security-admission-config-file: /etc/rancher/rke2/rke2-custom-pss.yaml

$ cat /etc/rancher/rke2/rke2-custom-pss.yaml
apiVersion: apiserver.config.k8s.io/v1
kind: AdmissionConfiguration
plugins:
- name: PodSecurity
  configuration:
    apiVersion: pod-security.admission.config.k8s.io/v1beta1
    kind: PodSecurityConfiguration
    defaults:
      enforce: "restricted"
      enforce-version: "latest"
      audit: "restricted"
      audit-version: "latest"
      warn: "restricted"
      warn-version: "latest"
    exemptions:
      usernames: []
      runtimeClasses: []
      namespaces: [kube-system, cis-operator-system, tigera-operator, cattle-system, cattle-fleet-system, cattle-fleet-local-system, cattle-monitoring-system]
```

release.yaml
joshmeranda: `charts-build-scripts` is really only checking that everything that is going to be released is included, it does not check that things are should not be released are not included. In that process it might add unexpected things to `release.yaml`

dharmit: IMO doing `make validate` is sorting stuff alphabetically. However, the YAML it's generating is broken. That is fixed by running `make check-release-yaml`. That's been my observation so far.

joshmeranda: Yeah that whole, `check-release-yaml` scripts stuff is annoying. If I remember correctly the yaml itself isn't "broken" just not in the desired format (whitespce vs no whitespace in array data).

> Well, those changes are based on make check-release-yaml.

`check-release.yaml` does not modify data, it basically just runs it through `yq` and checks if that changed release.yaml. Those changes are coming from `make validate` which does a whole lot more than just alphabetize `release.yaml` (although I can remember if thats done in `check-release-yaml` or `make validate`). `make validate` basically compares the local list of charts with those found in the last release, and reports on anything that should be in released and wasn't. If you wanna see the nitty gritty look here: https://github.com/rancher/charts-build-scripts/blob/master/pkg/validate/validate.go.

We should not be adding charts to `release.yaml` unless we made changes to that chart in the same PR. We've all been guilty of doing this before since they are easy to miss changes, so don't sweat it :)

dharmit: > check-release.yaml does not modify data, it basically just runs it through yq and checks if that changed release.yaml

The `scripts/check-release-yaml` has this `yq -i release.yaml` which modifies `release.yaml` in place. But you're right that it's coming from `make validate` in this case.

> make validate which does a whole lot more than just alphabetize release.yaml (although I can remember if thats done in check-release-yaml or make validate). make validate basically compares the local list of charts with those found in the last release, and reports on anything that should be in released and wasn't. 

Yep, makes sense. In case of this PR, it's alphabetizing as well as adding the rc chart to the `relesae.yaml`.

> We should not be adding charts to release.yaml unless we made changes to that chart in the same PR. We've all been guilty of doing this before since they are easy to miss changes, so don't sweat it :)

I'm lost. Should this PR have the changes to `release.yaml` or not? Anyway, let's discuss this over a call for quicker resolution.

joshmeranda: Realized this was my mistake getting confused / not looking carefully enough. Changes to / `release.yaml` were just reordering + adding monitoring charts versions.

packages/rancher-monitoring/rancher-monitoring/generated-changes/overlay/templates/rancher-monitoring/exporters/ingress-nginx/network-policy.yaml
mallardduck: So I think looking at other similar situations to this one, what we actually want to do here is:
1. define a new item in `_helpers.tpl` for this,
2. Call that helper here to reference that instead,
3. Allow the `values.yaml` to have the `namespaceOverride` value default be empty.

mallardduck: So in this file we would replicate:
https://github.com/dharmit/rancher-charts/blob/aa9a6b007735706c311588752309cda89139982f/charts/rancher-monitoring/104.1.1-rc1%2Bup57.0.3/charts/prometheus-node-exporter/templates/daemonset.yaml#L5C26-L5C60

But call it: `"rke2-ingress-nginx.namespace"`

mallardduck: And in the helpers file, I think we would mimic something like:
https://github.com/dharmit/rancher-charts/blob/aa9a6b007735706c311588752309cda89139982f/charts/rancher-monitoring/104.1.1-rc1%2Bup57.0.3/templates/_helpers.tpl#L301-L310

Since we know the default value should generally be `kube-system`, similar to how the default there is `kube-state-metrics`.

mallardduck:
Edit: Disregard original message for now - Prachi can dismiss the blocking review.
Will review the PR first thing tomorrow and review again.

dharmit:
> Edit: Disregard original message for now - Prachi can dismiss the blocking review.


TBH, I have been backing up old branch and force pushing each time. So, in a sense, this PR keeps getting fresh each time. 😉 

> Will review the PR first thing tomorrow and review again.

Before you spend time reviewing, I want to get the thing about RC clarified. Last time I pinged you on Slack, we agreed upon following steps mentioned in https://github.com/rancher/charts/pull/4404#issuecomment-2344160459 for this PR. However, a recent Slack thread started by @joshmeranda has got me concluding that I don't need to do those and only have a PR with the actual changes. Did I get this right?

mallardduck:
> I don't need to do those and only have a PR with the actual changes. Did I get this right?

Yes that's correct - with the clarification we got from @joshmeranda, there's no need for this to bump RC.1 to RC.2.

dharmit:
Closing this PR as it looks like @jbiers got this merged via #4545. The process for opening PR kept changing, and I got embedded into a different team, which led to work going unmerged and unnoticed. Disappointed. 😞

PR #4315, opened 2024-08-07
[dev-2.10] rancher-webhook 105.0.0+up0.6.0-rc.1 create

ericpromislow:
## Issue: 

For testing https://github.com/rancher/rancher/issues/42611


charts/rancher-webhook/105.0.0+up0.6.0-rc.1/Chart.yaml
```diff
@@ -0,0 +1,14 @@
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/hidden: "true"
+  catalog.cattle.io/kube-version: < 1.31.0-0
+  catalog.cattle.io/namespace: cattle-system
+  catalog.cattle.io/os: linux
+  catalog.cattle.io/permits-os: linux,windows
+  catalog.cattle.io/rancher-version: '>= 2.9.0-0 < 2.10.0-0'
```
MbolotSuse, line 8: At some point we need to fix this (you can't do this here, we need to do this in the WH repo) so that it only works for 2.10.0+

MbolotSuse (approved):

joshmeranda (approved):

PR #4354, opened 2024-08-16
[dev-v2.10] Include metrics options for rancher logging

jbiers:
## Issue: 
[#46591](https://github.com/rancher/rancher/issues/46591)

## Problem

In order to add monitoring to the rancher-logging stack, it is necessary to allow users to enable the serviceMonitor option via Helm Chart.


release.yaml
```diff
@@ -21,9 +21,12 @@ rancher-vsphere-csi:
 rancher-vsphere-cpi:
   - 104.0.1+up1.8.1
 rancher-logging:
-  - 104.0.1-rc1+up4.4.0
+  - 105.0.0-rc1+up4.4.0
+  - 105.0.0+up4.4.0
 rancher-logging-crd:
+  - 105.0.0-rc1+up4.4.0
   - 104.0.1-rc1+up4.4.0
```
joshmeranda: This bit can be removed

joshmeranda (changes requested):

release.yaml
jbiers: Good catch! Fixed :)

joshmeranda (approved):

PR #4376, opened 2024-08-26
Update webhook to v0.6.1-rc.1

ericpromislow:
Used for testing rancher-webhook PR https://github.com/rancher/webhook/pull/478

Note that due to a mixup this is the next RC in the v2.10 dev line after `v0.6.0-rc.1`.
There will not be a `v0.6.0` release for webhook -- we have jumped to `v0.6.1` and chart version `105.0.1+up0.6.1-rc.1`, to be eventually released as chart version `105.0.1+up0.6.1`


maxsokolovsky (approved):

charts/rancher-webhook/105.0.1+up0.6.1-rc.1/Chart.yaml
```diff
@@ -8,7 +8,7 @@ annotations:
   catalog.cattle.io/rancher-version: '>= 2.9.0-0 < 2.10.0-0'
   catalog.cattle.io/release-name: rancher-webhook
 apiVersion: v2
-appVersion: 0.6.0-rc.1
+appVersion: 0.6.1-rc.1
 description: ValidatingAdmissionWebhook for Rancher types
 name: rancher-webhook
-version: 105.0.0+up0.6.0-rc.1
+version: 105.0.1+up0.6.1-rc.1
```
MbolotSuse, line 14: Can you elaborate on why this is 105.0.1? This part is the rancher compat version, would should still be 105.0.0 for 2.10 AFAIK

MbolotSuse (changes requested):
One issue on the version which prefixes the WH version.

ericpromislow:
I assumed something in the chart version string had to be incremented because we incremented the minor-version on the underlying release. What should it be?

MbolotSuse:
@ericpromislow as far as I know since this targets the same version of Rancher it should stay at the current value.

ericpromislow:
Wrong chart version. Should be `105.0.0+up0.6.1-rc.1` , not `105.0.1+up0.6.1-rc.1`

PR #4381, opened 2024-08-29
[dev-2.9] Backport update fluentd image repository

jbiers:
## Issue: 
[#1366](https://github.com/rancherlabs/image-scanning/issues/1366)

## Problem
We are using an old version of the fluentd image which needs to be updated. It fixes some CVEs and also prevents the bug described in [this issue](https://github.com/rancher/rancher/issues/46862) from happening.


mallardduck (approved):
lgtm

PR #4385, opened 2024-08-30
[dev-v2.10] Update logging image dependencies

jbiers:
## Issue: 
[#2764](https://github.com/rancherlabs/image-scanning/issues/2764)


charts/rancher-logging/105.0.0-rc1+up4.8.0/values.yaml
```diff
@@ -6,7 +6,7 @@ replicaCount: 1
 
 image:
   repository: rancher/mirrored-kube-logging-logging-operator
-  tag: 4.8.0
```
joshmeranda: The linked CVE is for `4.4.0` are those CVEs not fixed yet in `4.8.0`? To fix issues in `4.4.0` you'll probably want to be targeting 2.9 / 2.8

joshmeranda (changes requested):

charts/rancher-logging/105.0.0-rc1+up4.8.0/values.yaml
jbiers: That's right the linked CVE is for 4.4.0. The idea here is, since the fix will be needed for 2.9, why not follow upstream and update to the latest used version in both branches?

mallardduck: I think if we do that, then we also need to bump the charts too.

Per my analysis here: https://github.com/rancher/image-mirror/pull/706#pullrequestreview-2261719867
(assuming we agree with that logic still)

Then we would want to use the following logic:

1. Always prefer to match the version of images used in upstream charts,
    -  Crucial because per that bug the other day, if we allow image versions to drift between each other then bugs happen.
2. When we determine upstream lags for no reason, update local and PR to upstream secondary priority.

So we'd also want to use 4.9.1 chart for logging if we're bumping images that high.

So either we start from these defaults targeting 2.8.0: https://github.com/kube-logging/logging-operator/blob/4.8.0/pkg/sdk/logging/api/v1beta1/logging_types.go#L166-L186

Or these defaults for 2.9.1: https://github.com/kube-logging/logging-operator/blob/4.9.1/pkg/sdk/logging/api/v1beta1/logging_types.go#L173-L192

jbiers: I actually made a mistake when interpreting the values from upstream, let's keep that value as it is then!

joshmeranda (approved):

PR #4389, opened 2024-09-02
[dev-v2.10] Remove prometheus adapter deprecated flag

jbiers:
## Issue: 
[#46919](https://github.com/rancher/rancher/issues/46919)

## Problem
The **logtostderr** flag was removed in prometheus-adapter [PR 586](https://github.com/kubernetes-sigs/prometheus-adapter/pull/586/commits/8579be6c7b1eb069eb4c5c6e435b6defca76393b). Since we recently updated prometheus-adapter to version v0.12.0, the flag should no longer be used.


charts/rancher-monitoring/105.0.0-rc1+up57.0.3/charts/prometheus-adapter/templates/deployment.yaml
```diff
@@ -60,7 +60,6 @@ spec:
         - --tls-private-key-file=/var/run/serving-cert/tls.key
         {{- end }}
         - --cert-dir=/tmp/cert
-        - --logtostderr=true
```
joshmeranda, line 63: Does this change the behavior at all? or are logs sent to stderr be default?

jbiers: According to [KEP-2845](https://github.com/kubernetes/enhancements/blob/master/keps/sig-instrumentation/2845-deprecate-klog-specific-flags-in-k8s-components/README.md#logging-defaults) the default is to split logs into two streams, stdout for info logs and stderr for error logs.

joshmeranda (approved):

PR #4391, opened 2024-09-02
[dev-v2.10] Expose prometheusRules value

jbiers:
## Issue: 
[#46878](https://github.com/rancher/rancher/issues/46878)

## Problem
The logging-operator chart provides by default a value which, if enabled, will create PrometheusRules to monitor both fluentd and fluentbit. This PR seeks to make it more obvious by setting it explicitly in the values file and adding a comment with a link to the upstream documentation.


joshmeranda (approved):
Remember to backport this fix

jbiers:
> Remember to backport this fix

This feature was only added to v2.10 so no backport needed

PR #4428, opened 2024-09-12
[release-v2.10] Charts release validation 2.10

nicholasSUSE:
Propagation from: https://github.com/rancher/charts/pull/4424


.github/workflows/charts-release.yaml
```diff
@@ -0,0 +1,28 @@
...
+            GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Checkout into branch
+        run: git checkout -b staging-pr-workflow
+
+      - name: Pull scripts
+        run: sudo make pull-scripts
+
+      - name: Validate Charts Release
+        run: sudo make validate-release-charts BRANCH=${{ github.event.pull_request.base.ref }} GH_TOKEN=${{ secrets.GITHUB_TOKEN }} PR_NUMBER=${{ github.event.pull_request.number }}
```
tashima42: Does these commands really need to use `sudo`? Maybe you can drop it by updating permissions on files

nicholasSUSE: I am going to test this now.
Good suggestion

PR #4429, opened 2024-09-12
Fix the syntax: the 'and' was missing.:

ericpromislow:
See issue [#47092](https://github.com/rancher/rancher/issues/47092)

This typo is in rancher-monitoring charts for v2.8 and v2.9


mallardduck:
Seems like we want to use https://github.com/rancher/charts/pull/4449 in favor of this PR?

PR #4433, opened 2024-09-13
[dev-v2.10] rancher-logging 105.0.0-rc.2+up4.8.0 support k8s 1.31

jbiers:
## Issue: 
[#46839](https://github.com/rancher/rancher/issues/46839)

PR #4437, opened 2024-09-13
[dev-v2.10] rancher-logging 105.0.0-rc.2+up4.8.0 drop support for k8s 1.27

jbiers:
## Issue: https://github.com/rancher/rancher/issues/46839

This is a fix for the previous PR https://github.com/rancher/charts/pull/4433 which dropped only one version instead of two


ericpromislow (approved):
looks fine

PR #4444, opened 2024-09-16
 [dev-v2.8] elemental 103.4.0+up1.6.4 update

fgiudici:


packages/elemental/elemental/generated-changes/overlay/values.yaml.orig
```diff
@@ -0,0 +1,43 @@
+image:
```
davidcassany: Was this file meant to be part of the commit? This `.orig` suffix smells a like a merging/comparing branches leftover

fgiudici: good catch, removed

davidcassany (approved):
LGTM

PR #4510, opened 2024-09-30
[dev-v2.8] backport rebase-logging

jbiers:
## Issue: https://github.com/rancher/rancher/issues/47323


packages/rancher-logging/package.yaml
```diff
@@ -1,5 +1,5 @@
 url: oci://ghcr.io/kube-logging/helm-charts/logging-operator:4.8.0
-version: 103.1.2
+version: 103.1.3
```
mallardduck: I think this should be an `-rc.1`.

Similar to how Josh's looks here: https://github.com/rancher/charts/pull/4311

Once that's corrected I think this looks good.

mallardduck (changes requested):
Just one minor version related thing.
We do want this to add an rc name, but we won't bump it for subsequent changes to the chart before release. We'll just reuse `-rc.1` since monitoring and logging have the exception.

mallardduck (approved):

PR #4516, opened 2024-10-01
Remove old sachet charts

### LINKED ISSUE #47332: [v2.9] Remove published versions of old/unused `rancher-sachet`
Per title, this is to track the removal of all files under `charts/rancher-sachet` ~and updating the package to ensure it doesn't publish~ (the package has been correctly configured as `doNotRelease` for multiple branches).

We will still keep the `rancher-sachet` package as it's necessary for building the alerting chart. However it is only needed as a package for sub-chart compilation. Rather than for releasing the chart standalone.

mallardduck:
## Issue: 
Will fix: https://github.com/rancher/rancher/issues/47332

## Problem
Old unused charts in chart repo.

## Solution
Delete old un-used charts.

## Backporting considerations
Probably could backport to 2.8 - need to apply to 2.10 branch too.


release.yaml
```diff
@@ -28,3 +28,7 @@ rancher-cis-benchmark:
   - 6.3.0-rc.1
 rancher-cis-benchmark-crd:
   - 6.3.0-rc.1
+rancher-sachet:
```
joshmeranda: If we are removing versions why are we adding them to release.yaml?

joshmeranda (changes requested):
Looks ok to me. Before I give approval I would like to get MAPPs to sign off on this. They typically don't like us to remove charts that have already been released

CC @nicholasSUSE

release.yaml
mallardduck: Weird - maybe an issue with the remove command? I just did:

```
CHART=rancher-sachet VERSION=* make remove
```

Sequentially for each affected version.

mallardduck: Not sure the cause - but I've corrected it now.

rohitsakala:
The below is assumming that noone uses these versions.

I think it is not a good practice to delete old released versions of a product. I checked [Kubernetes](https://github.com/kubernetes/kubernetes/tags?after=v1.0.0), [Rancher](https://github.com/rancher/rancher/tags?after=v1.0.0). I am sure nobody uses Rancher 1.0.0 or kubernetes 1.0.0. Also [semver](https://semver.org/) states we should not MODIFY released versions. 

I might be wrong here, but this is an inference from the practices of the upstream projects.

mallardduck:
@rohitsakala - These probably should have been removed long ago, or never released as standalone charts for Rancher 2.6+. Ultimately the package is meant to not be published as a chart but compiled into an actual product chart (`rancher-alert-drivers`) living as a subchart. Removal was requested by @recena - however we're awaiting input form Support org folks to get their feedback.

rohitsakala:
@mallardduck Thanks for the clarification. Approving since this chart should have never being released as standalone charts.

rohitsakala (approved):

diogoasouza (approved):

joshmeranda (approved):

PR #4517, opened 2024-10-01
[v2.9] Correct 404 errors causing Grafana display issues

mallardduck:
## Issue: 
https://github.com/rancher/rancher/issues/46567 / SURE-9001

## Problem
Chart is using image with bug.

## Solution
Update affected images to patched versions

## Testing
Yes - valid reproduction steps on issue and JIRA.

### Manual Testing

Was able to verify locally the bug existed, then verify that using updated images seems to resolve it.

### Automated Testing

n/a

## Backporting considerations

Should be backported to 2.8 as well.


jbiers (approved):
LGTM

PR #4527, opened 2024-10-07
Bump rancher-webhook to v0.4.12-rc1

maxsokolovsky:


charts/rancher-webhook/103.0.10+up0.4.12-rc1/Chart.yaml
```diff
@@ -0,0 +1,14 @@
...
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/hidden: "true"
+  catalog.cattle.io/kube-version: '>= 1.23.0-0 < 1.29.0-0'
+  catalog.cattle.io/namespace: cattle-system
+  catalog.cattle.io/os: linux
+  catalog.cattle.io/permits-os: linux,windows
+  catalog.cattle.io/rancher-version: '>= 2.8.0-0 < 2.9.0-0'
+  catalog.cattle.io/release-name: rancher-webhook
+apiVersion: v2
+appVersion: 0.4.12-rc1
```
tomleb: Shouldn't that be `0.4.12-rc.1`? We've moved away from this format, I see we had `0.4.11-rc.1`.

maxsokolovsky: Yup, I had noticed that right after pushing; updated.

maxsokolovsky: Yup, but I had pushed the webhook tag with the wrong format last week, and we want those immutable.
I noticed, though, that I had the wrong base version for the chart here, so updated.

tomleb: Right, so what I was suggesting was actually keeping the "wrong" format but still creating one with the right format and using that one. So in that case, you could create `0.4.12-rc.1` and use it, leaving `0.4.12-rc1` alone.

tomleb (approved):

prachidamle (approved):

PR #4567, opened 2024-10-15
[dev-v2.9] Bump harvester csi driver v0.1.20

Vicente-Cheng:
## Problem
We need to bump the new Harvester CSI driver to support RWX volume

## Solution
Bump the Harvester CSI driver


Vicente-Cheng:
Hi @w13915984028, @FrankYang0529, could you help review it?

FrankYang0529 (approved):

PR #4421, opened 2024-09-11
[WIP] Prom reorg example test

mallardduck:
Test of this: https://github.com/rancher/prometheus-federator/pull/93

PR #4588, opened 2024-10-17
[2.10] Prometheus Federator 105.0.0-rc.1+up0.4.3-rc.1

mallardduck:
## Issue: https://github.com/rancher/rancher/issues/46296

Note: This is related to that issue, but not going to solve it. This is just the first step, the overall fix that will solve that issue will come later which is why that is tagged for 2.11. However this portion is expected for 2.10 AFAIK.  

(~@MKlimuszka can correct if I'm wrong~ Verified with @prachidamle this part can go in 2.10)

## Problem
N/A; Doesn't directly seek to solve end-user problem.

## Solution
N/A; Doesn't directly seek to solve end-user problem.

### Manual Testing

General testing of the chart has been done and the same behaviour as before was found. (as expecte

### Automated Testing
I've recently re-enabled all the GH CI and E2E testing on this chart (and the subcomponents). So this build has a higher coverage of automated tests run on it before successfully merging PRs and creating releases than past versions.

## QA Testing Considerations

This has already been tested once via one-off builds/charts I created via: https://github.com/rancher/charts/pull/4421
This new PR is just because I wanted to start fresh instead of using that old branch - save overall idea and set of changes.
Ideally users will notice no significant difference in the features that PromFed provdies. I.e. if it works before it still works, if it's bugged before it's still bugged.

Same as the last time we tested this. Ideally we see no changes that would cause new issues or solve known existing ones.

## Backporting considerations
Eventually - likely after more has been built off this new foundation. That way the backport is more significant and includes fixes as well as this restructure of the project/codebase.


joshmeranda (approved):

pankajsqe:
validated Prometheus Federator 105.0.0-rc.1+up0.4.3-rc.1


Screenshot 2024-11-07 at 4 16 37 PM
Screenshot 2024-11-07 at 3 44 29 PM

PR #4633, opened 2024-10-22
[dev-2.10] Add system-upgrade-controller 105.0.0

jiaqiluo:
https://github.com/rancher/rancher/issues/46835 

This PR adds the new chart: 
- system-upgrade-controller 105.0.0

Changes:
- migrate the source chart from the [rancher/system-chart repo](https://github.com/rancher/system-charts/tree/release-v2.9/charts/rancher-k3s-upgrader) to this repo as the former is deprecated 
- update the manifest for the system-upgrade-controller deployment to match the one in the [upstream v0.14.2 tag](https://github.com/rancher/system-upgrade-controller/blob/v0.14.2/manifests/system-upgrade-controller.yaml#L30-L119) except the `tolerations` and `affinity` section
- update image tags in the `values.yaml` file
- update the `catalog.cattle.io/rancher-version` annotation to enable the chart for rancher v2.10.0+ only 
- update the chart version to a new major release line  (105.x.x)

Next step:
- bump the env var `CATTLE_SYSTEM_UPGRADE_CONTROLLER_CHART_VERSION` in the `package/Dockerfile` and `tests/v2/codecoverage/package/Dockerfile` file in rancher/rancher


HarrisonWAffel (approved):

packages/system-upgrade-controller/package.yaml
```diff
@@ -1,5 +1 @@
-url: https://github.com/rancher/system-charts.git
-subdirectory: charts/rancher-k3s-upgrader/0.7.0
-commit: 0e464d6f3afd709d8f3d9a528eb65973f696c5ae
-version: 104.0.0
-doNotRelease: false
+url: local
```
snasovich, line 1: Wherever I see `url: local` used there is generally `version` as well. Any specific reason we don't want to add it, at least for consistency?

packages/system-upgrade-controller/charts/Chart.yaml
```diff
@@ -0,0 +1,18 @@
...
+  catalog.cattle.io/rancher-version: '>= 2.10.0-0'
+  catalog.cattle.io/release-name: system-upgrade-controller
+apiVersion: v1
+appVersion: v0.14.2
+description: General purpose controller to make system level updates to nodes.
+home: https://github.com/rancher/system-upgrade-controller
+kubeVersion: '>= 1.25.0-0'
+name: system-upgrade-controller
+sources:
+- https://github.com/rancher/charts/tree/dev-v2.10/charts/system-upgrade-controller
```
snasovich: Checking a few charts I don't see us referencing `charts` in sources and reference to `system-upgrade-controller` is covered by `home` attribute already - so suggesting to remove `sources` here.

packages/system-upgrade-controller/charts/values.yaml
```diff
@@ -0,0 +1,13 @@
...
+
+systemUpgradeController:
+  image:
+    repository: rancher/system-upgrade-controller
+    tag: v0.14.2
+
+kubectl:
+  image:
+    repository: rancher/kubectl
+    tag: v1.31.1
```
snasovich, line 13: That's quite a jump from 1.23.3 we've used for a good while! Should be fine and actually help with CVEs and such.

packages/system-upgrade-controller/charts/templates/deployment.yaml
snasovich, line 1: Thank you for updating deployment to match latest manifest from "upstream" - https://github.com/rancher/system-upgrade-controller/blob/master/manifests/system-upgrade-controller.yaml
Looks like we've failed to update one in `system-charts`

snasovich (approved):
LGTM - only one comment to potentially address (about missing `version`) - others are more observations/kudos.

It's a bit weird to see chart version without `+up...` part but I see logic here - we don't really have upstream chart to reference there. We could reference SUC version there but I can see reasons for both doing and not doing that.

packages/system-upgrade-controller/package.yaml
jiaqiluo: I dropped it because the version will be read from the chart.yaml file in the local chart when it is not specified in the package.yaml file.

The file was updated to include the `version` for consistency.

packages/system-upgrade-controller/charts/Chart.yaml
jiaqiluo: It was removed.

snasovich (approved):

PR #4634, opened 2024-10-22
[dev-2.10] Add new vsphere-cpi and vsphere-csi charts

jiaqiluo:
https://github.com/rancher/rancher/issues/46836
https://github.com/rancher/rancher/issues/46837

This PR adds the following 2 new charts:
- rancher-vsphere-cpi  105.0.0+up1.9.1
- rancher-vsphere-csi  105.0.0+up3.3.1-rancher6

Both charts
- are generated based on the same commit in the upstream repo to support k8s 1.31 
- update the `catalog.cattle.io/rancher-version` annotation to enable the chart for rancher v2.10.0+ only 
- update the chart version to a new major release line  (105.x.x)


packages/rancher-vsphere/rancher-vsphere-cpi/generated-changes/patch/Chart.yaml.patch
```diff
@@ -1,11 +1,11 @@
...
+-  catalog.cattle.io/rancher-version: '>= 2.9.0-0'
++  catalog.cattle.io/rancher-version: '>= 2.10.0-0'
    catalog.cattle.io/release-name: vsphere-cpi
  apiVersion: v1
--appVersion: 1.30.1
-+appVersion: 1.8.0
- description: vSphere Cloud Provider Interface (CPI)
- icon: https://charts.rancher.io/assets/logos/vsphere-cpi.svg
- keywords:
+ appVersion: 1.31.0
```
HarrisonWAffel: Shouldn't this equal the `version` of the CPI chart, `1.9.1`? That's the convention the prior version of this chart used with `1.8.0`.

HarrisonWAffel (changes requested):

packages/rancher-vsphere/rancher-vsphere-cpi/generated-changes/patch/Chart.yaml.patch
jiaqiluo: I believe the value was set incorrectly in the past, as the `appVersion` should reflect the version of the application that the chart contains, according to the [Helm documentation](https://helm.sh/docs/topics/charts/#the-chartyaml-file). In this case, the upstream chart's value of 1.30.1 is also outdated, so I'll need to update it to 1.31.0.

I am OK with either value. Which one is the better choice here?

HarrisonWAffel: I think keeping it as `1.9.1` would be better, as it allows us to track minor changes to the chart which do not require  a bump to the appVersion. For instance, changing `questions.yaml`

jiaqiluo: The file was updated to use `1.9.1`.

HarrisonWAffel (approved):

release.yaml
```diff
@@ -56,7 +56,9 @@ rancher-provisioning-capi:
   - 105.0.0+up0.4.0
 rancher-vsphere-cpi:
   - 104.0.1+up1.8.1
```
snasovich, line 58: We probably want to remove 104.x.x charts from here - looks like they were added by https://github.com/rancher/charts/pull/4559 and I'm not sure it was intentional.
@nicholasSUSE , could you please I'm not missing a reason why we need those here.

release.yaml
```diff
@@ -56,7 +56,9 @@ rancher-provisioning-capi:
...
+  - 105.0.0+up1.9.1
 rancher-vsphere-csi:
   - 104.0.1+up3.3.0-rancher2
```
snasovich, line 61: Same as comment on `rancher-vsphere-cpi`

snasovich (approved):
LGTM provided `release.yaml` comments are resolved; please count this approval if that's the only change done.

nicholasSUSE (approved):
Approved.

If the Build CI passes, we are good. 
I will be doing all of the forward-ports to v2.10 from the last 2 releases this week.

release.yaml
snasovich: This is expected per https://github.com/rancher/charts/pull/4634#pullrequestreview-2393557550

PR #4695, opened 2024-11-05
[dev-v2.10] rancher-logging 105.1.0-rc.1+up4.10.0 create

jbiers:
## Issue: https://github.com/rancher/rancher/issues/47783

NOT TO BE MERGED UNTIL AFTER V2.10 RELEASE!


assets/rancher-logging-crd/rancher-logging-crd-105.0.0-rc.1+up4.8.0.tgz
joshmeranda, line 1: We want to unrc and leave the old version to help users with their upgrade path.

Since we are waiting until after this chart will be unrcd and released anyway, you can probably even remove the commits to remove these older charts since that will be basically handled for you anyway.

joshmeranda (changes requested):

assets/rancher-logging-crd/rancher-logging-crd-105.0.0-rc.1+up4.8.0.tgz
joshmeranda: Also will want to bump to `105.1.0`

charts/rancher-logging-crd/105.1.0-rc.1+up4.10.0/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+annotations:
+  catalog.cattle.io/certified: rancher
```
joshmeranda, line 2: looks like we're missing a log of the chart annotations I would expect to see here, see [105.0.0+up4.8.0](https://github.com/rancher/charts/blob/dev-v2.10/charts/rancher-logging/105.0.0%2Bup4.8.0/Chart.yaml#L2-L14) for example.

Looks like these are still included in `packages/rancher-logging/generated-changes/patch/Chart.yaml.patch` so not sure why these aren't being included here.

joshmeranda (changes requested):

charts/rancher-logging-crd/105.1.0-rc.1+up4.10.0/Chart.yaml
jbiers: Are you sure these annotations should be included for the crd chart? Looking [here](https://github.com/rancher/charts/blob/dev-v2.10/charts/rancher-logging-crd/105.0.0%2Bup4.8.0/Chart.yaml) doesn't look like they are. Let me know if I'm missing something ;)

joshmeranda: Ope my mistake, thought I was looking at the main chart, this is good as it is

joshmeranda (approved):

PR #4710, opened 2024-11-07
Adding Istio to 2.10

diogoasouza:
This PR adds the `rancher-istio` chart to 2.10. 
The charts are the same as the ones present in 2.9, with changes to annotations only.

versions:
 - 105.0.0+up1.19.6
  - 105.1.0+up1.20.3
  - 105.2.0+up1.21.1
  - 105.3.0+up1.22.1
  - 105.4.0+up1.23.2


charts/rancher-istio/104.4.1+up1.22.1/Chart.yaml
```diff
@@ -11,14 +11,14 @@ annotations:
...
+appVersion: 1.22.1
 description: A basic Istio setup that installs with the istioctl. Refer to https://istio.io/latest/
   for details.
 icon: https://charts.rancher.io/assets/logos/istio.svg
 keywords:
 - networking
 - infrastructure
 name: rancher-istio
-version: 1.18.2
+version: 104.4.1+up1.22.1
```
recena, line 24: Are we changing the versioning?

rohitsakala (approved):

charts/rancher-istio/104.4.1+up1.22.1/Chart.yaml
diogoasouza: Not really sure why git is comparing these two files, this is a forward-port of [this](https://github.com/rancher/charts/blob/dev-v2.9/charts/rancher-istio/104.4.1%2Bup1.22.1/Chart.yaml), so we are not changing the versioning.

PR #4750, opened 2024-11-13
[dev-v2.9] Bump harvester csi driver to v0.1.21

Vicente-Cheng:
## Problem

We need to bump harvester csi driver v0.1.21 to resolve some backward compatibility and corner cases.

## Solution

Bump Harvester CSI driver


FrankYang0529 (approved):
LGTM. Thanks.

PR #4760, opened 2024-11-13
[release-v2.10] adding slsa file for bypassing for chart slsa compliance

nicholasSUSE:
Adding `slsa.yaml` file because of this: https://github.com/rancher/charts-build-scripts/pull/153


tashima42 (approved):

pjbgf (approved):

pjbgf:
Should this be kept at the default branch instead of on a release branch? If this is branch-specific, we may have issues when that configuration is out of sync across all the release branches for a given branch - as the configuration is at image-level.

nicholasSUSE:
> Should this be kept at the default branch instead of on a release branch? If this is branch-specific, we may have issues when that configuration is out of sync across all the release branches for a given branch - as the configuration is at image-level.

No, this should only be kept in release branches. 

The regsync job is only run on release branches. 
We never sync to the prime registry at development level because there are lot of rc images going on there. 

The sync of the regsync.yaml file won't be lost across different branches like (release-v2.9 and release-v2.10). 
However, I am assuming that from this point on, all future cis-operator images will be SLSA. Therefore, they should never be synced to the prime registry again. 

Am I right in this assumption?

pjbgf:
> Am I right in this assumption?

Yes, that's correct. We sign and publish Prime images for all the currently supported release lines for cis-operator.

PR #4762, opened 2024-11-14
[dev-v2.10] k8s 1.31 support for harvester-cloud-provider

FrankYang0529:
## Issue: 

https://github.com/rancher/rancher/issues/46848

## Problem

Support k8s 1.31.


Vicente-Cheng (approved):
LGTM, thanks!

PR #4783, opened 2024-11-25
[dev-v2.10] rancher-pushprox image update, rancher-monitoring 105.1.1-rc.1+up61.3.2 update

mallardduck:
## Issue: 
This charts updates should address, this issue: https://github.com/rancher/rancher/issues/47766

We're opting to bump these minor dependency things before doing the next big rebase for monitoring.


mallardduck:
@joshmeranda - I can't recall right now exactly how commits should look and if this is ok. Normally I do a `make remove` before the `make charts`, however I recalled part of us not bumping RC version is to: a) make git diffs more reasonable to understand, and b) reduce creating many unnecessary git objects. I think this process created the most optimal for both those factors, but LMK if there's any changes needed.

joshmeranda (approved):

PR #4786, opened 2024-11-27
[dev-v2.10] rancher-cis-benchmark auto-bump 105.0.1

nicholasSUSE:
Auto bumping rancher-cis-benchmark


release.yaml
```diff
@@ -23,8 +23,10 @@ neuvector-monitor:
   - 105.0.0+up2.8.3
 rancher-cis-benchmark:
   - 105.0.0+up7.0.0
+  - 105.0.1+up7.0.1
 rancher-cis-benchmark-crd:
   - 105.0.0+up7.0.0
```
mitulshah-suse, line 27: @nicholasSUSE this version is already released right? shouldnt it be removed already?

nicholasSUSE: Yes it should be removed but it does not break anything. 
I will be removing it.

nicholasSUSE: Done and thanks for reviewing!

mitulshah-suse (approved):
LGTM

PR #4872, opened 2024-12-13
[dev-v2.11] prometheus-federator 105.1.0-rc.1+up0.4.5-rc.1

mallardduck:
Per title, this adds a new PromFed RC


packages/rancher-monitoring/prometheus-federator/generated-changes/patch/charts/helmProjectOperator/values.yaml.patch
```diff
@@ -1,17 +1,17 @@
...
  
  image:
--  repository: ghcr.io/rancher/helm-project-operator
--  tag: v0.3.0
-+  repository: rancher/prometheus-federator
-+  tag: ''
+-  registry: ghcr.io
+-  repository: rancher/prometheus-federator/helm-project-operator
++  registry: ''
++  repository: ''
```
alexandreLamarre, line 10: Could be looking at this wrong, but doesn't look right. We should be setting this (I think)

mallardduck: I opted to set these as empty - as this completely will exclude this image from release artifacts (like `rancher-image-origins.txt`). Since we never want end-users touching this helps remove those remaining traces of them.

For end users, the values for this should only come from the main defaults in `prometheus-federator/values.yaml` or their specific values. Where as the origin `helm-project-operator` chart defaults to the `ghcr.io` container for HPO.

mallardduck: Example: https://github.com/rancher/rancher/releases/tag/v2.10.1-alpha1
Grab `rancher-images-origins.txt` from the release and it will have the HPO image listed.

By removing these from values at `rancher/charts` repo it will prevent the HPO image from ending up in releases or security scans. However the PromFed images customers actually use will still be in both.

packages/rancher-monitoring/prometheus-federator/generated-changes/patch/Chart.yaml.patch
```diff
@@ -12,4 +12,8 @@
 +  catalog.cattle.io/rancher-version: '>= 2.10.0-0 < 2.11.0-0'
    catalog.cattle.io/release-name: prometheus-federator
  apiVersion: v2
- appVersion: 0.4.3
+-appVersion: v0.4.4-rc.1
++appVersion: v0.4.4-rc.2
```
joshmeranda, line 16: I'm confused by the versioning here. I thought we'd decided to not bump rc numbers for monitoring charts? Plus we have the `release.yaml` showing `105.1.0-rc.1+up0.4.5-rc.1` but all these are showing `-rc.2`

joshmeranda (changes requested):

packages/rancher-monitoring/prometheus-federator/generated-changes/patch/Chart.yaml.patch
mallardduck: `appVersion` in helm should represent the version of the image/binary, where as `version` is about the chart itself. These don't need to match for any technical reason. Note this is also why it's chart version `0.4.5` (with an RC) and image tag `0.4.4` (with an RC).

I don't believe we have a policy based requirement that App and Chart versions match 1:1. And if we do then it creates a chicken-egg problem in the current `prometheus-federator` repo - as charts values would need reference versions of images that don't exist.

As for comparing the chart version, `rancher/charts` appends the "upstream chart version" onto the end. So in cases where the image version (`appVersion`) and chart version do not match it's expected to see a difference there.

I was under the impression the "no bump RC for monitoring" was just for the monitoring charts from upstream. I thought that since PromFed is something we produce completely that we do full RC process?

alexandreLamarre (approved):

packages/rancher-monitoring/prometheus-federator/generated-changes/patch/Chart.yaml.patch
joshmeranda: Sounds good to me, I was confused / just wanted to confirm since the versions were so close to each other

joshmeranda (approved):

PR #4907, opened 2024-12-19
[dev-v2.10] rancher-logging 105.2.0-rc.1+up4.10.0 update

jbiers:
## Problem
This PR is a fix for the internal issue SURE-9146 which requests the possibility of setting a different number of replicas and resources for the fluentd Statefulset of each logging, as well as different resources and loggingRef for their fluentbitagent. 

This is currently not possible on a per-logging basis, a user can only update these values for all loggings at once using values.loggingOverlay or values.fluentbit+values.fluentd.

## Solution
My current solution is simply to make these values configurable for each logging via the additionalLoggingSource.logging-name.fluentd and additionalLoggingSource.logging-name.fluentbit fields. It seems to me like this has been done to some capacity in the past for fields like additionalLoggingSource.kubeAudit.fluentbit, so my choice was to continue with this mechanism of exposing values.

Other solutions are viable, for instance using a mechanism similar to what is done in the loggingServiceAccountAnnotations field. However I chose to continue with the idea that was already implemented in the past.

## Backporting considerations
Backport to 2.9 and forwardport to 2.11


mallardduck (approved):
Tested local in RKE2 and with help (thanks Dave) in AKS.

PR #4909, opened 2024-12-23
[dev-v2.11] prometheus-federator update

mallardduck:
Update the RC after initial QA validation.

NOTE: Just realized that I've been using a 2.10 version for the chart, I think 2.11 should have been a 106 chart? No sweat either way since 2.11 is pretty far out and this can be adjusted to correct version as needed.

Due to this when I backport for 2.10 I'll just reuse the same version as the one here. And if we do a 2.9 backport it would decrement to `104` base version.


mallardduck:
This chart did pass QA to be able to un-RC however as @joshmeranda pointed out I did the double RC thing on accident. Something we have an exception for with monitoring related charts. Also I noticed after the process that the version I picked would match 2.10 not 2.11.

Due to those hiccups I'm going to opt to remove the chart and add a non-RC'd one in a follow up PR. Follow up to remove charts here: https://github.com/rancher/charts/pull/4930

PR #4912, opened 2024-12-23
[dev-v2.10] rancher-vsphere-csi 105.1.0+up3.3.1-rancher7 add

pedromfcarvalho:
This incorporates this upstream change:

- https://github.com/rancher/vsphere-charts/pull/96

Issue: https://github.com/rancher/rancher/issues/48053

### Engineeering testing

For testing that this latest bump had an effect, I installed the csi chart in the local cluster, even though it wasn't a vsphere cluster, because it was still possible to verify that arguments could be passed. To do this, I customized the following values before installing the chart (directly in the value yaml view):

```
csiController:
  csiResizer:
    enabled: true
  image:
    additionalArgs: ["-version"]
    csiAttacher:
      additionalArgs: ["--logging-format=json"]
    csiProvisioner:
      additionalArgs: ["--version"]
    csiResizer:
      additionalArgs: ["--logging-format=json"]
    csiSnapshotter:
      additionalArgs: ["-version"
    livenessProbe:
      additionalArgs: ["--logging-format=json"]
    vsphereSyncer:
      additionalArgs: ["-version"]

csiNode:
  image:
    additionalArgs: ["-version"]
    livenessProbe:
      additionalArgs: ["--logging-format=json"]
    nodeDriverRegistrar:
      additionalArgs: ["--logging-format=json"]

blockVolumeSnapshot:
  enabled: true
```

Note that these values have to be configured individually, and the other ones left as they are. The point is to pass additional args to every container. It's also necessary to enable the csi resizer and block volume snapshots so two of the containers run.

After this, I checked the logs every container for one controller deployment and the single node deployment (a total of 10 containers). Those that were passed `--logging-format=json` had json output in their logs, and those that were passed `-version/--version` printed their version before exiting. Most containers failed because this was not run in a vsphere environment, but it was sufficient to verify that the arguments were being passed correctly.

### QA considerations

Test suggestions:

- The checks from engineering testing above, to verify the effects of this bump
- Any other usual vsphere CSI tests not specific to this bump to check for regressions


packages/rancher-vsphere/rancher-vsphere-csi/package.yaml
```diff
@@ -1,4 +1,4 @@
 url: https://github.com/rancher/vsphere-charts.git
 subdirectory: charts/rancher-vsphere-csi
-commit: a5c99d716df960dc50cf417d9ecffad6b55ca0ad
-version: 105.0.0
```
jiaqiluo, line 4: What do you think about bumping the version to `105.0.1`, since the upstream chart version has changed from `-rancher6` to `-rancher7`?

jiaqiluo (changes requested):

packages/rancher-vsphere/rancher-vsphere-csi/package.yaml
HarrisonWAffel: In his last commit he is bumping it to `105.1.0`, which follows his 2.9 bump to `104.1.0`

jiaqiluo: Yes, I noticed that he tried to maintain consistent version bumping across release lines. However, in the 2.9 line, the upstream chart was bumped from `3.3.0-rancher2` to `3.3.1-rancher7`, so it made sense to bump the minor version ( 104.**0**.1 -> 104.**1**.0). That's not the case here, so I suggest bumping the patch version instead.

HarrisonWAffel: Oh my bad, yes that makes more sense! Though technically this bump is introducing a new feature to the chart

pedromfcarvalho: Thanks!

I ended up changing the patch following the first comment, following the same rationale that Jack mentioned above.

But it is true that technically this is a feature, although I don't know how strictly we are following semver conventions here.

I can change it back if needed.

release.yaml
```diff
@@ -16,3 +16,5 @@ rancher-logging:
   - 105.2.0-rc.1+up4.10.0
 rancher-logging-crd:
   - 105.2.0-rc.1+up4.10.0
+rancher-vsphere-csi:
+  - 105.1.0+up3.3.1-rancher7
```
jiaqiluo:
```suggestion
rancher-vsphere-csi:
  - 105.0.1+up3.3.1-rancher7
```

jiaqiluo (changes requested):

release.yaml
pedromfcarvalho: Nice catch, I missed this!

jiaqiluo (approved):

HarrisonWAffel (approved):

PR #4935, opened 2025-01-09
[dev-v2.8] bump harvester cloud provider v0.2.9

FrankYang0529:
## Problem

Add system default registry for kube-vip.

## Solution

Bump harvester-cloud-provider


Vicente-Cheng (approved):
lgtm, thanks!

PR #4987, opened 2025-01-17
[dev-v2.11] rancher-logging 106.0.0-rc.1+up4.10.0 create

apoorvajagtap:
## Issue: https://github.com/rancher/rancher/issues/48443
ForwardPort of #4876 

## Solution
This PR updates the fluent-bit image for windows to v3.1.8.


charts/rancher-logging/106.0.0-rc.1+up4.10.0/Chart.yaml
```diff
@@ -0,0 +1,30 @@
+annotations:
+  catalog.cattle.io/auto-install: rancher-logging-crd=match
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/deploys-on-os: windows
+  catalog.cattle.io/display-name: Logging
+  catalog.cattle.io/kube-version: '>= 1.28.0-0 < 1.32.0-0'
+  catalog.cattle.io/namespace: cattle-logging-system
+  catalog.cattle.io/permits-os: linux,windows
+  catalog.cattle.io/provides-gvr: logging.banzaicloud.io.clusterflow/v1beta1
+  catalog.cattle.io/rancher-version: '>= 2.10.0-0 < 2.11.0-0'
```
mallardduck: Oh yeah - for 2.11 this forward port needs an additional change to patches to change this. Instead of this it should be:

```
>= 2.10.0-0 < 2.12.0-0
```

For now we can skip bumping the minimum side - that lets us test on 2.10 still. I think main shows as 2.11 now but just in case it's 2.10 this will allow it to work either way. Then closer to release - since we'll have more RCs likely anyway - we can bump the minimum to 2.11.

mallardduck (changes requested):

packages/rancher-logging/generated-changes/patch/values.yaml.patch
```diff
@@ -130,14 +130,14 @@
 +  nodeagent_fluentbit:
 +    os: "windows"
 +    repository: rancher/fluent-bit
-+    tag: 2.2.0
++    tag: 3.1.8
```
jbiers, line 133: Looks good overall. I see that this PR seems to include changes other than the fluent-bit tag itself, but it's only white spaces as it looks like, perhaps your IDE is doing some formatting?
After fixing what Dan is pointing out it should be good to go :smile:

apoorvajagtap: Hey, Thanks for catching that!
Indeed, it was my IDE. I have disabled the auto-indentation now, so shouldn't bring in unnecessary changes to the patch now.

apoorvajagtap: Ah! it still does, wasn't the scene when I checked locally. I guess something got reverted.

mallardduck: Oh yeah a lot better now - way less files. Great work.

There might still be some extra changes we're not expecting but that could be caused by something else. @jbiers looking at the 2.10 PR you did it should only be this tag and the `fluentbit_debug` one? 🤔

jbiers: Ideally the change should only include this tag - [here's](https://github.com/rancher/charts/pull/4876/commits/43abd84c421ad3a50e74859f9a4c88ac8fedbc4b) my commit to the 2.10 PR for reference

packages/rancher-logging/generated-changes/patch/Chart.yaml.patch
```diff
@@ -10,7 +10,7 @@
 +  catalog.cattle.io/namespace: cattle-logging-system
 +  catalog.cattle.io/permits-os: linux,windows
 +  catalog.cattle.io/provides-gvr: logging.banzaicloud.io.clusterflow/v1beta1
-+  catalog.cattle.io/rancher-version: '>= 2.10.0-0 < 2.11.0-0'
++  catalog.cattle.io/rancher-version: '>= 2.10.0-0 < 2.12.0-0'
```
nicholasSUSE:
```
+  catalog.cattle.io/rancher-version: '>= 2.11.0-0 < 2.12.0-0'
```

apoorvajagtap: @nicholasSUSE thanks for highlighting this, but we are considering to bump the min version in upcoming rc updates. ref: https://github.com/rancher/charts/pull/4987#discussion_r1927072419

@mallardduck let me know, if we'd like to bump the min now itself.

packages/rancher-logging/generated-changes/patch/values.yaml.patch
apoorvajagtap: Hello, finally I think the patch has no additional updates. PTAL.

jbiers (approved):
LGTM

packages/rancher-logging/generated-changes/patch/Chart.yaml.patch
mallardduck: @apoorvajagtap - Lets take @nicholasSUSE advice and update it now.

I think I was suspecting we'd follow up with more changes and additional PRs were we'd do it. However the current release cycle is nearly over and it's a small window for the next one. So we're probably less likely to do logging follow ups than I originally expected.

apoorvajagtap:
Hello @jbiers @mallardduck I have incorporated all the suggested changes. I belive the PR is in a good shape now. PTAL.

mallardduck (approved):

mallardduck (approved):

PR #5000, opened 2025-01-21
[dev-v2.10] Add chart for Supportability Review Operator

masap:
## Issue: 

This PR will add chart for Supportability Review Operator.


nicholasSUSE (approved):
I am giving my approval, this is the first chart, it will be used for testing. 

There is still a situation to resolve in the failing job because of an image stored at ghcr. 
We are going to investigate further on what do with it.

PR #5010, opened 2025-01-22
[dev-v2.9] rancher-logging 104.1.3+up4.8.0 update

jmeza-xyz:

`release.yaml`:

`Chart.yaml and index.yaml`:
  - kube-version annotation
  - rancher-version annotation

---

## Solution

This PR backports the changes from https://github.com/rancher/charts/pull/4907 to dev-2.9 branch.


packages/rancher-logging/generated-changes/patch/values.yaml.patch
```diff
@@ -126,7 +126,7 @@
 +  fluentd:
 +    repository: rancher/mirrored-kube-logging-fluentd
 +    tag: v1.16-4.8-full
```
mallardduck, line 128: Does this tag apply to this version of logging?

In this case the 2.9 chart is still targeting 4.8.0 of upstream charts. Whereas the 2.10 charts is tracking 4.10.0, so we may want to take care when backporting to avoid any changes there that are specific to 2.10's chart.

mallardduck (changes requested):
Just one small note to look into. Overall looks correct other than that. But @jbiers is the definitive authority here as this was originally her changes.

Note: One obvious issue is `release.yaml` but that doesn't matter right now. This can't merge until after current release cycle which will change that anyway. So fixing now will just mean needing to fix again later.

packages/rancher-logging/generated-changes/patch/values.yaml.patch
jbiers: It looks like this image tag is not actually being changed, as it's the same as the 2.9 branch. I think these (+) signs are only showing that we add these lines to the upstream chart, not that these are being added in this PR

jbiers (approved):

jbiers (changes requested):
I had actually approved the PR but for some reason CI is not passing. It seems like an issue with GA or Vault. I'm removing the approval temporarily just to prevent this from being merged while CI is not working.

Also, created an issue to track this backport so that QA can use it to test it and linked in the description.

jmeza-xyz:
Sounds good! Sadly the CI build error doesn't really help debug. Let me know if there is anything else I can do help here. Thanks!

jbiers:
Ended up mistakenly opening another PR to fix this issue. It was merged here: https://github.com/rancher/charts/pull/5037

jmeza-xyz:
> Ended up mistakenly opening another PR to fix this issue. It was merged here: #5037

No worries! Thanks for the update, will close this PR. Appreciate the backport work, thanks!

PR #5061, opened 2025-01-31
Fix duplicate fluentbitagent tolerations

skanakal:

`release.yaml`:

`Chart.yaml and index.yaml`:
  - kube-version annotation
  - rancher-version annotation

---

##### Issue: 

https://github.com/rancher/rancher/issues/48917

##### Solution

set the tolerations based on the kubernetes distribution

Already mentioned in the above issue


charts/rancher-logging/106.0.0-rc.1+up4.10.0/templates/_helpers.tpl
```diff
@@ -94,6 +94,24 @@ true
...
+{{/*
+Set tolerations based on Kubernetes distribution and merge with values.yaml
+*/}}
+{{- define "customTolerations" -}}
+{{- $isRKE2orK3S := or .Values.additionalLoggingSources.rke2.enabled .Values.additionalLoggingSources.k3s.enabled -}}
+{{- $defaultTolerations := list -}}
+{{- if $isRKE2orK3S }}
+  {{- $defaultTolerations = append $defaultTolerations (dict "key" "node-role.kubernetes.io/control-plane" "operator" "Exists" "effect" "NoSchedule") -}}
+{{- else }}
+  {{- $defaultTolerations = append $defaultTolerations (dict "key" "node-role.kubernetes.io/controlplane" "operator" "Exists" "effect" "NoSchedule") -}}
```
alexandreLamarre: This doesn't just include RKE1, but EKS and GKE deployments as well, but I don't know off the top of my head if this affects those types of deployments

skanakal: yeah that does as the DS managed by fluentbitagent... I have now modified the logic, and I believe it will now reflect the current behavior—using "control-plane" for the rest of the kubernetes distribution. only, thing extra "controlplane" for rke based clusters.

adamkpickering:
Hi @skanakal, a couple of points:
- As you might be aware, we now work on Rancher Logging from [rancher/ob-team-charts](https://github.com/rancher/ob-team-charts/tree/main). If you could resubmit this PR there, that would be great! Alternatively, I can put one together.
- I noticed that we completely misuse `.Values.tolerations`. In the upstream chart `.Values.Tolerations` is used to set tolerations on the rancher-logging operator Deployment, not fluentbit or anything else. That is, unless the logging operator picks up that value and propagates it to the resources that *it* creates. But I think that is unlikely. I think that we should consider returning `.Values.tolerations` to its default of `[]`, or at least not referencing it when setting tolerations on fluentbit.
- I think that rather than introducing logic into `_helpers.tpl`, we should stay as close to the templates being templates as possible, i.e. avoid introducing logic. The way to do this is to set `fluentbit.tolerations` in `values.yaml` as follows:
```yaml
fluentbit:
  ...
  tolerations:
    - key: node-role.kubernetes.io/controlplane
      value: "true"
      effect: NoSchedule
    - key: node-role.kubernetes.io/control-plane
      value: "true"
      effect: NoSchedule
    - key: node-role.kubernetes.io/etcd
      value: "true"
      effect: NoExecute
```
This does mean that we have the `node-role.kubernetes.io/controlplane` toleration in cases where we do not need it (this is most cases I think, because the community switched to `node-role.kubernetes.io/control-plane` a long time ago). But I think this is better than introducing logic and indirection.

Let me know what you think :smile:

skanakal:
> Hi @skanakal, a couple of points:
> 
> * As you might be aware, we now work on Rancher Logging from [rancher/ob-team-charts](https://github.com/rancher/ob-team-charts/tree/main). If you could resubmit this PR there, that would be great! Alternatively, I can put one together.

Here is the pr opened in ob-team-charts https://github.com/rancher/ob-team-charts/pull/74

> * I noticed that we completely misuse `.Values.tolerations`. In the upstream chart `.Values.Tolerations` is used to set tolerations on the rancher-logging operator Deployment, not fluentbit or anything else. That is, unless the logging operator picks up that value and propagates it to the resources that _it_ creates. But I think that is unlikely. I think that we should consider returning `.Values.tolerations` to its default of `[]`, or at least not referencing it when setting tolerations on fluentbit.
> * I think that rather than introducing logic into `_helpers.tpl`, we should stay as close to the templates being templates as possible, i.e. avoid introducing logic. The way to do this is to set `fluentbit.tolerations` in `values.yaml` as follows:
> 
> ```yaml
> fluentbit:
>   ...
>   tolerations:
>     - key: node-role.kubernetes.io/controlplane
>       value: "true"
>       effect: NoSchedule
>     - key: node-role.kubernetes.io/control-plane
>       value: "true"
>       effect: NoSchedule
>     - key: node-role.kubernetes.io/etcd
>       value: "true"
>       effect: NoExecute
> ```
> 
> This does mean that we have the `node-role.kubernetes.io/controlplane` toleration in cases where we do not need it (this is most cases I think, because the community switched to `node-role.kubernetes.io/control-plane` a long time ago). But I think this is better than introducing logic and indirection.
> 
> Let me know what you think 😄

The problem is, if you add `control-plane` and `cotnrolplane` tolerations , it adds both of them which is not needed for k3s/rke2(control-plane) or rke(needs only controlplane) and sets duplicate tolertions... see the issue point #3

PR #5062, opened 2025-01-31
[dev-v2.9] rancher-logging 104.1.3+up4.8.0 create

apoorvajagtap:

`release.yaml`:

`Chart.yaml and index.yaml`:
  - kube-version annotation
  - rancher-version annotation

---

##### Issue: https://github.com/rancher/rancher/issues/48805
Backport of #4752 

#### Additional Description:
- Fixes the unknown field 'spec.securityContext.seLinuxOptions'


mallardduck (approved):
LGTM

PR #5064, opened 2025-01-31
[dev-v2.11] rancher-monitoring 106.0.0+up61.3.2 update

apoorvajagtap:

`release.yaml`:

`Chart.yaml and index.yaml`:
  - kube-version annotation
  - rancher-version annotation

---

#### Additional detail:
Bumps k8s supported version from 1.30.0 to 1.32.0


apoorvajagtap:
I did not bump the k8s version for `prometheus-federator`, as I believe that should be done separately per [rancher/rancher#48828](https://github.com/rancher/rancher/issues/48828)?
Let me know, in case we would like to incorporate k8s bump for prometheus-federator chart in this PR itself.

alexandreLamarre (approved):

jbiers (approved):

charts/rancher-monitoring/106.0.0+up61.3.2/Chart.yaml
```diff
@@ -10,7 +10,7 @@ annotations:
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/deploys-on-os: windows
   catalog.cattle.io/display-name: Monitoring
-  catalog.cattle.io/kube-version: '>= 1.28.0-0 < 1.32.0-0'
+  catalog.cattle.io/kube-version: '>= 1.30.0-0 < 1.33.0-0'
```
jbiers, line 13: I had approved the PR but now I see something. I believe we should also be bumping the **catalog.cattle.io/rancher-version** annotation to be '>= 2.11.0-0 < 2.12.0-0' for all subcharts

jbiers (changes requested):
We should bump the rancher-version annotation too

jbiers (approved):
CI is failing but not due to changes in this PR. Approved.

PR #5013, opened 2025-01-22
[dev-v2.11] rancher-alerting 106.0.0 create

mallardduck:

PR #5073, opened 2025-02-03
[dev-v2.11] rancher-alerting-drivers 106.0.0-rc.1 update

alexandreLamarre:
Related issue : https://github.com/rancher/rancher/issues/48831

#### Pull Requests Rules

- Pull request title:
    ```
    ```

---

`release.yaml`:

`Chart.yaml and index.yaml`:
  - kube-version annotation
  - rancher-version annotation

---


charts/rancher-alerting-drivers/106.0.0-rc.1/Chart.yaml
```diff
@@ -0,0 +1,29 @@
+annotations:
+  catalog.cattle.io/certified: rancher
+  catalog.cattle.io/display-name: Alerting Drivers
+  catalog.cattle.io/kube-version: '>= 1.28.0-0 < 1.32.0-0'
```
jbiers: Maybe we could use this PR to also bump the kube-version and rancher-version annotations to fit the requirements for 2.11 closing https://github.com/rancher/rancher/issues/48831

alexandreLamarre: Sure, but requires some sanity testing, will update in a sec

jbiers: I'm assuming rancher-version wasn't updated to allow testing with rancher 2.10, is that right? I don't have a problem with that, just want to make sure I understood the thought process for not including it in this PR ;)

alexandreLamarre: ah no good catch it just slipped my mind

jbiers (approved):

charts/rancher-alerting-drivers/106.0.0-rc.1/values.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+global:
+  cattle:
+    psp:
+      enabled: false
+    # the registry where all images will be pulled from
+    systemDefaultRegistry: ""
+  kubectl:
+    repository: rancher/kubectl
+    tag: v1.20.2
```
mallardduck, line 13: This needs to be updated to use Kuberlr if we're going to use your PR instead of the one I prepared. I've closed the PR I had, so yours should use the same value: https://github.com/rancher/charts/pull/5013/commits/a01773164c4d481db86802bac91adcdc2c775ec3

mallardduck (changes requested):

mallardduck:
Because I closed the other PR that this was conflicting with, this PR should be made to cover this issue as well: https://github.com/rancher/rancher/issues/48849

@alexandreLamarre - I've assigned you as the owner of that issue now that the PR i created for it will no longer be used.

mallardduck:
Closing as the PR I created to incorporate this is merged instead.

PR #5107, opened 2025-02-06
[dev-v2.11] rancher-backup  106.0.0+up7.0.0-alpha.4 update

alexandreLamarre:
#### Related issues

https://github.com/rancher/rancher/issues/48832

##### QA Testing Considerations

- BRO on k8s 1.32, particular attention to regressions regarding encryption configs

PR #5108, opened 2025-02-06
[dev-v2.11] rancher-backup 106.0.0+up7.0.0-alpha.3 update

apoorvajagtap:

`release.yaml`:

`Chart.yaml and index.yaml`:
  - kube-version annotation
  - rancher-version annotation

---

#### Additional detail:
bumps k8s and rancher supported version for rancher-backup.


mallardduck:
Hey @apoorvajagtap - this looks very good, however it's best to do this change at https://github.com/rancher/backup-restore-operator directly and then release an RC. This is because we fully control/produce the project so doing it there will prevent us needing any patch here.

I believe that @alexandreLamarre just made a similar release and PR - so you can review that and make changes in a similar way in the BRO repo. https://github.com/rancher/backup-restore-operator/blob/337159ca1f7d40fadf264336f4fbbef10334ddcf/charts/rancher-backup/Chart.yaml#L10

Once you make a PR there and we merge it we can create another non-stable tag (alpha/rc), then the PR would look similar to: https://github.com/rancher/charts/pull/5107

mallardduck:
Hiya, I'm going to close this out per the comment I left above.

PR #5113, opened 2025-02-07
[dev-v2.11] Bump rancher webhook to 0.7.0-rc.3

crobby:
#### Pull Requests Rules

- Pull request title:
    ```
    ```

---

`release.yaml`:

`Chart.yaml and index.yaml`:
  - kube-version annotation
  - rancher-version annotation

---


tomleb (approved):

release.yaml
```diff
@@ -64,3 +64,5 @@ rancher-monitoring-crd:
   - 106.0.0+up61.3.2
 rancher-vsphere-csi:
   - 106.0.0+up3.3.1-rancher7
+rancher-webhook:
+  - 105.0.1+up0.7.0-rc.3
```
tomleb: Wrong version here

PR #5159, opened 2025-02-17
[v2.11] Remove Fleet alpha version

thardeck:
and clean up release.yaml.

This should allow auto-bump to work properly.


release.yaml
```diff
@@ -2,18 +2,6 @@ elemental:
   - 106.0.0+up1.6.5
 elemental-crd:
   - 106.0.0+up1.6.5
-fleet:
-  - 104.1.2+up0.10.6
-  - 105.0.0+up0.11.0
```
weyfonk, line 7: Why remove these 2 versions as well? Same remark about agent and CRDs charts.

thardeck: Because they should not be there. The release.yaml provides versions which should be released in the next run, which is usually the latest version.

The mentioned versions were already released, so it seems to be a migration issue, why they are still listed.

It should look like this: https://github.com/rancher/charts/blob/dev-v2.10/release.yaml

weyfonk: Fair enough. Thanks for clarifying.
As a side note, I could see on `dev-v2.11` that a few other charts (eg. `rancher-cis-benchmark`) also list multiple versions in `release.yaml`, so there may be a need to raise awareness about this org-wide.

weyfonk (approved):

PR #5213, opened 2025-02-24
[dev-v2.9] rancher-logging 104.2.1-rc.1+up4.8.0 bump

adamkpickering:
This PR simply updates `app-readme.md` for the `rancher-logging` chart.

This PR is part of the solution for https://github.com/rancher/ob-team-charts/issues/23.


mallardduck (commented):
This looks pretty good overall.

Only potential notes:
- Usually charts team wants us to break things into specific commits, (not sure if hard requirement)
- The `release.yaml` will need to be updated for CI to pass.

However, I think we're past dates for this to land in February so aiming for March; meaning these issues probably don't matter to fix until after the release cycle for Feb branch is complete anyway.

mallardduck (approved):
LGTM, just think this needs to correct `release.yaml`.

PR #5214, opened 2025-02-24
[dev-v2.11] Add previously released Istio versions to 2.11

diogoasouza:
Adding previously released Istio versions to 2.11


adamkpickering (changes requested):
I am under the impression that `catalog.cattle.io/kube-version` should be set to `>= 1.30.0-0 < 1.33.0-0` for 2.11 charts. This is something that @mallardduck told me, but I didn't get more details. I'll loop him in.

Looks good to me otherwise.

mallardduck:
Yeah - so technically @adamkpickering is correct about the `kube-version` annotation and the versions 2.11 are expected to support. So to be 100% exact it could be the value Adam shared there.

However, as long as the versions in the charts are wider than the expected range for the Rancher minor we're targeting that's OK too. And since the `catalog.cattle.io/rancher-version` annotation seems correct I think it won't cause issues.

charts/rancher-istio/106.0.0+up1.22.1/Chart.yaml
```diff
@@ -1,24 +1,24 @@
 annotations:
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/display-name: Istio
-  catalog.cattle.io/kube-version: '>= 1.26.0-0 < 1.30.0-0'
+  catalog.cattle.io/kube-version: '>= 1.27.0-0 < 1.31.0-0'
```
mallardduck, line 4: Yeah - I actually see my confusion now. @adamkpickering is right that this max version isn't high enough for 2.11.

As I mentioned I don't think there's an issue if the lower k8s version is even lower than the minimum 2.11 will support. However, there would be an issue if we don't support the maximum version of k8s that 2.11 will have.

Unless this is an intentional choice to limit k8s support scope for this chart based on upstream istio requirements. If that's the case then this point is moot - otherwise, this should be `< 1.33.0-0` instead of 1.31.

mallardduck (changes requested):

charts/rancher-istio/106.0.0+up1.22.1/Chart.yaml
diogoasouza: It was intentional, Istio 1.22 does not [support](https://istio.io/latest/docs/releases/supported-releases/#support-status-of-istio-releases) 1.31 or 1.32.
My understanding is that the chart can be included as long as it supports any k8s version supported by that particular version of Rancher. It won't show up in the UI unless the user is using a version specified in the annotation. I.e. it will show for a user using Rancher v2.11 + k8s 1.30 but won't for a user using Rancher v2.11 + k8s 1.31

mallardduck (approved):

mallardduck (approved):

adamkpickering (approved):

PR #5226, opened 2025-02-25
[dev-v2.10] rancher-logging 105.2.2-rc.1+up4.10.0-rancher.1 bump

adamkpickering:
This PR bumps `rancher-logging` to use the `4.10.0-rancher.1` version from rancher/ob-team-charts.

This is part of the solution for https://github.com/rancher/ob-team-charts/issues/23.


mallardduck (approved):
LGTM, this one also needs to correct `release.yaml` then likely good to go.

PR #5257, opened 2025-02-28
[dev-v2.10] rancher-backup 105.0.1+up6.0.1-rc.1 create

jbiers:


charts/rancher-logging/105.2.1+up4.10.0/values.yaml
```diff
@@ -408,15 +408,22 @@ additionalLoggingSources:
     container_engine: "systemd"
```
diogoasouza: Is modifying this chart intentional? Afair we are not supposed to change two different charts in one PR

jbiers: absolutely not, messed up the PR including some test stuff I had in my local fork lol. fixed now, thanks!

diogoasouza (approved):
lgtm

PR #5300, opened 2025-03-03
[dev-v2.11] Maintenance pull scripts

nicholasSUSE:
Maintenance of `make pull-scripts`


mallardduck:
I think this will be good to go now that https://github.com/rancher/charts/pull/5303 is merged to fix PromFed.

scripts/pull-scripts
```diff
@@ -9,12 +9,26 @@ if ls ../bin/charts-build-scripts 1>/dev/null 2>/dev/null; then
     # Use awk to split the version output correctly and extract the version number
     CURRENT_SCRIPT_VERSION="v$(../bin/charts-build-scripts --version | awk '{print $3}')"
     echo "Current version: ${CURRENT_SCRIPT_VERSION}, Expected version: ${CHARTS_BUILD_SCRIPT_VERSION}"
+
+    # If the version is "latest", fetch the latest release version
+    if [[ "${CHARTS_BUILD_SCRIPT_VERSION}" == "latest" ]]; then
+        CHARTS_BUILD_SCRIPT_VERSION=$(curl --silent "https://api.github.com/repos/rancher/charts-build-scripts/releases/latest" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/')
```
mallardduck: BTW - this is causing me errors in `dev-v2.10` so I suspect it should be fixed before merging in here (and then fixed in 2.10). When I run this line I get:

```
curl --silent "https://api.github.com/repos/rancher/charts-build-scripts/releases/latest" | grep '"tag_name":' | sed -E 's/.*"([^"]+)".*/\1/'
mentions_count
```

Which comes from a field name rather than being the value where field name is matching "tag_name". This seems to be because the JSON is a single line block of text rather than a well formatted human readable JSON file with line returns.

I wonder if we can just use JQ? Every dev should have that installed for other Charts required processes anyway?

mallardduck (changes requested):

scripts/pull-scripts
nicholasSUSE: @mallardduck I have implemented the changes that fixes this. 

summary:
- remove cat on binary file
- remove awk in favor of jq

mallardduck (approved):

PR #5359, opened 2025-03-19
[dev-v2.10] rancher-vsphere-csi 105.0.1+up3.3.1-rancher7 Add option priority class for csi node daemonset

epacke:

`release.yaml`:

`Chart.yaml and index.yaml`:
  - kube-version annotation
  - rancher-version annotation

---

##### Issue:
Nodes under memory pressure may evict the csi-node daemonset pods. Which in turn prevents pods depending on PVCs from starting.

##### Solution
My suggestion is to add a priority class option to the helm chart

Induce memory pressure on a node by creating a pod which uses upp all memory available for pods.


nicholasSUSE:
Hello @epacke I appreciate you wanting to help, however this chart is deprecated: https://github.com/rancher/rancher/issues/52642

We are no longer providing any new versions bumps for it at any version line.

brandond:
This chart is now maintained at https://github.com/rancher/vsphere-charts, and only for RKE2.

PR #5416, opened 2025-03-27
[dev-v2.11] rancher-webhook 106.0.0+up0.7.0 add

ericpromislow:

`release.yaml`:
- [ x] Each chart version in release.yaml DOES NOT modify an already released chart. If so, stop and modify the versions so that it releases a net-new chart.
- [ x] Each chart version in release.yaml IS exactly 1 more patch or minor version than the last released chart version. If not, stop and modify the versions so that it releases a net-new chart.

`Chart.yaml and index.yaml`:
- [x ] The `index.yaml` file has an entry for your new chart version.
- [x ] The `index.yaml` entries for each chart matches the `Chart.yaml` for each chart.
- [x ] Each chart has ALL required annotations
  - kube-version annotation
  - rancher-version annotation


gehrkefc (approved):

ericpromislow:
It's like this PR doesn't exist.  I'm going to create a new one.

PR #5511, opened 2025-04-14
Fix "nodeSelector" in prometheus-adapter

skanakal:
#### Pull Requests Rules

- Pull request title:

    ```
    ```

---

`release.yaml`:

`Chart.yaml and index.yaml`:

  - kube-version annotation
  - rancher-version annotation

---

##### Issue: 

https://github.com/rancher/rancher/issues/49833

##### Solution

aligned as similar to [105.0.0](https://github.com/rancher/charts/blob/release-v2.10/charts/rancher-monitoring/105.0.0%2Bup57.0.3/charts/prometheus-adapter/templates/deployment.yaml#L108C7-L111C11)


mallardduck:
Thank you for the PR @skanakal - however, this is not the ideal way to resolve this. The O&B team now maintains the Monitoring and Logging charts in: https://github.com/rancher/ob-team-charts

From what I can see the newest stable version from our new repo should have this issue fixed. As such a better fix will be for our team to backport that fixed version from 2.11 to 2.10.

Edit: For more details, in 2.10.6 our team plans to backport the newer versions, however those are tied to new versions of Prometheus Federator (V2 to be specific) which are still in progress.

Edit 2: I'm verifying that the version of this chart in `ob-team-charts` has this fix. If so we can look into if there's bandwidth to use that version to implement this fix.

mallardduck:
An alternative PR is here: https://github.com/rancher/charts/pull/5512

This alternative PR pulls from ob-team-charts but has the same fix. The fix changes can be reviewed here: https://github.com/rancher/ob-team-charts/pull/58 As we have a PR that will pull from `ob-team-charts` I would favor that option. As it gets 2.10 release onboarded to the new system.

PR #5583, opened 2025-04-22
[dev-v2.9] Forward ports from 2.8.15

nicholasSUSE:
Issue: https://github.com/rancher/ecm-distro-tools/issues/557
Forward porting released charts in 2.8.15 cycle


rafaelbreno (approved):

tomleb (approved):
LGTM but.. (And let me know if that's just me not understanding this forward porting process correctly) with webhook we have 1 release lines for each Rancher release line which means a version for v2.8 will always only be for v2.8, etc (Concretely, `103.0.17+up0.4.18` will never be needed for `dev-v2.9`, nor `dev-v2.10`, nor `dev-v2.11`)  

I'm not sure there's a point in forward porting rancher-webhook this way. Would it save time for future release not to do it? Or do we need to do it for reason I don't understand / know?

PR #5605, opened 2025-04-26
Update Chart.yaml

xiaofengzhang-1:
#### Pull Requests Rules

- Pull request title:

    ```
    ```

---

`release.yaml`:

`Chart.yaml and index.yaml`:

  - kube-version annotation
  - rancher-version annotation

---


mallardduck:
As mentioned here: https://github.com/rancher/charts/pull/5603#issuecomment-2832478121
> Rancher 2.9's k8s support range is 1.27-1.30. These changes will not be accepted because of that; additionally these are not the correct way to provide changes to charts for a number of reasons. Please update your Rancher version to something higher than 2.9 and use the version of Rancher logging created for that minor version.

PR #5645, opened 2025-05-08
[dev-v2.10] rancher-logging 105.3.0-rc1+up4.10.0-rancher.4 create

diogoasouza:

`release.yaml`:

`Chart.yaml and index.yaml`:

  - kube-version annotation
  - rancher-version annotation

---


adamkpickering (changes requested):
Since this is including the changes from https://github.com/rancher/ob-team-charts/pull/50, which is a removal, I'd argue that we should be bumping minor version rather than patch. I.e. `105.3.0`, rather than `105.2.3`.

Otherwise LGTM.

adamkpickering (approved):

PR #5649, opened 2025-05-09
[dev-v2.9] rancher-cis-benchmark 6.9.0 unrc

krunalhinguu:
#### Pull Requests Rules

- Pull request title:

    ```
    ```

---

`release.yaml`:

`Chart.yaml and index.yaml`:

  - kube-version annotation
  - rancher-version annotation

---


vardhaman22 (approved):

release.yaml
```diff
@@ -23,3 +23,7 @@ rancher-gke-operator-crd:
   - 104.9.0+up1.9.9-rc.1
 rancher-webhook:
   - 103.0.17+up0.4.18
+rancher-cis-benchmark:
+  - 104.9.0+up6.9.0
```
vardhaman22, line 27: chart version needs to be just 6.9.0 instead of 104.9.0+up6.9.0

release.yaml
```diff
@@ -23,3 +23,7 @@ rancher-gke-operator-crd:
...
+rancher-cis-benchmark-crd:
+  - 104.9.0+up6.9.0
```
vardhaman22: chart version needs to be just 6.9.0

vardhaman22 (commented):
need to update the chart version in release.yaml and PR title
LGTM otherwise

vardhaman22 (approved):

PR #5651, opened 2025-05-09
[dev-v2.11] rancher-monitoring 106.1.0-rc.2+up69.8.2-rancher.4 create

jbiers:


diogoasouza:
Is there a reason the chart uses both `v4.0.2` and `v4.0.1` versions of `rancher/kuberlr-kubectl`?

diogoasouza (approved):

jbiers:
@diogoasouza 

> Is there a reason the chart uses both `v4.0.2` and `v4.0.1` versions of `rancher/kuberlr-kubectl`?

Probably when the image versions for kuberlr were bumped, one of the places where it is mentioned went unnoticed :thinking:

PR #5767, opened 2025-06-13
[dev-v2.12] rancher-monitoring 106.1.1+up69.8.2-rancher.4 bump

SamuZad:

`release.yaml`:

`Chart.yaml and index.yaml`:

  - kube-version annotation
  - rancher-version annotation

---

##### Solution
The only change made to the chart is reworking how the nginx template works in `charts/rancher-monitoring/106.1.1+up69.8.2-rancher.4/charts/grafana/templates/nginx-config.yaml`

I added a new map in the nginx config:
```
      map $host $app_suburl {
        {{- if eq .Values.global.cattle.clusterId "local" -}}
        default "/api/v1/namespaces/{{ template "grafana.namespace" . }}/services/http:{{ template "grafana.fullname" . }}:{{ .Values.service.port }}/proxy"';
        {{- else -}}
        default "/k8s/clusters/{{ .Values.global.cattle.clusterId }}/api/v1/namespaces/{{ template "grafana.namespace" . }}/services/http:{{ template "grafana.fullname" . }}:{{ .Values.service.port }}/proxy"';
        {{- end -}}
        {{- if .Values.ingress.enabled }}
        {{- range $host := .Values.ingress.host -}}
        $host "";
        {{- end -}}
        {{- end -}}
      }
  ...
  sub_filter '"appSubUrl":""' '"appSubUrl":"$app_suburl"';
```

The default behaviour remains the same: if the target not called through an ingress, the behaviour will not change.
If, however, the url is one of the ingress hosts, it will no longer append the "proxy path", making for more manageable URLs.

N/A


mallardduck:
To resolve this issue please submit the PR with patch changes (within the `package` folders) to the repo here: https://github.com/rancher/ob-team-charts

once Patched and tested there, our team will handle releasing the fixes to this repo.

PR #5768, opened 2025-06-13
[dev-v2.12] rancher-monitoring 107.0.0-rc.1+up69.8.2-rancher.8 create

jbiers:


index.yaml
```diff
@@ -1062,28 +1062,6 @@ entries:
     urls:
     - assets/fleet/fleet-107.0.0+up0.13.0-alpha.5.tgz
     version: 107.0.0+up0.13.0-alpha.5
-  - annotations:
```
diogoasouza, line 1065: This shouldn't be removed from the index.yaml in this pr

diogoasouza (changes requested):

index.yaml
jbiers: I think this is caused by [this previous PR](https://github.com/rancher/charts/pull/5761) to dev-v2.12 not including changes to index.yaml when removing this old fleet chart, so the stale references in index.yaml are now getting removed when we run `make charts`

jbiers: These changes essentially remove references to the fleet 107.0.0+up0.13.0-alpha.3, which no longer exists and was replaced by 107.0.0+up0.13.0-alpha.5

diogoasouza: I see, thanks for the context. 
Tagging @nicholasSUSE because I think auto-bumps should be updating the index.yaml file correctly, right?

diogoasouza (approved):

PR #5769, opened 2025-06-13
[dev-v2.12] prometheus-federator 107.0.0+up4.0.0-rc.3 create

jbiers:


index.yaml
```diff
@@ -1062,28 +1062,6 @@ entries:
     urls:
     - assets/fleet/fleet-107.0.0+up0.13.0-alpha.5.tgz
     version: 107.0.0+up0.13.0-alpha.5
-  - annotations:
```
diogoasouza: I don't think this should be getting removed in this PR

diogoasouza (changes requested):

index.yaml
jbiers: Same case as discussed in the rancher-monitoring PR :)

diogoasouza (approved):

PR #5941, opened 2025-07-17
[release-v2.9] fleet 104.1.1+up0.10.5 remove(hiding)

nicholasSUSE:
Issue: https://github.com/rancher/ecm-distro-tools/issues/524

Hiding buggy fleet version that was previously released


nicholasSUSE:
I am skipping the validation; it will throw a failure because we are modifying an already released chart. 

But this version needs to be hidden from the user because it has a severe bug.

PR #5945, opened 2025-07-17
[release-v2.11] Forward-port hide fleet

nicholasSUSE:
Issue: https://github.com/rancher/ecm-distro-tools/issues/524

Forward-porting hidden version of fleet.
Parent PR: https://github.com/rancher/charts/pull/5940


nicholasSUSE:
The validation says we can't modify an already released chart.
However, we need to hide this buggy fleet version.
I am bypassing and merging.

PR #5978, opened 2025-07-23
[dev-v2.12] rancher-backup 107.0.0+up8.0.0 create

jbiers:


release.yaml
```diff
@@ -30,11 +30,13 @@ rancher-aks-operator-crd:
...
 rancher-backup:
-  - 107.0.0+up8.0.0-rc.9
+  - 107.0.0+up8.0.0
   - 106.0.1-rc.1+up7.0.1-rc.1
-rancher-backup-crd:
   - 107.0.0+up8.0.0-rc.9
+rancher-backup-crd:
+  - 107.0.0+up8.0.0
   - 106.0.1-rc.1+up7.0.1-rc.1
+  - 107.0.0+up8.0.0-rc.9
```
diogoasouza: Why are we adding/keeping the rc versions?

index.yaml
```diff
@@ -16544,27 +16589,6 @@ entries:
     - assets/rancher-compliance-crd/rancher-compliance-crd-107.0.0+up1.0.0-rc.5.tgz
     version: 107.0.0+up1.0.0-rc.5
   rancher-csp-adapter:
-  - annotations:
```
diogoasouza, line 16547: ~Was this intentional or is the side-effect of `autobump`?~

diogoasouza (changes requested):

index.yaml
diogoasouza: Nvm, it seems that the unrc [PR](https://github.com/rancher/charts/pull/5977) for the csp-adapter didn't update the index.yaml.

release.yaml
jbiers: def not intentional, thanks for pointing that out. fixed :D

diogoasouza (approved):

PR #5980, opened 2025-07-24
[dev-v2.12] prometheus-federator 107.0.0+up4.0.0 create 

susesamu:


index.yaml
```diff
@@ -12030,6 +12030,38 @@ entries:
     - assets/neuvector-monitor/neuvector-monitor-102.0.3+up2.6.0.tgz
     version: 102.0.3+up2.6.0
   prometheus-federator:
+  - annotations:
```
jbiers, line 12033: For some reason the script only added the new version to the index.yaml without removing the previous one, and the same thing happened with the BRO PR (which is why we see this PR removing the old BRO references). 

I'll approve it as I have a logging PR in line which will do the job once I run the scripts there.

jbiers (approved):

PR #5981, opened 2025-07-24
[dev-v2.12] rancher-logging 107.0.0+up4.10.0-rancher.6 create

jbiers:


Jono-SUSE-Rancher (approved):
I don't know a whole lot about un-rcing, but if it means removing the -rc, this looks good to me ;)

diogoasouza (approved):
lgtm

PR #6093, opened 2025-08-11
[release-v2.12] removing rancher-windows-gmsa

nicholasSUSE:
Issue: https://github.com/rancher/ecm-distro-tools/issues/629


nicholasSUSE:
Safe to remove chart, no one is using and this branch was recenlty released only for community

PR #6353, opened 2025-10-14
[dev-v2.12] rancher-monitoring - 106 and 107 - Fixed indentation issue with metricRelabelings in multiple servicemonitor manifests

belgaied2:
Signed-off-by: Mohamed Belgaied <mohamed.belgaied@suse.com>

#### Pull Requests Rules

- Pull request title:

    ```
    ```

---

`release.yaml`:

- [X] Each chart version in release.yaml DOES NOT modify an already released chart. If so, stop and modify the versions so that it releases a net-new chart.
- [X] Each chart version in release.yaml IS exactly 1 more patch or minor version than the last released chart version. If not, stop and modify the versions so that it releases a net-new chart.

`Chart.yaml and index.yaml`:

  - kube-version annotation
  - rancher-version annotation

---

##### Issue: 
https://github.com/rancher/rancher/issues/43392

##### Solution
The `metricRelabelings` attribute in `ServiceMonitor` takes in a list of maps coming from the `values.yaml`.
In order to template that in helm, the previous implementation did the following :\
```
  {{ tpl (toYaml .Values.XXX.metricRelabelings ) . | indent 4 }}
```
This does not work with multiline relabelings, such as:
```
  - action: keep
    sourceLabels: [__name__]
    regex: <some_regex>
```
which would generate:
```
  metricRelabelings:
      - action: keep
    sourceLabels: [__name__]
    regex: <some_regex>
```
AND throw an error!

We need to remove first space characters using `{{- tpl ...` and do multiline indentation using `nindent` instead of `indent`.

That's what this PR does for each location of `metricRelabelings` in versions 106 and 107 of the `rancher-monitoring` chart.
  

Add automatic testing for multiline indentations, specifically around `metricRelabelings` which is a VERY frequent use case.

### DIFF
--- a/charts/rancher-monitoring/106.0.0+up66.7.1-rancher.2/templates/exporters/core-dns/servicemonitor.yaml
+++ b/charts/rancher-monitoring/106.0.0+up66.7.1-rancher.2/templates/exporters/core-dns/servicemonitor.yaml
@@ -43,7 +43,7 @@ spec:
     bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
     metricRelabelings:
     {{- if .Values.coreDns.serviceMonitor.metricRelabelings }}
-    {{ tpl (toYaml .Values.coreDns.serviceMonitor.metricRelabelings | indent 4) . }}
+    {{- tpl (toYaml .Values.coreDns.serviceMonitor.metricRelabelings | nindent 4) . }}
     {{- end }}
     {{ if .Values.global.cattle.clusterId }}
     - sourceLabels: [__address__]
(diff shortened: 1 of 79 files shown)


susesamu:
@belgaied2 changes on rancher-monitoring should be first applied at [ob-team-charts](https://github.com/rancher/ob-team-charts) following this [pattern](https://github.com/rancher/charts/wiki/1%E2%80%90Developing).
What we need to do is to patch this fix on the next release of 2.12 (rancher-monitoring 69.8.2-rancher.26), and create a backport for 2.11.
We should also check for this on 2.13 which will be the next minor.

PR #6412, opened 2025-10-28
[dev-v2.12] removing wrong RC for fleet

nicholasSUSE:
```
make remove CHART="fleet" VERSION=107.1.0+up0.13.4-rc.2
make remove CHART="fleet-crd VERSION=107.1.0+up0.13.4-rc.2
make remove CHART="fleet-agent" VERSION=107.1.0+up0.13.4-rc.2
```


weyfonk:
What is wrong about this RC?
I would be tempted to remove rc.1, or even both RCs now that v0.13.4 has been released, but not sure I understand why rc.2 in particular would need to be removed :thinking:

thardeck:
> What is wrong about this RC?

Because of an out of band release I have bumped the Rancher chart minor version for the next Rancher release with auto bump, following the policy, but the un-rc auto-bump run decreased the minor-number again without us noticing.

For the end user that is not a problem, since the final Rancher chart  versions are still in proper succession. That's why this development release were supposed to be deleted.

This can be closed, because it was fixed in another pr.

PR #6513, opened 2025-11-20
[dev-v2.13] rancher-monitoring + prometheus-federator unRCs

mallardduck:


index.yaml
```diff
@@ -1428,6 +1406,28 @@ entries:
     urls:
     - assets/fleet/fleet-108.0.0+up0.14.0-alpha.3.tgz
     version: 108.0.0+up0.14.0-alpha.3
+  - annotations:
```
adamkpickering, line 1409: This is a little weird, but doesn't appear to change anything from a functional perspective. Just calling this out in case we can avoid it.

adamkpickering (approved):

PR #6528, opened 2025-11-21
Remove Fleet 107.1.0+up0.13.4-rc.2

thardeck:
we initially bumped to 107.1.0 because there was an inbetween Rancher versions Fleet release.

But when we used chart autobump to release the following development versions the minor version was decreased again, which went unnoticed initially.

To prevent issues in the release script, we are removing this release. It is an old development version anyway which final release is long out.


mmartin24 (approved):
Thanks for the cleansing

PR #6650, opened 2025-12-16
[dev-v2.13] elemental 107.1.0+up1.7.3 update

ipetrov117:
#### Pull Requests Rules

- Pull request title:

    ```
    ```

---

`release.yaml`:

`Chart.yaml and index.yaml`:

  - kube-version annotation
  - rancher-version annotation

---


charts/elemental/107.1.0+up1.7.3/Chart.yaml
```diff
@@ -7,7 +7,7 @@ annotations:
   catalog.cattle.io/os: linux
   catalog.cattle.io/permits-os: linux
   catalog.cattle.io/provides-gvr: elemental.cattle.io/v1beta1
-  catalog.cattle.io/rancher-version: '>= 2.12.0-0 < 2.13.0-0'
+  catalog.cattle.io/rancher-version: '>= 2.13.0-0 < 2.14.0-0'
```
ldevulder, line 10: Does this mean that we cannot install anymore this version of Elemental on Rancher v2.12?

ipetrov117: No, this change is only for the `dev-v2.13` branch, which will be transferred to `release-v2.13`, which is only for the Rancher `v2.13.X` versions, IIUC. This should not reflect the previous versions.

ldevulder (approved):

davidcassany (approved):

PR #7041, opened 2026-03-13
[dev-v2.14] rancher-backup 109.0.0+up10.0.0-rc.3 bump

mallardduck:


mallardduck:
Note that this PR needed to have the `Chart.yaml` patch updated to remove the k8s version parts now that the new RC has that fix by default (so we don't need to apply it here too).

diogoasouza (approved):

PR #7117, opened 2026-03-24
Copy/Forward harvester-cloud-provider on v214

w13915984028:
#### Pull Requests Rules

- Pull request title:

    ```
    ```

---

`release.yaml`:

`Chart.yaml and index.yaml`:

  - kube-version annotation
  - rancher-version annotation

---

##### Issue: 

https://github.com/rancher/charts/pull/7108 Did not forward harvester-cloud-provider to dev-2.14 branch, not sure we need  a `harvester-cloud-provider v109.0.0+up0.2.11`.


release.yaml
```diff
@@ -236,3 +236,5 @@ rancher-supportability-review:
   - 108.0.0+up0.1.5
 rancher-supportability-review-crd:
   - 108.0.0+up0.1.5
+harvester-cloud-provider:
```
wheatdog, line 239: This list seems sorted alphabetically. Suggest to move it above `harvester-csi-driver`.

charts/harvester-cloud-provider/109.0.0+up0.2.11/Chart.yaml.rej
```diff
@@ -0,0 +1,27 @@
+--- Chart.yaml
```
ihcsim, line 1: how come this rejected patch got included in the chart?

packages/harvester/harvester-cloud-provider/generated-changes/patch/Chart.yaml.patch
```diff
@@ -5,12 +5,12 @@
    catalog.cattle.io/certified: rancher
    catalog.cattle.io/display-name: Harvester Cloud Provider
 -  catalog.cattle.io/kube-version: '>= 1.23.0-0'
-+  catalog.cattle.io/kube-version: '>= 1.23.0-0 < 1.35.0-0'
++  catalog.cattle.io/kube-version: '>= 1.23.0-0 < 1.36.0-0'
```
ihcsim, line 8: Are these changes manually generated by `make prepare`, modify the temporary `Chart.yaml` at `packages/harvestr-cloud-provider/charts`, and then `make patch`? If yes, how do we know what the `rancher-version`, `kube-version` ranges to use in the temporary `Chart.yaml`?

I am assuming that if all these annotations are present in the chart in `h/charts`, we can just run `make charts` without `make prepare` and `make patch`?

ihcsim (commented):
Some questions to help me better understand the workflow here - thanks.

w13915984028:
rework, close for now.

PR #7226, opened 2026-04-14
feat(ci): add github token auth and improve script logging

nicholasSUSE:
Issue: https://github.com/rancher/ecm-distro-tools/issues/817

### Goal

`automation-core` branch will be the source of truth for GitHub Actions and operation Scripts to all active branches. 

## Summary
- Add github_token input to build composite action for authenticated gh api calls in pull-scripts
- Add curl fallback in pull-scripts when gh api fails without token
- Add colored logging ([PASS]/[FAIL]/[SKIP]/[INFO]) to package-ci with counters and summary output
- Add auto: true package filtering in package-ci to skip packages that track upstream branches
- Replace hardcoded package list with dynamic PORCELAIN=1 make list

### DIFF
--- a/scripts/package-ci
+++ b/scripts/package-ci
@@ -29,7 +29,7 @@ log_skip() {
 }
 
 log_error() {
-    echo -e "${RED}[FAIL]${RESET} $1"
+    echo -e "${RED}[FAIL]${RESET} $1" >&2
 }
 
 # Validate clean git state
--- a/scripts/package-ci
+++ b/scripts/package-ci
@@ -1,44 +1,129 @@
 #!/bin/bash
+set -e
+
+# ============================================================================
+# Package CI - Validates all packages pass the build workflow
+# ============================================================================
 
 cd $(dirname $0)
 cd ..
 
+# Colors
+RESET='\033[0m'
+GREEN='\033[0;32m'
+YELLOW='\033[0;33m'
+RED='\033[0;31m'
+BLUE='\033[0;34m'
+
+# Logging functions
+log_info() {
+    echo -e "${BLUE}[INFO]${RESET} $1"
+}
+
+log_success() {
+    echo -e "${GREEN}[PASS]${RESET} $1"
+}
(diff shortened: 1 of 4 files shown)


scripts/package-ci
```diff
@@ -1,44 +1,129 @@
...
+log_success() {
+    echo -e "${GREEN}[PASS]${RESET} $1"
+}
+
+log_skip() {
+    echo -e "${YELLOW}[SKIP]${RESET} $1"
+}
+
+log_error() {
+    echo -e "${RED}[FAIL]${RESET} $1"
```
tashima42:
```suggestion
log_error() {
    echo -e "${RED}[FAIL]${RESET} $1" >&2
```
Write to stderr. Ignore if you don't think this is necessary

scripts/pull-scripts
```diff
@@ -47,42 +77,76 @@ detect_platform() {
...
+        log "Script already exists, skipping: ${script_path}"
+        return 0
+    fi
+
+    log "Downloading script: ${script_path}"
+    mkdir -p "$(dirname "$target_path")"
+    git show upstream/automation-core:scripts/${script_path} > "$target_path"
+    chmod +x "$target_path"
+    log "Downloaded: ${script_path}"
+}
```
tashima42, line 133: Wouldn't be easier to checkout the whole automation-core branch at a temporary location outside of the main workspace? I think this would maintain the scripts permissions.

briandowns (approved):

scripts/pull-scripts
nicholasSUSE: Thanks for the suggestion! I considered using git worktree, but the current git show approach is simpler and achieves the same result. The explicit chmod +x handles permissions reliably, and we avoid the complexity of managing temporary worktrees. Since this runs frequently (every make pull-scripts), keeping it lightweight is preferable.

tashima42 (approved):

PR #7246, opened 2026-04-16
[automation-core] improve pull-scripts and build job

nicholasSUSE:
Issue: https://github.com/rancher/ecm-distro-tools/issues/817

## Summary
- Add SKIP_PS mode to pull-scripts to skip version resolution and binary download when binary already exists
- Set SKIP_PS=true in all build validation steps to prevent rate-limiting from repeated GitHub API calls
- Optimize resolve_version to return immediately after successful gh api call instead of attempting curl fallback
- Restore correct function order (resolve_version before check_existing_binary) so version comparison works
- Add pull-requests write permission to workflow and make PR comment posting fork-safe with graceful error handling
- Improve package-ci error reporting by capturing and displaying stderr when make commands fail
- Add binary existence check in package-ci as safety guard before setting SKIP_PS

### DIFF
--- a/.github/workflows/build-core.yaml
+++ b/.github/workflows/build-core.yaml
@@ -20,6 +20,7 @@ jobs:
     permissions:
       contents: read
       id-token: write  # Required for OIDC authentication with Vault
+      pull-requests: write  # Required for posting build validation summary comment
     steps:
       # ====================================================================
       # WORKFLOW CONTEXT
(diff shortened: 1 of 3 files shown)


.github/workflows/build-core.yaml
```diff
@@ -20,6 +20,7 @@ jobs:
     permissions:
       contents: read
       id-token: write  # Required for OIDC authentication with Vault
+      pull-requests: write  # Required for posting build validation summary comment
```
nicholasSUSE, line 23: @tashima42

tashima42 (approved):

PR #7392, opened 2026-04-27
[release-v2.13] 2.13.5 release

nicholasSUSE:
Issue: https://github.com/rancher/ecm-distro-tools/issues/865

### Checkpoint 0: Validate **release.yaml**

Validation steps:
- [x] Each chart version in **release.yaml** DOES NOT modify an already released chart. If so, stop and modify the versions so that it releases a net-new chart.
- [x] Each chart version in **release.yaml** IS exactly 1 more patch version than the last released chart version. If not, stop and modify the versions so that it releases a net-new chart.

Validation steps:
- [x] Running **make unzip** to regenerate the **charts/** from scratch, then **git diff** to check differences between **assets/** and **charts/** yields NO differences or innocuous differences.

Validation steps:
- [x] The **index.yaml** file has an entry for each chart version.
- [x] The **index.yaml** entries for each chart matches the **Chart.yaml** for each chart.
- kube-version annotation
- rancher-version annotation


briandowns (approved):

rafaelbreno (approved):

tashima42 (approved):

daleckystepan:
Looks like nginx config broke embedded Metrics tabs inside Rancher v2.13.3. it is stuck in "Loading..."

  chart_name    = "rancher-monitoring"
  chart_version = "108.0.5+up77.9.1-rancher.16"
  
  wo fixes were needed, applied to the live ConfigMap grafana-nginx-proxy-config:

Fix 1 — nginx rewrite double-slash bug (blocked Grafana API calls)
```
# Before (bug): captures "/api/health" → produces "//api/health" → Grafana 404
rewrite ^/k8s/clusters/.*/proxy(.*) /$1 break;

# After: /? consumes the slash → captures "api/health" → produces "/api/health"
rewrite ^/k8s/clusters/.*/proxy/?(.*) /$1 break;
```
Fix 2 — gzip on blocking sub_filter (broke appSubUrl substitution)
```
location / {
  # Added:
  gzip off;
  
  # Removed duplicate no-op sub_filter, kept only:
  sub_filter '"appSubUrl":""' '"appSubUrl":"/k8s/clusters/c-m-pmvwjj6n/api/v1/namespaces/cattle-monitoring-system/services/http:rancher-monitoring-grafana:80/proxy"';
  sub_filter ':"/avatar/' ':"avatar/';
  ...
}
```
gzip on in the server block compressed HTML responses before sub_filter could run, so appSubUrl stayed "". Grafana then tried to load plugin JS from bare paths like /public/plugins/… (without the Rancher proxy prefix) → 404s → dashboard failed to initialize.

Fix 1 unblocked /api/health. Fix 2 made the dashboard actually initialize correctly. Both are in the upstream Helm chart template

PR #7402, opened 2026-04-27
[dev-v2.15] system-upgrade-controller Add `imagePullSecret` support

HarrisonWAffel:
Relates to: https://github.com/rancher/rancher/issues/54806

This PR adds support for image pull secrets in `values.yaml`. It uses a helm helper template to accept pull secrets as either a plain string array, or an array of object references. This allows the SUC to be pulled from an authenticated registry in environments where credentials cannot be configured at the container runtime level (e.g. imported, hosted clusters).


jiaqiluo (commented):
I’m curious about the need to support two formats for the same field—it could lead to ~inconsistent or~ confusing behavior if they’re used interchangeably. One option is to split this into two similar fields: one that accepts a list of strings and another that accepts a map, although that still feels somewhat redundant. Alternatively, we could simplify the design by standardizing on a single format.
Open to discussion.

HarrisonWAffel:
I agree that if someone passed both formats at the same time it may be harder to understand, but I don't think it would lead to inconsistent behavior, unless there's a bug in the helper file that you can see. I also don't think splitting this field into two would be worth while.

I opted to support both formats since there are differing expectations on the format of this field in different Rancher components. For instance, the primary rancher and remotedialer-proxy charts expect local object references, but turtles expects a list of strings. It's not immediately obvious what you need to pass without looking at how the value is used, so supporting both can help eliminate that guess work. 

I'm not married to the idea though, especially since this is not a user-configurable chart. But, supporting both was only ~10 LOC extra, and this same helper logic could be used in other charts which are user configurable.

jiaqiluo:
> I agree that if someone passed both formats at the same time it may be harder to understand, but I don't think it would lead to inconsistent behavior, unless there's a bug in the helper file that you can see. I also don't think splitting this field into two would be worth while.
> 
> I opted to support both formats since there are differing expectations on the format of this field in different Rancher components. For instance, the primary rancher and remotedialer-proxy charts expect local object references, but turtles expects a list of strings. It's not immediately obvious what you need to pass without looking at how the value is used, so supporting both can help eliminate that guess work.
> 
> I'm not married to the idea though, especially since this is not a user-configurable chart. But, supporting both was only ~10 LOC extra, and this same helper logic could be used in other charts which are user configurable.

Thanks for the explanation. The current approach makes more sense now.

jiaqiluo (approved):

thatmidwesterncoder (approved):

PR #7415, opened 2026-04-28
[automation-core] improve make remove 

nicholasSUSE:
## Summary

- Add colored logging with structured output (info/pass/warn/fail)
- Validate chart and asset paths exist before removal  
- Verify index.yaml updates after regeneration to detect silent failures

### DIFF
--- a/scripts/remove-asset
+++ b/scripts/remove-asset
@@ -1,34 +1,141 @@
 #!/bin/bash
-
 set -e
 
-cd $(dirname $0)
+# ============================================================================
+# Remove Asset - Remove chart version from repository
+# ============================================================================
+
+cd "$(dirname "$0")"
 cd ..
 
+# Colors
+RESET='\033[0m'
+GREEN='\033[0;32m'
+YELLOW='\033[0;33m'
+RED='\033[0;31m'
+BLUE='\033[0;34m'
+
+# Logging functions
+log_info() {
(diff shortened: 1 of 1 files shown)


nicholasSUSE:
Only to  be merged after the release finishes

mallardduck:
This looks good to me and should prevent me from adding `make index` to the BRO automation.

I did realize that I need to adjust the BRO one a bit to ensure I capture the errors instead of ignoring them like I do now: https://github.com/rancher/backup-restore-operator/blob/d72dbda2105b489872bde81dda78863e085ab5f8/.github/workflows/push-to-charts/remove-previous-rc.sh#L45

But that's not related to this issue and this looks like a solid fix!

nicholasSUSE:
> This looks good to me and should prevent me from adding `make index` to the BRO automation.
> 
> I did realize that I need to adjust the BRO one a bit to ensure I capture the errors instead of ignoring them like I do now: https://github.com/rancher/backup-restore-operator/blob/d72dbda2105b489872bde81dda78863e085ab5f8/.github/workflows/push-to-charts/remove-previous-rc.sh#L45
> 
> But that's not related to this issue and this looks like a solid fix!

At least when `make index` fails to remove the entry, it will throw an error, and I will have some reproducible scenario. 
Please, if it fails with you, open a PR with it at the upstream so I can fetch and try to reproduce.

tashima42 (approved):

PR #7430, opened 2026-04-29
Backup new rc 2.14

diogoasouza:


mallardduck (approved):

release.yaml
```diff
@@ -115,8 +115,10 @@ rancher-alerting-drivers:
   - 109.0.0
 rancher-backup:
   - 109.0.2+up10.0.2
+  - 109.0.3+up10.0.3-rc.1
```
prachidamle, line 118: Do we need to remove the earlier version `109.0.2+up10.0.2`?

mallardduck: yes - but we can in next PR when we unRC since we need to remove all of them then.

mallardduck: We don't need to bump `109.0.3` either IMHO if our goal is to expect that the previous releases are not released and therefor not supported. (can also correct in PR to unRC)

PR #7614, opened 2026-05-28
[dev-v2.14] system-upgrade-controller 109.0.2-rc.1 add

vardhaman22:
#### Pull Requests Rules

- Pull request title:

    ```
    ```

---

`release.yaml`:

`Chart.yaml and index.yaml`:

  - kube-version annotation
  - rancher-version annotation

---

##### Issue: 

https://github.com/rancher/rancher/issues/55307
##### Solution

added support for SYSTEM_UPGRADE_JOB_KUBECTL_IMAGE_WINDOWS env variable in system-upgrade-controller chart.
the default values for this env variable will be **undefined** . used helpers to achieve this and kept the entries in values.yaml empty


packages/system-upgrade-controller/charts/values.yaml
```diff
@@ -11,3 +11,6 @@ kubectl:
   image:
     repository: rancher/kuberlr-kubectl
     tag: v7.0.3 # For compatibility and versioning, see https://github.com/rancher/kuberlr-kubectl
+  windowsImage:
```
HarrisonWAffel: `windowsImage` is a bit too generic IMO, could we rename this to `windowsKubectlImage`?  Or even just pass the existing image to the new SUC env var and create a follow up issue in the kuberlr repo to add support for windows images?

vardhaman22: named it windowsImage since it was already inside the kubectl section....but yes the idea of passing the same kubectl image seems good to me. If we decide to host windows kubectl image somewhere else then we can change later i guess.
I have updated the PR based on this suggestion.

HarrisonWAffel (commented):
LGTM, though I think the version should be `109.0.2-rc.1`?

vardhaman22:
> LGTM, though I think the version should be `109.0.2-rc.1`?

updated the version.
Just thought that it is a small change so did not create the rc version.
I will unrc it later after the QA verification

HarrisonWAffel (approved):
LGTM for windows related changes

jakefhyde (approved):

PR #7628, opened 2026-06-01
[dev-v2.15] system-upgrade-controller 110.0.1 update

vardhaman22:
#### Pull Requests Rules

- Pull request title:

    ```
    ```

---

`release.yaml`:

- [X] Each chart version in release.yaml DOES NOT modify an already released chart. If so, stop and modify the versions so that it releases a net-new chart.
- [X] Each chart version in release.yaml IS exactly 1 more patch or minor version than the last released chart version. If not, stop and modify the versions so that it releases a net-new chart.

`Chart.yaml and index.yaml`:

- [X] The `index.yaml` file has an entry for your new chart version.
- [X] The `index.yaml` entries for each chart matches the `Chart.yaml` for each chart.
- [X] Each chart has ALL required annotations
  - kube-version annotation
  - rancher-version annotation

---

##### Issue: 

https://github.com/rancher/rancher/issues/55298
and https://github.com/rancher/rancher/issues/54377
##### Solution

- added the kubectl windows image env variable 
- also update annotation and app version to add 1.36 support
- updated an existing chart since 110.0.1 is still not released it just exist in the dev branch.
also did not create and rc version of the chart to not break CI on rancher main since this version is updated in rancher.

- the app version and image needs to be unrced when an unrced tag for suc is available.


HarrisonWAffel (approved):
LGTM for windows related changes

jakefhyde (approved):

PR #7718, opened 2026-06-11
[dev-v2.15] prometheus-federator 110.0.0+up7.0.0-rc.1 bump

alemorcuq:

`release.yaml`:

`Chart.yaml and index.yaml`:

  - kube-version annotation
  - rancher-version annotation


mallardduck (approved):
LGTM; since our main concern is an RC for QA to test the Helm stuff.
We can still do follow up RCs to bump images and such as needed too btw.

PR #7965, opened 2026-06-24
[dev-v2.15] harvester-cloud-provider 0.2.12 bump

w13915984028:
#### Pull Requests Rules

- Pull request title:

    ```
    ```

---

`release.yaml`:

`Chart.yaml and index.yaml`:

  - kube-version annotation
  - rancher-version annotation

---

##### Issue: 

https://github.com/harvester/harvester/issues/10068

##### Solution

Bump harvester-cloud-provider to upstream Rancher and RKE2.


packages/harvester/harvester-cloud-provider/generated-changes/overlay/Chart.yaml.rej
w13915984028, line 1: note: this was a legacy accidentally added file, it does not affect the chart, but is better to be removed.

I tried to remove it on this PR (with some addtional efforts and tests), some other branches also have this file, they will be removed on later PRs, thanks.

ihcsim (approved):

packages/harvester/harvester-cloud-provider/generated-changes/patch/dependency_charts/kube-vip/values.yaml.patch
```diff
@@ -9,3 +9,10 @@
    pullPolicy: IfNotPresent
    # Overrides the image tag whose default is the chart appVersion.
    # tag: "v0.7.0"
++  # Explicitly uncomment and update this tag to ensure it pass upstrem `rancher/charts` check
++  # which complains:
++  # - orphanImage:  rancher/orphan-image, no tag → orphan_repository
++  tag: "v1.1.2"
```
bk201, line 15: @w13915984028 Can you share how we determine which version to use? thanks.

bk201 (approved):

packages/harvester/harvester-cloud-provider/generated-changes/patch/dependency_charts/kube-vip/values.yaml.patch
w13915984028: The version was tested & fixed when HCP chart was released, e.g.

https://github.com/harvester/charts/blob/67dd00b59335705d9c4f488a173de8748e17906f/charts/harvester-cloud-provider/values.yaml#L139-L141

bk201: Thank you!

PR #7982, opened 2026-06-29
[dev-v2.15] update annotation harvester csi driver

WebberHuang1118:
#### Pull Requests Rules

- Pull request title:

    ```
    ```

---

`release.yaml`:

`Chart.yaml and index.yaml`:

  - kube-version annotation
  - rancher-version annotation

---


Vicente-Cheng (approved):

charts/harvester-csi-driver/110.0.1+up0.1.30/Chart.yaml
```diff
@@ -1,11 +1,11 @@
 annotations:
   catalog.cattle.io/certified: rancher
   catalog.cattle.io/display-name: Harvester CSI Driver
-  catalog.cattle.io/kube-version: '>= 1.23.0-0 < 1.36.0-0'
+  catalog.cattle.io/kube-version: '>= 1.23.0-0 < 1.37.0-0'
```
nicholasSUSE, line 4: This looks good, every new Rancher version line (in this case: v2.15) has a different 
Kubernetes version range supported. 

Just make sure this is aligned with Rancher/Rancher. 

Rule of thumb here: 
- **Ideally**: kub-version range from the charts `==` Rancher/Rancher version range 
- **Maybe**: kube-version range from the charts `inside` the range of Rancher/Rancher version range
- **Not recommended**: kube-version range from the charts `beyond` the range of Rancher/Rancher supported version range

index.yaml
```diff
@@ -7777,19 +7777,19 @@ entries:
...
       catalog.cattle.io/release-name: harvester-csi-driver
       catalog.cattle.io/ui-component: harvester-csi-driver
       catalog.cattle.io/upstream-version: 0.1.30
     apiVersion: v2
     appVersion: v0.2.9
-    created: "2026-06-22T03:39:26.162442369Z"
+    created: "2026-06-29T09:50:09.099285144Z"
     description: A Helm chart for Harvester CSI driver
-    digest: 318f6dc832e1fe49ba80a9d5e521961167b04f3018ab50fd9fa74a04105e67fd
+    digest: 067dc797ce865254e1d82ae2e7573adc8d766357199745c1d57ea7825ac1feb0
```
nicholasSUSE, line 7792: This is very important, I will leave here a deep explanation and also the rule of thumb. 

This is one of the reasons **we never modify already released charts**. 

This PR is modifying **an already bumped chart**!


It is ok to modify an already **bumped chart** **IF and ONLY IF** that chart was not already released. 

Otherwise, that digest changes. 
I just checked, this chart was not released, so that is why this is ok. 

--- 

### Rule of thumb:
**Naming convention**
- bumped charts = new versions at **dev-v2.*** branches
- released charts = new versions at **release-v2.*** branches

**Digest changed?** = Warning, double the attention. 
- Chart already released? stop everything
- Chart not released already? it's ok

nicholasSUSE (approved):

WebberHuang1118:
@nicholasSUSE Thanks for your detailed explanation.

PR #8355, opened 2026-07-24
[automation-core] harvester 2.13,2.14,2.15 UnRC

ihcsim:
#### Pull Requests Rules

- Pull request title:

    ```
    ```

---

`release.yaml`:

`Chart.yaml and index.yaml`:

  - kube-version annotation
  - rancher-version annotation

---


release/2.13.8.yaml
```diff
@@ -4,7 +4,7 @@ elemental:
     QA: true
     UnRC: true
     Released: false
-  
+
```
nicholasSUSE: Don't jump the line

release/2.13.8.yaml
```diff
@@ -16,7 +16,7 @@ elemental-crd:
...
```
nicholasSUSE: Don't jump the line

ihcsim:
@nicholasSUSE per our discussion, i've updated the release files to include chart versions found in the `charts/harvester-*` folders in the different `dev-v2.N` branches; for now, they are marked as `ToRelease: true` - chart owners of the csi driver and cloud provider can chime in.

release/2.14.4.yaml
```diff
@@ -42,26 +42,36 @@ fleet-crd:
...
     Released: false
   109.0.0+up0.2.12:
-    ToRelease: false
-    QA: false
-    UnRC: false
+    ToRelease: true
+    QA: true
+    UnRC: true
     Released: false
 harvester-csi-driver:
```
nicholasSUSE, line 54: These 3 versions are being marked to release; however, they were all already released!
Don't mark an already released version to be released again. 
This harvester-csi-driver could be just empty as there is nothing to release. 

Image

release/2.14.4.yaml
```diff
@@ -42,26 +42,36 @@ fleet-crd:
...
+    ToRelease: true
+    QA: true
+    UnRC: true
     Released: false
 harvester-rbac:
-  <version>:
-    ToRelease: false
-    QA: false
-    UnRC: false
+  109.0.0+up0.1.1:
```
nicholasSUSE: This version was already released, don't mark it to release again:

Image

release/2.13.8.yaml
```diff
@@ -54,20 +54,40 @@ fleet-crd:
...
     Released: false
   108.0.1+up0.2.12:
-    ToRelease: false
-    QA: false
-    UnRC: false
+    ToRelease: true
+    QA: true
+    UnRC: true
+    Released: false
+  108.0.0+up0.2.11:
```
nicholasSUSE: 108.0.0 was already released; just remove the entire entry for it. The other ones were not released yet, so they are right. 

Image

nicholasSUSE (changes requested):

release/2.14.4.yaml
ihcsim: won't we want the release files to match the release branches? the [wiki](https://github.com/rancher/charts/wiki/9-%E2%80%90-GitOps-Release-Tracking) should be updated to mention the `release-v2.*` branches mark the final release states of the charts.

nicholasSUSE: They will match at the end of the file. The Wiki was updated. 
The goal here is to replace the previous existing system of the Confluence Tables first. 
This is exactly what it did.

nicholasSUSE (approved):

PR #8445, opened 2026-08-05
[dev-v2.12] Supportability Review 0.1.8 bump

masap:
#### Pull Requests Rules

- Pull request title:

    ```
    ```

---

`release.yaml`:

`Chart.yaml and index.yaml`:

  - kube-version annotation
  - rancher-version annotation

---


packages/rancher-supportability-review/charts/questions.yaml
```diff
@@ -45,14 +45,14 @@ questions:
...
       default: "rancher/mirrored-sonobuoy-sonobuoy"
       description: "The repository of sonobuoy image"
       type: string
       required: false
     - variable: image.sonobuoy.tag
-      label: Collector image Tag
-      default: "v0.57.3"
+      label: Sonobuoy image Tag
+      default: "v0.57.5"
       description: "The repository of sonobuoy Tag"
```
piyushkrana1, line 56: Change Description To : "The tag of sonobuoy repository"

release.yaml
```diff
@@ -83,9 +83,9 @@ fleet-crd:
   - 106.1.15+up0.12.17
   - 106.1.16+up0.12.18
 rancher-supportability-review:
-  - 107.0.5+up0.1.7
+  - 107.0.6+up0.1.8
```
piyushkrana1, line 87: This conflicts with the PR’s rule:

1. Never remove an already released chart.

release.yaml should preserve the old version and append the new one

release.yaml
```diff
@@ -83,9 +83,9 @@ fleet-crd:
...
 rancher-supportability-review-crd:
-  - 107.0.5+up0.1.7
+  - 107.0.6+up0.1.8
```
piyushkrana1, line 90: This conflicts with the PR’s rule:

Never remove an already released chart.
release.yaml should preserve the old version and append the new one

piyushkrana1 (approved):
LGTM overall 
Left a couple of minor grammar suggestions.

packages/rancher-supportability-review/charts/questions.yaml
masap: Thank you. Yes, I won't fix this for this release. I already fixed it for next release.

release.yaml
masap: Yes, fixed.

release.yaml
masap: Fixed as well.

PR #8563, opened 2026-08-19
[dev-v2.15] feat test auto-merge

nicholasSUSE:
Issue: https://github.com/rancher/ecm-distro-tools/issues/972


briandowns (approved):

.github/workflows/auto-bump-merge.yaml
```diff
@@ -40,13 +40,45 @@ jobs:
...
+
+          while true; do
+            STATUS=$(gh pr checks "$PR_NUMBER" --repo "$REPO" --json state --jq 'map(.state) | unique | join(" ")')
+
+            # If any check is PENDING or IN_PROGRESS, keep waiting
+            if echo "$STATUS" | grep -qE "PENDING|IN_PROGRESS"; then
+              echo "Checks still running: $STATUS"
+              sleep 30
+              continue
+            fi
```
tashima42, line 59: Isn't this going to spend a lot of billable hours just waiting for other CI? Would it be possible to run something like the publish provisioning test results workflow that is picked up after other workflows are completed? https://github.com/rancher/rancher/blob/main/.github/workflows/publish-provisioning-test-results.yaml

nicholasSUSE: To be honest, here in Charts we basically have one CI that takes only 10-20 minutes. 
Your idea is good, but it would be more effort for a low result; however, I will keep this in mind once this Proof of concept is working.

thardeck: We can also move the auto-merge part as separate job to the build.yaml, then it is only triggered when CI is green, but can be skipped with the same criteria like I have implemented in this [pr](https://github.com/rancher/charts/pull/8620).

PR #8845, opened 2026-09-14
Add ORBS 2.15 chart releases

mallardduck:


release/2.15.2.yaml
rmoioliveira, line 92: `prometheus-federator` has not released the last cycle due to this: https://github.com/rancher/rancher/issues/56795

We have to release this cycle.

alemorcuq: There are some RCs for Prometheus Federator that are ready to be pushed to `rancher/charts`: https://github.com/rancher/prometheus-federator/releases

rmoioliveira: We need to UnRC version [v7.0.1-rc.3](https://github.com/rancher/prometheus-federator/releases/tag/v7.0.1-rc.3)

diogoasouza (approved):

